PR #74, opened 2013-04-13
Subtitle priority dropdowns

SubJunk:


Nadahar:
@SubJunk I think the idea is good, so I've taken the liberty to rebase it onto the current master and force pushed it. I have a local copy of the original branch if it's needed in any way, but all the changes are here - I just had to rearrange the labels and rows somewhat to make it fit into the current master.

I don't think this is the way to implement it though, it must be made dynamic with a table or something of that kind so that the user can add and remove rows, and move them up and down. Maybe there is such a component for swing already?

SubJunk:
@Nadahar cool, yeah I don't think I was implementing it in a great way here so feel free to make any changes you want :) It would be a great feature to have.
Our folder selection GUI on the Navigation tab is similar to that so I wonder if that could work

Sami32:
@SubJunk Doesn't it already implemented ? or i misunderstood...

SubJunk:
@Nadahar We should either advance this or close the PR

Nadahar:
@SubJunk Yeah, I have way too much on my plate already, just close it for now.

PR #77, opened 2013-04-16
Filename folders

SubJunk:
This creates virtual folders for TV shows inside a real folder.
For now it works pretty well but it will need some tweaking before it is ready to use I think.

There are some different directions to take this. So right now it works like this:
It looks for TV episodes in the current folder, and puts them in virtual subfolders. The files that are not matched as TV episodes are added to the folder like normal. I'm not sure if this is the most intuitive way to do it though, so I have a couple of possible ideas:

1) When this feature is enabled, add files from subfolders into the main folder too, so it's all on the same level. I wonder if we can do this quickly enough to not rely on building a cache.
2) Put the TV series virtual subfolders within another virtual subfolder called something like "Television", and then put the rest of the files in other virtual subfolders like "Movies" and a catchall for ones that aren't matched as a series or movie like "Unsorted" or something.

Basically my goal is to offer our users 2 main options:
1) View your files and folders in the same structure as you would on your hard drive, or
2) Let us make an intelligent virtual order for your files, so you don't have to manage your folders. This way seems to be the way of the future, with the newer (and most popular) media servers offering this option.

What are your thoughts, @SharkHunter and @valib ?

[Review on Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/77)


SubJunk:
This feature is implemented in 7.0.0

PR #265, opened 2014-03-25
Create OPPOBDP103.conf

DeFlanko:
For @AYColumbia :)


SubJunk:
@DeFlanko can you give logs for this player using the latest UMS? We may be able to identify it more accurately now

PR #327, opened 2014-05-31
Rem srv1

SharkHunter:
Hi There,

This allows you to connect two UMSes. There is an issue. I would like to just dump the stream from the "server" UMS and let the "client" USM do any transcoding. This sadly fails with MP4s since the whole file must be transfered. Any ideas here? A simple solution could be to transcode at the "server" UMS, but you don't know if it is needed and it wastes resources...

[Review on Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/327)


SubJunk:
@SharkHunter this is cool, any interest in resolving the conflicts so we can merge? If so, please reopen

Sami32:
@SharkHunter That's a smart idea :)

PR #340, opened 2014-07-17
4.1.0

skeptical:
Bleeding-edge branch for advanced stuff including upnp as suggested by @SharkHunter

### DIFF
--- a/src/main/java/net/pms/encoders/FFMpegVideo.java
+++ b/src/main/java/net/pms/encoders/FFMpegVideo.java
@@ -488,6 +488,7 @@ public class FFMpegVideo extends Player {
 			if (isNotBlank(mpeg2OptionsRenderer)) {
 				mpeg2Options = mpeg2OptionsRenderer;
 			} else if (mpeg2Options.contains("Automatic")) {
+				boolean isWireless = mpeg2Options.contains("Wireless");
 				mpeg2Options = "-g 5 -q:v 1 -qmin 2 -qmax 3";
 
 				// It has been reported that non-PS3 renderers prefer keyint 5 but prefer it for PS3 because it lowers the average bitrate
@@ -495,7 +496,7 @@ public class FFMpegVideo extends Player {
 					mpeg2Options = "-g 25 -q:v 1 -qmin 2 -qmax 3";
 				}
 
-				if (mpeg2Options.contains("Wireless") || defaultMaxBitrates[0] < 70) {
+				if (isWireless || defaultMaxBitrates[0] < 70) {
 					// Lower quality for 720p+ content
 					if (media.getWidth() > 1280) {
 						mpeg2Options = "-g 25 -qmax 7 -qmin 2";


SubJunk:
@SharkHunter I know this was a VERY long time ago now but did you get anywhere with the HLS investigation? I have been considering it too while I look into making seeking work in the web interface

SharkHunter:
@SubJunk 
Can't say that I ever did. But I did do an video streaming app. ANd it wasn't that hard. The ps4player that is linked above does the trick real nice. You spawn one process running ffmpeg to create/generate the samples and then one that just manages the m3u8 file.

SubJunk:
@SharkHunter cool! What is the video streaming app? I'll have a play around with that ps4player code

A feature I have been playing with is opening up the web interface over the internet so I think using HLS with newer codecs in supporting browsers will be good preparation for that.

SharkHunter:
@SubJunk it was a private app were I wnated to be able to stream something (plus some other overlay info) back home. IIt worked quite nicely. I used som lib on the phone which used ffmpeg in the back and generated the segements (plus an m3u8) and then I just uploaded those to my server. There were some trickery in handling the m3u8 files as the client need to fetch it again and again, I'll look and see what I did.
As for exposing the web interface to the internet I have done that and it works quite ok for the videos I tried.

SharkHunter:
I took a quick look in my code. And the ffmpeg generator produces both a m3u8 + the actual ts segement. This needed to be handled some how to make the correct m3u8 file be seen by the client. If you create a file "index.m3u8" this file needs to be constantly updated so the client can fetch over and over again (or one can keep this in memory of course and not generate the file physically). Thats the only trick. 
HLS can solve seeking but it will have a price since you need to store all those segements on disk to allow a client to fetch them. The m3u8 files are so small in comparisson so thats not an issue

Nadahar:
I've been thinking about this too, and the biggest hurdle as far as I can tell is that I can't find any way to "direct" FFmpeg to encode certain segments on any given time. I assume it will start from the beginning and encode segments sequentially, which means that if a client seeks/skips far ahead, the client will have to wait until the encoding process "gets there". If the transcoding speed is somewhere close to playback speed (1.5x, 2x etc), this will be a long wait and would no doubt time out.

I imagine that starting a playback and then immediately skipping ahead to where one stopped watching or to a part one wishes to see again is common  and such a solution would fail to handle this. So, in a way, all you really gain is the ability to reverse or to forward until you catch up with the transcoding, but you have no way to know where that "limit" is and going beyond it would lead to time out and aborted playback.

Am I missing something?

SharkHunter:
@Nadahar  
1) you can instruct ffmpeg to start at time X, 
2) fwd/rev has never worked very well in UMS/PMS
3) " all you really gain is the ability to reverse or to forward until you catch up " which is better than today? Thus if someone wants to do exactly that then they could with do it if this is implemented which they can't do today. If people wants to skip an hour ahead they cant do that now and this might nit do the trick fror them either so they have not lost anything nor gained anything. Ergo status quo. 
4) "I imagine that starting a playback and then immediately skipping ahead to where one stopped watching" this feature called resume was implemented way back. 
5) HLS only works if you split the whole video into small chunks thats the idea. The client can then download then chunks however it feels best. To make this work one would use ffmpeg to split the file and then obviously keep the .ts files around. If that is done then you only need to do that process once and the next time you can fwd/rev best you feel like it. Will this use disk? Yes.
6) Note what the L in HLS stands for. It is best for live streams. Where fwd is rater pointless.

SubJunk:
Yeah I'm also interested in exploring ways to enable seeking via the web interface, since that is one of my biggest motivations for this. I have a branch where I am playing around with HLS. That might not be the solution for this though. I also have a performant webm (vp9+vorbis) transcoding working, I'm not sure if that is a good idea to use but it seems supported by the browsers I have tested, and it would potentially allow us to use lower bandwidth than H.264 when serving over the internet.

One other good thing about HLS is that it is designed to support different video streams of varying bitrates so we could have adaptive quality on the web interface. So for each video we could pass e.g. 3 streams to support different network speeds. That could be useful for internet streaming. Our web video player (video.js) seems to have good support for HLS too.

For a few days I have also been playing around with the byte seeking that we currently have, since I noticed that seeking works in Firefox but not Chrome. I think that particular problem is because our byte-range serving isn't quite right. One thing I noticed is that we are not sending the content-length header when we send our partial-content (206) responses, and Firefox seems to be forgiving of that but Chrome might not be.

I am new to all of this stuff so I have been on a steep learning curve but it is becoming clearer the more I play around.

Nadahar:
My concern was with the web interface specifically, and the restrictions of the HTML5 player/videojs. The fundamental problem is that they have no "time-seek" concept, only byte-seek. Thus, seeking works fine for local content that doesn't need transcoding, but doesn't work at all when transcoding (since the byte range requests can't be honored).

My goal has been to try to find a way to make seeking work in the web-interface, and I interpreted it as that was @SubJunk's intention too. As such, I think 3) is a big problem, although the ability to reverse is an "improvement", the same could be achieved by transcoding to file and then streaming from the (incomplete) file without using HLS. To me, seeking is still not "supported" in this scenario.

Regarding 4) I'm aware of this, but it has many issues and many users disable it as a result. Also, this was just an example, it might just be that you want to seek to "your favorite part" etc.

6) A lot of non-live content use HLS (or the MPEG equivalent). Think of all the streaming services like Netflix, iPlayer etc. I works perfectly fine for non-live content as well, and has several "advantages". The problem is only that for non-live content the chunks (in different qualities etc) are pre-encoded so they are all available on request.

I'm not so worried about the disk use, there would need to be cleanup handling of course, the real issue as I see it is that it's impossible to direct FFmpeg to "focus" on certain chunks. I guess it wouldn't be possible either, as FFmpeg probably doesn't know where the chunks starts and ends until the moment it encodes that part. HLS assumes that the chunks fits seamlessly together so that the client can cache and "glue" the chunks together, and I don't see any way to "redirect" FFmpeg without breaking this.

SubJunk:
One idea I had, particularly if we use VP9 for better performance at lower bitrates, is that we could use fixed bitrates on the web interface instead of crf, then we can manually parse the byte request into a time request on the server by doing the math, and then I think FFmpeg supports using `-ss` for HLS like any other format.

Another option is that there may be a video.js plugin that supports sending a custom request while seeking so we could make it just send the desired time to us.

SharkHunter:
@SubJunk - how HLS works is that you essentially splits your video into a whole bunch of a small 5-10 seconds segments per resolution you want. It is not any automatic transcoding or rescaling it is multiple versions of the same file. How it/was used by professional services last time I checked was that you at the top level had one index.m3u8 file with multiple bandwith/resolution options and then the client picks one wich then leads to a new index_640x320.m3u8 which contains the segment. So you will have something like
moviexxx:
640:
   segment
   segment
720:
   segment
   segment
etc.
@Nadahar 
You can instruct ffmpeg to start at time X and stop at time Y. So you can focus on how small chunks you want. 
"HLS assumes that the chunks fits seamlessly together so that the client can cache and "glue" the chunks together, and I don't see any way to "redirect" FFmpeg without breaking this" ??? ffmpeg will split the file into a bunch of small segment files. These are normallay of fixed size which is an option to ffmpeg. FFmepg is not streaming anything it is in this scenario just spitting out a bunch of small files what you're streaming is the m3u8 file and then the client downloads the segements just as any regular http file download.

Nadahar:
@SharkHunter I guess we're talking about slightly different things. My thought was to use the same M3M8 file and populate it with segments that FFmpeg generated for the entire playback. That would require the chunks to "fit perfectly", making skipping forward a problem as we discussed earlier. Starting a new transcoding session at a different time would "invalidate" all the old chunks, in fact FFmpeg would generate an entirely new M3M8 file. Clients buffer the segments, it seems to me that videojs/the browsers (I'm not sure at what level this is controlled) caches quite a lot, and won't even request a segment that's already in the buffer if it is needed. 

Let's say that we have 3 segments, A, B and C. Playback starts and segment A is transcoded. The user then skips to content that would be in segment C. We would then have to abort the current FFmpeg session, leaving potentially A only partly complete and B not started, and then instruct FFmpeg to start generating a new segment that is somewhere within segment C.  If the user then seeks back to either A, the player would reuse the cached A segment. When it's done playing this, it would request the non-existing B segment - or maybe it wouldn't because it no longer exists in the M3M8 file. Regardless, this wouldn't work. There would be no way to create the B segment so that if fits perfectly with the already buffered A and "C2" segments.

If you don't understand what I mean, never mind. I think I've kind of answered my own question, I think my "idea" would be impossible under any circumstances, unless everything (input and output, all codecs) was CBR and somebody had implemented a "special solution" for this scenario.

SubJunk:
Maybe that idea could work if the interval that is being requested is always the same as the HLS interval. So if each TS file is 5 seconds long, then maybe the player could be rigged to allow only seeking per 5 seconds, or the server can respond as if that were the case

SubJunk:
@SharkHunter yes that's what I'm seeing too, it's cool

SharkHunter:
@SubJunk 
ffmpeg has switches for the split up + the generation of index files. From the ps4player
"-segment_time 5 -segment_list /tmp/ps4player/" + hash + "/playlist.m3u8 -segment_format"
This would produce a bunch of files and would be vanilla HLS. It would work fast and would be easy to implement.
A more exotic variant would be to generate each segment on the fly and never create the file. Just spawn a new ffmpeg with -ss and maybe -t for each segment and just pass it along.The problem here will be the index file since it must contain all indexes from the start. This could be solved if you know how long the file/movie is than you can calculate how many segments you would have. This will not work for web based dynamic content, the first approach will.

PR #445, opened 2014-12-28
5.0.0

skeptical:
Just regenerating the 5.0.0 PR to prune old comments and make the page load faster, with mixed results apparently. See #362 for previous discussion.


valib:
@SubJunk I have no idea how to fix it. Only to reverted them fix it.

PR #452, opened 2015-01-09
ignore obvious bad lowranges (ie >100GB)

wolcano:
this fixes timeout for PhilipsTV (fw QN141E_012.003.026.128) when using dlna transcoding.

### DIFF
--- a/src/main/java/net/pms/network/RequestV2.java
+++ b/src/main/java/net/pms/network/RequestV2.java
@@ -114,7 +114,13 @@ public class RequestV2 extends HTTPResource {
 	 * @param lowRange The byte to start from.
 	 */
 	public void setLowRange(long lowRange) {
-		this.lowRange = lowRange;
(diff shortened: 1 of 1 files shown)


SubJunk:
Thanks for this :)

wolcano:
well, this is only a hotfix for PhilipsTVs. I have dug around and it seems, that TV tries to purposefully (bug in TV?) read far in stream maybe to obtain index part of the movie. I suspect all this behavior is caused by UMS lying about transcoded stream size (looks like UMS sends in Content-range upper range Long.MAX_VALUE).

This behavior is also present with not-transoded streams, and works well with UMS/TV without any hiccups.

SubJunk:
@wolcano ah right, yeah we lie about it since we don't know how big the transcoded stream will be before it's done. Most renderers don't need that info but Philips Aurea does, so we created the TranscodedVideoFileSize setting which is set to 100GB

PR #453, opened 2015-01-20
A few small improvements

mfulgo:
These are the parts of my local changes I thought worth keeping around.  See commit message bodies for more details.


SubJunk:
@mfulgo thanks for this!

PR #461, opened 2015-01-23
New plugin type FileProvider

taconaut:
I've been maintaining a separate build for the media library extension (mlx) for the last years for pms and have plugged it on top of ums lately.
To avoid the hassle of merging new code all the time I got the idea of creating a new plugin type, which would allow to choose the way files and folders are being presented on the renderer. By default, the current one, where the files and folders would represent the file system would be available. If the user wants to change, a plugin could be dropped into the plugins folder and selected in the 'Navigation/Share settings' tab.
Before starting I wanted to have your opinion to see if you'd be willing to merge it, if I get everything working. Architecture-wise I think it would be nice to decouple how files and folders are being presented from how files are being streamed.

For now, only the interface for FileProvider exists and has been chaecked in.

How it would work in the GUI:
In the 'Navigation/Share settings' tab, an additional combo-box would be displayed if there is more than one FileProvider available, allowing to choose which one to use. Underneath, a panel to configure the properties would be shown.
The plugins for AdditionalFoldersAtRoot won't be part of the global plugins anymore, as they are dependent on the current implementation. They would still be available though.

For the end-user nothing would change after a fresh install. AdditionalFoldersAtRoot would be shown somewhere else, I haven't figured out where to put them for now though.

I hope you like the idea :)

Tasks:
- [ ] Create FileSystemFileProvider in ums code. It will contain the current GUI from 'Navigation/Share settings' tab as well as implementation specific parts of DLNAResource (e.g. resume option). Use RootFolder code to create final DLNAResource
- [ ] Remove RootFolder from ums and do required code changes
- [ ] Move AdditionalFolder(s)AtRoot interfaces to FileSystemFileProvider 
- [ ] Show instanciated AdditionalFolder(s)AtRoot in FileSystemFileProvider GUI
- [ ] Add loading mechanism for FileProvider plugins
- [ ] Add GUI to select FileProvider plugins
- [ ] Persist selected FileProvider plugin
- [ ] Add code to properly activate/deactivate FileProvider plugins upon change

### DIFF
--- a/src/main/java/net/pms/external/FileProvider.java
+++ b/src/main/java/net/pms/external/FileProvider.java
@@ -1,0 +1,54 @@
+package net.pms.external;
+
+import javax.swing.JPanel;
+
+import net.pms.dlna.DLNAResource;
+
+/**
+ * All classes implementing the FileProvider Interface will be loaded when UMS
+ * is being started and can be selected as file providers in the
+ * 'Navigation/Share Settings' tab.
+ */
+public interface FileProvider {
+
+	/**
+	 * Gets the configuration panel which will be displayed in the
+	 * 'Navigation/Share Settings' tab if the file provider is the active one.
+	 *
+	 * @return the configuration panel
+	 */
+	public JPanel getConfigurationPanel();
+
+	/**
+	 * Gets the root folder.</br></br> All children of the root folder (files
+	 * and folders) will be shown on the first level when browsing on the
+	 * renderer.
+	 *
+	 * @return the root folder
+	 */
+	public DLNAResource getRootFolder();
+
+	/**
+	 * Gets the name which will be displayed in the 'Navigation/Share Settings'
+	 * tab.</br></br> This method has to be functional before
+	 * {@link #activate()} is being called.
+	 * 
+	 * @return the name of the file provider
+	 */
+	public String getName();
+
+	/**
+	 * Activates the file provider.</br></br> This method is being called when
+	 * UMS has been completely initialized and the file provider is the active
+	 * one.</br> If the plugin has to access UMS functionality, do it here and
+	 * not in the constructor.
+	 */
+	public void activate();
+
+	/**
+	 * Deactivates the file provider.</br></br> This method is being called when
+	 * UMS shuts down or when the file provider is currently the active one and
+	 * another one is being activated,
+	 */
+	public void deactivate();
+}


SubJunk:
@taconaut great to see you :) It's a promising idea and I'm for it, depending on what the other developers say.

taconaut:
Cool :) Waiting for some more feedback and the 5.0 branch to be merged into master and then I'll see what can be done

skeptical:
@taconaut how would this impact other plugins (e.g. jumpy, channels) that provide mostly non-realfile content and are already internally manipulating/hacking the existing rootfolder api a bit to achieve this?

taconaut:
@skeptical good question. I didn't think plugins would directly access the root folder. I'd welcome input from the plugin developers to see if the adaptations could be done in the plugins or if the idea should be abandoned here.
Doing the proposed changes would definitely be a breaking change.

skeptical:
@taconaut Is it necessarily all or nothing? Can FileProvider become a late "PathConverter" filter instead, so the rootfolder api continues to be the underlying "file system"?

taconaut:
@skeptical I'll have to give this a closer look. A PathConverter wouldn't be enough for the media library though. If you want to have a look at its functionalities, a snapshot build can be downloaded here http://universalmediaserver-ci.hopto.org:9999/job/ums-mlx%20trunk/ws/packaging/target/

SharkHunter:
I'm not sure I understand what the problem is here. What is you're trying to achieve that can't be done by the current system? Or is it that you want to modify what is shown on the root folder? I suspect (after some more reading) that is it the later. So lets see if I got it right. You want to modify what is shown on the rootfolder. To do so you must make a change that would render all plugins unusable? That sounds like it is the wrong path to take. If you want to modify what is shown on rootfolder (which is what the FileProvider interface you supplied only allows you to do) I would add a RootFolderFilter instead which gets the RootFolder and modifies it according to whatever rules you feel like. And the best part is that if you don't have a RootFolderFilter nothing has changed. Pseudo code:

RootFolderFilter {
    public void filter(RootFolder f) {
       // do stuff here
       // most likely something like
       /*for(DLNAResource r : f.getChildren())
                // do stuff with r
    }
}

And in RootFolder

at the end of discoverChildren:

for(RootFolderFilter rf : getRootFilters())
  rf.filter(this)

BTW. We already have a bunch of root folder filters implemented in UMS where you can modify the rootfolder based on render name,IP etc.

taconaut:
@skeptical Actually, the RootFolder would be a good spot to implement this and guarantee backwards compatibility.

discoverChildren() would forward the call to the FileProvider DLNAResource.
The private methods would be moved to the FileProviderFileProvider (based on the current implementation) and used to return the same structure as RootFolder currently does.

I got some questions about other functionalities which are specific to the current implementation:
Can the scan functions be moved outside RootFolder?
Is the folder limit functionality generic for all DLNAResources?
How do tags work?

SharkHunter:
@taconaut  - still I vote for a filter that is called after. This way nothing needs to be moved at all!

taconaut:
@SharkHunter 

> What is you're trying to achieve that can't be done by the current system?

Offer a different way to configure the files and folders being shown on the renderer, which has nothing to do with file system (as it currently works).

> To do so you must make a change that would render all plugins unusable?

No. 

> If you want to modify what is shown on rootfolder (which is what the FileProvider interface you supplied only allows you to do) I would add a RootFolderFilter instead which gets the RootFolder and modifies it according to whatever rules you feel like.

How can I provide a GUI to configure it?
As all properties being set in the Navigation/Share settings tab wouldn't be used, I think it'll be confusing for the users to let them there. That's why I proposed to replace them by a UI which is actually applied.

SharkHunter:
@taconaut - Add a gui option to the RootFolderFilter if you feel an urge. But still I thinks a filter functions is the best way to do this. You're handed the rootfolder and you modify it any way you feel like.Simple and clean, and this is actually more flexible as wee since you can make even more complicated filters.

taconaut:
I'd rather keep my own build to have a more consistent UI then.
Cheers

skeptical:
@taconaut MLX is very slick, glad I finally looked into it :). The thing is that at this point UMS is targeting 2 basic types of users
- traditional, i.e. people who maintain a media collection, for whom MLX would be a big plus obviously.
- ad-hoc, i.e. people who just want a fast easy way to put random stuff from wherever (web, hdd) on their tvs, who are better suited to a passive organization model. A lot of their media has no accessible mediainfo anyway.

Extending rootfolder as you suggest, e.g. LibraryFolder, and letting both types coexist is doable, and I'm sure any compatibility wrinkles can all be worked out to everyone's satisfaction. Fwiw I don't think the UI(s) would be necessarily inconsistent if framed as dealing with different areas of the xmb (e.g. a top-level 'Library' folder or something).

SharkHunter:
@taconaut , @skeptical - I still don't understand the need to modfiy RootFolder completly. The FileProvider only has a getRootFolder function. This can be replaced by a filter function and nothing would change except for those who uses these types of plugins. And the GUI has no connection to RootFolder at all. RootFolder (as the rest of UMS) is controlled via UMS.conf which is written from the GUI (or from a text editor). So you can do whatever you want in the GUI as long as it writes it's changes to UMS.conf. If not then your conf won't survive a restart.

skeptical:
@SharkHunter I guess I'm suggesting letting RootFolder continue to exist as-is but extending it for "library" folders only (which internally discoverChildren() in cool new ways as I understood @taconaut to mean). No different than how a plugin might do it, come to think of it. 

Which raises the (probably naive) question of why MLX can't be encapsulated as a "library" plugin (which I'm sure you have good answers to @taconaut :-). I suppose the bigger point here is that every plugin is effectively a "FileProvider" already with its own presumably valid and sensible organizing rules/logic.

SharkHunter:
@skeptical - yeah thats what I'm proposing too.  MLX could be wrapped in/integrated with a minimal effort in the resource part (read small modification to RootFolder). What ever GUI changes you need is outside of that scope.

taconaut:
@SharkHunter , @skeptical  Sorry for the late reply, I was a bit busy the last few days.

> MLX is very slick, glad I finally looked into it :)

Thanks :)

> The thing is that at this point UMS is targeting 2 basic types of user

Sure, that's why I propose to make a plugin for it. This way nothing changes for the user after a fresh install. For those who want, other ways to display the files would be available.

> I still don't understand the need to modfiy RootFolder completly. The FileProvider only has a getRootFolder function. This can be replaced by a filter function and nothing would change except for those who uses these types of plugins. And the GUI has no connection to RootFolder at all. RootFolder (as the rest of UMS) is controlled via UMS.conf which is written from the GUI (or from a text editor). So you can do whatever you want in the GUI as long as it writes it's changes to UMS.conf. If not then your conf won't survive a restart.

I'd rather entierly replace the rootfolder than to extend and filter the existing one for the following resons:

* It avoids loading unused resources. If e.g. iTunes library is enabled 'Navigation/Share Settings' tab, initializing UMS can take a lot longer than without (for large iTunes libraries), it also uses quite a lot of RAM and wont be used anyway. 
* There are features like 'Hide file extensions' which interfere with the displayed names. If this feature is enabled all names which contain a dot will be truncated. The cache and video resuming are other features which are bound to the current implementation.
* Keeping the current 'Navigation/Share Settings' options will be confusing for the user, as none of the set properties will be used by mlx. The additional folders like e.g. iTunes or iPhoto are available as plugins in mlx.

When I proposed to move the AdditionalFolder(s)AtRoot, it was to be coherent with the functionalities and only show them where they will be used. They can of course stay where they are and not be used by other FileProviders.

skeptical:
@taconaut sorry, I've been meaning to respond but see 29 days have gone by :).  My main issue is that you're proposing a breaking change and therefore forcing every other plugin to be rewritten even though there won't be any tangible benefits for them.  As you know better than me, for better or worse here (and in later PMS) the approach has become pragmatic: avoid making too many waves code-wise. This doesn't make for great code since it usually means tacking on new things on top of old stuff in hacky less-than-ideal ways, but at least the boat still floats. If we were to refactor everything that needs it (which is pretty much everything), we would be better off starting from scratch.

I don't really see why MLX can't be a regular plugin managing a 'Library' folder in the xmb root.  Personally I don't have any problem with a plugin that removes the navigation/share settings tab (or manipulates the gui in any reasonable way) since this is would ultimately be by user choice, and would be in favor of adding that functionality to ums.

taconaut:
No worries ;) You're right, what I initially proposed would have contained quite big breaking changes, this could all be done without breaking anything, but imo it would get pretty confusing for the user.

I'll continue going with my own build, this way I've got the freedom of changing what I want.

Cheers :)

PR #469, opened 2015-01-26
Removes extra close on InputStream

mfulgo:
The ChunkedWriteHandler already automatically closes the InputStream
that we send to it when the Channel goes inactive (i.e. the client
disconnects) and when the InputStream is at its end.

If we also call .close(), it causes a warning to be logged by the
ChunkedWriteHandler when it tries to close the stream because an
IOException is thrown.


SubJunk:
Thanks!

PR #470, opened 2015-01-26
FFmpeg internal subtitle handling

SubJunk:
Resurrecting the work done by @valib

Some renderers… don't wait long enough for the parsing to be done so that needs to be fixed before a merge


SubJunk:
@UniversalMediaServer/developers the advantage of getting this working is that it would mean FFmpeg can use embedded fonts, meaning we don't have to defer to MEncoder for them anymore.

valib:
@SubJunk it works but I am affraid that we can't do anythink with the FFmpeg delay especially for the time seeking. In that case we stop the current instance and start the new one which always does subtitles parsing. If FFmpeg could be a dll not the exe it could be possible to manage it dynamically and not start a new instance of it like we do now.

SubJunk:
@valib darn, so this won't really work unless FFmpeg implements an option to use internal subtitles without parsing them at the beginning. It would be cool if there were a flag to tell it to parse subtitles as it goes instead.

valib:
@SubJunk we can use that embedded subtitles implementation because FFmpeg works exactly the same way like for external subs. There is not a difference. FFmpeg developers take that feature as an enhancement but it lasts more than 2 years when it was reported. I think that use FFmpeg as an library could solve everything. I saw somewhere that it was produced before but I don't know if it is available now.

valib:
@SubJunk  I have checked the FFmpeg source code and there is writen `/* sub2video hack:
   Convert subtitles to video with alpha to insert them in filter graphs.
   This is a temporary solution until libavfilter gets real subtitles support.
 */` so we should wait until they improve it.

SubJunk:
I made an issue on the FFmpeg trac for this https://trac.ffmpeg.org/ticket/4499#ticket

ExSport:
Great! Thx @SubJunk

Nadahar:
@valib @ExSport @SubJunk Does closing this mean that we give up, or that it's fixed? Have you tried the workaround given in [FFmpeg ticket 2067] (https://trac.ffmpeg.org/ticket/2067)?
>  Ricky1252's workaround gave me some quality issues for some reason.
This is another workaround for this issue.

> ```ffmpeg -ss 102.042 -t 24.250 -i sintel-1024-surround.mp4 -c:v libvpx -vf scale=640:-1,setpts=PTS+102.042/TB,subtitles=sintel_en.srt,setpts=PTS-STARTPTS out.webm```

> Works nice for me, even in 2pass mode.

PR #484, opened 2015-02-12
fix Italian translation

bartsimp:


SubJunk:
@bartsimp thanks for the contribution. What effect does the tick marks part have?

bartsimp:
for machines with lot of memory (I've 3625 MB (3,6Gb)), a tick every 100mb result in too many tick. My propose is to scale the tick based on real memory

SubJunk:
Thanks for this :)

skeptical:
@bartsimp thanks, it's a valid issue re tickmarks, but solution is way overkill computationally speaking, plus I don't think this is desireable, for example 1028 mb memory:
1028mb

SubJunk:
@skeptical feel free to revert that part of it, or change it

skeptical:
Already changed in f87f744 when I posted the comment above, should've mentioned it.

SubJunk:
Nice one

bartsimp:
happy to contribute :)
@skeptical good idea

PR #500, opened 2015-03-07
Update messages_fr.properties

Kirvx:


SubJunk:
Thanks a lot, @Kirvx !

PR #508, opened 2015-03-20
Hyperlink tool tip

taconaut:
I wanted to add hyperlinks to tooltips to make it easier for users to find some info and thought this could be added to UMS itself.

This pull request includes:
* Allow to configure tooltip foreground and background colors in the configuration (Keys are tooltip_foreground and tooltip_background). This has been added to have a common configuration for default and hyperlink tooltips
* Added a new method getRGBColor to RendererConfiguration
* Added a new class HyperLinkToolTip
* Added an example usage in 'Subtitles language priority'

A HyperLinkToolTip can be used like this:

    JButton btn = new JButton() {
    			private static final long serialVersionUID = -2927951764552780686L;
    
    			public JToolTip createToolTip() {
    				JToolTip tip = new HyperLinkToolTip();
    				tip.setComponent(this);
    				return tip;
    			}
    
    			// Set tooltip location
    			public Point getToolTipLocation(MouseEvent event) {
    				return new Point(getWidth() / 2, getHeight() / 2);
    			}
    		};
    
    		btn.setText("Tooltip Test");
    		btn.setToolTipText("<html><div width=\"300\">Specifies the languages to search for subtitles. E.g. 'eng', 'eng,fra,esp'.<br><br>See <a href=\"http://en.wikipedia.org/wiki/List_of_ISO_639-2_codes\">this article</a> for a full list of languages.</div></html>");


valib:
@taconaut nice idea. I have one improvement to this. what about to create JTextField2.class with 

    package net.pms.newgui;

    import java.awt.Point;
    import java.awt.event.MouseEvent;
    import javax.swing.JTextField;
    import javax.swing.JToolTip;

    public class JTextField2 extends JTextField {
	private static final long serialVersionUID = -5697643578934331831L;

	public JTextField2(String text) {
        super(text);
    }

	public JToolTip createToolTip() {
		JToolTip tip = new HyperLinkToolTip();
		tip.setComponent(this);
		return tip;
	}

	public Point getToolTipLocation(MouseEvent event) {
		return new Point(getWidth() / 2, getHeight() / 2);
	}
}

skeptical:
@taconaut :+1: very nice, now we just need to write all the tooltips (he said optimistically). My only comment is that it could be rolled into ```GuiUtil``` with all the other generic swing widgets instead of being standalone.

PS Extra points for allowing background color configuration, ideally we would have an overall skinning solution for ums :).

taconaut:
@valib Good idea, this way, only the types of the UI components has to be changed. It should be done for JTextFields, JLabels and JButtons. Others? Into which package should they go and how should they be called? (Adding a 2 postfix doesn't really tell a lot about their additional functionality, but I can't think of a better name..)

@skeptical No need to change all the tooltips right away ;) but new ones can easily use the hyperlink functionality. I wasn't aware of `GuiUtil`, looks like a good place to put it.

Additionally, getRGBColor returns a java.awt.Color. Could this be an issue when running headless? I haven't tested it.

valib:
@taconaut now we use CustomButton in the GuiUtil so names could be the CustomLabel and CustomTextField.

valib:
@SubJunk @taconaut I merged this commit and updated it. It seems to me that it is better to use single instances of the components than put them to the GuiUtil because there are problems to instantiate the inner clases.

taconaut:
Cool, thanks!

PR #511, opened 2015-03-25
added onlytray checkbox

bartsimp:
if the checkbox is checked the minimize action use only the traybar (if supported)


SubJunk:
@bartsimp thanks for the PR! I'm not sure about this, since it doesn't add any extra functionality - you can minimize to tray by clicking x already, so this just makes two buttons do the same thing... I guess we can add the option to the config but not the GUI, maybe.
What do @UniversalMediaServer/developers think?

Issue #503, opened 2015-03-10
[Suggestion] Disable defer to tsMuxeR option by default


SubJunk:
I'm unsure about this one so I want your opinions @UniversalMediaServer/developers 

Here is a pros and cons list for using the defer to tsMuxeR option:

Pros:
Less CPU-usage - this is the only big pro but it's a big one.
Likely to use less bandwidth - renderers that transcode to H.264 will use similar bandwidth.

Cons:
Audio sync is often off on PS3, especially after seeking
Seems to cause mysterious bugs like streams quitting half way (the forum is littered with it in logs)
Files don't play to the end (a few seconds is cut off)

I think for overall stability we should disable the option by default but what do you guys think?

WolfganP:
You probably know by now that I'm a big proponent of avoiding video transcode at all costs (quality, resources utilization), so re-muxing is a must for me. 
If tsmuxer introduces side effects, can we force the remuxing features by using ffmpeg or mencoder instead? (ie enforce -v copy in global or renderer specific settings)

SubJunk:
@WolfganP this isn't about removing the feature, it's about disabling it by default, so whatever we do it won't affect your current functionality :)
I tried having FFmpeg remux things before (released in version 2.6.1) but it was problematic. I guess I can try it again with a newer FFmpeg version.

WolfganP:
@SubJunk yes, understood. I just wanted to cheer in and add my 2cents in making UMS a mean and lean remuxing video machine :-)
That's why I always look for ways to forcing it w/o having to resort to TRANSCODE folder or the like; and suggested adding "-v copy" option to the engines to enforce it (at main setting? as a transcoding quality profile? at renderer level?)
Count me in for the remuxing tests!

SubJunk:
@WolfganP I've made a quick build that muxes via FFmpeg whenever it would normally mux via tsMuxeR, please test it http://www.spirton.com/uploads/UMS/UMS-199876e.exe
I tested it on my TV and it worked for a file but it will need testing on many renderers and with many files before I will consider making it a permanent change - we have created enough bugs lately so we don't need more :) Thanks for the help

ExSport:
Nice @SubJunk !
Will test it in few weeks on real renderers or days on Android Mobile or
with some local Windows sw renderer...
Dne 18. 3. 2015 0:51 "SubJunk" <notifications@github.com> napsal(a):

SubJunk:
[...]

PR #517, opened 2015-04-04
FFmpeg remux

SubJunk:
Mux to MPEG-TS via FFmpeg instead of tsMuxeR.
This code shouldn't be permanent since it's hackish, it is just for testing. It can easily be written in a better way if we conclude that it can be done.

So far it is working for some but not all files.


WolfganP:
Thx @SubJunk . I tested the version you posted on issue #503 and noticed no major errors on common media files. Which files did you notice it didn't work?
Would it make sense to promote the code as a config option for broader testing? (ie [ ] force remux with ffmpeg, besides the [ ] defer remux with txmuxer)

SubJunk:
@WolfganP one file that doesn't work with one of my devices using this method is Banshee.S03E08.720p.HDTV.DD5.1.x264-DON. I'm looking into it

ExSport:
@SubJunk any news about stability? I will try this code with some of my files as I get home hopefully this weekend so post tips what should I test specially for you :+1: 
Anyway we have here options like remux with tsmuxer, mux when possible for MEncoder and FFmpeg engines, so one additional "beta option" doesn't kill anyone :+1: 
I vote for implementing it to the master, configurable via ums.conf or in GUI so anybody can test all engines which best work for them, easily changeable on-the-fly. What do you think?

SubJunk:
@ExSport tsMuxeR has some cool abiliities that FFmpeg doesn't have, so I'm starting to think that our best option will be to continue to use tsMuxeR but via FFmpeg instead of via MEncoder. In my testing, that option is more stable than this one, because tsMuxeR is so good at it.

ExSport:
@SubJunk What cool abilities exactly do you mean?
I thought FFmpeg is more robust and bug free when comparing to tsMuxeR but maybe it is only my opinion without any real evidence. I only remember communication with Roman about incompatible tsMuxeR output when used file created by FFmpeg so great you found FFmpeg is now more compatible than before :+1:
Thx

SubJunk:
@ExSport tsMuxeR gives a lot of control over aspects of the output that FFmpeg doesn't give, with settings like "no-pcr-on-video-pid", "new-audio-pes", "no-asyncio", and other stream-specific ones that control/rewrite the timestamp information to make the stream more compatible.
I think FFmpeg has the timestamp options but not the PCR and PES ones.

The file I listed above (Banshee.S03E08.720p.HDTV.DD5.1.x264-DON) is a good example of it - it works with FFmpeg via tsMuxeR, but not with just FFmpeg.

Also I should note that all options are as stable as each other on my Panasonic TV, and probably other renderers. PS3 just seems very picky about how it likes MPEG-TS.

SubJunk:
I'm closing this for now because it is more stable to use FFmpeg>tsMuxeR than FFmpeg by itself. It can be reopened if FFmpeg does some work on it.

PR #519, opened 2015-04-07
Make the tsMuxeR video engine use FFmpeg instead of MEncoder

SubJunk:
In my tests this has improved tsMuxeR compatibility on PS3.
I tried this years ago and it negatively affected audio sync, but FFmpeg have since improved their audio sync so that may be enough to make this work.


SubJunk:
This is possibly ready to merge if it gets enough testing. It seems that FFmpeg passes along streams that tsMuxeR likes better than MEncoder does, so this could really improve our muxing stability.

SubJunk:
@UniversalMediaServer/developers can you guys please test this branch with your devices and let me know if it's stable? I have tested it on mine and it improved compatibility, but I want to be really sure about other devices before merging

PR #521, opened 2015-04-11
Subtitle handling fixes and simplification

SubJunk:
The main reason for this is to prevent the needless muxing of files that happens all the time.
The most common example of this error is if there's a file that is compatible with the renderer, but it has subtitles that are incompatible with the renderer, our current code marks it as needing transcoding. So far so good, except that usually when files have subtitles they're not wanted, so we end up muxing the file without subtitles in the end, which is pointless and complicated.
This PR fixes the above case by making sure we check language preferences before deciding which player we will use.

This will need a lot of testing. We will need to confirm that it handles files with external and embedded subtitles, when they are compatible and incompatible with the renderer, for renderers that use MediaInfo and those that don't.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -8,9 +8,15 @@
 		Added the ability to restart the program (not just the server)
 		Various minor fixes/improvements
 	Renderers:
+		Added DSD/DFF streaming support to Cambridge Audio Blu-ray Disc players
+		Added 3gp and 3g2 tags to DefaultRenderer.conf
 		Improved renderer detection/handling
 		Improved Android device detection
+		Improved detection of Panasonic AS600 Series TVs
 		Improved PS3 muxing via tsMuxeR
+		Improved support for Samsung D6400 TVs
+		Improved support for Samsung EH5300 TVs
+		Fixed support for MP3 files on some Samsung TVs and Blu-ray Disc players
 	External Components:
 		Updated JDom to 2.0.6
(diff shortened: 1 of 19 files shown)


SubJunk:
At this stage the code is probably incomplete but this is a good start and has been stable in my tests so far.

SubJunk:
@valib it would be good to get your input on this since you've written a lot of our subtitle handling code. I'm hoping to merge this pretty quickly

SubJunk:
I've posted a test build to http://www.universalmediaserver.com/forum/viewtopic.php?f=9&t=3278 to hopefully get some testing on a renderer that supports internal and external subtitles, since I don't have one

valib:
@SubJunk it works for me for external and also for embedded subtitles. This idea is good and could be merged.

SubJunk:
@valib great, thanks for testing

SubJunk:
@valib did you test on a renderer that supported the subtitles, or both? I want to make sure we have covered all use cases

valib:
@SubJunk  I fixed a bug but what is more interesting that the child.media.mimeType in the DLNAResource.addChild() is always null. There should be a bug to recognize the mime type from the media.

SubJunk:
@valib is that just with this branch or in master too?

valib:
@SubJunk I fixed the NPE also in master so you can merge this PR and it will be OK.

PR #522, opened 2015-04-12
Convert all renderers to use MediaInfo = true

SubJunk:
This would allow for better accuracy when analyzing files

[Review on Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/522)


Nadahar:
@SubJunk I restored the branch. Even though the PR is outdated and would need work now, I'm sure that many of the renderer configurations in here is better than the ones in master..? Shouldn't they at least be compared to current master and the obvious improvements be merged?

SubJunk:
I posted builds of it in some forum threads while I was working on it and it just seemed to cause more problems so I'm not confident of it. Maybe there are some good changes in it amongst the problems, I'm not sure.

I think my alternate approach in the future will be to work on these files one at a time with a user testing them, instead of doing all of them at once.

Nadahar:
Ok, but don't you agree that the branch can stay? It would be easy to use it for getting "templates" to use as a starting point for converting a renderer configuration - or isn't it as useful as I think?

SubJunk:
Yeah that sounds like a good idea

PR #526, opened 2015-04-15
Document chromecast_extension option in UMS.conf

ExSport:
Adding undocumented option "chromecast_extension"
Useful when troubleshooting possible UMS problems.


SubJunk:
Thanks!

PR #529, opened 2015-04-18
Second attempt to use port 80 for WEB interface

valib:


SubJunk:
@valib @skeptical I agree with where the discussion ended up. It's a shame we can't use port 80 by default because it would have the stated benefits. Oh well.

@valib Feel free to add a log message to inform users they can try using port 80

PR #544, opened 2015-04-28
Improve codepage detection

valib:


valib:
It seems to me that the Apache Tika is more accurate than UniversalDetector which also doesn't have any development. It need more testing especially for asian languages but for european languages the Apache Tika is better and should substitute the current implementation.

valib:
@SubJunk what about this PR? I think it works. This has only one disadvantege that the final souce code is rised to almost 100 MB.

SubJunk:
@valib if it works then that's okay I think, what do @UniversalMediaServer/developers think?

SubJunk:
I wonder if there's a way we can reduce that size

valib:
@SubJunk I think the only way how to redude the size is to take the original code and rewrite it to use only what we need.

src/main/java/net/pms/util/FileUtil.java
```diff
@@ -668,37 +669,55 @@ public boolean accept(File dir, String name) {
 		return found;
 	}
 
+	private static String externalSubsLang;
```
Nadahar: @valib I don't understand why you made this variable. It seems to me to simply return the same result as the last call to getFileCharset. You may have made this for efficiency, but you can't do it like this in a multithreaded application, especially since the methods are static. To avoid making the test multiple times, you must keep the result in a variable in the calling method instead.

Another thread could have changed the value of ```externalSubsLang``` since the last time ```getFileCharset``` was called from the thread in question. It's not treadsafe either, so you risk getting garbage back.

I'm working on #689 and even though it's not related to that PR; I'm going to remove it there as it has other changes to ```FileUtil```.

valib:
@Nadahar I don't think that that the return of the CharsetMatch is a good idea. The main thing is the getConfidence() which decides which CharsetMatch [] result is the best match of the charset. I don't know if other thread can change that.

src/main/java/net/pms/util/FileUtil.java
Nadahar: @valib df475b3b39403c36bc80ca5096c98f26df2c0b66 is my proposed fix.

PR #552, opened 2015-05-05
OSX Java 6/7/8 Build Update

Optimusprimeums:
Updates pom.xml with a Java 6 and Java7/8 profiles, Adds appbundler (custom for UMS) to build/launch Java 7/8 UMS and adds a assembly.xml file for Java 7/8.
Tested (extensively) and working correct


SubJunk:
@Optimusprimeums looking good mate! But can we separate 7 and 8 too? With this as it is now, the Java 8 build compiles in Java 7 mode...?

Optimusprimeums:
sure @SubJunk , just means a bigger pom.xml if your happy i'll close this one and re-pull

SubJunk:
@Optimusprimeums what do you mean manually change? Can't you set the JAVA_HOME variable to reference the JDK you need? For example:
```SET JAVA_HOME=%CD%\JDK-1.7.67

Optimusprimeums:
@SubJunk I'll try that but i usually just change it to the JDK in the build settings then build the branch, 6/7/8, that i've checked out. Maybe i've been doing it the hard way all this time

SubJunk:
This is my build folder:

build folder

and my process is to change the "Code" folder to the tag I want, then double-click the appropriate .bat file.

It could be much more streamlined if I combined the 6 build files into one called "Release.bat" that included git commands to switch branches. I should probably do that, but anyway this process already only takes 5-10 minutes.
The time-consuming part is making the tags in the first place because of merge conflicts.

Optimusprimeums:
Well @SubJunk i use netbeans on OSX to do everything. One reason is the the JDK for Java 6 is built in to apple. The others are straight from oracle. I'm using multiple build profiles with Netbeans to control what i'm building. I'm currently looking at the pom.xml now to solve those issues but i have two JDK's to manage which means switching between both Java 6 and Java 8. I build with Java 8 for 7 as it's excellent backwards compatible

SubJunk:
@Optimusprimeums I just made the release.bat file like I described and it's working great now, so building only requires a double-click. I'd be happy to help you set up a similar script, it's definitely possible since OS X has a much better command prompt than Windows

Optimusprimeums:
Sure @SubJunk  it would help, how do you want to get it to me?

Optimusprimeums:
Ok @SubJunk i'm going to submit this pull request (with changes) that need to into the build of UMS. Now i've added a plugin that will require the correct Java version to be used for the correct build. That should prevent anyone using the incorrect JDK. Now i think its complete and ready to merge, if not no one else can build the OSX releases that wants too

SubJunk:
@Optimusprimeums this looks great :)

@UniversalMediaServer/developers does it look ready to merge?

PR #556, opened 2015-05-07
Change check box state only on check box click

taconaut:
Every once in a while, I do a configuration change without wanting to do so. The reason is that many check boxes span multiple columns and clicking on a visually empty spot changes the state of the check box at the beginning of the line.

This pull request wraps all check boxes into a JPanel for them to have their preferred size.

In the below print screens you can see the clickable area for a check box.

Before:
Before the change

After:
After the change

### DIFF
--- a/src/main/java/net/pms/encoders/AviSynthFFmpeg.java
+++ b/src/main/java/net/pms/encoders/AviSynthFFmpeg.java
@@ -45,6 +45,7 @@ import net.pms.dlna.DLNAMediaSubtitle;
 import net.pms.dlna.DLNAResource;
 import net.pms.formats.Format;
 import net.pms.formats.v2.SubtitleType;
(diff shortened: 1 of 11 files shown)


SubJunk:
@taconaut great idea! hopefully we can test and merge

taconaut:
Sure :) It's pretty straight forward. It should be checked that the panel behaves the same way on all platforms, e.g. not adding unwanted padding and margin.

skeptical:
:+1: especially since we autosave settings now.

valib:
@taconaut looks good and when tested it I found the old bug in UMS fixed in 5792f10 :-)
@SubJunk I think It could be merged.

taconaut:
@valib hehe I must admit I haven't tested right to left languages in the GUI

PR #558, opened 2015-05-19
Add aspect ratio distortion tolerance

ExSport:
For avoiding unnecessary transcoding.
TODO: Panasonic TV and maybe others don't require to fix AR for all files as e.g. streamed XviD/DivX files are correctly shown without transcoding.

### DIFF
--- a/src/main/external-resources/renderers/DefaultRenderer.conf
+++ b/src/main/external-resources/renderers/DefaultRenderer.conf
@@ -407,8 +407,18 @@ SubtitleHttpHeader =
 # If some videos on your renderer looked either stretched or squashed, enabling
 # this setting may fix that.
 # This is only reliable when MediaInfo = true.
+# TODO: PanTV doesn't require AR fix for streamed DivX/XviD files!
 # Default: false
 KeepAspectRatio = 
+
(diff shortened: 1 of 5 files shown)


ExSport:
@SharkHunter @SubJunk @skeptical @valib : What do you think? Can it be useful for others?
Btw. if it will be accepted the question is if default value should stay 0 or 5?
Sorry for dirty pull as I am not programmer :+1: and also had no time to test it :-1:

SubJunk:
@ExSport it seems like a good option. The default could be maybe 3. I think the current code is around 2 (we call anything between 1.7-1.8 16/9)
I wonder if it would make more sense as a user option instead of renderer option, since it is more a matter of preference.

ExSport:
@SubJunk Maybe. I didn't wanted to influence other renderers badly but it is true that other renderers don't force AR transcoding as e.g. Panasonic so seems you are right.
I saw code if AR==16:9 as string so maybe it is on another place in code rounded all 1.7-1.8 to this string.
Anyway I parsed quite a lot of my files to see used AR and most of them are:
```
1.78[16:9], 1.85 (~5%), 2.35 (~33%), 2.40 (~36%)
```
So in most cases 3 will be as 0. Because of that I recommended 5 :+1:

PR #560, opened 2015-05-30
Update myGrid repository URL

Jensaarai:
www.mygrid.org.uk was unresponsive; found here http://stackoverflow.com/a/23640746 that the URL changed to build.mygrid.co.uk, which allowed my build to continue.


Optimusprimeums:
Nice work @Jensaarai This was really annoying me. Now i can build

PR #567, opened 2015-06-06
Fix issue 475 - hidden empty folder refresh

Jensaarai:
Add logic to MapFile that keeps track of empty folders we've ignored when configuration.isHideEmptyFolders is true. The empty folders are re-checked when isRefreshNeeded is called to see if they are now relevant, and if so, isRefreshNeeded will return True.

### DIFF
--- a/src/main/java/net/pms/dlna/DLNAResource.java
+++ b/src/main/java/net/pms/dlna/DLNAResource.java
@@ -1191,6 +1191,9 @@ public abstract class DLNAResource extends HTTPResource implements Cloneable, Ru
 				notifyRefresh();
 			} */
 			if (forced) {
+				// This seems to follow the same code path as the else below in the case of MapFile, because
+				// refreshChildren calls shouldRefresh -> isRefreshNeeded -> doRefreshChildren, which is what happens below
+				// (refreshChildren is not overridden in MapFile) 
 				if (refreshChildren(searchStr)) {
 					notifyRefresh();
 				}
--- a/src/main/java/net/pms/dlna/MapFile.java
+++ b/src/main/java/net/pms/dlna/MapFile.java
@@ -40,6 +40,7 @@ import org.slf4j.LoggerFactory;
 public class MapFile extends DLNAResource {
 	private static final Logger LOGGER = LoggerFactory.getLogger(MapFile.class);
 	private List<File> discoverable;
+	private List<File> emptyFoldersToRescan;
 	private String forcedName;
 
 	private ArrayList<RealFile> searchList;
@@ -98,6 +99,14 @@ public class MapFile extends DLNAResource {
 					/* Optionally ignore empty directories */
 					if (f.isDirectory() && configuration.isHideEmptyFolders() && !FileUtil.isFolderRelevant(f, configuration)) {
 						LOGGER.debug("Ignoring empty/non-relevant directory: " + f.getName());
+						// Keep track of the fact that we have empty folders, so when we're asked if we should refresh,
+						// we can re-scan the folders in this list to see if they contain something relevant
+						if (emptyFoldersToRescan == null) {
+							emptyFoldersToRescan = new ArrayList<>();
+						}
+						if (!emptyFoldersToRescan.contains(f)) {
+							emptyFoldersToRescan.add(f);
+						}									
 					} else { // Otherwise add the file
 						RealFile rf = new RealFile(f);
 						if (searchList != null) {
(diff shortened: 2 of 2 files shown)


SubJunk:
@Jensaarai this looks good, thanks. Is it stable enough to merge?

Jensaarai:
I tested it with my PS3 and did not encounter any issues, so I think it's good to go. I wouldn't be offended if you needed to pull it down and check it out yourself though :)
- If a folder was empty, it was ignored and not displayed in the folder list
- If a file was added to a folder that was empty, the folder would appear after I re-selected the parent folder
- Files added to the empty folder as well as files in other folders played

SubJunk:
@Jensaarai I just tested this and it seems great, and does everything you claimed above. However it doesn't re-hide empty folders if you add a file and then remove it, which it seems like it should do to be complete. It is still better than what we currently have, but do you have any interest in adding that functionality before I merge?

Jensaarai:
Ah; I'll try to fix that, yeah, sorry about that.

SubJunk:
@Jensaarai that would be great! But keep us updated, since this is still a good feature to merge as it is :)

skeptical:
@Jensaarai @SubJunk Not to make your life too difficult, but just fyi if we're willing to ditch Java6 (which I saw mention of elsewhere) there's a much more efficient/low-overhead solution to this problem using [FileWatcher](https://github.com/UniversalMediaServer/UniversalMediaServer/blob/master/src/main/java/net/pms/util/FileWatcher.java), where the filesystem notifies us of any changes in directories we request to be watched.  Usage is like so:
```java
FileWatcher.Listener myListener = new FileWatcher.Listener() {
	@Override
	public void notify(String filename, String event, FileWatcher.Watch watch, boolean isDir) {
		if (event.equals("ENTRY_CREATE")) {
			// ...
		} else if (event.equals("ENTRY_MODIFY")) {
			// ...
		} else if (event.equals("ENTRY_DELETE")) {
			// ...
		}
	}
};

// Watch all files in myDirPath and below
PMS.getFileWatcher().add(new FileWatcher.Watch(myDirPath + "/**", myListener));
```

SubJunk:
@skeptical what are your thoughts on ditching Java 6?
I think at least we need to release PS4 support for Java 6, but after that I'm open to ideas.

taconaut:
@skeptical Just be aware that the nio FileWatcher fires events multiple times. I wrestled with it (http://stackoverflow.com/questions/10063742/use-java-nio-to-raise-create-move-delete-events-on-file-changes) and ended up using the jnotify library.

skeptical:
@taconaut I also found that (in linux) I would get thousands of modify events during file copy, but solved it by building-in a notification delay into ```FileWatcher``` so that subsequent modifies would cancel previous ones and the notification is actually sent only after the commotion dies down.  This tested ok for giant files too, and it's been working fine so far in UMS for (admittedly tiny :-) confs and web templates so far.

skeptical:
@SubJunk in general I'm extremely retrograde as you know, but this is the first java7 feature I've come across that tempts me otherwise :-).

skeptical:
@taconaut no doubt I must have seen your SO thread on this (without following links otherwise I would have figured out who Phillipe was :-), since it seems you also tried essentially the same approach.

SubJunk:
Cool well it looks like we can last out on Java 6 a little bit longer then :) Good work

skeptical:
@taconaut With a big queue delay (2.5 seconds) it seems to be working right in linux, i.e. one ENTRY_CREATE and one ENTRY_MODIFY per file. Making the queue volatile doesn't seem to affect anything, since I'm only seeing one thread per file operation-type, but seems like that would be good to do nonetheless.

taconaut:
@skeptical I had tried pretty hard to find a stable solution without having to use timers and such as there will always be border cases which won't be covered and wanted to let you know the path might be slippery ;)

skeptical:
@taconaut fools rush in where angels fear to tread :).

taconaut:
@skeptical :beers:

Jensaarai:
@SubJunk I haven't had a lot of time lately to work on this, been busy with work :/

I think it might work out better if you pulled this as-is, and I made another pull request at a later time to fix re-hiding empty folders (or fixing it might have to wait for a more robust solution like FileWatcher above). In my understanding of the current system (which is definitely limited), there's nothing that will tell us that a folder is now empty, so without rescanning each time I'm not yet sure how we would solve that :/

SubJunk:
Done, thanks a lot @Jensaarai :)

PR #573, opened 2015-06-16
Add aspect ratio distortion tolerance in % for avoiding unnecessary transcoding (globally defined)

ExSport:
Code simplified and reworked from renderer based to globally based configuration.
TODO: Panasonic TV and maybe others don't require to fix AR for every file as e.g. streamed XviD/DivX files are correctly shown without transcoding.
TODO2: Not possible to configure it via GUI right now.

### DIFF
--- a/src/main/external-resources/UMS.conf
+++ b/src/main/external-resources/UMS.conf
@@ -1046,6 +1046,15 @@ vlc_path =
 
 # ---< Unsorted >-------------------------------------------------------------
 
+# The number represents percentage tolerance from ideal 16:9 AR. Within this
+# tolerance file will not be transcoded due to wrong AR.
+# Useful for files with almost invisible AR distortion like AR = 1.85 (~5%).
+# User can tune value between acceptable distortion and overhead
+# invoked by transcoding.
+# Mostly used wide AR: 1.78[16:9], 1.85 (~5%), 2.35 (~33%), 2.40 (~36%)
+# Default: 5
+keep_aspect_ratio_tolerance =
+
 # Default: false
 transcode_block_multiple_connections =
 
--- a/src/main/external-resources/renderers/DefaultRenderer.conf
+++ b/src/main/external-resources/renderers/DefaultRenderer.conf
@@ -407,6 +407,7 @@ SubtitleHttpHeader =
 # If some videos on your renderer looked either stretched or squashed, enabling
 # this setting may fix that.
 # This is only reliable when MediaInfo = true.
+# TODO: PanTV doesn't require AR fix for streamed DivX/XviD files!
 # Default: false
 KeepAspectRatio = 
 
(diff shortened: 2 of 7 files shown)


ExSport:
Still incomplete because now I see in log that UMS gets DAR also as 320:173 etc. instead of 1.84xyz...
Will make changes...

ExSport:
Ok now I understand this function:-) I thought AR is in a format of double/float number.
Now I checked that in almost all cases MediaInfo already sends ?:? format of AR so there is no need to convert it from 1.78 etc.
But Mediainfo returns also AR format like 1.85:1, 2.40:1 etc. and is already automatically rounded (1280x536=2.388 but presented as 2.40:1, 1280x694=1.84438 and 1280x692=1.8497 presented as 1.85:1) so there is already some rounding inbuilt.
Newly implemented tolerance will allow round AR like 1.85:1 to 16:9 (~5% distortion) which is acceptable for most users but also there is a place to use higher distortion tolerance to convert 2.40:1 to 16:9 so transcoding will be omitted as users can on some rendereres manually force AR.
Or is it overhead with defining tolerance and use only one 5% distortion [AR 1.85=1.78]? {like AcceptSmallARDistortion=True/False}?
Any thoughts welcomed (as 5% distortion tolerance setting is enough for me...)
Digging more and found out that MediaInfo always returns AR in x:y format as UMS is using __DisplayAspectRatio/String__ which returns e.g. 2.40:1. Numeric value of AR is stored in __DisplayAspectRatio__ = 2.388.
Will think about rounding by new code with tolerance implemented (using ___DisplayAspectRatio_) or using already coded ___DisplayAspectRatio/String_ and adding possibility to round 1.85-->1.78...

ExSport:
In LibMediaInfoParser.java there is:
```
media.setAspectRatioVideoTrack(MI.Get(video, i, "DisplayAspectRatio_Original/Stri"));
```
I suppose it should be:
```
media.setAspectRatioVideoTrack(MI.Get(video, i, "DisplayAspectRatio_Original/String"));
```
Anyway I tested what this property returns for some of my files (20?) and it was always empty string :-)
Btw. I found one file which returned 1.714 for __DisplayAspectRatio/String__ property, so not always it is in x:y format (but for common files all are in x:y format, this one was some crappy video clip from net)

SubJunk:
@ExSport @valib I think that string was truncated for a reason, please test that kind of thing before committing :)

valib:
@SubJunk this is a real bug. It was truncated by mistake when the command was taken from the GUI not from the code. When I exported all commands to the CSV format the parameter is 
`DisplayAspectRatio_Original/String;Original (in the raw stream) Display Aspect ratio`

ExSport:
Approving it is a bug. Without fix it returns null but fixing the code it returns correct value.

ExSport:
As MediaInfo partially rounds aspect ratio itself I reconsidered patch.
Also tried to limit configuration confusion to true/false option only as definding % is overkill.
https://github.com/UniversalMediaServer/UniversalMediaServer/pull/578

PR #578, opened 2015-06-21
Allow 5% distortion tolerance for AR=1.85 for avoiding unnecessary transcoding

ExSport:
Take 1.85:1 aspect ratio as 1.78:1 [16:9]. This will avoid transcoding for AR=1.85:1 if KeepAspectRatio=true is defined in renderer.conf and Force_Exact_Aspect_Ratio=false in UMS.conf.
This feature is disabled by default as for someone 5% distortion (1.78~1.85) can be confusing.
Or should it be enabled to avoid transcoding/network bandwidth overhead and for better experience of FF/RW?
aspectratio

[Review on Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/578)

### DIFF
--- a/src/main/external-resources/UMS.conf
+++ b/src/main/external-resources/UMS.conf
@@ -1046,6 +1046,14 @@ vlc_path =
 
 # ---< Unsorted >-------------------------------------------------------------
 
+# Renderer.conf "KeepAspectRatio" option allows forcing transcoding
+# for fixing non 16:9 aspect ratios.
+# Skipping transcoding for files with almost invisible AR distortion
+# like AR = 1.85 (~5%) can be accepted by most users to minimize CPU and
+# network bandwidth overhead or to not loose flawless FF/RW experience.
+# Default: true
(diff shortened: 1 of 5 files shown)


ExSport:
Or Force_Strict_Aspect_Ratio/Strict_Aspect_Ratio will be better? Any thoughts for changes? :-)

SubJunk:
@ExSport this seems like a good idea. I think we should make it a bit more general by making the configuration option aspect_ratio_tolerance where it can be a number between 1-100 where it's the percentage of allowed variation from the renderer's aspect ratio. 100 would be the same as just using keep_aspect_ratio=false

ExSport:
@SubJunk check linked discussion. If I remember right in meantime I reconsidered it to use 5% only as UMS right now retrieves AR from MefiaInfo already rounded and also to minimally confuse user with new settings :) But my memory already lost details about this patch 😀

SubJunk:
@ExSport cool, does that mean we can close this PR?

PR #575, opened 2015-06-19
Toggle the state of the install service button

taconaut:
Instead of having two active buttons to install and uninstall UMS as a windows service, have a single one, which shows install or uninstall depending on the installation state of the service.

I have tested it only on win7.
~~I might have put the plugin to copy the wrapper in the wrong spot which might lead to errors on other platforms.~~ Please test :)

Changes:
- Added WindowsUtil class with method isUmsServiceInstalled()
- Moved install and unistall service methods from PMS to WindowsUtil
- Added constant NEWLINE_CHARACTER to StringUtil
- Added empty constructor to CustomJButton
- Removed the unistall button
- Update text, tooltip and executed action for installService button in new method refreshInstallServiceButtonState() depending on the installation state of the service
- Copy the service wrapper into the working directory in order to be able to debug this functionality and add thisdirectory to the git ignore list

PR #583, opened 2015-06-25
Copy web resources into the working dir while debugging

taconaut:
This pull request modifies the maven config to copy the web resources from external resources into the working dir during the process resource phase for all environments (win, osx, linux).
With this, the website will be available while debubugging from the IDE without having to manually copy files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ target
 *.orig
 win32/
 *.sublime-*
+web/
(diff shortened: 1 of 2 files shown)


valib:
@taconaut nice one. It helps a lot.

skeptical:
@taconaut +1:.

PS not a big deal, but some of us are dinosaurs without IDEs and don't actually need this, so might be good to make it optional, e.g. ```<target if="copyresources">```, enabled by ```-Dcopyresources```.

SubJunk:
Yeah would be cool to see it made optional :)

taconaut:
To be coherent, the MediaInfo DLLs as well as the win32 folder (I've added with #575) should also be copied conditionally into the working dir, right?

PR #590, opened 2015-07-06
Rewrite of PMS.isHeadless() so that the actual check is done only once

Nadahar:
On top of being called several times during init from PMS and DLNAMediaDatabase, it's called repeatedly from DLNAMediaInfo.parse(). I changed the logic so that the check is only done once during an application instance. This is not a bug, it's just for efficiency.

### DIFF
--- a/src/main/java/net/pms/PMS.java
+++ b/src/main/java/net/pms/PMS.java
@@ -1050,7 +1050,6 @@ public class PMS {
 
 	public static void main(String args[]) {
 		boolean displayProfileChooser = false;
-		boolean headless = true;
 
 		if (args.length > 0) {
 			for (String arg : args) {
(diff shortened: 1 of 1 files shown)


src/main/java/net/pms/PMS.java
```diff
@@ -1086,9 +1085,7 @@ public static void main(String args[]) {
 				if (System.getProperty(NOCONSOLE) == null) {
 					System.setProperty(CONSOLE, Boolean.toString(true));
 				}
-			} else {
-				headless = false;
```
SubJunk: @Nadahar why was this removed?

Nadahar: @SubJunk The author of PMS.main() had made a local variable ```headless``` probably to avoid calling ```isHeadless()``` unnecessary. It was set on line 1090 based on the result of ```isHeadless``` and then read once in line 1100: 
```java
		if (!headless && displayProfileChooser) {
			ProfileChooser.display();
		}
```
 Since ```isHeadless()``` is now rewritten so that it only makes the test once, and after that just returns the previously estabilished status, I simply replaced the test at line 1100 (now 1097) to use isHeadless() instead:
```java
		if (!isHeadless() && displayProfileChooser) {
			ProfileChooser.display();
		}
```
I then deleted the variable and its remaining reference, the line above.

SubJunk: @Nadahar cool, thanks for the explanation :)

Issue #579, opened 2015-06-21
Problems with xml encoding of predefined entities in element names (artists, titles, albums etc) during ContentDirectory1#Browse


Nadahar:
I started this "case" on your [help forum](http://www.universalmediaserver.com/forum/viewtopic.php?f=9&t=3496&p=16679#p16679). My original problem was that some music titles and artists names where shown wrong to the extent that it's hard to read. Since then I've studied UMS, Java, Eclipse, and Maven much more than I'd want to :open_mouth:

Nadahar:
I even got @valib to commit 2258692a944a7f86874b12ec6846d2f6b5d5118f and then later revert ab04b6cede2085c9854ac9354637cb3c8805427b. I'm sorry about that, it seemed like an obvious bug AND it seemed to work on my Onkyo renderer (even though I've figured out that it's not that simple now).

Nadahar:
I just did a Wireshark trace for Plex as well, and it has the same behaviour as UMS - and is not presenting titles etc. correctly on the Onkyo:
```
&lt;dc:title&gt;Konstant i Opel &amp;apos;34&amp;apos; ds &amp;quot;3&amp;amp;&amp;quot; &amp;gt;&amp;gt; ds&lt;/dc:title&gt;
```

Nadahar:
After much looking, I finally found that [UPnP-av-ContentDirectory-v1-Service] (http://upnp.org/specs/av/UPnP-av-ContentDirectory-v1-Service.pdf) (which UMS uses) refers to [XML Section 2.4 Character Data and Markup] (http://www.w3.org/TR/REC-xml/#syntax):
> Note that since the DIDL-Lite format of Result is based on XML, it needs to be escaped (using the normal XML rules: [XML] Section 2.4 Character Data and Markup) before embedding in a SOAP response message. In addition, when a variable of type A_ARG_TYPE_Result is employed as a parameter in a comma-separated list, commas (‘,’) that appear within XML CDATA must be escaped (as \,), see section 2.3.1.

[XML Section 2.4 Character Data and Markup] (http://www.w3.org/TR/REC-xml/#syntax) states:
> Text consists of intermingled character data and markup. [Definition: Markup takes the form of start-tags, end-tags, empty-element tags, entity references, character references, comments, CDATA section delimiters, document type declarations, processing instructions, XML declarations, text declarations, and any white space that is at the top level of the document entity (that is, outside the document element and not inside any other markup).]
> 
[...]

PR #592, opened 2015-07-06
Fix issue #579 XML encoding of predefined entities

Nadahar:
The explanation can be found in #579. Some renderers can't handle that single ( ' ) and double ( " ) quotes are XML encoded/escaped. After a lot of searching DLNA/uPnP specs i finally found a reference to the XML spec that applies. It states that encoding/escaping the characters in question is optional, which means to renderers should cope with them being both encoded/escaped and not.

Even though the renderers should accept them encoded/escaped, some don't. When checking some media servers, I found that neither Serviio not Intel's Media Server (which is supposed to be a reference implementation) encode/escape them. Plex on the other hand does. Logically I find having them unencoded/unescaped most likely to be compatible since even if the renderer decodes/unescapes them the result is likely to be correct as they are simply left untouched. Since they are optional having them unencoded still produces a valid document.

I've left the omitted lines commented out in the freak case that some renderer actually require them to be encoded/escaped, in which case it would have to be implemented as a renderer configuration and having the comment there makes it easier to remember what needs to change. If you don't like such commented blocks let me know and I'll remove it since the history is provided by git anyway.


SubJunk:
@Nadahar you're right that in general committing commented out code is a bit redundant with versioning, but in this case I think it may be useful to future developers to have that note to let them know the omission is purposeful.

I'm not a fan of changing things that work, especially after all these years of observing that simple things like this can break renderer support, so I think this would be better as a renderer option, but then considering that Serviio doesn't escape them probably means it's safe since they're a fairly stable product.

@UniversalMediaServer/developers any opinions on this?

Nadahar:
@SubJunk I agree with your scepticism about changing things that work, and if you want to I can make it a renderer option. But, I'm also thinking that creating options that's not really needed is confusing to users "for no reason".

I think it's a very good chance that this won't break anything, because how the parsing at the renderer must be done. When it e.g "decodes" ```&quot;``` to ```"``` in a string, an already existing ```"``` in that string will remain untouched, while a ```&quot```; would be changed to a ". Either way, you end up having a string containing the ```"```. ```&<>``` must be encoded simply because ```&``` is the escape character, and ```<>``` are used for tags. But, since they made single- and double-quote encoding optional, it means that having them unencoded doesn't break the syntax of the XML document itself, it's still a valid document. The parser know that it's not part of a tag, but is a string/value.

It's not problem for me to make the renderer option, my only worry is if it complicates things (for the users configuring it) when I consider the chances for it creating trouble to be very slim. Another thing to remember is that this "problem" is only seen if you have titles/artist/album names etc containing these characters. That's probably not that common, so I'm betting that there are more renderers out there that don't expect them to be encoded, but that people might not have noticed or reported it because it's rare.

SubJunk:
Thanks :)

Nadahar:
:+1:

PR #593, opened 2015-07-06
Rewritten Norwegian translation

Nadahar:
I rewrote the Norwegian translation, only a very few lines are kept, because it was badly outdated and the translations that were there often didn't mean the same as the corresponding english sentence. I suspect that ```message.properties``` have had quite a few edits since the translation was done.

It's not always easy to figure out how to translate a given sentence, as i often depends on context. To try to get this as correct as possible, I relied heavily on searching the entire project and seeing where the label in question was used. Doing this, I discovered some labels that were not in use in ```message.properties``` and I've also deleted them from there instead of creating translations that will never be used. That doesn't mean that I've searched all labels and done a complete cleanup of ```message.properties```, I've only corrected those I stumbled upon. I've also done a very few corrections where I found obvious mistakes in the english version.

I'd like to comment specificly on the change I did at ```TrTab2.24```. I changed it from:
> CPU threads to use when enabled for engine: (it seems you have %d)

to:
> CPU threads to use when enabled for engine: (you seem to have %d cores)

The reason is that I think it was confusing what the setting actually did, and looking at the naming in the underlying code it seems like there's some confusion there as well. As I'm sure you know there are no real link between the number of cores and threads. You can spawn as many threads as you like, and the OS scheduler runs them where it sees fit. If you have 4 cores and spawn 4 threads, you are in no way guaranteed that these will be run on separate cores. The only corrolation between the two is that there's no way to use all available cores for a given task with fewer threads than available cores as one thread will only run at one core at any given moment. UMS defaults to one thread per CPU core here, and that might be sensible from that perspective, but the way the text was written indicated that you had *%d threads available* which simply doesn't make any sense and thus is confusing.


SubJunk:
@Nadahar this seems great, thanks a lot :)

PR #596, opened 2015-07-09
Fixed a non-breaking bug in PmsConfiguration

Nadahar:
PropertiesConfiguration.load() expects a valid URI as a parameter, while a path was sent. PropertiesConfiguration would figure it out and load correctly, but threw a couple of ```DEBUG``` messages complaining about it.

This is fixed.

### DIFF
--- a/src/main/java/net/pms/configuration/PmsConfiguration.java
+++ b/src/main/java/net/pms/configuration/PmsConfiguration.java
@@ -507,7 +507,7 @@ public class PmsConfiguration extends RendererConfiguration {
 
 			if (pmsConfFile.isFile()) {
 				if (FileUtil.isFileReadable(pmsConfFile)) {
-					((PropertiesConfiguration)configuration).load(pmsConfFile.toURI().toString());
+					((PropertiesConfiguration)configuration).load(pmsConfFile);
 				} else {
 					LOGGER.warn("Can't load {}", PROFILE_PATH);
 				}
--- a/src/main/java/net/pms/configuration/PmsConfiguration.java
+++ b/src/main/java/net/pms/configuration/PmsConfiguration.java
@@ -507,7 +507,7 @@ public class PmsConfiguration extends RendererConfiguration {
 
 			if (pmsConfFile.isFile()) {
 				if (FileUtil.isFileReadable(pmsConfFile)) {
-					((PropertiesConfiguration)configuration).load(PROFILE_PATH);
+					((PropertiesConfiguration)configuration).load(pmsConfFile.toURI().toString());
 				} else {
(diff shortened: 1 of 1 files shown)


skeptical:
@Nadahar thx, can you give an example of what exactly is fixed and how it qualifies as a bug? Also ```toURI().toString()``` is relatively expensive, how about just [```load(pmsConfFile)```](http://commons.apache.org/proper/commons-configuration/javadocs/v1.10/apidocs/org/apache/commons/configuration/AbstractFileConfiguration.html#load(java.io.File))?

Nadahar:
@skeptical Looking at the overloaded methods of load() i see that ```File``` is also a valid parameter and that works just fine. I don't know java well enough to know the cost of things that well I'm afraid, and I wasn't that focused on that since I'm guessing the configuration file isn't loaded that often. I've changed it anyhow.

I guess it's debatable whether or not it constitutes a bug since it worked the way it was, that was why I wrote "non-breaking": My wording should maybe have been different, but I don't really know what else to call it. The behaviour that's changed by this is that the following lines no longer appear in the log each time you start UMS:
```
18:38:35.854 [main] DEBUG o.a.c.c.ConfigurationUtils - ConfigurationUtils.locate(): base is null, name is C:\ProgramData\UMS\UMS.conf
18:38:35.858 [main] DEBUG o.a.c.c.DefaultFileSystem - Could not locate file C:\ProgramData\UMS\UMS.conf at null: unknown protocol: c
18:38:35.858 [main] DEBUG o.a.c.c.ConfigurationUtils - Loading configuration from the absolute path C:\ProgramData\UMS\UMS.conf
```

skeptical:
@Nadahar thx, squashed + merged: c56990d :+1: . I've never seen those messages before, but it's likely more efficient regardless, since ```load(String)``` probably calls ```load(File)``` internally.

Nadahar:
@skeptical 
> I've never seen those messages before

It depends on the environment, but they aren't logged to ```debug.log``` because they occur before ```logback.xml```/```logback.headless.xml``` is loaded. In this state logback is in a "default" state basicly searching for any ```logback.xml``` it can find in your classpath, and if that fails it loads a basic configuration that only logs to console. Whether you see them or not therefore a bit random.

skeptical:
@Nadahar I take it back then,  I must have seen them since I'm usually running ums from a console (for 'printf' debugging), but I generally ignore the cascade of 'pre-logging' messages that go by at startup :).

Nadahar:
@skeptical I've been looking through the startup logs some times looking for other things, and being slightly OCD I couldn't leave them there in the end ;)

PR #605, opened 2015-07-15
Made ProcessUtil.run() accept expected exit codes as ok

Nadahar:
When UMS checks if it's installed as a Windows Service the exit code from the process is 1060 if the service is not installed. ProcessUtil.run() logs this as a warning, while it's actually a legit result. I've added a parameter so that the caller can choose to accept non-zero exit code, and I've added an overloaded version to accept the current parameters with a warning (that is, behaving exactly like before).

I've also changed the log level from debug to warn on the log message, now that it's possible to filter out "false positives".

### DIFF
--- a/src/main/java/net/pms/util/ProcessUtil.java
+++ b/src/main/java/net/pms/util/ProcessUtil.java
@@ -149,7 +149,7 @@ public class ProcessUtil {
 							p.exitValue();
 						} catch (IllegalThreadStateException itse) { // still running: nuke it
 							// kill -14 (ALRM) works (for MEncoder) and is less dangerous than kill -9
-							// so try that first 
+							// so try that first
 							if (!kill(pid, 14)) {
 								try {
 									// This is a last resort, so let's not be too eager
@@ -176,7 +176,7 @@ public class ProcessUtil {
 	}
 
 	// Run cmd and return combined stdout/stderr
-	public static String run(String... cmd) {
+	public static String run(int[] expectedExitCodes, String... cmd) {
 		try {
 			ProcessBuilder pb = new ProcessBuilder(cmd);
 			pb.redirectErrorStream(true);
@@ -190,7 +190,16 @@ public class ProcessUtil {
 				}
 			}
 			p.waitFor();
-			if (p.exitValue() != 0) {
+			boolean expected = false;
+			if (expectedExitCodes != null) {
+				for (int expectedCode : expectedExitCodes) {
+					if (expectedCode == p.exitValue()) {
+						expected = true;
+						break;
+					}
+				}
+			}
+			if (!expected) {
 				LOGGER.debug("Warning: command {} returned {}", Arrays.toString(cmd), p.exitValue());
 			}
 			return output.toString();
@@ -198,6 +207,11 @@ public class ProcessUtil {
 			LOGGER.error("Error running command " + Arrays.toString(cmd), e);
 		}
 		return "";
+	}
+
+	public static String run(String... cmd) {
+		int[] zeroExpected = { 0 };
+		return run(zeroExpected, cmd);
 	}
 
 	// Whitewash any arguments not suitable to display in dbg messages
@@ -235,7 +249,7 @@ public class ProcessUtil {
 			// We're doing a straight reboot
 			cmd = reboot;
 		} else {
-			// We're running a script that will eventually restart UMS 
+			// We're running a script that will eventually restart UMS
 			if (env == null) {
 				env = new HashMap<>();
 			}
--- a/src/main/java/net/pms/util/WindowsUtil.java
+++ b/src/main/java/net/pms/util/WindowsUtil.java
@@ -20,7 +20,8 @@ public class WindowsUtil {
 	 */
 	public static boolean isUmsServiceInstalled() {
 		String[] commands = new String[]{ "sc", "query", "\"Universal Media Server\"" };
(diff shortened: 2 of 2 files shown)


Nadahar:
Is there something wrong/missing/unclear about this?

Nadahar:
@SubJunk Is there something wrong/missing/unclear about this? It's just to remove an error message that's not actually an error at all. Maybe The message should stay at ```debug```, but my reasoning was that for external processes that's supposed to return 0, something requiring attention is probably going on if it's non-zero which should be shown also to "normal" users not running debug.

Nadahar:
I'd really like to know if anyone has any problems with this, I've never had any feedback on this one. What it changes is simply the assumption in ProcessUtil.run() that a non-zero exit code should always be 0 unless an error occurred. 

Even though this is often the case, Microsoft see it differently and use the exit code to return the process output state. Exit code ```1060``` literally mean that the service is not installed when you run ```sc query "Universal Media Server"```, it is _not_ an error in any way. This simply modifies the ```run()``` to accept an additional parameter: ```boolean logNonZeroExitCode```. If this is false, a non-zero error code will not be logged as an error, which is correct e.g. in the case when you check the existance of a Windows service. I've kept the original call as an overloaded method that substitute the "missing parameter" with ```true```. 

This means that no code need to be changed anywhere else, the only difference is that the caller now has the ability to choose whether or not a non-zero exit code should be considered an error.

Nadahar:
@skeptical I feel like I'm explaining myself badly, so I'll try again (remember that English isn't native to me):

My original reason to do something about this was that when I look through the log and see a warning that this process returned exit code 1060, I interpret this as "Ah, something went wrong, I have to look into this". I don't remember that ```sc``` returns 1060 if it the service doesn't exist.

When I found the reason for the code, and saw how this was constructed, I concluded that the author of ```ProcessUtil``` didn't plan on exit code being used this way. Normally a process will give you an output which you might need to parse, but the exit code will be zero unless something went wrong. My solution to this was simply that the caller should decide if the usage case was the "normal" (exit code != 0 means an error) or a situation where a non-zero exit code would be expected.

I then thought further, that if we by doing this filtered out the cases where the exit code was expected to possibly be non-zero, the remaining cases might actually be something to alert the user about. In that way, we might pick up things that went wrong with external processes that normally went by unnoticed, or the user could detect something wrong (like a wrong version, permission problem or whatever) instead of just experiencing "strange" behaviour for no apparant reason.

I still feel that the idea to hide away everything that might go wrong to not unduly worry people is a bit like the crew on the Titanic telling everyone that all is fine until the water started flowing in and it was more of a "every man save himself". If things are to be logged just as a detailed reference of what's happening, I think ```trace``` is the more appropriate level for that, and in that case the exit code should be logged whatever it might be.

skeptical:
@Nadahar I'm the author in this case, and my intent/reasoning is what I've explained above (short version: non-zero exit code is always good to know).  We can have a long discussion about what it appears I knew or didn't know based on reading between the lines or following breadcrumbs, but it's not really productive.

Nadahar:
@skeptical My point wasn't to say anything about what the author actually thought, it was to explain how I reasoned at the time I did this. I simply tried to explain 1) Why I though a change was needed 2) Why I opted to change the logging level.

skeptical:
@Nadahar and in my first comment I agreed with you on the necessity for the change, just added some thoughts as to what the appropriate fix was.  I didn't misunderstand your critique or thought process, you stated it very well :).

Nadahar:
@skeptical Ok, maybe it's me who doesn't understand then, because I'm still not quite sure about what you mean except that you don't like the ```warn``` logging level.

skeptical:
@Nadahar loglevel WARN appears (by default) in the gui, and looks a little scary. In this case the exit code is significant enough to log, but almost always not a cause for great concern.  That's the reasoning I tried to convey in the first comment.  This is why I had originally used the milder approach of just using the word "warning" in a debug message. Your critique caused me to rethink this, and now I agree on making the message more neutral.

Nadahar:
@skeptical Logging level changed to ```debug```. Any other issues?

skeptical:
@Nadahar my main issue was that ```logNonZeroExitCode``` isn't the right fix, we should always log non-zero exit code, it can't hurt. This is true for ```sc``` too, since it isn't guaranteed that the non-zero will always be 1060.

The minor issue (alarm) is most easily fixed by just omitting the scary word 'warning' from the debug message, e.g.
```logtalk
[...] command sc query "Universal Media Server" returned 1060
```
You convinced me it's simplest not to use the word 'warning' at all here, ever.

Nadahar:
@skeptical What about this approach? It lets ut log anything unexpected and at the same time not log what's expected looking like a potential problem.

skeptical:
@Nadahar :+1: To me this is a clean solution without any conceptual flaw, though we can (hopefully not :-) have another debate about whether it's worth the trouble here.

To simplify it a bit I would pass ```null``` instead of creating ``zeroExpected``` in the default case, and test with ```if (p.exitValue() != 0 && expectedExitCodes != null)``` (unless of course you want to preserve the possibility where 0 would be unexpected :-).

Nadahar:
@skeptical My idea was that there might be a situation where you'd want to log any exit code (0 were "unexpected").

skeptical:
@Nadahar theoretically valid, I guess, I don't really object. Tho it's not necessary yet, or probably ever.

PR #606, opened 2015-07-16
Add Eclipse codeStyle.xml

Nadahar:
This is not ready to merge, it's more of a suggestion. I haven't really figured out the "UMS code style" yet other than that you prefer less blank lines than me and that you put else, catch etc on the same line as }.

I've simply exported my current Eclipse profile as a starting point. It would really make it easier if you agreed on what the desired settings are and then included the file in the project. In that way, anyone using Eclipse could get it right from the beginning.

### DIFF
--- a/CheckStyle.xml
+++ b/CheckStyle.xml
@@ -1,0 +1,171 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE module PUBLIC "-//Puppy Crawl//DTD Check Configuration 1.3//EN" "http://www.puppycrawl.com/dtds/configuration_1_3.dtd">
+
+<!--
+    This configuration file was written by the eclipse-cs plugin configuration editor
+-->
+<!--
+    Checkstyle-Configuration: UMS
+    Description:
+Checkstyle configuration that checks the sun coding conventions.
+-->
+<module name="Checker">
+  <property name="severity" value="warning"/>
+  <property name="fileExtensions" value="java, properties, xml"/>
+  <module name="TreeWalker">
+    <module name="RegexpSinglelineJava">
+      <property name="format" value="^\t*(?! \*) "/>
+      <property name="message" value="Indent must use tab characters"/>
+      <property name="ignoreComments" value="true"/>
+    </module>
+    <property name="tabWidth" value="4"/>
+    <module name="JavadocMethod">
+      <property name="scope" value="protected"/>
+    </module>
+    <module name="JavadocType">
+      <property name="scope" value="protected"/>
+    </module>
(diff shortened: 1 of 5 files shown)


SubJunk:
We aren't very strict about code style but it would be nice to define an ideal standard.
A few things that I usually change in formatting commits are:

Use tab-indentation, but use space-indentation for inline-indentation. This is because it allows developers to change their tab display to match the level of indentation they prefer (i.e. developers who like using 2-space indentation can just change their IDE settings to use 2-space tab width), and using spaces for inline-indentation means the indentation is always in sync with itself no matter which tab-width is set by the developer.

if {
} else {
}

Use switches instead of ifs for 3 or more conditions.

Use blank lines for grouping. So if you initialize a variable just before the statement it's use in, there's no blank line between them. But if you're initializing a variable to be used later in the code, a blank space is used. Similarly if if statements are related they don't have a blank space but if they are separate they do.

Those are some ranty thoughts on code style and I'm happy for people to disagree and it would be cool if we can define a document that explains what we agree on

Nadahar:
I've made certain changes to my ```codeStyle.xml``` since then to better fit UMS' style (UMS' style isn't homogeneous, so it's hard to be precise). It's good enough to use for automatic formatting of simple files, but I wouldn't dare to use it on more complex classes.

I also have a ```checkstyle.xml``` that I'm using some, but I think it still needs some more work to fit properly. We could always make updates to them as we see fit later though, so it might be an idea to start with those I have.

Sami32:
:+1:
Better, yes ;)

Nadahar:
I've added more "formatting" setting files I use, and updated the one that was already here. They are far from perfect, but they're much better than nothing and should ideally evolve with the project.

Nadahar:
@SubJunk @valib Do you have any reservations against merging this? None of this is applied automatically, it's just template files that provides a default that matches UMS a bit better than the default settings.

SubJunk:
I haven't reviewed the changes but since I don't use Eclipse I don't mind :) Happy for it to be merged

valib:
@Nadahar no objection but I reccomend to add somewhere how to set Eclipse to use those settings.

Nadahar:
@valib Is that better?

valib:
@Nadahar mybe I am doing something wrong but the setting is not automatically enabled. This works only when I manually change the Eclipse setting.

Nadahar:
@valib Yes, I described how to use them in the Eclipse.README file. Nothing of this is automatically applied. If we want them to be automatically applied we have to check in the ```.settings``` folder as well. My thought wasn't to enforce any of this, but to make these settings easily available for those who wanted them.

PR #601, opened 2015-07-12
Cleanup messages.properties

ler0y:
As a continuation of the work Nadahar already did:
https://github.com/Nadahar/UniversalMediaServer/commit/814fa1f0a5f7d641a3d734037520cfab169661dc

I have checked all the labels in messages.properties whether they are still in use in the code, and deleted the unused.

PR #611, opened 2015-07-23
Polish language update

K4r0lSz:
Polish language updated from english scratch.
Currently all lines translated without polish letters.
Next step will be including polish letters, projected delivery date week 34 of 2015.


Nadahar:
@K4r0lSz Great job, only too bad that #601 hasn't been merged yet - it removes about 50 of the english lines that's no longer in use.

Regarding the special characters, it might be that you know this already, but if you use Eclipse and e.g. download the additional [Properties Editor] (http://marketplace.eclipse.org/content/properties-editor) you can just write or paste a document with the unicode characters in it, and the plugin will convert to and from escaped unicode codes transparantly. That means that the file you save will be correctly coded.

ExSport:
Or without Eclipse:
http://www.universalmediaserver.com/forum/viewtopic.php?f=13&p=17415#p17398

K4r0lSz:
Yeap, I know from UMS forum, but first I'd use WinMerge to work PL EN side-by-side.
I'll put PL chars in coming weeks.
Currently checked length and fixed some too long strings.

SubJunk:
Thanks!

PR #612, opened 2015-07-23
Flash video adpcm parse error

ler0y:
Flash video with ADPCM audio are wrongly parsed as LPCM.
This commit fix this but maybe breaks something else.
Please verify this change because I am not a skilled coder :)

Test files:
http://samples.mplayerhq.hu/FLV/flash_flv_adpcm_testfiles/

### DIFF
--- a/src/main/java/net/pms/dlna/LibMediaInfoParser.java
+++ b/src/main/java/net/pms/dlna/LibMediaInfoParser.java
@@ -489,7 +489,7 @@ public class LibMediaInfoParser {
 			format = FormatConfiguration.AAC_HE;
 		} else if (value.startsWith("adpcm")) {
 			format = FormatConfiguration.ADPCM;
-		} else if (value.equals("pcm") || (value.equals("1") && (audio.getCodecA() == null || !audio.getCodecA().equals(FormatConfiguration.DTS)))) {
+		} else if (value.equals("pcm") || (value.equals("1") && (audio.getCodecA() == null))) {
 			format = FormatConfiguration.LPCM;
 		} else if (value.equals("alac")) {
 			format = FormatConfiguration.ALAC;


Sami32:
@Nadahar Any idea on that ?

I tried with your latest audio improvement, still i don't understannd how is handled FLV...

Nadahar:
@Sami32 My audio format commits shouldn't impact FLV handling as FLV is video.

What @ler0y removed seems like a bug in any case - what a strange idea to assume that something is LPCM if it *isn'* DTS.

Trying to track this down, it originates from 2cd57ec4a0ed1df3379ad92b0d942265161e5b7b. It's quite old so a lot has changed since then, but IMO the real problem is this whole approach to parsing format from MediaInfo. It's as good as impossible to know the impact of changing things here, in this case we need to know what situation the original author tried to match (as the match needs to be narrowed but still match what it was intended to). I don't know a way to figure that out...

Sami32:
@Nadahar I does ;) as i'm working on a PR with some audio fixes, and i added a FlashAudio class.

I'm wondering why it was done this was as well !

Nadahar:
@Sami32 It makes absolutely no sense to me, and it really needs a refactoring. The most important thing would be to parse the different parameters at the same time, not in sequence **using the same parsing code**. It's completely insane imo.

If the parsing was done with all parameters at the same time, you would have much more control over the combination of values you were trying to match.

But, as this code is critical to the operation of UMS I don't know if anyone dares to touch it. Trying to refactor ```getFormat()``` (even the name is wrong) would without a doubt break a whole lot of things because the information about all the cases currently being handled by the code isn't documented anywhere, and is impossible to read from the code. It would probably need beta testing and a lot of follow-up tweaks to get things right. At the same time, the current code is a constant source of bugs...

Sami32:
Good idea ! i didn't even think of that.

But as you know my low level skill, for now i'll just add and patch ;)

2017-01-28 17:19 GMT+01:00 Nadar <notifications@github.com>:

Sami32:
@Nadahar Now i remember, i did fixed DTS in WAV detection some months ago.

Nadahar:
@Sami32 I don't quite follow you with how this ties together, but if you've tested it and it works I guess it's ok. The code in question is still highly likely to cause bugs AFAICS.

ler0y:
@Sami32 did you fixed this? imo the bug is still present in the latest UMS code..

stereo_11k.flv (sample Flash video from link see first message, with ADPCM audio)

debug.log gives:
Audio codec "lpcm" failed to match support line f:flv v:sor a:adpcm m:video/x-flv
Matched support line f:flv v:sor a:lpcm m:video/x-flv

Sami32:
@ler0y Yes, but it has not been merged.

Though you can build [some_AUDIO_fixes](https://github.com/UniversalMediaServer/UniversalMediaServer/compare/some_AUDIO_fixes)

ler0y:
@Sami32 Would it not be better to re-open this issue pull request until this issue is really resolved in the master branch?

SubJunk:
If this bug still exists, there should be an issue created for it, and if someone wants to make a better fix they can submit a new PR

PR #589, opened 2015-07-03
Encoding of xml predefined entities, Norwegian translation, new renderer profiles, audio formats ++

Nadahar:
**Notice**

This is not ready to merge, I just needed a place to discuss issues I have with this instead of posting comments and creating issues all around the place. 

**Preamble**

This is the first time I do a fork/pull request (I'm new to GitHub), and I've already realized that I've made some mistakes. The main problem is that this has grown to be a mixture of differnt things which makes is somewhat disorganized. In the future I will definitly create separate request for separate issues. That said, I do want to try to get this through as I've put quite a lot of effort into this.

It all started with me wanting to fix the xml predefined entities (xml encoding of "><&') for some renderers like my Onkyo TX-NR807. During that I found a couple other small things I thought needed improving. Then I figured I should re-do the Norwegian translation as it was so bad that it was hardly comprehensible, and so I retranslated everything, not only adding what was missing. As the existing renderer profile that was used on my Onkyo (Onkyo TX-NR717) didn't really work very well with my Onkyo (the only thing that would play was MP3), I wanted to add a better profile as well. Being slightly OCD, that led me to the conclution that I'd have to test every format Onkyo claimed to be compatible. That again led to a big overhaul of the audio formats as much was missing or interpreted wrong.

This is also the first time I'm coding Java, although I've used quite a few other languages before. Pascal is my primary language though, so there's a lot of unknowns here to me (Eclipse, dependencies, Maven, formatting +++).

Whether or not you wish to use my changes remains to be seen, but I hope atleast some of it is of interest. But, being new to the project, new to the language and new to the environment there's a lot of things that's unclear or unknown to me, so navigating this all has been somewhat difficult and I will just have to excuse if I've done some newbie mistakes. I certainly have tried my best to verify things, but sometimes I just can't seem to figure things out (as you've seen by my postings around here).

**Main**

These are the issues I have at the moment:

* While trying to build UMS, it fails a test: ``` org.junit.ComparisonFailure: "test.wma" is expected to match expected:<[M4]A> but was:<[WM]A> ```. I do understand why that fails, as I have reworked a lot of the audio formats and updated extensions. What I don't understand is why "test.wma" is supposed to be recognized as M4A when it clearly isn't. Is this some kind of special PS3 trick, and if so, shouldn't it be handled elsewhere (limited to the PS3, not globally)? I understand the importance of testing for backwards compability, but does that mean you want to keep something that is plain wrong, or should I just adjust the test to expect WMA?

[...]

PR #613, opened 2015-07-23
Added opus extenstion

ler0y:
Ogg files with Opus audio and .opus extension were not used in UMS.
Now with the opus extension added, these files can be opened.

Test files:
http://xiph.org/~giles/2012/opus/

### DIFF
--- a/src/main/java/net/pms/formats/OGG.java
+++ b/src/main/java/net/pms/formats/OGG.java
@@ -53,6 +53,7 @@ public class OGG extends MP3 {
 			"mpc",
 			"ogg",
 			"oma",
+			"opus",
 			"ra",
 			"shn",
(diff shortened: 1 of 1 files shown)


Nadahar:
@ler0y I have already added a lot of audio formats, including opus and adts in #589. I have also restructured the audio formats quite a bit, so audio recognition is much better in general. The problem was that the PL was too complex, so I'm working on dividing the PL into smaller pieces. The problem is that the different parts depend on eachother, so I have to make PL's in the right order and then wait for them to be merged before I do the next. Therefore the process can take some time, but I dare to say that once my audio format changes are merged, audio support will improve significantly.

Nadahar:
Closing as this is covered in #735

PR #617, opened 2015-07-25
Audio Selection: More Options and all Album sorts by Track

Executioner0:
There is also a change in here commenting out -nosync for ISO container as it caused Audio sync problems when I finally got DVD ISO playing working on my PS3 (PS3 won't play converted ISO with vrc_buf_size=7000, fails with unsupported data).

[Review on Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/617)

PR #615, opened 2015-07-23
Synced other language files with messages.properties

Nadahar:
This is based on 468b0f372d6507137115cfab742e8b2ea5ea0bda (#601). 2b28c3c89b7077f9ccff9f2483dfbf21cf370dc3 (#614) and 23d45fc3367068f282887b4cbbe5179cdf02724f (#611) are also merged and manually synced (that is, the same lines are deleted).

@SubJunk I think it's important to merge this before some more translations are done, as it's getting hard to keep track of.

Unfortunately the diffs are pretty much useless because my eclipse language file editor has rearranged the labels in alphabetical order. Because of that, most of the files shows as changed, but I've only deleted the same lines that was removed in 468b0f372d6507137115cfab742e8b2ea5ea0bda. Once I discovered this, I opened and resaved all language files so that as of this PR they are all sorted the same. That was far from the case, and I figured that it only makes things easier in the future.

@K4r0lSz You should use this version when you insert the polish characters - because this file has about 50 lines less.

PR #622, opened 2015-07-27
Audio Selection: More Options and all Album sorts by Track *rebased*

Nadahar:
I did a rebase of #617 where I removed the google translations as they caused all kind og merge problems because of the recent language cleanup (#601 & #615). In addition I don't believe google translatation is a good idea, as it often is completely off.

[Review on Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/622)


Executioner0:
Thanks @Nadahar, does that mean this is now merged? Is there some way I can (or do I have to) delete my previous Pull Request?

Nadahar:
@Executioner0 No, it's not merged until this (#622) is marked as merged. You can close your request anytime you want on the bottom to the left of the "Comment" button, but I'd let it stay open until a merge is done.

Nadahar:
This PR haven't gotten much attention, the things in here might be worth a look. The change in ```MEncoderVideo``` is particularly interesting as I've read about synchronization issues in the forum that could be caused by this. Any idea why it's there in the first place?

Also, the sorting of the categories in the media library isn't something that usually gets much attention, and it looks to me to be for the better. The fact that it's still mergable after this time, speaks to how seldom this code is touched.

I'm going to fast forward the branch so it can be tested with the current incarnation of UMS.

SubJunk:
@Nadahar I think I get notified when you tag the developers group so no need to tag me separately :)

I don't get sync issues with MEncoder anymore, but then we should hardly be using it anymore. We used to have to use special sauce parameters to get it to work perfectly but over time it became more stable. DVD playback is and has always been occasionally troublesome but mostly works.

The PR looks okay, nice and simple. If it's tested it can be merged

Nadahar:
@SubJunk I haven't tested it, but I can look at it. It's not long since I read on the support forum about missinc sync and DVD ISO.

About the tagging, I'm not so sure. I started doing that because you didn't seem to notice when I only tagged the dev group, and when I look at its members you're not in it. Can't you just add yourself to the group, and I can really stop?

SubJunk:
I get notified about every post on this org regardless of who gets tagged

PR #624, opened 2015-07-28
Created LanguageFileFormatter

Nadahar:
This isn't a ready pull request by any means, but I need to ask some questions.

After all the merging and working with the language files, I came to the conclusion that there's just too many variables that gives wrong diffs. To remedy this I've thrown together a standalone "tool" called LanguageFileFormatter. Error handling, overwrite functionality etc isn't done yet, but the core logic do work. The thought is that it could be run on any language file after manipulation so that we're certain that the files are formatted the same way.

What the tool does is:
- It's hardcoded to the `src/main/resources/i18n/` folder so it won't mess with other files. 
- It's hardcoded to only work with files starting with `messages` and ending with `.properties`.
- It can sort the file by `group`, `name`. If `name` is numerical for the given group, it will sort numerical, so that 9 < 10. Otherwise it will sort `name` alphanumerical.
- It can indent the file so that every group is indented equal among it's members with minimum one space on either side of the equal sign. If you look at the `messages_pli.properties` in the diff of this PR, you can see how that looks.
- It can unindent the file so that no spaces exist on either side of the equal sign. If you look at the `messages_plu.properties` in the diff of this PR, you can see how that looks.
- It will automaticly convert any non latin character in the file to `\uXXXX` representation, so there's no need to use other tools for that.
- It will automaticly change `\uXXXX` to `\uxxxx` for consistency.
- It will automaticly change the line ending to `\n` only (as opposed to `\r\n` or `\r`)..
- It's command line only, making a GUI is more work than to make the formatter itself.

First of all, do you think it's a good idea? Second, if so, where should I put it? I just made a new package "net.pms.tools" now where I put the class. I'm guessing that is wrong. I've tried to read up on Maven and it's structure, but I havn't been successful. The idea is that this should not be packaged with the binaries, and I'm not even sure if it should be compiled. In short, I need help on where to put it and how to configure it. It should be done in such a way that translaters cloning the repo should have easy access to it - or maybe it should be compiled and packaged separatly for download on the forum?


Nadahar:
This is no longer relevant.

PR #625, opened 2015-07-29
Minor improvement: Logged information if help file is not found.

Nadahar:
In some cases (Java 6 at least) an exception will be thrown if the help file is not found. In Java 7 it seems to be just silently ignored. In any case, I find it useful that this is mentioned in the log (since the help tab will be blank).


SubJunk:
Thanks!

PR #629, opened 2015-08-02
Uninstaller fixes

ler0y:
After uninstalling UMS 5.2.0
Some renderer config files were still remaining in the renderers folder:

DLink-DSM510.conf
Samsung-ES6100.conf
Sony-PlayStation4.conf
Thomson-U3.conf
Yamaha-RN500.conf

Some config files were not added, and others had the wrong name according to the recent renderer config files.


Nadahar:
@SubJunk @SharkHunter @valib @skeptical Isn't there a better way to handle this than to have to constantly keep the nsis scipts updated with every included file. What about [this] (http://nsis.sourceforge.net/Uninstall_only_installed_files) for example, could that be a good starting point for an automated system that both backups any files it replaces and undoes any changes it has done when uninstalling?

SubJunk:
@Nadahar yes, automating it would be great

SubJunk:
@ler0y thanks!

PR #632, opened 2015-08-04
Bugfix for faking file extension when transcoding

Nadahar:
@SubJunk @valib 3b25f7a686381b644761ba5205940c65174baa8e didn't take into account that not all files are videos, and changed the extension to the renderer's video transcoding extension also for e.g audio files. This broke audio transcoding, as my two audio renderers wouldn't even try to play the audio files ending with .mpg. 

This is not quite ready to merge though, because I need your input on which engines to include for video and audio files. As there is no "good" way to determine if a file is a video or audio file in UMS (Format simply isn't enough as several formats can be both), I decided to separate them by transcoding/muxing engine. I don't know much about them though, as I have only used FFmpegAudio, so I need you to tell me which ones should use a faked extension for video and audio files, and where to place (if at all) ```TsMuxeRAudio, FFmpegDVRMSRemux, RAWThumbnailer```. I'm also unsure if the "web" formats should have the faked extension as I don't know what they do, but it's likely that they should.

This should make it before release as it breaks audio transcoding, so if you get back to me on the engines fast, I'll adjust them quickly.

### DIFF
--- a/src/main/java/net/pms/dlna/DLNAResource.java
+++ b/src/main/java/net/pms/dlna/DLNAResource.java
@@ -1222,7 +1222,7 @@ public abstract class DLNAResource extends HTTPResource implements Cloneable, Ru
 			if (forced) {
 				// This seems to follow the same code path as the else below in the case of MapFile, because
 				// refreshChildren calls shouldRefresh -> isRefreshNeeded -> doRefreshChildren, which is what happens below
-				// (refreshChildren is not overridden in MapFile) 
+				// (refreshChildren is not overridden in MapFile)
 				if (refreshChildren(searchStr)) {
 					notifyRefresh();
 				}
@@ -2388,7 +2388,7 @@ public abstract class DLNAResource extends HTTPResource implements Cloneable, Ru
 						} else {
 							addAttribute(sb, "resolution", media.getResolution());
 						}
-						
+
 					}
 
 					addAttribute(sb, "bitrate", media.getRealVideoBitrate());
@@ -2469,13 +2469,24 @@ public abstract class DLNAResource extends HTTPResource implements Cloneable, Ru
 				endTag(sb);
 				// Add transcoded format extension to the output stream URL.
 				String transcodedExtension = "";
-				if (player != null) {
-					if (mediaRenderer.isTranscodeToMPEGPSMPEG2AC3()) {
-						transcodedExtension = "_transcoded_to.mpg";
-					} else if (mediaRenderer.isTranscodeToMPEGTS()) {
-						transcodedExtension = "_transcoded_to.ts";
-					} else if (mediaRenderer.isTranscodeToWMV() && !xbox360) {
-						transcodedExtension = "_transcoded_to.wmv";
+				if (player != null && media != null) {
+					// Note: Can't use instanceof below because the audio classes inherit the corresponding video class
+					if (media.isVideo()) {
+						if (mediaRenderer.isTranscodeToMPEGTS()) {
+							transcodedExtension = "_transcoded_to.ts";
+						} else if (mediaRenderer.isTranscodeToWMV() && !xbox360) {
+							transcodedExtension = "_transcoded_to.wmv";
+						} else {
+							transcodedExtension = "_transcoded_to.mpg";
+						}
+					} else if (media.isAudio()) {
+						if (mediaRenderer.isTranscodeToMP3()) {
+							transcodedExtension = "_transcoded_to.mp3";
+						} else if (mediaRenderer.isTranscodeToWAV()) {
+							transcodedExtension = "_transcoded_to.wav";
+						} else {
+							transcodedExtension = "_transcoded_to.pcm";
+						}
 					}
 				}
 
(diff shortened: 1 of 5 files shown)


SubJunk:
Okay time for a bit of a rant:

This is why I did a localized commit in the first place - it only affected PS4. The whole reason for having UMS instead of PMS is to avoid breaking features and one good way to avoid that is to make specific changes instead of sweeping ones.

I don't have time right now to go through this but will check it out soon

SubJunk:
Another note is that unit/integration tests could have easily picked up this bug so I am even more motivated to do more automated testing, especially around things like this which our tests hardly touch.

skeptical:
@Nadahar like @SubJunk I'm pressed for time these days (disclaimer: this is just untested off the top of my head :-), but what about using mimetype as the determinant for fake extension instead, something like this: make a note of mimetype at [DLNAResource.java#L2353](https://github.com/Nadahar/UniversalMediaServer/blob/1915e9a9ade07ee6bd3b686370f04752de8b7c47/src/main/java/net/pms/dlna/DLNAResource.java#L2353):
```java
				String mimetype = getRendererMimeType(mediaRenderer);
				String tempString = "http-get:*:" + mimetype + ...
```
and use it in the ```transcodedExtension``` block
```java
				if (mimetype.startsWith("video/")) {
					// ...
				else if (mimetype.startsWith("audio/")) {
					// ...
```
Using player as the determinant is potentially problematic because plugins are free to add players (and do, e.g. jumpy, channel), so we don't know enough beforehand to hardcode by player class as proposed here.

Nadahar:
@skeptical Always the plugins... :disappointed_relieved:

I don't think mime type is suitable either, as it has the same basic problem as UMS has in general: The logic from the beginning has been to use format (that is: extension) to separate between video, audio and image files. Some formats (container formats) can be more than one type, and thus the logics that's used to set mimetype is flawed as well. The reason I'm so aware of this is form working with audio support, and I've met this problem numerous times. Ambiguous files are always classified as video files, and thus they fail as audio files.

I've tried another approach now (I've updated this PR) that I've had in the back of my mind for a long time, that is to try to define, once and for all, what is an audio, video or image file. Once this definition is correct, that information can be used here and other places where that information is needed. I've made the definition in ```DLNAMediaInfo``` with a number of new "is" or "has" methods, and I've had to gather some extra information from the media parser to be able to define it. 

From my testing it seems to work fine, but I must admit that the parserV1 is a bit illusive to me and I'm not sure if I've defined how to add a videotrack correctly. My tests however indicate that it works as intended, but there are so many cases to test that it's hard to be certain. The parserV2 implementation is pretty safe though.

I've also reversed the changes 3b25f7a686381b644761ba5205940c65174baa8e did to ```RendererConfiguration ```. I'm not 100% sure why this was changed in the first place, but it breaks the "default" transcoding (if ```TranscodeVideo``` is undefined or illegal in renderer configuration). I've adjusted the changes in ```DLNAResource``` to also treat the default as the default both for audio and video.

src/main/java/net/pms/dlna/DLNAResource.java
```diff
@@ -2469,13 +2469,24 @@ public final String getDidlString(RendererConfiguration mediaRenderer) {
...
-					if (mediaRenderer.isTranscodeToMPEGPSMPEG2AC3()) {
-						transcodedExtension = "_transcoded_to.mpg";
-					} else if (mediaRenderer.isTranscodeToMPEGTS()) {
-						transcodedExtension = "_transcoded_to.ts";
-					} else if (mediaRenderer.isTranscodeToWMV() && !xbox360) {
-						transcodedExtension = "_transcoded_to.wmv";
+				if (player != null && media != null) {
+					// Note: Can't use instanceof below because the audio classes inherit the corresponding video class
+					if (media.isVideo()) {
+						if (mediaRenderer.isTranscodeToMPEGTS()) {
```
valib: @Nadahar it doesn't work when the cache is enabled. It works only for the first time when a media is parsed but the next time the video is not recognized and the media is treated as an audio.

Nadahar: Ah, forgot about the cache. It's easy enough to store those two int fields there, but that means altering the table definition for table "FILES". Is that controversial?

Nadahar:
@valib It wasn't much job, so I just did it - this PR is updated. I added two fields to the ```FILES``` table in the database: ```VIDEOTRACKCOUNT``` and ```IMAGECOUNT```, and bumped the database version to 5. I also edited the ```SELECT``` (read) and ```INSERT``` (write) methods to include the fields, and from what I can see that works as expected.

@SubJunk You mentioned further up that the issue was only for PS4. That's not strictly the case, but the PS4 was the most pressing issue for sure. I've had multiple problems with audio playback caused by wrong extensions, and I think it must be said to be "good behaviour" to serve an extension matching the format. My guess is that doing this can help reduce "random errors" and "strange behaviour" in general.

If you don't want this way of determining what kind of file it is, I guess the only logical way is to fall back to the format ```type```. There are already methods for checking by format if a file is audio or video, even though it's not always correct.

SubJunk:
@Nadahar this is looking good, thanks. Well I think it can be a good idea to roll things out for one renderer first and then expand it to others after stability is confirmed in the wild. But now we have released the unstable thing so we should apply this fix :)

Issue #633, opened 2015-08-05
Yamaha renderers GUI "Select renderers" bug


ler0y:
The Yamaha renderers are not listed in the GUI "Select renderers" option.
With default UMS settings (all renderers enabled) the Yamaha renderers are loaded and used as normal:

DEBUG 19:17:27.736 [main] Added file watch at renderers: renderers\\Yamaha-RN500.conf
DEBUG 19:17:27.738 [main] Added file watch at renderers: renderers\\Yamaha-RXA1010.conf
DEBUG 19:17:27.739 [main] Added file watch at renderers: renderers\\Yamaha-RXV3900.conf
DEBUG 19:17:27.740 [main] Added file watch at renderers: renderers\\Yamaha-RXV671.conf

But if for example the Xbox360 renderer is disabled in the GUI, than the Xbox360 renderer is ignored,
but also the Yamaha renderers be ignored:

DEBUG 19:35:40.703 [main] Added file watch at renderers: renderers\\Microsoft-Xbox360.conf
DEBUG 19:35:40.703 [main] Ignored Xbox 360 configuration
DEBUG 19:35:40.871 [main] Added file watch at renderers: renderers\\Yamaha-RN500.conf
DEBUG 19:35:40.871 [main] Ignored Yamaha R-N500 configuration
DEBUG 19:35:40.872 [main] Added file watch at renderers: renderers\\Yamaha-RXA1010.conf
DEBUG 19:35:40.872 [main] Ignored Yamaha RX-A1010 configuration
DEBUG 19:35:40.878 [main] Added file watch at renderers: renderers\\Yamaha-RXV3900.conf
DEBUG 19:35:40.879 [main] Ignored Yamaha RX-V3900 configuration
DEBUG 19:35:40.880 [main] Added file watch at renderers: renderers\\Yamaha-RXV671.conf
DEBUG 19:35:40.880 [main] Ignored Yamaha RX-V671 configuration

After some testing, I found some workaround to this issue which hopefully can give you guys more insight about this.

It seems that UMS does not like the "Y" character in the naming Yamaha, as renderer name.
If in the Yamaha config files the RendererName is changed from "Yamaha" to "Jamaha" like:

Yamaha-RN500.conf:
#RendererName = Yamaha R-N500
RendererName = Jamaha R-N500

Yamaha-RXA1010.conf:
#RendererName = Yamaha RX-A1010
RendererName = Jamaha RX-A1010

[...]

PR #635, opened 2015-08-07
Fixes to SelectRenderers inspired by #633

Nadahar:
When @valib posted 1182f560e534e1f6e40eb9a17e33cfa3f5e6d55e I was already looking at it and had discovered an additional problem: The logic building the groups relied on the incoming data being sorted. I always find that a bit scary and had written a different logic that wasn't hurt by it being unsorted.

Further testing revealed a couple of other issues as well:
* When selecting only some members of a group, the next time one opened the Select Renderers window that whole group would be cleared.
* If selecting no renderers (from some strange reason), the effect would be that all renderers would be selected (because the default, all renderers, would kick in).

Further logics rewrite have fixed that too. To try to move some "low level" logic out of ```SelectRenderers``` I created a new class ```SearchableMutableTreeNode``` extending ```DefaultMutableTreeNode``` that can be reused for similar tasks. The class is general except one "special logic" (controlled by a parameter) for handling search on both the current node and the parent node simultaneously (because names are split by the first space in the renderer name). I don't like this, but found it to be to most practical approach.

I also added "Apple" to the name of AirPlayer and iDevice renderers, so that they are grouped under "Apple" and documented ```Selected renderers``` in ```UMS.conf```.

### DIFF
--- a/src/main/java/net/pms/newgui/SelectRenderers.java
+++ b/src/main/java/net/pms/newgui/SelectRenderers.java
@@ -1,7 +1,7 @@
 /*
  * Universal Media Server, for streaming any medias to DLNA
  * compatible renderers based on the http://www.ps3mediaserver.org.
- * Copyright (C) 2012  UMS developers.
+ * Copyright (C) 2012 UMS developers.
  *
  * This program is a free software; you can redistribute it and/or
  * modify it under the terms of the GNU General Public License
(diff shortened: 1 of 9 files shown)


valib:
@Nadahar there are missing one word renderer names like DirecTV, Kodi, WMP etc.

src/main/java/net/pms/newgui/SelectRenderers.java
```diff
@@ -52,45 +58,41 @@ public void build() {
...
-					renderersGroup.add(new DefaultMutableTreeNode(renderer.substring(renderer.indexOf(" "))));
-				} else {
-					renderersGroup.add(new DefaultMutableTreeNode(renderer));
-				}
+		allRenderers = new SearchableMutableTreeNode(Messages.getString("GeneralTab.13"));
 
-			} else {
-				if (!firstLoop) {
-					allRenderers.add(renderersGroup);
+		Pattern pattern = Pattern.compile("^\\s*([^\\s]*) ?([^\\s].*?)?\\s*$");
```
Nadahar: @valib Aargh! The original, tested regex was ```^\s*([^\s]*)\s*([^\s].*?)?\s*$```. During "final touches" I changed this to ```^\s*([^\s]*) ([^\s].*?)?\s*$``` to force a single space to comply with the "grouping" rules. I assumed this to be safe, and did not retest. It obviously isn't as it requires a space to get a match at all. The correct regex is ```^\s*([^\s]*) ?([^\s].*?)?\s*$```.

Nadahar:
I've updated to PR with the corrected regex, and I've added the tests. As I assumed the last minute change to the regex to be safe, I assumes that this wouldn't affect tests and didn't try to build.

Note to self: **NEVER** assume (anything to be safe)!

SubJunk:
@Nadahar I did a bit of testing on this and it looks good. Do you think it's ready to merge? What about @UniversalMediaServer/developers ?

Nadahar:
@SubJunk As far as I can tell it's ready.

PR #643, opened 2015-08-15
Minor refinement of PMS.java

Nadahar:
Updated PMS.java to remove warnings from FindBugs. Most is of marginal importance, but there were some synchronization issues that potentially could cause problems.

I also removed some deprecated static thread unsafe variables. I've checked, and they are NOT used by UMS code and not by any of the plugins I've gone through, and the fields have been deprecated since PMS version 1.5. I'd say it should be safe to just drop them now.

The other changes are either syncronization, speed optimalization or charset handling.

### DIFF
--- a/src/main/java/net/pms/PMS.java
+++ b/src/main/java/net/pms/PMS.java
@@ -24,6 +24,7 @@ import com.sun.net.httpserver.HttpServer;
 import java.awt.*;
 import java.io.*;
 import java.net.BindException;
+import java.nio.charset.StandardCharsets;
 import java.text.SimpleDateFormat;
 import java.util.*;
 import java.util.List;
@@ -135,18 +136,6 @@ public class PMS {
 	 * Pointer to a running PMS server.
 	 */
 	private static PMS instance = null;
-
-	/**
-	 * @deprecated This field is not used and will be removed in the future.
-	 */
-	@Deprecated
-	public final static SimpleDateFormat sdfDate = new SimpleDateFormat("HH:mm:ss.SSS", Locale.US);
-
-	/**
-	 * @deprecated This field is not used and will be removed in the future.
-	 */
-	@Deprecated
-	public final static SimpleDateFormat sdfHour = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss", Locale.US);
 
 	/**
 	 * Array of {@link net.pms.configuration.RendererConfiguration} that have been found by PMS.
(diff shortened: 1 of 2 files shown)


src/main/java/net/pms/PMS.java
```diff
@@ -366,7 +355,11 @@ private void displayBanner() throws IOException {
 		 * usually done by the installer
 		 */
 		File dDir = new File(configuration.getDataDir());
-		dDir.mkdirs();
+		if (!dDir.exists()) {
```
SubJunk: @Nadahar I assume there's a reason for the if statements being on two separate lines instead of one statement, what is it?

Nadahar: @SubJunk There's a reason but that has nothing to do with the final result and everything to do with how I ended up with that solution :wink: 

The "requirement" from FindBugs was that the value from .mkdirs wasn't ignored, so I added a log message if it didn't work. I then testen and found out that it "failed" if the folders already existed, and thus I wrapped it with the first ```if```. 

I'll fix :innocent:

Nadahar:
@SubJunk It's corrected.

SubJunk:
@Nadahar cool, thanks :) by the way, I've talked with the other developers and we have been impressed with your contributions to UMS, we would like to invite you to join the dev team.

It wouldn't really change much - we still try to use PRs for significant changes, although simple/stable bugfixes can go straight to master, but there is no added pressure to contribute code.

What do you think?

Nadahar:
Sure, I'd be honored ;) 

I must admit that when I downloaded the source my intention was simply to fix the XML encoding problem. I had no intention to get this involved, but I haven't really done much coding the last couple of years, and once I started I just couldn't stop. Plans and ideas pop up "by themselves", so my conclusion is simply that I must miss it much more than I realized. What's better than to spend such a "drive" on an open source project? I can't promise that the drive will stay at the same level all the time, but I generally like to do some coding now and then.

skeptical:
@Nadahar :+1: welcome

Optimusprimeums:
@Nadahar Welcome to the Team

SubJunk:
@Nadahar that's great news, and no problem if your drive wanes, we all have times like that too. I've added you to the project on GitHub now, and have made an announcement post on our forum and social media accounts. If you have/make a forum account I can set it to developer status too

Nadahar:
Thank you all :smile: 

I already have a forum account, but it's my "real" nick Nadar - which was occupied on Github.

PR #616, opened 2015-07-25
Logging system overhaul

Nadahar:
@SubJunk @valib @SharkHunter @skeptical 
First of all, I know this is big and somewhat complex, but it would be almost impossible to make changes like this with many small commits, and it would if nothing else have taken ages. This is confined to the logging system only. I hope you will take the time to pull and test/verify it and then discuss anything that is unclear or that you disagree with. I have tested everything to the best of my ability, tracing code while testing out the different scenarios I could think of. I believe there are many useful additions/improvements here.

* Renamed ```LoggingConfigFileLoader``` to ```LoggingConfig``` and extended it into a utility class for logging configuration:
  * Added support for prioritized multiple logback configuration file locations similar to that of renderer configurations. UMS will now search first in the users profile folder, then in working folder and finally in src/main/external-resources/ . This is defined in ```project.properties```: ```project.logback=[PROFILE_DIR]/logback.xml, logback.xml, src/main/external-resources/logback.xml```. This enables users to have a custom ```logback.xml``` that isn't overwritten each time UMS is upgraded, and it also enables the files to be found while running from an IDE.
  * Added support for buffered file logging and syslog logging instead of file logging without writing a custom logback configuration file. Both options are first and foremost made with SSD's in mind, trying to limit the disk writes to prolong SSD life. All options are configured in UMF.conf or the GUI, description is created in UMF.conf. Both are aware of ```tracemode == 2``` or "forced trace", so that both syslog and buffered logging is turned off when UMS is started with ```trace``` argument.
  * Added configurable filters to both the GUI log window/"traces" and the console. These used to be set to ```info``` in the logback configuration thus being "invisible" to anyone not into logback. The filtering in the logback configuration is removed, and instead UMS inserts filters on the fly during startup. The filters are configurable in UMS.conf, and they are visible so that users can know that the filtering takes place. The GUI filter can also be configured from the GUI, which makes it possible to see debug and trace messages from there.
  * Parameterized logfile name and removed all hardcoded references to "debug.log" except for one instance in some web stuff that I haven't studied enough to change. The logfile name can be configured in UMS.conf, and the default is "debug.log". The main reason for this is not because it is important that users can change the log file name, but to get better control over what the name actually is. This used to depend entirely on the logback config, and if a user changed this it would fail many places in the code that was hardcoded to that filename.
  * Added log messages if things go wrong various places in ```LoggingConfig```.
* DbgPacker:
  * Improved error handling if OS can't open file or file type.
[...]

PR #660, opened 2015-08-31
A different solution for wrappable CustomJLabel

Nadahar:
@skeptical 

I had to try to find a non-html way to do this, and after a lot of complicated tracing in swing and paint methods etc. -> etc. which led nowhere, I decided to check out how ```JTextArea``` did it. It's actually mindbogglingly simple, yet it seems to work. I haven't done any extensive testing, and I haven't looked at the Checkbox situation, but I bet it can be solved the same way.


skeptical:
@Nadahar great :+1:. More direct and avoids potential parsing/syntax pitfalls. It's probably more or less what's happening under the hood for wrapping in ```<html>``` anyway.

skeptical:
@Nadahar Actually on looking deeper this code isn't actually doing anything, the wrapping you see is via ```CustomPanelBuilder._addLabel()```, which is still calling ```htmlify()```.

skeptical:
@Nadahar no, ```ProcessBuilder.addLabel()``` is declared ```final``` so it can't be overriden to use ```CustomJLabel```, hence ```_addLabel()``` hack.

Nadahar:
@skeptical ```final``` is pure evil in Java, yet it's encouraged for immutability/thread safety. It can really make one wonder what or if they were thinking when they designed this. There should rather be a way to enforce or atleast encourage any overridden methods that they have to stay immutable while still maintaining the possibility to overload. That and the very stupid "package private" functionality really hinders any effective subclassing of many objects.

Yet, my code actually does exactly the same as ```JTextArea``` does down to their common anchestor ```JComponent```, so if it doesn't work that means that the difference is in the "renderers". They call different "renderer" classes that actually does the "look and feel", drawing etc, so as far as I can understand that must be where it's not being picked up.

skeptical:
@Nadahar your code seems to be implementing the [Scrollable](http://docs.oracle.com/javase/7/docs/api/javax/swing/Scrollable.html) interface, which has no effect in JLabel
```
All Known Implementing Classes:
DefaultTreeCellEditor.DefaultTextField, JEditorPane, JFormattedTextField, JLayer, JList, JPasswordField, JTable, JTextArea, JTextComponent, JTextField, JTextPane, JTree
```

skeptical:
@Nadahar btw if you're in the mood for more deep diving (and a huge amount of refactoring), would be interesting to switch from ```JTextArea``` to ```JEditorPane``` in #616 for syntax highlighting of the log via html with css/regexes borrowed from the web interface logviewer. One issue is that you would need to append/remove lines correctly in html, but it seems doable in [HTMLDocument](http://docs.oracle.com/javase/7/docs/api/javax/swing/text/html/HTMLDocument.html)

skeptical:
@Nadahar ```JEditorPane``` doesn't seem to manage the css (supports html 3):
jeditorpane
but ```JavaFX``` (java 8) can do it :-):
javafx

Nadahar:
@skeptical I've looked at it a bit, and as always I'm reluctant to use HTML since I find it too "random" for my taste. Both ```JEditorPane``` and ```JTextPane``` can have multiple text colors one one line from what I can see, so it should be possitble to implement. I was wondering if RTF (which is also supported) would create less problems than HTML, but that seems to be a dead end with similar parsing problems as HTML (even though you have more formatting control) and NO UTF-8 support. UTF-8 or UTF-16 is a must imo, working with codepages or unicode encoding is just something I'm not willing to do anymore.

Adding and deleting lines seems to be quite manageable, what really worries me is how I'd be able to do the searching. I could maintain one plantext and one HTML version of the document and use the plaintext for searching, but it would waste a lot of memory and resources and probably create some deadaces when it comes to highlighting the found text.

IDE's like Eclipse must have solved this exact issue, so I guess I could start by looking at their solution, but I'm afraid it could be a rather large project and could take a lot of time.

While I like the idea of syntax highlighting, I'm wondering if that should be a separate project from #616. It's getting complicated and teadious keeping the PR and the Java 6 and Java 8 versions synchonized with master, and I need the changes to language/locale done in #616 for the Crowdin intergration, so I'm wondering if it would be better to just get #616 and its siblings merged and then take it from there. Personally I'd like to press on with the Crowdin integration as soon as possible, but I don't fancy doing almost exactly the same job with languages/locale separatly for this, and then have another merge nightmare for #616 later.

Nadahar:
@skeptical Regarding wrappable ```JCustomLabel``` I still don't like going into the HTML minefield, but no other solution can be found (I've tried some more, but with no results this far), I already know two things that must be addressed:

* HTML encoding of at the very minimum ```<```, ```>``` and ```&```.
* A trailing space is used as a separator between the label and the field in UMS many places. As HTML trims text, those are not displayed anymore. The solution is probably to add a dummy object like an ```EmptyBorder``` as a separator instead (and it's a more proper solution anyway).

PR #661, opened 2015-09-02
Comparing computer clock with a NTP server during startup

Nadahar:
@UniversalMediaServer/developers Based on [this forum thread](http://www.universalmediaserver.com/forum/viewtopic.php?f=14&t=3940&sid=f9091a23bf11166201425f4bfec8154f) I created this as a suggested solution.

There are some issues though:
- Before I understood the now solved bug, I started patching potential NPE's in `UPNPControl` and `UPNPHelper` (since they were the sources of the NPE's). I've left them in here for now, but I'm not really sure if these "fixes" are relevant to keep or should just be discarded.
- We have to decide a time difference threshold for when UMS generates a warning. It's currently set to 5 seconds, but that's just to make testing easy.
- What if the renderer's clock is off? The issue should probably be handled on a different level, does uPnP or DLNA have some mechanizm to handle this that's not being properly employed by UMS?
- `org.apache.commons.net` is added to dependencies to provide access to `org.apache.commons.net.ntp`. I'm unsure if this poses a problem or not.
- I'm unsure of my english wording on this issue, what's the correct terms (too slow, too late, running late etc.)?
- ~~There is a major bug somewhere that means that the dialog created by this sometimes makes UMS create two GUI's. When this happens NPE's start raining in from all over the place. I haven't been able to figure out why this happens, and it's not every time - which indicates some kind of a synchronization issue. I had to do this in a separate thread since the process of checking the NTP server can take anything from a few milliseconds to a minute or more - depending on network connectivity and NTP server response times, and the dialog is spawned from the thread. It tried it first with a callback to a static synchronized `PMS` method, but that obviously made no sense and gave the same results except that it also blocked the `main` thread until the dialog was closed.~~

[Review on Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/661)


Nadahar:
@skeptical It's now implemented with SwingUtilities.invokeLater(), the result is the same :confused: 

@SubJunk I didn't realize that you're not a part of the ```UniversalMediaServer/developers``` group, but now you're mentioned here as well :wink:

Nadahar:
It seems it the line ```SwingUtilities.getWindowAncestor((Component) PMS.get().getFrame())``` that's creating the problems. If I replace that with null, the bug is gone (but so are the look and feel of the dialog).

skeptical:
@Nadahar the check may be finishing too soon, try launching it later on (towards end of ```PMS.init()```).

Optimusprimeums:
@Nadahar can you input your own NTP server to check? If not could we add it to the conf file to enable users (and me) to use a local NTP Server (I run a domain and have a master NTP server that all machines call for time)

Nadahar:
@Optimusprimeums Sure, that's no problem to make - provided this whole idea is worthwhile. I'm still not really sure if this solves that much...

Optimusprimeums:
@Nadahar I think its a solid idea, like you said only if it solves some issues

skeptical:
@Nadahar Afaict the bug reporter didn't verify his own hypothesis, i.e. would be good to double-check by intentionally changing system time and see if the problem reappears on the PS3, or other renderers, or are you way ahead of me there too :-)?

Nadahar:
@skeptical I've only concentrated on the NTP check, and I've looked a little bit in the DLNA spec and found that there's methods for comparing clock sources. More reading in the spec is needed, but I think the most useful thing would be to compare the renderer's clock with the server's. 

The biggest problem however is that I don't really understand the cause of the problem (exactly what goes wrong when clocks are out of sync), and thus it's very hard to judge how much out of sync they must be for it to pose a problem.

Nadahar:
I think I've fixed the "strange" bug. You were partly correct @skeptical that it finished too soon, but it turns out if also started to soon, before the PMS instance was created. That means that it sometimes created a PMS instance via ```PMS.get()```, and then main created another one or something like that. Looking at ```PMS.createInstance()``` that shouldn't be possible (because of the ```assert```), but something strange happened.

I created a waiting logic in the thread that waits until ```LooksFrame``` is initialized before queuing the dialog, but there was still some trange behaviour until I moved the start call to after ```PMS.createInstance()```. In any case that bug seems dealth with.

I'll ask the forum thread poster if he can try to replicate the problem to shed some more light on what is happening.

And btw, thanks for the tip about [SwingUtilities.invokeLater](http://docs.oracle.com/javase/7/docs/api/javax/swing/SwingUtilities.html#invokeLater(java.lang.Runnable)). I wasn't aware of this arrangement, and I think that's also the key to another bug I've been tinkering with (the magic progressbar on the status page).

Nadahar:
@skeptical I still occasionally get an NPE during shutdown in this PR. It's always from ```p.exitValue();``` in the ```PMS Listeners Stopper``` shutdown hook. Do I have to do any cleanup after the runnable has finished ```run()```? Won't the thread cleaned up automatically then?

Nadahar:
@skeptical The tip about [CountDownLatch] (http://docs.oracle.com/javase/7/docs/api/java/util/concurrent/CountDownLatch.html) was great, thanks :+1: 

I think I have fixed the shutdown NPE too. It seems it didn't have to do with this PR, but the way I tested this PR (starting UMS, clicking the dialog and then immediately quitting). I added synchronization to PMS.currentProcess, and that seems to have taken care of that problem. My guess is that I quit while things were still in "startup", and that some of the processes fetched by the ```ShutdownHook``` was being removed/destroyed before it tried to access them. I didn't think that was possible with Java's reference counting, but strange things can happen when threads are unsynchronized.

skeptical:
@Nadahar I'd call the latch ```guiReady``` or something like that, since it can be used just as well by any other concurrent process that needs it.

Re. shudown NPE, I often see it when I ctrl-c my ums process from a console, it's never bothered me though :).

Nadahar:
@skeptical 
>  I'd call the latch guiReady or something like that, since it can be used just as well by any other concurrent process that needs it.

I agree - didn't think that far. Done.

> Re shudown NPE, I often see it when I ctrl-c my ums process from a console, it's never bothered me though :).

In my book every unhandled exception is an error. Why even write a ```ShutdownHook``` if it doesn't matter if it crashes or not :wink: 

I'm wondering if you could take a look at the changes I did to ```UPNPHelper``` and ```UPNPControl``` in this branch. Is any of them even remotely useful? I wrote them as a consequence of the GUI synchronization bug because I got NPE's there, but that was pretty artificial circumstances. As these are "your" code, you're probably the one that can best judge if any of them can have any real life value. If not I'll just discard them.

skeptical:
@Nadahar I personally would prefer to see a lot of NPEs etc in oddball cases like this, since it grabs one's attention and maybe even narrows the scope for us if we're lucky.  We should catch/hide/minimize expected and harmless stuff, but let it bleed profusely otherwise imo. I've never seen those NPEs in 'ordinary' use, they would indicate something is seriously wrong.

Another slightly related side note: like you I often automatically start throwing around ```volatile``` and ```synchronized``` everywhere before I discover the true problem, which 99% of the time isn't really related to synchronicity.  My personal policy is (usually) to remove these, even if I think there's a hypothetical/potential necessity, untiI someone actually experiences a problem.

Also part 3, if you permit a little friendly advice.  Often your commits are a chronicle of the session that produced them, i.e. a circumstantial mix of changes (with an impressively meticulous changelog attached).  Imo it's better to go through the exercise of chopping things up and committing them separately, i.e. as they should have happened in an ideal session rather than as they actually happened in real life.  My personal way (just as an example, yours would no doubt be different, plus I don't mean to imply that you don't already do tons of similar self-editing, you're hyper-aware of everything afaict :-) is something like this:
```sh
# hack hack hack for minutes/hours/days/weeks till ready to commit.
# At this point my diff is an incoherent mess to anyone but me,
# so I take a look at it
git diff > tmp.diff

# and chop it up by hand into separate logical chunks
# e.g. a.diff b.diff c.diff e.diff etc
# throwing out any junk, and then rewrite history
git reset --hard

# commit the new stuff as if it happened logically
patch -p1 < a.diff
git commit -a -m "Implement A"
patch -p1 < b.diff
git commit -a -m "Add B"

# c.diff is just a fix for some earlier unpushed commit
# so go back and correct it now
git rebase -i 123ABC4^
patch -p1 < c.diff
git commit -a --amend
# fixing any conflicts as they appear
git rebase --continue

# d.diff is a general fix, or a separate topic
# so patch/commit in another branch as appropriate

# e.diff relates to something of possible interest/use but not
# strictly speaking required/relevant right now, so give it a
# better name and save it for later
```
A lot of this can also be done via ```git stash```, ```git cherry-pick```, squashing intermediate commits, etc if one is good at mental juggling :-).

Nadahar:
> I personally would prefer to see a lot of NPEs etc in oddball cases like this, since it grabs one's attention and maybe even narrows the scope for us if we're lucky. We should catch/hide/minimize expected and harmless stuff, but let it bleed profusely otherwise imo. I've never seen those NPEs in 'ordinary' use, they would indicate something is seriously wrong.

I have a slightly different philosophy: Any error that I can imagine could happen, no matter how remote, should be handled. Things that can't on the other hand shouldn't, because that's only a waste of time/energy/size. I like applications to be robust and handle most things thrown at them, and I think applications that only handle things that go exactly like the author has meant for them to go, is low quality and often borderline useless. The problem is when "patching" code I don't know, partly because I don't have enough overview over what's happening to judge what could and could not happen, and partly because I don't know what is the proper response to an error. Catch and log is appropriate some times, but far from always, some times it should actually be taken actions as a response. These things are impossible to judge in "foreign" code, and that's why I asked you it these had any potential value.

> Another slightly related side note: like you I often automatically start throwing around volatile and synchronized everywhere before I discover the true problem, which 99% of the time isn't really related to synchronicity. My personal policy is (usually) to remove these, even if I think there's a hypothetical/potential necessity, untiI someone actually experiences a problem.

I generally don't commit testing, I wait until I have solved the problem or implemented some isolated goal. I go through the diffs before I commit, and simply reverse any such temporary test in diff view. If I have to save in the middle of such testing to tend to another branch, I use stash. If I for some reason still have had to (or by mistake) commit such code, I'll amend or rebase it away. I don't remove the things I judge to be a theoretical necessity though, because I'm not that fond of debugging that I actually have to experience every pitfall to acknowledge it. As I see it, a bug is still av bug even if current usage pattern doesn't expose it. Future changes can change that, triggering such bugs, and that can be very hard and troublesome to find then. I therefore prefer to "do it right" the first time around.

> Also part 3, if you permit a little friendly advice. Often your commits are a chronicle of the session that produced them, i.e. a circumstantial mix of changes (with an impressively meticulous changelog attached). Imo it's better to go through the exercise of chopping things up and committing them separately, i.e. as they should have happened in an ideal session rather than as they actually happened in real life. My personal way (just as an example, yours would no doubt be different, plus I don't mean to imply that you don't already do tons of similar self-editing, you're hyper-aware of everything afaict :-) is something like this:

Thanks for the tips on rebasing, I mostly use TortoiseGit so I do things a little bit different, but it's good to have the command line to fall back to when TortoiseGit can do what I want. Regarding the commits, it's partly logical and partly effort based evaluation behind how I commit. I generally try to have the code in a "working state" between commits, so any changes that's tied together will be in the same commit. The more complex the changes, the bigger each commit will become. However I will admit that I potentiall have overseen some such "working states" where I could have commited making them unnecessarily big lately, and I'll try to be more aware of that. The thing that also is appealing is that I can ammend the last commit (in TortoiseGit) but have to rebase the eariler ones. That means that I'm reluctant to make a new commit until I'm confident about the previous. Splitting them later is labour intensive if you I to keep the "working state" between commits principle, and it's hard to see that it's worth the effort.

SubJunk:
Feel free to resurrect this

PR #669, opened 2015-09-11
File provider

taconaut:
Even though the idea of a new file provider plugin hasn't been welcomed with open arms (#461) I wanted to see if it would be possible to implement it, as I truly think it could be a great addition to ums. I've taken the comments into account and tried to minimize breaking changes. Giving the default plugins a go, there were no issues with them.

When ums is started without additional plugins, you wont see any diference:
UMS without file provider plugins

As a POC, I've created two dummy plugins, which show a simple Swing UI und show an empty VirtualFolder when browsed on a render. The plugins as well as their sources can be downloaded [here](https://sourceforge.net/projects/ums-mlx/files/temp/FileProviderPlugins/). Drop them in the plugins folder to make them available.
Once they are included a pane to select the active file provider shows up:
UMS with file provider plugins

For the implementation, one of the changes which might have an impact is that getRootFolder in PMS and RendererConfiguration now returns a DLNAResource instead of a RootFolder. Some unfinished code is included, where the RootFolder is being called from different places and has now to be casted e.g. in DLNAMediaDataBase:

	public void stopScanLibrary() {
		if (isScanLibraryRunning()) {
			DLNAResource root = PMS.get().getRootFolder(null);
			if(root instanceof RootFolder){
				((RootFolder)root).stopScan();
			}
		}
	}

This can be solved in a clean way if the idea gets accepted.

With this pull request, the feature is functional but isn't finished. I hope you will give it a try and consider including it in the master.

I'd be happy to elaborate a bit more if you are interested.


Nadahar:
@taconaut I'm new to this, could you give me a short description of what MLX actually does?

taconaut:
@Nadahar I'll give you the description added to the [GitHub wiki](https://github.com/taconaut/ums-mlx/wiki) (which I ment to use to describe the features but haven't found the time to do..). The [thread in the forum](http://www.universalmediaserver.com/forum/viewtopic.php?f=6&t=3355) includes some pictures of the main functionality.

ums-mlx is a modified version of the [Universal Media Server](http://www.universalmediaserver.com/) (UMS). mlx stands for **M**edia **L**ibrary e**X**tension. It has been started to allow more flexibility to configure what shows up on a renderer while keeping the powerful transcoding possibilities it offers.

Its functionality includes:
- Original Universal Media Server features
- Configure what files will show up in a folder by setting conditions (e.g. all action movies with kurt russell)
- Define how files will be sorted in a folder
- Flexible way to define the name of the file which will be shown on the renderer (e.g. '%name (%year) - %rating_percent/100 [%rating_voters]' will show 'The Dark Knight (2008) 89/100 [1480975]')
- Import video properties from [imdb](http://www.imdb.com), [tmdb](http://www.themoviedb.org) and [thetvdb](http://thetvdb.com) (extensible through a plugin system)
- Show files either as single files or with more information in a folder (similar to the transcode folder included in UMS)
- Track the files you've played
- Reworked plugin system allowing full customization

Nadahar:
I know that the will for changes isn't that great here, but personally MLX looks like a much better solution than the current. I haven't look at this in enough depth to really see what it does, but I find the current Media Library to be severly lacking. 

The only thing that worries me is complexity for users. First of all, I think it could be difficult for users to understand the concept of different file providers. Would it be possible for the MLX functionality to scale to different "user levels" so that "basic mode" would result in a structure basicly mirroring the file system and then add manipulation flexibility with more advanced levels? If so, I personally could see it replacing the current system.

What is the link to the current MLX source?

taconaut:
> I know that the will for changes isn't that great here, but personally MLX looks like a much better solution than the current. I haven't look at this in enough depth to really see what it does, but I find the current Media Library to be severly lacking.

mlx is very powerful but with the possibilities, the complexity increases a lot. The original implementation was also lacking a lot of functionality for me, that's why I have started this (about 6 years ago ;).

> The only thing that worries me is complexity for users. First of all, I think it could be difficult for users to understand the concept of different file providers. Would it be possible for the MLX functionality to scale to different "user levels" so that "basic mode" would result in a structure basically mirroring the file system and then add manipulation flexibility with more advanced levels? If so, I personally could see it replacing the current system.

I fully agree for the complexity issue. The other big issue is the zero config ums offers by exposing the entire file system on the renderer. Even though it would easily be possible to do this in mlx, the other possibilities would make it difficult for users used to the current interface to understand how to only share some folders (as they can do now easily). This is why I like the idea of the file provider plugin as the user has to consciously download the plugin and put it into the plugins directory. If he doesn't, he won't notice any changes compared to now.
As I've described in the other pull request, I'd like to provide functionality to configure the DLNA tree while keeping the streaming/transcoding functionalities untouched with this pull request.

> What is the link to the current MLX source?

It's in [my master](https://github.com/taconaut/ums-mlx). Besides a few changes in the base code, everything related to mlx lives in the [net.pms.medialibrary](https://github.com/taconaut/ums-mlx/tree/master/core/src/main/java/net/pms/medialibrary) package.

Nadahar:
> This is why I like the idea of the file provider plugin as the user has to consciously download the plugin and put it into the plugins directory. If he doesn't, he won't notice any changes compared to now.

I think most users won't install any plugins, atleast not anyone that can't be installed from the GUI. The plugin installer in the GUI is also realtively buggy and hard to understand imo. How are normal users to even know about this plugin, not to mention what it does. I think having to install plugins raises the complexity quite a bit too.

taconaut:
Agreed again :) I'm not a big fan of the plugins tab either (the biggest issue being that it only works if run as admin on windows).
An icon :information_source: with a descriptive tooltip linking to a download page could help.

skeptical:
@taconaut I tested a little, some issues:

* Choosing 'Mr Blue' hides any folders/media added by other plugins.  It's hard to imagine that a user would actually want to lose access to their other plugins' content as a consequence of activating Mr. Blue.  It doesn't seem to be undone without a reboot, see following item.

* Each renderer continues to see whatever "root folder" it initially received from the active file provider for the remainder of that UMS session, changing the file provider on the fly doesn't seem to make a corresponding change in what the renderer sees until we reboot UMS.  Rebooting the renderer itself is a mixed bag (for me Kodi can't see anything if it's rebooted after changing file providers, my Sony Bluray on reboot still sees the old view), but the point is it's not an artifact of renderer-caching, since the view is still wrong when the renderer is rebooted.

The second issue is just a matter of further refinement I guess, but the first issue is (as you can imagine :-) very problematic for other plugins.  Any content-providing plugin is essentially a de-facto file-provider, "discovering" its children according to its own internal logic whatever that may be. With ```FileProvider``` we're effectively allowing the internal logic of one plugin to nullify the others.  I can't see how this could be desirable or productive.

I still wonder why Mr. Blue (or an MLX plugin) can't work the same magic acting as a normal plugin, i.e. add a folder to the xmb root (e.g. 'Library') and do wondrous things inside it (and even remove/replace the Navigation tab and the xmb 'shared' folders if it wants to, since that's its intended purpose).  This would be a little friendlier to the other plugins.

taconaut:
@skeptical Thanks for giving it a go :)

Your first issue is intentional. I didn't want to put constraints like that onto a FileProvider. It is easy enough to access the ExternalListeners when creating the RootFolder returned by the FileProvider. With mlx, these plugins are supported too and they can be added anywhere in the sub-tree and not only at root. 

The second issue is a bug. I've only tested this branch through the web interface. There I've noticed that the GlobalId was continuously incremented even after another FileProvider has been activated; this could well be the cause of the issue your describing. But I'm confident this can be fixed :)

> I still wonder why Mr. Blue (or an MLX plugin) can't work the same magic acting as a normal plugin, i.e. add a folder to the xmb root (e.g. 'Library') and do wondrous things inside it (and even remove/replace the Navigation tab and the xmb 'shared' folders if it wants to, since that's its intended purpose). This would be a little friendlier to the other plugins.

It could, but there are different reasons why I prefer the FileProvider approach. One is on the GUI side, that the configuration can be done from the main window and not from behind a click on the plugin button which can be missed. And the more important one is that the entire DLNA tree can be managed without having to remove unwanted nods which are part of the default implementation. E.g. Having checked the iTunes library can result in huge amounts of RAM being used as the resolution is being done on application start and not when it is browsed on the renderer.

skeptical:
@taconaut 

> Your first issue is intentional. I didn't want to put constraints like that onto a FileProvider.

But it seems to me you're instead putting constraints on the other plugins, which now have to worry about what exactly the active file provider is doing (and likely how to get around it :-).  ```FileProvider``` is tailored to the specific issues of library management but imposed as a global constraint on everything, even plugins that don't need or receive any special advantage from its features. What's the upside for them?

> With mlx, these plugins are supported too...

I meant to ask you about that.  When I run mlx I can see via logging that jumpy and channels are loaded as usual, but their content is nowhere to be found on the xmb. What am I missing?

> ...and they can be added anywhere in the sub-tree and not only at root.

Another plugin may have implemented its own solution to this already.  For example jumpy understands linux path syntax and will 'mkdir' any absolute or relative path on the xmb, including top level folders.  Users are free to rearrange/customize (the jumpy-created portions of) their specific tree by editing paths in ```jumpy-scripts.ini```.  I was actually curious to see how this fared under mlx.

> there are different reasons why I prefer the FileProvider approach. One is on the GUI side, that the configuration can be done from the main window and not from behind a click on the plugin button which can be missed.

As mentioned above I personally have no objection to a plugin replacing the Navigation/Share tab and removing the shared folders on the xmb, etc (and would favor implementing whatever changes are required in ums to make that possible).  This is actually a reason not to make FileProvider a globally controlling feature, since the issue itself isn't global but specific to Navigation/Share.

> And the more important one is that the entire DLNA tree can be managed without having to remove unwanted nods which are part of the default implementation.

This is already possible, you can always remove a child from its parent and add it back later, etc, and could do it just within the relevant scope, in this case Navigation/Share. Why should one plugin be granted the ability to manage/gatekeep/control/veto the entire DLNA tree?

SharkHunter:
I said it before and I'll say it again I don't understand the need to change everything. You can make this happen with two lines of code in rootfolder simply add something like
if(some_conf_option)
return mlx_root_folder();

This allows the rest of the world to stay the same and you can do whatever you feel like. What ever you try to do I would almost require a conf option that makes EVERYTHING (including plugins/etc) to work EXACTLY as today.  You're not a fan of the plugin folder which is fine don't use it, I'm not found of this don't force me to use it.

taconaut:
Sorry for the late reply, I've been a bit busy those days.

@skeptical 

> But it seems to me you're instead putting constraints on the other plugins [...] FileProvider is tailored to the specific issues of library management but imposed as a global constraint on everything [...]

I've got the feeling we're not talking about the same thing. What I'm proposing is to decouple the way files and folders are being presented on a renderer from the way it is now. Currently I've got to live with folders (at least one) reflecting my file system which I don't want to be there as they get managed another way.

> I meant to ask you about that. When I run mlx I can see via logging that jumpy and channels are loaded as usual, but their content is nowhere to be found on the xmb. What am I missing?

I would have bet I had implemented that. Checking the source, I've noticed it isn't :/ I've started implementing support for AdditionalFolder(s)AtRoot.

> Another plugin may have implemented its own solution to this already. For example jumpy understands linux path syntax and will 'mkdir' any absolute or relative path on the xmb, including top level folders. Users are free to rearrange/customize (the jumpy-created portions of) their specific tree by editing paths in jumpy-scripts.ini. I was actually curious to see how this fared under mlx.

Ok, we'll see when I've finished the above :)

> As mentioned above I personally have no objection to a plugin replacing the Navigation/Share tab and removing the shared folders on the xmb, etc (and would favor implementing whatever changes are required in ums to make that possible). This is actually a reason not to make FileProvider a globally controlling feature, since the issue itself isn't global but specific to Navigation/Share.

Cool :) In Navigation/Share there are global features like e.g. thumbnail generation that should remain available throughout the application and others like e.g. iTunes support (which IMO should actually be a AdditionalFolderAtRoot) which shouldn't.

> Why should one plugin be granted the ability to manage/gatekeep/control/veto the entire DLNA tree?

I'll quote the description text for mlx to respond to that :) :
[...] It has been started to allow more flexibility to configure what shows up on a renderer while keeping the powerful transcoding possibilities it offers.

@SharkHunter 

> You can make this happen with two lines of code in rootfolder

Including the GUI will add some more lines ;) But why prefer a hard-coded solution where the user has to edit the conf file instead of a generic one where you can do it through the GUI?

> What ever you try to do I would almost require a conf option that makes EVERYTHING (including plugins/etc) to work EXACTLY as today

I've put quite some effort into that. If you don't add a FileProvider plugin into the plugins folder you won't see any difference in the GUI.
The only major difference in code is that the RootFolder is now a DLNAResource and no RootFolder anymore. With the plugins I've tested, none were affected though.

Issue #668, opened 2015-09-07
RunSingleInstance issues


Nadahar:
@UniversalMediaServer/developers @SubJunk 
While testing the isAdmin() under Windows 10 and Linux, I stumbled over something else (as usual). While I think having a mechanizm to ensure only a single instance is smart, especially since UMS claims several ports (http server and web interface that I can think of), I don't really get the implementation:
- Why stop the old process instead of refusing to start the new?
- Why, if we are to stop the old process, do we forcibly terminate it so that is has no chance to close connections, files, the database etc. in a controlled manner. It's asking for corruption of data imo.
- Why the PID file implementation? I know it's a common way to do it under Linux, but placing the .pid file under the users home directory doesn't preven UMS from being started multiple times, it only prevents it being started multiple times by the same user. I will still have the same port conflicts if started by different users. Under Windows it makes even less sense.

I'd recommend that we did a rethink of this, and prevented UMS starting up it an instance is already running in the same OS (that is, independant of by which user) if "single instance" is selected.

I'm not quite sure how to do it under Linux, and I have no idea regarding OS X, but under Windows it seems like the simple way would be to run:

```
tasklist /FI "WINDOWTITLE eq Universal Media Serve*" /V /NH /FO CSV
```

From my testing this seems to work for non admin users as well, so there's no requirement that the user is admin for this to work. If such a process exists, UMS should simply generate a message that it's already running and then die.

The pidfile system under Linux doesn't allow non admin users to do this check, but one might be able to parse `ps` output to do the same. The problem, if I'm not mistaken, is that `ps` syntax isn't standardized under Linux, so it could be hard to find a way to make that work. Another option is to use [D-Bus](https://en.wikipedia.org/wiki/D-Bus), but I'm unsure if this is widely enough adopted.

I have no idea about OS X, but it might be that the Linux solution could also be used on OS X.

What do you guys think?

SharkHunter:
The whole idea of the single instance is to solve a problem that if you restart UMS you're sure that the old instance doesn't hang around. The pid file stuff is standard in Linux (and will work in OSX since it is BSD based) and it is a common way to deal with this problem. The pid file wasn't placed under the users home directory before and the location of the file can be changed but I doubt that is a problem very few people uses this in some multi-user environment so stopping and starting is most likely done from the same user.  

[...]

Issue #640, opened 2015-08-13
Test Windows 10 compatibility


Nadahar:
I doubt there are much changes needed for Windows 10, but if somebody has been crazy enough to upgrade, UMS should be tested. 

I've picked up one place where a change is needed, ```PmsConfiguration.isAdmin()``` (the behaviour is probably the same as for Windows 7 & 8, but it needs testing). There might be other places where it matters as well.

SubJunk:
@Nadahar good point. I think @Optimusprimeums uses it and said it works well with UMS, is that right?

valib:
@SubJunk I am also testing UMS on WIN 10 and it works without any problems.

Nadahar:
@valib You should check out ```PmsConfiguration.isAdmin()``` then and make sure that the admin permissions check works also for Windows 10:
```java
	public boolean isAdmin() {
		if (
			"Windows 8".equals(System.getProperty("os.name")) ||
			"Windows 7".equals(System.getProperty("os.name")) ||
			"Windows Vista".equals(System.getProperty("os.name"))
		) {
			try {
				String command = "reg query \"HKU\\S-1-5-19\"";
				Process p = Runtime.getRuntime().exec(command);
				p.waitFor();
				int exitValue = p.exitValue();

				if (0 == exitValue) {
					return true;
				}

				return false;
			} catch (IOException | InterruptedException e) {
				LOGGER.error("Something prevented UMS from checking Windows permissions", e);
			}
		}

		return true;
```

Nadahar:
@valib If you make a commit regarding that, you could fix the exception logging as well, as it is not it doesn't print the exception. Change:
```java
LOGGER.error("Something prevented UMS from checking Windows permissions", e);
```
to:
```java
LOGGER.error("Something prevented UMS from checking Windows permissions: {}", e);
```

Optimusprimeums:
Windows 10 works fine for me, been running UMS on it since the the inside preview

Nadahar:
Thanks @Optimusprimeums, I'll probably pick it up if nobody else does.

@valib @Optimusprimeums I'm not surprised that it works in general, but things like install service and any other "special Windows function" needs to be tested specifically. Have any of you tried to install UMS as a service in Windows 10?

valib:
@Nadahar will try it when will be at home

Nadahar:
@valib Did you ever get around to test this?

Optimusprimeums:
[...]

PR #670, opened 2015-09-11
Addressing #640 Windows 10 compatibility 

Nadahar:
As far as I can see, ```PmsConfiguration.isAdmin()``` is the only part of UMS directly incompatible with Windows 10 - if anyone knows of any other issues please let me know.

I've updated ```PmsConfiguration.isAdmin()``` to work for Windows 10, or actually to work for any Windows version from Windows XP and upwards. I've tested this on Windows XP, Windows 7 and Windows 10, so I assume it's correct also for Windows Vista and Windows 8. When looking at Windows version instead of matching a specific OS name, it will work for Windows servers as well a future Windows versions as long as the method work. The day it doesn't work anymore, the Windows version simply has to be limited upwards.

But, since I was already in that method, I wanted to complete it by adding support for Linux and OS X as well. This is done and tested (don't let me start on how much trouble it was getting a OS X VM to run and to get git and maven to work on OS X).

This has created problems with the "single instance" logic though, since ```isAdmin()``` has always returned ```true``` for Linux and OS X until now. I've tried to suggest fixes in #668, but the only feedback has been negative. Until that is solved I think it's better to leave this unmerged.

Any suggested resolutions are welcome.

### DIFF
--- a/src/main/external-resources/documentation/installation.html
+++ b/src/main/external-resources/documentation/installation.html
@@ -261,8 +261,8 @@
 					permissions on the UMS install folder.
 					<ul>
 						<li>Check the base file system permissions.</li>
-						<li>On Windows Vista or Windows 7, reinstall with &quot;Run
-							as Administrator&quot; on the installer.</li>
+						<li>On Windows Vista, Windows 7, Windows 8 and Windows 10 reinstall 
+							with &quot;Run as Administrator&quot; on the installer.</li>
 						<li>Check if any filesystem protection programs are blocking
 							access (some antivirus or malware programs have this feature).</li>
 					</ul>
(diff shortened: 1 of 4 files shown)


SubJunk:
This looks good. We could always comment out the Linux/OS X logic so it can merge now, and uncomment that when the single instance logic issues are fixed

Optimusprimeums:
Great fix, i agree with @SubJunk

valib:
Agree too.

skeptical:
+1

Nadahar:
Ok, I've disabled the check for Linux/OS X and will merge this, but then I really hope we can continue the discussion in #668 .

Issue #564, opened 2015-06-02
Don't actualize "Shared folders" while scanning after a server restart; it's need a UMS restart.


rd1:
default list is all drives.

i add 1 local partition D: , press restart server, reset cache then scan (magnifier icon), it still scans C: drive.

close ums and relaunch it for changes to take effect.

dogancelik:
I had a similar problem.

This is what happened in my case:
1. Default is All drives, I can't remove it so I add E:\Music to get rid of it.
2. Start the scan
3. It still scans C: drive, D: drive.

SubJunk:
Seems like a bug, but for the record there is no need to manually scan. The cache is automatically populated when you browse.

Nadahar:
@SubJunk It's not a bug, I've noticed the same behaviour and have in the back of my mind to try to find a more inituitive way this could be handled. What happens is: When you add or remove folders, the "restart server" goes green. It's not very easy to notice imo, but the changes aren't actually "saved" until you press restart server. If you then click scan, it will scan all the folders, not just the new ones. If you press scan before "restart server", it will scan the list as it were before it was modified.

About browsing to populate: I guess that depends on the use case, but when you have thousand of music files you don't want to have to browse them all to get them scanned and categorized by artist, album, title etc.

SubJunk:
@Nadahar okay so maybe we should make it more obvious that a server restart is required

The iTunes integration data doesn't get cached anyway, so it doesn't make a difference for those folders.

Nadahar:
@SubJunk Yes, as a minimum. Disabling (gray out) the scan button from any path is changed until restart would probably also be smart. But, the elegant solutions would be to figure out if the server really needed a restart, or if it was a way to reinitialize only what was needed to avoid having to restart, and then automaticly start a queued scan for added folders?

I didn't think of ITunes, but if you have a large collection og media on disk like many people do.

dogancelik:
@Nadahar what you described in your first post is a bug.
* If you **press** green "Restart Server" button then press "Scan all shared folders", it doesn't scan new folders.
* If you **don't press** green "Restart Server" button then press "Scan all shared folders", it still doesn't scan new folders.

Either way it doesn't scan new folders.

Nadahar:
[...]

PR #673, opened 2015-09-20
Watched media integration

SubJunk:
This is not ready to merge.

This branch is for improving our support for tracking and dealing with the watched status of videos.

It should add a user option to hide watched videos, and if that option is disabled then the thumbnail will display "Watched" on it.


SubJunk:
@UniversalMediaServer/developers I'd love to see this get some attention if anyone is looking for something to work on, it's my dream feature at the moment

valib:
@SubJunk I tried to test it but I don't understand how it works. I set the Enable video resuming and Show watched videos as true but I don't see  the "Watched" on the thumbnails.

SubJunk:
@valib it only happens if a video is marked as "watched", as in videos that do not show up in the New Media folder

SubJunk:
@UniversalMediaServer/developers this is working in all folders now, instead of just the New Media folder. I also made the decision to only mark videos as watched if it has been 30 seconds since they started playing. This is because many renderers don't seem capable of giving us enough information about their file position, so the previous thing we were doing, which was seeing if 92% of the file had been played, was unreliable.

This way we do more of a YouTube style approach (marking the video as viewed immediately) but with a 30 second buffer to allow for accidental plays.

If anyone has any opinions on this approach please let me know because I want to develop this branch and eventually release it as version 6.

Nadahar:
@SubJunk I think it sounds good, but I'm a little sceptical about the 30 seconds - I'm guessing there's no easy way to mark it as "unwatched" again, especially if it's auto hidden. Maybe an option for the duration would be the way to go. Since we're lacking reliable position information, is the 30 second threshold simply that the file has to be played for that amount of time continuosly, or is it some kind of total so it's the sum of breif looks?

SubJunk:
Yeah there isn't currently a way to mark as unwatched, so I guess for now we shouldn't implement the automatic hiding of watched videos until we have a GUI to unhide files.

As it is now, it's just the thumbnail that changes, so it's not so important, but possibly a good goal for the 6.0.0 release could be a GUI to mark files as watched or unwatched.

The 30 second threshold starts fresh each time - it simply compares the time elapsed since the file started playing. One possible issue with it is that if a user accidentally plays a file but pauses it instead of stops it, the timer would continue and the video would be marked as watched.

Nadahar:
@SubJunk Isn't there a way to even detect a pause? I don't know how the streaming is done, but UMS must be involved in that in some way, and can't just push on filling the buffer at the renderer during a pause?

SubJunk:
I'm not sure if we can do it that way, I guess it depends on how much buffer the renderer likes to use.

SubJunk:
I've been playing with the VLC app for iOS lately and it has settings to adjust its internal buffer so I'm guessing that would throw off our ability to tell when it's paused

Nadahar:
Ok, so the way it works is that the renderer has all controle and just keeps requesting new chunks when it feels like?

SubJunk:
I think so, yeah. Some renderers report statuses back to us and others (most) don't. DLNA is pretty poorly implemented across the board so we can't rely on much.

Nadahar:
That means that any playing, pausing, seeking etc will simply count as play time. I'd say 30 seconds is a short time then, but it all depends on the actual length of the video. Maybe 5% of the video file's length would be a suitable threshold? In any case, I think it should be a configurable value with a sane default.

SubJunk:
@Nadahar maybe 5% or 30 seconds, whichever is larger. There's already a configuration value which is used for video resuming, "min_playtime", so we can probably reuse it, possibly with a rename to "minimum_playtime".

SubJunk:
Okay guys I've expanded upon the previous "show_watched_videos" option by removing that and adding "watched_video_action". It has several values that describe different scenarios that I think will all be useful to people. I'm interested to get opinions on these!

The options are:
1. Do nothing
2. Write "Watched" on the thumbnail (default)
3. Hide the video in UMS
4. Move the video to a different folder
5. Delete the video (not recommended on operating systems without recycle bins)

Options 2 and 3 are currently working.

taconaut:
For option 5, all code is available here if you want to use it https://github.com/taconaut/UMS-Plugin-AutoDeleteFileAfterPlay

Nadahar:
@SubJunk 

2) Will this use a renderer configuration parameter to decide between icon or text, or should textual indication be a separate option? I'm thinking about renderers without thumbnail support. Ideally I think it should be a renderer configuration, since that would enable the same server to create thumbnail overlay on supported renderers and textual indication on others. But, that requires that we figure out which renderers that displays thumbnails. The renderer configuration should be a general one, like "Show thumbnails" or "Generate thumbnails", and that options could be used for other purposes as well.
5) I'm not convinced that 5) is needed as long as 4) is there as long as the detection of what is played is as inaccurate as it is. It is after all, extremely easy to manually empty the folder (from 4)) in question. On Windows, files will not be added to the recycle bin if they are too large to fit, or if the bin is already "full". Video files can often be quite large, so this could easily happen here. I checked two drives now, on a 563 GB drive Windows had chosen a 25 GB recycle bin size, and on a 98 GB drive it was 7 GB.

SubJunk:
@taconaut that's really cool! I didn't know about that plugin. Have you considered adding it to the official plugin list so users can install it from within UMS?

@Nadahar yes I think there should be a renderer option like you described. I agree about the danger of 5, but I think it still has use for users who don't care if a file is mistakenly deleted, for example me. Also I have made this branch use the same strategy for marking files as watched as we do for resume files, so now we are using the 92% rule instead of 30 seconds. It hopefully makes it so it will be more likely to get a false-negative than a false-positive. We could possibly change the default to something like 80% to try to catch more.

taconaut:
@SubJunk I intend to add it but wanted to get some feedback for Linux and OS X before doing so (http://www.universalmediaserver.com/forum/viewtopic.php?f=6&t=4047&sid=adcc042f92a16e95ab49c0ca84e64bc4)

Nadahar:
@taconaut Lucky me, I've been stupid enough to set up 3 relevant VM's for tests by now, one with OS X 10.8, one with Centos 7 and one with Ubuntu 12.04.5 LTS (the exact version used by Travis CI). That means that I can test it for you, like I haven't got enough things to do already :wink: 

But, if you explain to me exactly how to test it, how to install/configure the plugin etc I'm sure I can do it.

taconaut:
@Nadahar Thanks for helping out testing! You can just follow the instructions from the GitHub readme. The part which interests me most is if the files get deleted to the recycle bin for all OSs if the option is checked.

Nadahar:
@taconaut It wasn't so easy to test as I was first thinking, since I had no renderer available on the site where I ran the VM's. I've moved the VM's to my home server now, and I'm just finished testing OS X.

The good news is that "move to trash" worked like it should on OS X. I first tried on some sample files just 15-20 seconds long, and nothing happened. So, I had to download a longer sample and then it worked. From the log I can see that there's a minimum threshold of 45 seconds. When such a value exists, I think that should be a part of the configuration as well. 

The other thing is that it seemed to want to delete a MP3 file I tested as well, but it failed because UMS would register it as multiple start/stop events during playback so 45 seconds were never reached. I'm not sure if people will love having their music collection deleted. There's a simple way to determine what is what now, with ```DLNAMediaInfo.isVideo()``` - I doubt deletion is relevant for anything but video files.

I'm attaching the logs. @SubJunk There were some missing symbol errors reported as well, maybe you should look at the log files and see if you know what that is about:

[debug1.log.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/24218/debug1.log.txt)
[debug3.log.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/24219/debug3.log.txt)
[debug2.log.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/24220/debug2.log.txt)

Ubuntu test coming up next...

Nadahar:
@taconaut Ubunto test didn't work with trash can:

```logtalk
WARN  2015-10-30 15:53:50.671 [StopPlaying Event for Auto-delete after play] Failed to move file '/home/user/Downloads/share/SampleVideo_1080x720_20mb.mp4' to the recycle bin after 10 retries
java.io.IOException: No trash location found (define fileutils.trash to be the path to the trash)
	at com.sun.jna.platform.FileUtils$DefaultFileUtils.moveToTrash(FileUtils.java:91) ~[na:na]
	at github.com.taconaut.plugin.autodeleteafterplay.AutoDeleteFileAfterPlayPlugin.donePlaying(AutoDeleteFileAfterPlayPlugin.java:139) ~[na:na]
	at net.pms.dlna.DLNAResource$2$1$1.run(DLNAResource.java:2777) [classes/:na]
	at java.lang.Thread.run(Thread.java:745) [na:1.7.0_80]
```

[debug.log.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/24268/debug.log.txt)

Nadahar:
@taconaut Centos 7 had the same result for trash can:
```logtalk
WARN  2015-10-30 16:36:19.153 [StopPlaying Event for Auto-delete after play] Failed to move file '/home/user/Downloads/share/big_buck_bunny_720p_20mb.mp4' to the recycle bin after 10 retries
java.io.IOException: No trash location found (define fileutils.trash to be the path to the trash)
	at com.sun.jna.platform.FileUtils$DefaultFileUtils.moveToTrash(FileUtils.java:91) ~[na:na]
	at github.com.taconaut.plugin.autodeleteafterplay.AutoDeleteFileAfterPlayPlugin.donePlaying(AutoDeleteFileAfterPlayPlugin.java:139) ~[na:na]
	at net.pms.dlna.DLNAResource$2$1$1.run(DLNAResource.java:2777) [classes/:na]
	at java.lang.Thread.run(Thread.java:745) [na:1.7.0_85]
```

[debug.log.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/24329/debug.log.txt)

taconaut:
@Nadahar Thanks a lot for testing :+1:

The required play length before deleting is calculated according to the play percentage and the duration of the media and has no hard coded value ([click](https://github.com/taconaut/UMS-Plugin-AutoDeleteFileAfterPlay/blob/master/src/main/java/github/com/taconaut/plugin/autodeleteafterplay/AutoDeleteFileAfterPlayPlugin.java#L115)).

Can the error you get on UNIX systems be avoided by setting a system parameter (fileutils.trash)? Otherwise the option to move files to the recycle bin could be disabled for UNIX systems.

For now you can specify in which folders (and sub-folders) files will be deleted, but it would be a good idea to add three check boxes to define if video, audio and/or pictures should be deleted (with default being only for videos).

The multiple notification sent by UMS issue had been fixed for videos quite a long time ago, it seems it's still there for audio files though. There should be special handling for these cases in the plugin, but from what you're telling, it's not working. I'll check that when I get some spare time (after holidays ;)

Nadahar:
@taconaut 
> The required play length before deleting is calculated according to the play percentage and the duration of the media and has no hard coded value (click).

Something must be wrong with the calculation then (or a bug somewhere in UMS) since a 15 seconds long clip required 45 seconds playtime with a 80% setting. Have you tried it on very short clips? That said, I think a minimum length setting would make sense, the likelihood is that it's TV programs/series/movies that you'd want to delete after viewing, not short clips.

> Can the error you get on UNIX systems be avoided by setting a system parameter (fileutils.trash)? Otherwise the option to move files to the recycle bin could be disabled for UNIX systems.

That might be, I just tested and didn't debug this (since that requires a lot more setup and the VM's aren't that great to work with). I think the work will be to figure out what to put there though, as there's no such thing as a proper standard on Linux. You'd probably have to probe for environmental variables and other settings to figure out what to use. 

You could install VirtualBox or one of the VMWare flavours (Player, Server, Workstation) and set up a Linux installation for testing. I've installed Git, Java, Maven and Eclipse without much hazzle on both Ubuntu and Centos. You'd have to install Maven from Maven and not one of the packages though since the packaged Maven's are too old for some of our Maven plugins. I installed the latest Eclipse as well, they have even made an installer for it. Once you have this in place, figuring out how to do it shouldn't be too hard at least for the most common graphical environments.

> For now you can specify in which folders (and sub-folders) files will be deleted, but it would be a good idea to add three check boxes to define if video, audio and/or pictures should be deleted (with default being only for videos).

I don't really know if I can see any situation where you'd want audio and images to be deleted after playing, I would probably just limit it to videos. I know you can configure which folders, but that's for the more advanced user and I'm thinking that it's worse to delete something that the users don't want to delete than the opposite.

> The multiple notification sent by UMS issue had been fixed for videos quite a long time ago, it seems it's still there for audio files though.

If you, or anyone else, remember where to find this fix, we should look at it and figure out why it only works for videos.

taconaut:
@Nadahar 

I've released the [plugin](https://github.com/taconaut/UMS-Plugin-AutoDeleteFileAfterPlay) version 0.0.3 which addresses all your comments and updated the plugin README to reflect the changes; I haven't been able to test the UNIX issues though.

> Something must be wrong with the calculation then (or a bug somewhere in UMS) since a 15 seconds long clip required 45 seconds playtime with a 80% setting. Have you tried it on very short clips? That said, I think a minimum length setting would make sense, the likelihood is that it's TV programs/series/movies that you'd want to delete after viewing, not short clips.

I haven't tried playing short clips yet. The length of the media as it is being returned by `DLNAMediaInfo.getDurationInSeconds()` is now included in the logs. I'm pretty sure it is this length which is wrong; correct me if I'm wrong ;)
I haven't added a absolute minimum length parameter as this can be easily handled by setting folder paths for auto delete. If a request for a real scenario will pop up, I'll reconsider it.

> That might be, I just tested and didn't debug this (since that requires a lot more setup and the VM's aren't that great to work with). I think the work will be to figure out what to put there though, as there's no such thing as a proper standard on Linux. You'd probably have to probe for environmental variables and other settings to figure out what to use.

The option to move to recycle bin will now be disabled if the system has no recycle bin; as returned by `fileUtils.hasTrash()`.

> I don't really know if I can see any situation where you'd want audio and images to be deleted after playing, I would probably just limit it to videos. I know you can configure which folders, but that's for the more advanced user and I'm thinking that it's worse to delete something that the users don't want to delete than the opposite.

I've added this configuration option, as it was easy enough to do ;)

> If you, or anyone else, remember where to find this fix, we should look at it and figure out why it only works for videos.

Sorry I don't and haven't been able to find it. I'm pretty sure this has been fixed in pms (before ums existed)

Where should we continue this discussion? This pull request is probably not the right place for it.

Nadahar:
@taconaut 
> Where should we continue this discussion? This pull request is probably not the right place for it.

I don't really know, we could discuss it on the forum or create an issue in the plugin repository, but as I see it this is highly relevant for this PR as this PR will need the exact same functionality and most issues with your plugin will apply to this PR as well. If you disagree @SubJunk, please say so.

> The option to move to recycle bin will now be disabled if the system has no recycle bin; as returned by fileUtils.hasTrash().

Firing up the new version on Centos I can see that ```hasTrash()``` has worked in a way. The result however is wrong. The OS has a trashcan, and it's being used as normal by other software. What the ```hasTrash()``` option returns, is whether or not ```FileUtils``` knows how to recognize it. Telling the user that "Your operation system does not support moving files to the recycle bin" is wrong, except that it's called trash and not recycle bin. If you don't bother implementing Linux trash, you should atleast change the message to something like "Java doesn't support move to trash on your...."

The most commonly used Linux desktops support the [freedesktop.org trash specification] (http://www.ramendik.ru/docs/trashspec.html) though, so a check for this could be considered a minimum effort. The "problem" though is implementing the functionality with the "info" file that has to be created for each trashed item. It's not much job, but it's hard to do without access to the platform. I'd still give a Linux VM a go if I were you.

> I haven't tried playing short clips yet.

It's tedious work for me to test this since I don't have a capable renderer available from my computer's location, and have to do a lot of running. I doubt this "problem" is specific to Linux, so you could probably debug this yourself.

> I've added this configuration option, as it was easy enough to do ;)

Obviously, but I don't see the use scenario unless someone is so fan of snapchat that they want to replicate this "view once" functionality on their pictures.

Nadahar:
@taconaut If you look at the implementation in [com.sun.jna.platform.FileUtils] (https://github.com/java-native-access/jna/blob/master/contrib/platform/src/com/sun/jna/platform/FileUtils.java) you can see that it more or less has to fail. ```DefaultFileUtils``` will be used if ```Platform``` is neither Windows nor Mac. This implementation is extremely lazy and doesn't look in the locations defined by the [freedesktop.org trash specification] (http://www.ramendik.ru/docs/trashspec.html) at all, nor does it implement creation of the "index" files. I don't know if there's any Linux desktop environment out there that will work with the ```DefaultFileUtils``` implementation, but it can't be many if any. You could almost just test for ```Platform.isWindows()``` or ```Platform.isMac``` instead, and then just reject the rest. The fact that nobody has written such a simple basic class to handle the most common Linux desktops hints to me that it's extremely few that actually use Java for writing desktop applications.

SubJunk:
I'm happy for the discussion to happen in this PR and I hope to implement the code from the plugin in this PR when we're happy with it.

I'm getting happy with this branch now, I'm using it as my main branch and it's going pretty well. There's still some work to do to improve stability but it's mostly usable now.

taconaut:
@Nadahar 

> Firing up the new version on Centos I can see that hasTrash() has worked in a way. The result however is wrong. The OS has a trashcan, and it's being used as normal by other software. What the hasTrash() option returns, is whether or not FileUtils knows how to recognize it. Telling the user that "Your operation system does not support moving files to the recycle bin" is wrong, except that it's called trash and not recycle bin. If you don't bother implementing Linux trash, you should atleast change the message to something like "Java doesn't support move to trash on your...."

Good remark, I'll change the tooltip text.

> The most commonly used Linux desktops support the freedesktop.org trash specification though, so a check for this could be considered a minimum effort.

This should be part of the jna FileUtils rather than custom code in ums IMO.

> You could almost just test for Platform.isWindows() or Platform.isMac instead, and then just reject the rest.

I'll keep the code as it is and keep fingers crossed that it will work with a future update of FileUtils.

I'm in the process of setting up an Ubuntu VM, I should be able to test it by using the web interface at some point.

taconaut:
@Nadahar I've now changed the tooltip text to `Java does not support moving files to the recycle bin for your operating system, they will always be permanently deleted.`
For me it's now good enough to use :)

@SubJunk Is it worth releasing the plugin or are you integrating the functionality into ums soon anyway?

Nadahar:
@SubJunk I'm just trying to translate ```Mark all as watched``` to Norwegian, and while looking for an appropriate replacement for "watched" I was thinking that this doesn't make sense for Audio files. Do you plan on offering this exclusively for video files, if not "played" might be a better word?

SubJunk:
@taconaut I'm not sure how long this branch will be in development before release so it's up to you :) Thanks for your work on it. I would welcome a PR to this branch (or I could give you access to commit directly) to merge your deletion functionality

@Nadahar I think this functionality is only useful for videos so I think we should ignore the "watched" status of everything else. It's probably a good idea to make sure that's clear everywhere in the program too. We could rename "New Media" to "New Videos" or "Unwatched Videos" to be more clear.

Nadahar:
@SubJunk I think marking watched images or played audio could be useful, even though moving/deleting them hardly seems relevant to me.

taconaut:
@SubJunk I see what I can do. Working with pull requests is fine for me.

taconaut:
@SubJunk Which configuration options do you want to use from the [plugin](https://github.com/taconaut/UMS-Plugin-AutoDeleteFileAfterPlay)? percentPlayedRequired, autoDeleteFolderPaths and moveToRecycleBin?
You haven't done any UI for now, right?

SubJunk:
@Nadahar hmm, I'm not sure about that. I can think of reasons for and against.

@taconaut cool! We have a percentPlayedRequired, called resume_back_factor, but it could probably be renamed to something like percentage_until_watched or something.
moveToRecycleBin should always be true IMO so I don't think we need a config for it.
What is autoDeleteFolderPaths?

taconaut:
@SubJunk 

> moveToRecycleBin should always be true IMO so I don't think we need a config for it.

As the used library doesn't support move to recycle bin for unix systems, the option will always be greyed out for linux builds. OK?

> What is autoDeleteFolderPaths?

This allows to define in which folders files will be automatically deleted. E.g. if setting `C:\video_tmp;D:\video_tmp` only files in these folders and sub-folders will be deleted.

SubJunk:
@taconaut I think for the UMS integration we should make it so that we always use the recycle bin if it's available. I don't think there's much point adding a config setting for it, because there will be cases of false-positives with playback (like if the person pauses the video for 30 minutes while they make dinner, then come back and decide to watch something else) so it's important to always let them have a recovery step when possible. As always though feel free to disagree, this is just my opinion.

For autoDeleteFolderPaths we can use folders_monitored

ExSport:
@SubJunk I don't understand why UMS is blind to pause?
I know I read here that e.g. PS3 doesn't send state of played file  so UMS doesn't know if file was stopped, paused or anything else but what I remember when pause is used on PS3 - UMS transcodes file until buffer is full. When it is full,  CPU drops from 100% to zero so PS3 must know that PS3 doesn't ask for data and also knows that stream should not be closed so unpause resumes file streaming.
Maybe I lost in translation but I expected that UMS knows what is going on the renderer so resume function in UMS or pause on PS3 should be reliable:)
Btw. I am not sure if delete function will be welcomed by users as only 1% possible angry users with accidentally deleted files will make more negative advertisement on net than 90% happy users. As an optional plugin I don't see any harm to UMS itself but as a built in feature...I am not sure here :))

SubJunk:
@ExSport by default, the text "Watched" is written on the thumbnails, so no deletion happens. The user has to select the setting which will be a dropdown in the GUI. I agree it shouldn't be set by default.

I agree it should in theory be possible to make us detect pauses, I would welcome that addition to the code

taconaut:
@SubJunk 

> I think for the UMS integration we should make it so that we always use the recycle bin if it's available.

The option to delete (to recycle bin) shouldn't available in the combo box on systems where `FileUtils.hasTrash()` is false then.

> For autoDeleteFolderPaths we can use folders_monitored

We can, but I don't see a direct link between the two. Someone might want to only auto-delete e.g. TV series.

Nadahar:
@taconaut @SubJunk 
First I'd like to say that I share @ExSport's worry about deleting files. I'd think that moving them to a separate folder should be sufficient, it's just as easy to delete everything in a folder as emptying the trash. That said, it's a tiny little extra step, so if it's important to have deletion I'd prefer that we hide that choice for the "non-advanced" view.

Beside that, I get the feeling that you talk a little bit past eachother because you speak from separate point of views. So, I'll try to pull them together :wink: 

@taconaut's plugin only deletes/moves to recycle/trash bin, and it does so only for specified folders. If works for all media types.
@SubJunk's feature gives a range of different options as to how to treat watched videos, and is only indended to be used for video media:
* Do nothing
* Write "Watched" on the thumbnail (default)
* Hide the video in UMS
* Move the video to a different folder
* Delete the video (not recommended on operating systems without recycle bins)
This is meant to be an extension/improvement/replacement for the "new media" feature (not exactly sure which). The existing system with "monitored folders" are meant to be what indicates if this functionality is enabled or not for that particular shared folder. This means there's not an option to enable it only for subfolders in a share, shares will have to be organized slightly different (than with @taconaut's model) if you only want selected folders to use this.

If I forgot something, please edit or correct me...

SubJunk:
@Nadahar I think that's an accurate summary

If you guys think we should leave the deletion option for the plugin we can do that. I'll usually defer to the majority for these decisions.

Also I should note that I recently broke some things by refactoring, so this branch took a little step backward but the implementation will be nicer.

Nadahar:
@SubJunk I guess it doesn't hurt to have the options as long as people really understand what they are doing. I'm working on trash bin support for Linux right now, so I'd hate for that to be a waste :stuck_out_tongue: 

I didn't understand that last part, what did you break?

SubJunk:
@Nadahar Yeah I think we should do everything we can to make the user aware it's a dangerous setting. Maybe we could have a confirmation dialog open up when the user selects the option from the dropdown. That's exciting about the trash bin support!

I switched from using our normal folder implementation to using MediaMonitor folders since I think that's a cleaner use of code, but it means that all watched files are hidden regardless of setting. I've already fixed it but haven't committed it yet. It has made me aware of another issue though which might make me go back to my previous implementation, which is that file ordering doesn't seem to work in MediaMonitor folders. I think there are already bug reports for this but this branch will make it much more likely that users will encounter the issue, since it means that the "Ignore the word "the" while sorting" setting doesn't work on monitored folders.

I wonder how easy it would be to implement sorting on these folders, @SharkHunter I think you have mentioned this issue before?

SubJunk:
Also I'm happy for us to open up watched media support to audio and images if people think that would be useful. Do we have any ideas for which word to use instead of "watched"?

Nadahar:
I was thinking of replacing the "Include in new media" ```Checkbox``` with a ```ComboBox``` with the given options (none, mark, hide, move, delete) in the shared folders table. If so, it could be useful to mark images and/or audio as "watched"/"played" and e.g movie folders for hide, move etc. Then support for audio and images could serve a purpose.

I would think "watched" would do fine for videos and images, while "played" would be better for audio? It's shouldn't be difficult to differentiate this based on ```isVideo()```, ```isAudio()``` and ```isImage```, so it doesn't have to be one word.

There's one thing though, I don't know how you plan to make the thumbnail overlay, but please keep in mind that it needs to support translation (and thereby different number of letters) as well.

SubJunk:
@Nadahar do you mean that each folder could have its own setting? I think I like that. It's still almost as simple for the user but more powerful.

The thumbnail overlay is implemented in a way that allows translation and automatically resizes the text if it is too wide for the image. I've tested it and it works well but I welcome more testing.

Nadahar:
@SubJunk 
> do you mean that each folder could have its own setting?
Yes - each shared folder entry could have it's own setting, but subfolders would have to inherit. I don't know how we should handle in one shared folder have a subfolder that is specified as a separate shared folder though. That's probably not handled to well today either?

SubJunk:
@Nadahar I guess it would be best in that case to use the subfolder's setting. Sounds cool.

@skeptical are you still around? I'm looking to make it so that the timeseek value propagates to the Status tab timer, as well as being able to use it for this, and I'm not sure what the best way to affect the Status tab timer is. Help would be appreciated :)

SubJunk:
We now have lastStartPosition and lastStartSystemTime which expose seeking activity, so we can use them to keep the Status tab timers up to date when users rewind or fast-forward. I've also made it so that it's taken into account when marking files as watched, so if you fast-forward straight to the end of a file it should be marked as watched.

Nadahar:
Sounds like a big improvement :+1:

Nadahar:
@SubJunk @taconaut The support for Freedesktop.org/Linux trash is available as soon as #722 is merged. 

Using it is very easy:
Call ```FreedesktopTrash.moveToTrash(File or Path)```. It will either succeed or throw an error, so all error handling is to be done by ```catch```. There is also a method ```FreedesktopTrash.hasTrash()``` method, but the more I think of it the more I'm against using it. 

There are multiple reasons why using ```hasTrash()``` is a bad idea:
* The answer might not be the same for all partitions/drives. It depends on the current user's permissions on the drive in question and on the file system used.
* There's no way to actually check if the OS supports the Freedesktop.org trash specification, the specification dictates that if the necessary folders aren't there they should be created. That means that ```hasTrash()``` will create the trash folder if it's not already there, even if the system itself does not use this standard. I also consider creating folders on a users system just for checking for trashability is bad behaviour.
* The work done by ```hasTrash()``` duplicates much of the work done by ```moveToTrash()```, so it's better to just try and handle it if it fails. 

In the end, I don't know of a good way to decide if this should be attempted on a particular system. It will succeed on any Unix based system with the permissions to create the folders needed. We probably just have to use this if ```Platform.isLinux()```. Some of the considerations done here (that trashability is dependent on partition/file system) is probably true on OS X and Windows as well, so the logic for using trash/recycle bin probably needs to be modified some. The user should probably be able to choose between:
* if possible trash, else leave in place
* if possible trash, else delete
* delete (on headless systems for example)

taconaut:
@Nadahar I'd suggest using the same structure as [jna.FileUtils](https://github.com/java-native-access/jna/blob/master/contrib/platform/src/com/sun/jna/platform/FileUtils.java) uses to allow using it the same way ([FileUtils.getInstance()](https://github.com/UniversalMediaServer/UniversalMediaServer/blob/WatchedMediaIntegrationV3/src/main/java/net/pms/dlna/MediaMonitor.java#L23) static and hasTrash and moveToTrash being instance methods). This way it would be easy to instantiate either your implementation or the `jna.FileUtils` one depending on the detected operating system.
Or even better include the functionality in `jna.FileUtils` by submitting a pull request; the project is active.

Nadahar:
@taconaut I started out that way, but their implementation has too many shortcomings so I abandoned it. Like I said, there's no way to give a general answer to if a trash bin is available, and that should be true for their implementation as well. That must be concluded on a per-file basis, as it depends on where the individual file is stored and the file system used on that device. 

I also dropped accepting arrays of files since the caller then has no way of knowing which files were moved and which were not. The caller has to do the looping to handle exceptions for individual files, and thus the logic can't be the same. I don't know how they've approaced the same issues, I'm guessing that they just throw an exception and leaves some files in the trash bin and some not, but that's not what I consider to be good practice.

taconaut:
@Nadahar Ok, this makes sence

SubJunk:
Nice one :)

SubJunk:
This is now using moveToTrash and synchronized with master. I think that considering the issues with MediaMonitor folders (they don't work on Panasonic TVs and they don't support sorting of files) I will go back to my original implementation of this feature, which monitored regular folders. When those issues are solved it can be reformatted easily.

SubJunk:
@taconaut @Nadahar I just did the suggested changes to moveToTrash, hopefully. I also reverted some code so it no longer uses MediaMonitor folders, therefore fixing file sorting and Panasonic TV support. I haven't tested all modes with these changes yet but I'm using mode 3 (moving the video after it's watched) and it's working.

taconaut:
@SubJunk  :+1:

SubJunk:
@UniversalMediaServer/developers this seems complete now, it would be great to get you guys to test it before it's released as 6.0.0-a1 or b1.

Nadahar:
@SubJunk I updated this branch on crowdin (so that the new strings are possible to translate). From the text strings I see that you've only kept the functionality for video files. I have some questions to some of the strings:
* ```Add an overlay to the thumbnail```: Does this mean that this doesn't work for renderers without thumbnail support? If it's indicated in text for renderers without thumbnail support, maybe ```Tag media``` or something like that would be better?
* ```Hide the video in UMS``` I think ```in UMS``` is superfluous. If you're thinking that some could interpret it as making the file hidden in the file system, I don't find that very likely. 
* ```Send the video to the Recycle Bin/trash``` First of all wouldn't ```move``` be a better word than ```send``` even though I know send is sometimes used by OS'es? Secondly, why the term ```Recycle Bin/trash```? I'd prefer ```recycle/trash bin```.

I've checked out your branch now and will take a closer look.

Nadahar:
@SubJunk Another thing is that I don't think we should make a 6.0 release with new functionality to shared folders and still ignore #564 which is a major bug imo. From what I've understood this has been a bug since the ```Save``` button was removed.

Nadahar:
@SubJunk I'm looking at the change to "Include in new media" to "Monitor status of videos", and I'm not sure. Does this mean that the "New media" feature has been reduced from working on all medias to only videos? If so, I think it's a bad idea.

I'm not sure about the whole "include in new media" feature either. Why do we need that on a shared folder level? I'd rather replace the "Hide new media folder" with "Monitor new media" and then use that for turning on or off the "New media" functionality as a whole.

SubJunk:
@Nadahar thanks for the feedback :)

1. I'm in favor of the enhancement you suggested to add a thumbnail support setting to renderer configs and make this feature use it to determine whether to modify the thumbnail or prepend a string.

2. Yeah that was the reason I said "in UMS", but maybe you're right

3. Yeah I like the suggestions. "Move the video to the recycle/trash bin" seems good.

4. I've just added it to the 6.0.0 Milestone

5. I think it would be good to make this functionality extend to audio and images. It was important to me to iron out the feature with videos to start with since that's the most tricky, but I think I'll extend it before release.

So I'll have a look at these 5 points before the release.

SubJunk:
Oh and the other point about not having include in new media at a shared folder level, I think it's important, especially if they're using option 3, as I am. I have a folder of unwatched videos, which I have monitored, and I have a folder of watched videos which is where UMS moves them to after watching them. It wouldn't make sense to monitor that folder. Also people might not want to add other folders to New Media, like porn for example.

SubJunk:
@Nadahar I think those 5 points are addressed now

SubJunk:
Overlays for images and audio are working now, but not perfectly. I've added scaleImage() to UMSUtils but it is unpredictable because it only outputs a file within the bounds you give it, due to Thumbnailator working that way. So ideally we should modify scaleImage() to work like FFmpeg does, which is to pad the thumbnail with a black background. Then we can output reliable sizes for all our thumbnails.

valib:
@SubJunk I think that the Thumbnailator ```Canvas(int width, int height, Position position, boolean crop, Color fillColor)``` can be used for that.

Nadahar:
@SubJunk I know you are the native english speaker, but are "heard" really better than "played" for audio?

Nadahar:
@SubJunk I did a tiny commit to this branch updating the jna-platform version.

SubJunk:
@valib cool, I'll look at that when I get a chance. Might not be until next weekend

@Nadahar the scaling works no matter what the length of the word is, but if it's a very long word then it may be unreadable because the text will be so small

Maybe it's better to use "played" for audio, we should change that. I also noticed that on my PS3 there's no event like "Started playing" when I view an image. I think maybe the PS3 loads the images with the folder so you can't tell when one has been viewed. So at least on the PS3 it doesn't seem possible to mark them as "seen", though it may be different for transcoded images, not sure.

By the way anyone can feel free to commit to this branch or any of mine if you have additions/fixes

SubJunk:
Another thing to add is that currently we only send audio thumbnails if they have cover art, so an enhancement to this should be to add the "played" thumbnail with a plain black background in that case

Nadahar:
@SubJunk Maybe we should look at that as well. From what I can see, the download coverart function doesn't do much. There's almost no thumbnails on the music I've tested at least.

SubJunk:
The thumbnails should be working for all media types now. We didn't output a consistent aspect ratio for images or audio before, so for those formats I've made the thumbnails square (renderer.thumbnailWidth x renderer.thumbnailWidth)
I based that on the way the PS3 displays them, but it may be better to add a renderer config for it like squareAudioThumbnails and squareImageThumbnails

Anyway since before we didn't have any strategy for that, it is still an improvement without the renderer option.

@UniversalMediaServer/developers I think that was the last thing I was going to do on this branch unless I've forgotten something?

Nadahar:
@SubJunk It depends a bit on how you want to do it. I have several suggestions above that's not been addressed or discussed, but I don't know how soon you want to release 6.0.0 and if you want some of those things done in different branches (if we should do them at all).

I'd like to have a go at some things in this branch, like specifying the handling on a per share level. You'd have to give me some time though, as I'm still trying to finish the #679. I've also let myself be distracted with #729 and #733 lately further slowing down the progress on #679. 

I have a hard time focusing on one thing at a time some times, but I've also been struggeling with some things that's taken me some time to figure out.

SubJunk:
@Nadahar I think this is a good place to release and get feedback on it, then we can improve upon it later on, especially if you're preoccupied with other branches. I think as it is now, this is already a big improvement, so it makes sense to give users the benefit of using it.

Also just leaving a note to myself, we should make the resume feature use the new lastStartedPosition stuff

SubJunk:
The status tab now updates its elapsed time when fast-forwarding or rewinding. I was going to commit it to master but it's probably best to get some other renderers tested first, so for now it's just in this branch.

SubJunk:
This PR is being replaced by https://github.com/UniversalMediaServer/UniversalMediaServer/pull/736

Issue #551, opened 2015-05-05
NullPointer when updating progress bar


taconaut:
When leaving ums idle for a long time, the progress bar in the status tab visually changes from the custom one to a default progress bar. When updating it, a NullPointerException is being thrown. Can't tell what the trigger for this status bar change is.

```
05-05 08:08:48  Exception in thread "Thread-4" java.lang.NullPointerException
05-05 08:08:48  	at net.pms.newgui.GuiUtil$SegmentedProgressBarUI.setValues(GuiUtil.java:285)
05-05 08:08:48  	at net.pms.newgui.StatusTab.updateMemoryUsage(StatusTab.java:457)
05-05 08:08:48  	at net.pms.newgui.StatusTab$3.run(StatusTab.java:465)
05-05 08:08:48  	at java.lang.Thread.run(Unknown Source)
```

ExSport:
Spotted same thing with latest official v5.1.3 on Windows:
```
TRACE 18:00:37.450 [New I/O worker #8] Ignoring self-originating request from /192.168.1.3:51799
TRACE 18:00:37.450 [New I/O worker #6] Sent to socket: Content-Type: text/xml; charset="utf-8"
TRACE 18:00:37.451 [New I/O worker #6] Sent to socket: Cache-Control: no-cache
TRACE 18:00:37.451 [New I/O worker #6] Sent to socket: Expires: 0
TRACE 18:00:37.451 [New I/O worker #6] Sent to socket: Accept-Ranges: bytes
TRACE 18:00:37.451 [New I/O worker #6] Sent to socket: Connection: keep-alive
TRACE 18:00:37.451 [New I/O worker #6] Sent to socket: Server: Windows_Server_2008_R2-amd64-6.1, UPnP/1.0, UMS/5.1.3
TRACE 18:00:37.451 [New I/O worker #6] Sent to socket: Content-Length: 3027
INFO  18:01:22.563 [Thread-4] Exception in thread "Thread-4" java.lang.NullPointerException
INFO  18:01:22.576 [Thread-4] 	at net.pms.newgui.GuiUtil$SegmentedProgressBarUI.setValues(GuiUtil.java:267)
INFO  18:01:22.577 [Thread-4] 	at net.pms.newgui.StatusTab.updateMemoryUsage(StatusTab.java:459)
INFO  18:01:22.578 [Thread-4] 	at net.pms.newgui.StatusTab$3.run(StatusTab.java:467)
[...]

PR #677, opened 2015-09-20
Prevent memory bar ui from reverting to LAF default

skeptical:
This is a fix for #551 based on the observation that the error is easily reproduced by just calling [JProgressBar.updateUI()](http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/6-b14/javax/swing/JProgressBar.java#JProgressBar.updateUI%28%29), at which point the ProgressBarUI is reset to whatever the active LAF's default happens to be.  The assumption is that an eventual invocation of ```updateUI()``` in real life causes the bug.

To reproduce the error in the current master add the (hacky-ish :-) 'updateUI' button below and press it  on the gui:
```diff
diff --git a/src/main/java/net/pms/newgui/StatusTab.java b/src/main/java/net/pms/newgui/StatusTab.java
index 9dc1b19..e746b72 100644
--- a/src/main/java/net/pms/newgui/StatusTab.java
+++ b/src/main/java/net/pms/newgui/StatusTab.java
@@ -263,6 +263,13 @@ public class StatusTab {
 		mem.setForeground(fgColor);
 		builder.add(memoryProgressBar, FormLayoutUtil.flip(cc.xyw(3, 9, 1), colSpec, orientation));
 
+builder.add(new JButton(new AbstractAction("updateUI") {
+	@Override
+	public void actionPerformed(ActionEvent e) {
+		memoryProgressBar.updateUI();
+	}
+}), FormLayoutUtil.flip(cc.xy(3, 7), colSpec, orientation));
+
 		// Bitrate
 		String bitColSpec = "left:pref, 3dlu, right:pref:grow";
 		PanelBuilder bitrateBuilder = new PanelBuilder(new FormLayout(bitColSpec, "p, 1dlu, p, 1dlu, p"));
```

### DIFF
--- a/src/main/java/net/pms/newgui/GuiUtil.java
+++ b/src/main/java/net/pms/newgui/GuiUtil.java
@@ -5,6 +5,7 @@ import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
 import java.util.ArrayList;
 import javax.swing.*;
+import javax.swing.plaf.ProgressBarUI;
 import org.apache.commons.lang3.StringUtils;
 
 public final class GuiUtil {
@@ -383,6 +384,26 @@ public final class GuiUtil {
 		@Override
 		protected synchronized Color getSelectionBackground() {
 			return bg;
+		}
+	}
+
+	// A JProgressBar with a persistent custom ProgressBarUI.
+	// This is to prevent replacement of the initial custom ui with
+	// the laf's default ProgressBarUI as a result of future
+	// invocations of JProgressBar.UpdateUI()
+	public static class CustomUIProgressBar extends JProgressBar {
+		private ProgressBarUI ui;
+
+		public CustomUIProgressBar(int min, int max, ProgressBarUI ui) {
+			super(min, max);
+			this.ui = ui;
+			setUI(ui);
+		}
+
+		@Override
+		public void setUI(javax.swing.plaf.ProgressBarUI ui) {
+			// Always prefer our own ui if we have one
+			super.setUI(this.ui != null ? this.ui : ui);
 		}
 	}
 
--- a/src/main/java/net/pms/newgui/GuiUtil.java
+++ b/src/main/java/net/pms/newgui/GuiUtil.java
@@ -43,11 +43,15 @@ public final class GuiUtil {
 	}
 
 	// A progress bar with smooth transitions
-	public static class SmoothProgressBar extends JProgressBar {
+	public static class SmoothProgressBar extends CustomUIProgressBar {
 		private static final long serialVersionUID = 4418306779403459913L;
 
 		public SmoothProgressBar(int min, int max) {
-			super(min, max);
+			super(min, max, null);
+		}
+
+		public SmoothProgressBar(int min, int max, ProgressBarUI ui) {
+			super(min, max, ui);
 		}
 
 		@Override
--- a/src/main/java/net/pms/newgui/StatusTab.java
+++ b/src/main/java/net/pms/newgui/StatusTab.java
@@ -82,8 +82,7 @@ public class StatusTab {
 			playing.add(playingLabel);
 			time = new JLabel(" ");
 			time.setForeground(Color.gray);
-			rendererProgressBar = new GuiUtil.SmoothProgressBar(0, 100);
-			rendererProgressBar.setUI(new GuiUtil.SimpleProgressUI(Color.gray, Color.gray));
+			rendererProgressBar = new GuiUtil.SmoothProgressBar(0, 100, new GuiUtil.SimpleProgressUI(Color.gray, Color.gray));
 			rendererProgressBar.setStringPainted(true);
 			rendererProgressBar.setBorderPainted(false);
 			rendererProgressBar.setString(r.getAddress().getHostAddress());
--- a/src/main/java/net/pms/newgui/StatusTab.java
+++ b/src/main/java/net/pms/newgui/StatusTab.java
@@ -243,17 +243,16 @@ public class StatusTab {
 		builder.add(imagePanel, FormLayoutUtil.flip(cc.xy(1, 9), colSpec, orientation));
 
 		// Memory
-		memoryProgressBar = new JProgressBar(0, 100);
-		memoryProgressBar.setStringPainted(true);
-		memoryProgressBar.setForeground(new Color(75, 140, 181));
-		memoryProgressBar.setString(Messages.getString("StatusTab.5"));
 		memBarUI = new GuiUtil.SegmentedProgressBarUI(Color.white, Color.gray);
 		memBarUI.setActiveLabel("{}", Color.white, 0);
 		memBarUI.setActiveLabel("{}", Color.red, 90);
 		memBarUI.addSegment("", memColor);
 		memBarUI.addSegment("", bufColor);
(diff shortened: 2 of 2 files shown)


Nadahar:
@skeptical Great find. I suppose e.g. exiting a screen saver could trigger an update and thus trigger the bug. I have tested by connecting with remote desktop to a computer running UMS, and that can trigger the bug.

But, I don't understand why you reversed b3a2df216364c38d100e7e7f8605cfe5d6e38430. It only had general improvements, and even though synchronization wasn't the cause of this (albeight a bit strange my testing indicated that it got much more rare), it doesn't do any damage. The renaming of the variables definitly makes the code more readable imo, and the synchronization penalty is negligible since the memory update happens so infrequent. It would be nice to have the day some of the methods is called from anywhere else than the ```AWT-EventQueue```.

skeptical:
@Nadahar thx. You're right, I was mostly insulating this fix from b3a2df2 as a first step.

* Restored the formatting changes, which are harmless/positive as you say (e.g. making implicit ```private``` explicit, making functions ```public```, renaming variables, though this part did raise my eyebrows a little: '*because I have a small suspicion that this common name could trigger some underlying java context bug under certain circumstances*').

* Discarded the synchronization parts because it's better to know the efficacy of this fix by itself. If we ever want to call these functions from anywhere else we can put those parts in at that time, right now they don't pay their way imo. There was one relevant synchronization issue here, ```SwingUtilities.invokeLater()```, which you most excellently fixed already in 5815849.

As we've discussed at length before, I don't think going on synchronization witch-hunts is a good debugging strategy (among other things it's very likely to lead one astray :-). Weren't you throwing them in mostly in hopes that something would stick?

Nadahar:
@skeptical 
> 'because I have a small suspicion that this common name could trigger some underlying java context bug under certain circumstances'

This one is obviously invalid now, but I was wondering if some context confusion under certain circumstances could be a cause - one is a class variable, the other a local variable.

Regarding synchronization I disagree completely. First of all, the cost of doing uncontested synchronization is minimal, in most circumstances negligible. The only place the performance loss counts is where you do millions of iterations over one. If you mean performance loss because of blocking, that's a completely different matter, but in that case it is still better to take the performance hit than to suffer from memory corruption. If you suffer from performance loss from blocking, you need to redesign the logic so that the data flow is different.

Second, I never go on any synchronization witch-hunt. You've stated that you do, but I don't. My position is that any object that is accessed by multiple threads MUST be synchronized or otherwise made thread safe (e.g via immutability). The only issue there is that if I don't know the code good enough, I'm not always sure which objects is actually accessed by multiple threads, or that's likely to be in the future. Considering that the penalty is so minor that it's completely negligible in this case, I consider leaving the synchronization in place because it's not unthinkable that some other thread will want to call some of these methods in response to events.

Most importantly though, I wonder where you get the idea that synchronization is so expensive from. I did read that earlier java versions had a flawed synchronization implementation that made it quite expensive, but that hasn't been the case since Java 4 from what I understand. I'm not coming from Java, and I'm not used to consider synchronization to be a cost at all. The way I reason is that: Synchronization (in the broad sense, either one uses the keyword, other lock mechanisms or immutability) is a mandatory cost of multithreading. If the cost of this is unacceptable, multithreading is unnacceptable. In absolutely most cases however, these costs will be negligible compared to both performance gains from using multithreading (utilizing multiple cores and avoiding single thread blocking when waiting for slow loops or external API's etc) and code maintainability. It's so much easier to write threaded code than to write asynchronous code that handles everything in one thread in most cases that this alone makes it worth it (the complicated code will quickly become very bug prone, mutithreading allows us to separate or duplicate logic). The very idea of doing multithreading without synchronization is flawed however, the two goes hand in hand.

skeptical:
@Nadahar is it not possible to simply disagree without the long lecture? You are constantly finding straw men to knock down, e.g. "they don't pay their way" != "synchronization is so expensive". To me it's a fix looking for a bug, which is backwards (although here it, meaning lack of any established link between fix and bug, caused me to dig in more likely spots and find something, which is a plus I guess - time will tell).

However I'll restore all of b3a2df2 here if you feel strongly about it, since there's no real inherent conflict and no new bug and you're the author, not me.

Nadahar:
@skeptical I agree that you found the real bug, and there's a reason why I wrote "possible fix". All I could do was test, and I don't know what made it run so much longer without triggering the bug after synchronization. Normally when I've tested this, it would happen within 2 hours or so. When I said great find, it's because I haven't only looked at synchronization either, but I didn't manage to figure out what was going wrong and since there's threads involved and no synchronization, that's a natural suspect for me. I was in no way certain that I found the bug, but I couldn't see any harm in the things I did and considered them to be "general improvements" in any case.

I'm pretty sure you nailed it, because I've been suspecting something of that sort as well, without managing to fint it. I still find the swing component <-> UI <-> LAF correlation to be somewhat logically confusing and I have a hard time to figure out what belongs where.

I'm not meaning to lecture anyone, but I think the thread safety issue is an extremely important one and I'm only trying to explain myself. I'm used to getting TLDR's so that's just me, I have a hard time finding a way to say what I want to say in a short way. You can't judge my level of conviction by the length of the text, I can go into long journeys on subjects I'm really unsure about as well.

I believe UMS is suffering from a general lack of thread safety, which leads to all kinds of rare bugs. They are rare because the frequent ones will get fixed, but the rare ones will live on since it's hard to prove the link. I strongly feel that this must be addressed to fix the general stability, and I feel like a lone voice talking into a dark void. This is the case I'm arguing for, not this single synchronization which I agree probably does nothing.

My main point is that the very unpredictabile nature of concurrent thread access is the reason why a normal proof <-> fix correlation is impossible. Consider references that's corrupted by a concurrent read and write - which will point to an arbitrary memory location. The result of this won't be the same every time, you'd probably have to do it millions of times to even stand a small chance of replicating the exact resulting behavior. That's why I argue that the only way to handle concurrency is to protect any object that has the potential to conflict. This can lead to reduced "liveliness" or even deadlocks, but those are much easier to find in my opinion.

There's one thing that I'd like to know that's relevant for all GUI issues though: By which thread is the normal GUI code run? I mean, events, constructors, every code that belongs to the GUI objects. Is that all run by ```AWT-EventQueue```?

Nadahar:
@skeptical Even though I can't seem to trigger the bug simply by waiting after the synchronization commits, I found a way to trigger it. Rightclick desktop -> Personalization and then choose e.g ```Windows Classic```. This triggers the updateUI and also the bug. This PR fixes that, so I definitly think #551 is taken care of now.

I've started UMS on one of my "test VMs" (Windows Server 2008 r2) with the latest build now, and I will let it stay running for days if possible just to be extra sure.

skeptical:
> I found a way to trigger it... This triggers the updateUI and also the bug. This PR fixes that

Nothing like reproducing the error, I always say.  Extra points to you for persistence/integrity/cleverness. :+1: 

> My main point is that the very unpredictabile nature of concurrent thread access is the reason why a normal proof <-> fix correlation is impossible.

Yes, it's a magic bullet that can't ever be shot down, and all resistance is futile (sorry, couldn't resist :-).

Nadahar:
-> Yes, it's a magic bullet that can't ever be shot down, and all resistance is futile (sorry, couldn't resist :-)

...but the reason why I mean it must be treated in a "paranoid" fashion.

PR #689, opened 2015-10-08
Updates to checking file permissions instead of admin status

Nadahar:
Here is my take on the checking file permissions instead of admin status where relevant. This is what I've done:
* Made new inner class ```FilePermissions``` in FileUtil to hold file permissions information (I think information exchange in ```String``` is bad and error prone and should be avoided where possible. It gives no type safety whatsoever).
* ```FileUtil.getPathPermissions``` renamed to ```FileUtil.getFilePermissions``` and rewritten to use FilePermissions class
* Rewrote ```PMS.killOld()``` and sub methods' error handling
* Updated ```PluginTab``` to use new ```getFilePermissions()``` logic
* Made ```PluginTab``` verify that ```plugins``` folder exists
* Updated ```AutoUpdateDialog``` to use new ```getFilePermissions()``` logic
* Updated ```FileUtil.isFileExecutable()``` to also check for root permission on Linux
* Moved ```isAdmin()``` from ```PmsConfiguration``` to ```FileUtil``` and made it static
* Rearranged translations strings to use new strings instead of modify old ones
* Modified the windows installer to make the ```plugins``` folder writable for all users


Nadahar:
This became much, much bigger than intended. The reason is that I did some debugging/testing in preparation for a merge. That made me find all kind of issues, most not caused by this PR.

I've included them here anyway because the alternative would be to make a "Misc fixes" PR - which wouldn't make much sense either.

The root problem I discovered was that the ```canRead```, ```canWrite``` and ```canExecute``` methods is completely fubar. Some genious at Java decided once upon a time that these didn't have to check ACL's. While that makes a little bit sense in Linux, it's completely useless in Windows since NTFS permissions are pure ACL's. That again means that under Windows, these methods only report the FAT permissions. Since the only permission that existed under FAT was "read only", that's allso all they can report, and that flag isn't even used at NTFS. In short, they will always return true on NTFS.

This had been dealt with in ```FileUtils``` with ```isReadable``` and ```isWritable``` methods that actually attempted to open files for reading or writing, and attempted to list folder contents or write a temp file to a folder. This is far from ideal though, creating unnecessary IO and having side effects like wearing out SSD's for no reason and changing "last access time" on files and folders. It wasn't possible to do an executable test in a similar fashion that was the least bit reasonable either.

As a consequence, and since Java 6 is declared abandoned by us, I decided to rewrite the permission checks to use the ```java.nio``` methods available in Java 7 that actually can get this right. I've tested the new ```FilePermissions``` class on Windows 7, Centos 7 and OS X 10.8 without finding any issues. **This means that after this is merged there's no going back to Java 6**.

I've removed the old methods and rerouted all code to the new. They were'n very much used in any case, and they shouldn't be since this permissions shouldn't be used to determine if an upcoming read or write will succeed since things can change in the meanwhile. In principle neither should ```File.exists()``` but that's to big a job to rewrite to even contemplate. Proper ```Exception``` handling from a failed operation seems to be the "correct" way in most circumstances. Permission checking does have it's uses though, and with this PR I hope that what we get back will actually be accurate.

The other fixes to things I've discovered while testing is in here in no particular order (or, by order of discovery) but I've at least managed to get them into separate commits most of the time.

What I've done is:
* Fix concurrency issues in ```FileUtil.getFileCharset()```:
  - Expanded with new methods ```getFileCharsetMatch()``` and ```getFileCharsetName()```
  - Changed the return type of ```getFileCharset()``` to ```Charset```
  - Created overloaded methods ```isCharsetUTF8()``` and ```isCharsetUTF16()``` that accept ```Charset``` as argument
  - Updated all calls to the original ```getFileCharset()``` with one of the three methods
* Moved actual permission checking from ```getFilePermissions()``` into class ```FilePermissions```.
* Renamed ```FilePermissions``` methods from ```canRead()```, ```canWrite()``` and ```canExecute()``` to ```isReadable()```, ```isWritable()``` and ```isExecutable()``` to emphasize that they don't suffer from the same flaws as the ```can...()``` named Java methods
* Removed the following methods from ```FileUtil``` and rewrote all calls to use one of the ```FileUtil.getFilePermissions()``` variants or exception handling instead:
  - ```isFileReadable()```
  - ```isFileWritable()```
  - ```isDirectoryReadable()```
  - ```isDirectoryWritable()```
* Removed the corresponding tests from ```FileUtilTest```
* Created tests for ```getFilePermissions()``` in ```FileUtilTest```
* Fixed a bug where any issues while getting the ```Locale``` could be logged multiple times
* Created constant ```PmsConfiguration.DEFAULT_CREDENTIALS_FILENAME``` for consistancy
* Rewrote ```PmsConfiguration.InitCred()```:
  - Made it work with ```File``` instead of path of ```String```
  - Made it save credentials path only if template file is successfully written
  - Made it warn if credentials path can't be saved
  - Made it use system dependent newline so that credentials file can be edited in text editors on any OS
  - Made credentials file ```UTF-8```
* Moved setting of credentials default filename from initCred() to getCredFile() using ```DEFAULT_CREDENTIALS_FILENAME```
* Deprecated ```PmsConfiguration.getCredPath()``` as it's no longer needed
* Modified ```RootFolder.addWebFolder()``` to handle ```FileNotFoundException``` instead if ```File.exists()``` to better handle permission issues
* Fixed an instance call to a static method in ```RootFolder.loadWebConf()```
* Fixed an NPE during shutdown if ```upnp_enable = false``` in configuration
* Fixed a bug where ```LooksFrame.quit()``` would throw an ```Exception``` if configuration couldn't be saved.
* Moved inner class ```FileUtil.FilePermissions``` to independent class
* Created a ```FilePermissions``` constructor with a ```Path``` as argument
* Created ```FilePermissionsTest``` with tests for the new class
* Minor finishing touch to ```FileUtilTest.testGetFilePermissions()```

I'll try to do some more testing, but a lot of this is difficult to test since the references are from very different places in the code where I simply don't know how to simulate some of the situations. I'll appreciate any help with testing - that is anyone that would give this PR a go and see if they have some (new) issues.

SubJunk:
@Nadahar great! It looks like you did a lot of work here so thanks for that. I'll test it out. Do you have any particular behaviors/actions you want tested?

SubJunk:
I get a test failure when building:

Running net.pms.configuration.PmsConfigurationTest
Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.307 sec <<< FAILURE! - in net.pms.configuration.PmsConfigurationTest
testLoggingConfigurationDefaults(net.pms.configuration.PmsConfigurationTest)  Time elapsed: 0.306 sec  <<< FAILURE!
java.lang.AssertionError: DefaultLogFilePath
        at org.junit.Assert.fail(Assert.java:88)
        at org.junit.Assert.assertTrue(Assert.java:41)
        at net.pms.configuration.PmsConfigurationTest.testLoggingConfigurationDefaults(PmsConfigurationTest.java:62)

Failed tests:
  PmsConfigurationTest.testLoggingConfigurationDefaults:62 DefaultLogFilePath

Nadahar:
@SubJunk I just tried again, and I get no failure at any test. That means it must be environment dependent, which is bad in itself. Can you see what is the actual and expected result of the failing test? Looking at the test in question, the test itself is flawed. I didn't know at the time I wrote that, but ```File.isFile()``` actually requires that the file exists on the file system. You probably simply doesn't have a file ```debug.log``` in the expected (default) location. I think line 62 simply must be deleted, that test doesn't make much sense anyway.

Regarding what to test: I have tried to test any specific case I can find, so I'm mostly looking for "anything out of the ordinary". The best thing to do to be specific is probably to look at the PR diff and see which parts of the code is altered and test things that relate to that. I'm not done looking at the testing myself, so if I find anything in particular that needs testing that I'm not able to test myself, I'll report it here.

Nadahar:
@SubJunk I've removed the test in question - it made no sense anyway. When I wrote it I was thinking that ```isFile()``` somehow magically analysed if it were a file or a folder without actually accessing the file system, but when I think about it that's not possible. Tests shouldn't rely on external circumstances, and it wasn't important any way.

Nadahar:
To anyone that wants to help test, this is a list of what's touched and could be affected by a new bug. Most of the things are relatively uncomplicated and doesn't need extensive testing as I see it. I have tested the reading of permissions itself relatively thoroughly. This is where most of the new code is. 

If anyone test any particular list item, please let me know and I'll update the list with the results.

* Reported permissions early in the log (by ```displayBanner()```). I think I've tested this fairly well.
* Any problems with the temp folder. Not that easy to test, I haven't experienced any issues on my test installations.
* Any issues with reading or writing the credentials file. I don't have anything to use the credentials file for, so I haven't tested this further than seeing that the entries are displayed in the GUI.
* Any *additional* issues with "run single instance". This has issues also without this PR, so it's about to spot if anything has gotten worse. Not tested by me.
* Any problems displaying virtual folders from file? Minor changes in ```parseVirtualFolders``` and ```FileSerializer.deserialize()```, but I'm not sure how it's being used or how to test it. Not tested by me.
* Any issues with ```NameFilter``` (UMS.deny). Not tested by me.
* Any issues with loading/reading the profile from file. I haven't observed any issues, but it's hard to simulate possible scenarios.
* Any issues finding the default log file folder. I have observed no issues, but the failed test above could indicate that the modified logic doesn't always come to the same conclusion. It would be very interesting if you @SubJunk could verify that it does find the same folder as earlier on your computer, since that's where the test failed.
* Any issues with subtitles from external file. Not tested by me, not sure exactly how it's being used. Do we stream subtitles from external files in some occations, or are they always transcoded?
* Any issues loading/reading the web configuration. Modification is unlikely to cause trouble, I haven't observed any issues.
* Any issues with the permission check for auto update. Not tested by me.
* Any issues with charset/encoding detection in external subtitles. Not tested by me.
* General search for Warn and Error log entries.

Nadahar:
@SubJunk Have you tried to build the updated version? I'm curious if you have any issues with the log file location?

SubJunk:
I just checked and the log file and profile locations are the same as master

Nadahar:
I've added a couple more things I came across during testing:
* Updated ```ExternalFactory``` to use ```FilePermissions``` when checking for access rights to the plugins folder when looking up plugins.
* Updated ```PluginTab```: 
  * Rearranged the layout some to try to use more of the tab. The credentials table will now grow if the windows is resized, and it will fill out the form. This isn't ideal as the plugins table won't scale if more plugins are added, but it's better than it was. I know this is a stretch to put into this PR, but I was testing the credentials functions and was annoyed by bugs in the UI, and one thing led to another.
  * Display vertical table scrollbars only when needed.
  * Hide the "installed plugins" section if no plugins are installed, simply because it seems like a bug when there's a header with no content.
  * Made it save any changes made to the credentials before opening the credentials editor, and to reread the file and refresh the table if the credentials editor is closed with "save". This is the bug that got me started messing with ```PluginTab```.
  * Added some exception logging

Nadahar:
@SubJunk 
> I just checked and the log file and profile locations are the same as master

Great, thanks.

Nadahar:
If any of you feel like testing any of that listed above, this PR is ready as soon as it is tested.

Nadahar:
@SubJunk I'm working on the Linux trash functionality in the [freedesktop.org branch] (https://github.com/UniversalMediaServer/UniversalMediaServer/tree/freedesktop.org), but I need methods from this branch to get any further. 

As stated before, I've done the testing I know how to, is this ok to merge (provided I fix the merge conflicts)?

Nadahar:
@SubJunk Merge conflicts solved. That merge wasn't fun, this PR has been lying around too long.

SubJunk:
I'm okay with it merging if no one objects

Nadahar:
@SubJunk It hasn't been any objections here for a long time, I've discussed some details both with @skeptical and @valib here eariler and we came to agreement about those changes. This is mostly small changes and bugfixes, and some minor GUI changes.

Issue #694, opened 2015-10-11
Renderer config files that hopefully can be added


ler0y:
Hello UMS Developers,

I just want to mention here on Github the renderer config files (profiles) that I have wrote for the 'BubbleUPnP Media Renderer-MX Player' and 'Windows Media Player Renderer' posted on the forum.
I hope they can be included into UMS.

BubbleUPnP Media Renderer-MX Player:
http://www.universalmediaserver.com/forum/viewtopic.php?f=5&t=2574&start=10#p18473

Windows Media Player Renderer:
http://www.universalmediaserver.com/forum/viewtopic.php?f=5&t=4149

They are though fully tested by me and works great :-)
All the 'Supported =  ' formats and codecs were tested with the media files that I have collected in years.
All related notes/issues can be found in the config files.

Sami32:
@ler0y @SubJunk Since these PR https://github.com/UniversalMediaServer/UniversalMediaServer/pull/697 include it already and this recent PR https://github.com/UniversalMediaServer/UniversalMediaServer/pull/853 update the WMP renderer, could this issue be closed ?
And what about the PR https://github.com/UniversalMediaServer/UniversalMediaServer/pull/697, do you want merge as it is now ?

PR #697, opened 2015-10-15
Add BubbleUPnP and WMP renderer profiles (#694)

Nadahar:
I've made this PR based on #694. This is work in progress so far.

Renderer profiles added:
- `Google-Android-BubbleUPnP-MXPlayer.conf`
- `Microsoft-WindowsMediaPlayer.conf`


Nadahar:
@ler0y I have some questions:
- The BubbleUPnP-MXPlayer is based on using those two together, right? I don't know this well enough, but what does BubbleUPnP play "natively"? I'm wondering if it would be an idea to also include a file based solely on BubblesUPnP as well, where most things are transcoded. Or is that a useless idea? I've made a quick example icon that could be used for the combined profile, scroll to the bottom of [Files changed] (https://github.com/UniversalMediaServer/UniversalMediaServer/pull/697/files) to see it.
- I installed the free version of both programs on a Samsung tablet with Android 5 something. It kindof worked, but the navigation was sluggish and a little confusing to me. I got into a never ending loop a couple of times when I played short audio samples. It would finish playing the file and restart so quickly that I couldn't abort it, and had to "kill" BubbleUPnP to get out of it. I realize that this has nothing to do with the UMS profile, but my question is: Is this combination very unstable/unmature, or do you have to have the payed versions for it to work properly?
- Windows Media Player: Is this profile only for playing via push? I can't find any media when browsing in WMP. I can't see much happening in the log either. I don't know what the deal is with WMP, but the old profile don't work either. It seems to me like WMP, just like VLC, asks to receive the whole media library first and then sort and categorize it themselves. Is the problem that UMS doesn't handle this, or is it some other problem?

ler0y:
@Nadahar Thanks for creating this pull request :)

> The BubbleUPnP-MXPlayer is based on using those two together, right? I don't know this well enough, but what does BubbleUPnP play "natively"? I'm wondering if it would be an idea to also include a file based solely on BubblesUPnP as well, where most things are transcoded. Or is that a useless idea? I've made a quick example icon that could be used for the combined profile, scroll to the bottom of Files changed to see it.

BubbleUPnP can only play most audio formats natively with its own audio player.
The audio formats that are natively supported with this player, will be played with this audio player.

Audio formats that are not natively supported with the bubbleupnp audio player can be played with the MX Player when an unknown video MIME type is sent. (supported lines with "video/blabla" MIME type)
Here for, the "Mime-type check" (Settings >> UPnP Tweaks) in the BubbleUPnP app must be unchecked.  (BubbleUPnP will auto ask to change this setting)

Note that if the "Supported = " audio format lines with the video/blabla MIME type are removed from the BubbleUPnP-MXPlayer config, UMS transcode the not natively supported audio and they are played with the BubbleUPnP audio player. 
However seek/FF while playing then fail.

Because BubbleUPnP can only play audio formats, it pass through the video files to an additional software video player installed on your Android device.
Where the basic Android video player and most other software players failed, the MX Player can handle almost every media type, including internal and external subtitles.
Also seeking through video/audio while it is played works nicely.

Video transcoding by UMS to the MX Player fails. After some time the video playback somehow freezes.
However DVD ISO files (not natively supported by MXPlayer) transcoded by UMS play fine without this issue. (however no seek support)

I am not sure the combined icon is a good idea, because UMS don't know if MX Player is installed on the device, and maybe also other software video players can be used.

> I installed the free version of both programs on a Samsung tablet with Android 5 something. It kindof worked, but the navigation was sluggish and a little confusing to me. I got into a never ending loop a couple of times when I played short audio samples. It would finish playing the file and restart so quickly that I couldn't abort it, and had to "kill" BubbleUPnP to get out of it. I realize that this has nothing to do with the UMS profile, but my question is: Is this combination very unstable/unmature, or do you have to have the payed versions for it to work properly?

The free versions should work fine, the paid versions only removes advertisements.

The issue with short audio samples might be because the same file is played over and over again, instead of that playback stops after the track is finished.
Maybe you have enabled repeat?

> Windows Media Player: Is this profile only for playing via push? I can't find any media when browsing in WMP. I can't see much happening in the log either. I don't know what the deal is with WMP, but the old profile don't work either. It seems to me like WMP, just like VLC, asks to receive the whole media library first and then sort and categorize it themselves. Is the problem that UMS doesn't handle this, or is it some other problem?

This profile indeed is only for playing via push to the "Windows Media Player Renderer". The "Windows Media Player Sharing" has never worked with PMS/UMS.

Nadahar:
@ler0y 
> I am not sure the combined icon is a good idea, because UMS don't know if MX Player is installed on the device, and maybe also other software video players can be used.

UMS obviously can't tell the difference, and one profile would need to have priority over the other (I'm not sure which should be "default"), but if both were there it would be simply to disable the profile one didn't want to use. The reason I would want the tie to MX player to be indicated is the fact that the supported lines presuppose it - it won't work very well for anyone just using BubbleUPnP. The "BubbleUPnP only" profile should just stream the audio formats supported by BubbleUPnP and whatever video formats supported "natively" by Android (if any), and the rest would be transcoded. I know that means no seek support, but that's better than files not playing at all because the renderer can't play them.

ler0y:
@Nadahar 
> UMS obviously can't tell the difference, and one profile would need to have priority over the other (I'm not sure which should be "default"), but if both were there it would be simply to disable the profile one didn't want to use. The reason I would want the tie to MX player to be indicated is the fact that the supported lines presuppose it - it won't work very well for anyone just using BubbleUPnP. The "BubbleUPnP only" profile should just stream the audio formats supported by BubbleUPnP and whatever video formats supported "natively" by Android (if any), and the rest would be transcoded. I know that means no seek support, but that's better than files not playing at all because the renderer can't play them.

Originally I was started with creating/testing profiles for both the "Windows Media Player Renderer" and the "BubbleUPnP Media Renderer" without any additional software or codec packs installed.
But for the best natively support and performance in general, I ended with the mentioned renderer profiles and the additional software installed.

Renderer profiles for the BubbleUPnP, compatible with the by default on device installed Android video player, for the devices I own, I posted already in past on forum.
I have not tested them since then, because I found that the MX Player with custom codec for MX Player installed, works best on my newer (transcode capable) mobile Android devices.
http://www.universalmediaserver.com/forum/viewtopic.php?f=5&t=2574#p13446

The by Android "natively" supported formats can be found here:
http://developer.android.com/guide/appendix/media-formats.html

According the supported media formats mentioned in this API guide, these differs per Android platform version.
Also note that any given mobile device may provide support for additional formats or file types.
So in the ideal world, I guess there should be renderer profiles available for:
* Any Android platform version.
* Android 3.1+ (additional FLAC audio support)
* Android 4.0+ (additional VP8 video support in Matroska)
* Android 4.1+ (additional AAC ELD support)
* Android 4.4+ (additional VP9 video support with WebM)
* Android 5.0+ (additional Opus audio support with Matroska and H.265 video support)
* One for any mobile device that has support for additional formats or file types
* MX Player (for devices capable to do on device transcoding)
* MX Player with the custom codec pack installed (for additional natively AC-3/DTS support)

> but a sluggish touch interface full of ad banners

The licence for BubbleUPnP can be bought for the price of less than a couple of beers.
I however never experienced "sluggishness" using the BubbleUPnP.
Sure, it would be great if there is a (open source) free program (like UMS) be available for Android with same capabilities BubbleUPnP provide. However I haven't found any yet.

> The current WMP profile makes no sense whatsoever, it has audio, video and image set to false so no content is actually available. I'd like to hear the history behind this.

I guess the WMP profile was added to prevent that WMP is recognized it as an Unknown renderer.

Maybe there should also be a "Windows Media Player Renderer" profile available that works without any additional codec pack installed?
Because that's perhaps also only installed by the more advanced PC user.

Nadahar:
@ler0y I can only speak for what I think, since the other guys are very quiet about this. My goal is to make things reasonable simple to use and work. As I see it, compromises must be made, the advanced user can easily add support for a particular format/codec if he/she wishes to do so.

> So in the ideal world, I guess there should be renderer profiles available for:

That many profiles would just confuse people. If you look at the current [Android version distribution] (https://developer.android.com/about/dashboards/index.html) you will see that anything prior to 4.1 is irrelevant. If you also look at the [distribution 1,5 years earlier] (http://androidandme.com/wp-content/uploads/2014/01/androiddistributionjanaam.jpg) it would be reasonable to believe that most user would use 4.4 and higher within 6 months or so.

I think one version for Android 4 and one for Android 5 would do, the tricky part is if the Android 4 version should include VP9 or not.

> The licence for BubbleUPnP can be bought for the price of less than a couple of beers. I however never experienced "sluggishness" using the BubbleUPnP.

Sure, if I was going to use it on a regular basis I'd buy it. But I don't fancy buying software just for testing something like I was now, and it was more the placement of the banners than the banners themselves that bothered me. I think Android is sluggish in general, but I haven't tested that many different devices. I do think it behaves strangely though because if often flows just fine 2/3 of the time, and then for no apparent reason is very sluggish at other times. I'm was using a Galaxy Tab 4 (SM-T535) with Android 5.02 when doing the testing, and that was quite laggy. I think Android 5 is a bit too heavy for that device. In any case, all put together, it made it difficult to actuall know if I had activated some kind of repeat function by accident.

> I guess the WMP profile was added to prevent that WMP is recognized it as an Unknown renderer.

If that is actually the case, we could just replace the current profile with one that supports push at least.

> Maybe there should also be a "Windows Media Player Renderer" profile available that works without any additional codec pack installed?
Because that's perhaps also only installed by the more advanced PC user.

I never actually use WMP for anything, but I never play media on Windows without installing codecs either, so I don't know what WMP can play "out of the box". Not a lot I suspect.

The reasonable thing to be would probably be to have one WMP profile without any additional codecs and one with one of the known codec packs installed and nothing more. That would make it manageable for users to use, and again, it's easy to modify an existing working profile to support a couple of extra codecs.

About the BubbleUPnP+MX setup, I think the threshold for going to a forum at downloading and installing a custom codec pack on an Android device is somewhat high. Mobile device users are generally much more clueless than computer users, that's the whole appeal of the mobile platform. It's hard to know if AC-3/DTS should be included - one option would be to include it but comment it out and write a comment that explains how to enable it in the profile itself. What's the story behind MX player ditching AC-3/DTS? Will it return? That should also be taken into consideration.

ler0y:
@Nadahar
>What's the story behind MX player ditching AC-3/DTS? Will it return? 

Support for dts, dtshd, ac3, eac3 or mlp audio was removed from the official MX-Player in version 1.7.33 due to license issue:
http://techposts.org/mx-player-support-dts-and-ac3-audio-using-custom-codec-how-to/

>one option would be to include it but comment it out and write a comment that explains how to enable it in the profile itself.

I have updated the BubbleUPnP/MX Player profile.
It is now compatible with the official, in the Google Play Store available codec:
http://www.universalmediaserver.com/forum/viewtopic.php?f=5&t=2574&p=20858#p20858

I posted the profiles on forum so they can be tested by users and issues with it might be reported.
Shall I remove, the MX Player custom codec compatible profiles posted on the forum?
Because it is absolutely not my intention to harm the UMS project in any manner or way by posting illegal? profiles.

>I never actually use WMP for anything, but I never play media on Windows without installing codecs either, so I don't know what WMP can play "out of the box". Not a lot I suspect.

The file types supported by Windows Media Player can be found here:
https://support.microsoft.com/en-us/kb/316992

Soon I will test again Windows Media Player without any codecs installed.

Nadahar:
@ler0y I can't see how UMS is doing anything illegal by not transcoding those codecs even if MX Player lack the license. I find it odd though, there are many players out there that can decode those formats that doesn't pay gigantic licensing fees to Dolby and DTS. If anything, it's probably transcoding not streaming the formats that's illegal from UMS's side, since transcoding involves decoding them. But then again, I'd think they'd have to go after the decoding libs themselves in such a case. Maybe MX Player is US based, they have to most insane protection in place to protect the wealth of the big companies.

atamariya:
@Nadahar, WMP expects the CDS to support UPnP search capabilities (as does Xbox). I've been able to access UMS via "Other Libraries" in WMP. You might want to use my code.  https://github.com/atamariya/UniversalMediaServer.git

Demo video http://lifeofpenguin.blogspot.com/2016/08/streaming-home-media-via-upnp.html

untitled

Nadahar:
@atamariya Great, I haven't really looked into what's been missing for WMP since the team haven't shown a lot of interest for it. I know that the renderer profile for WMP is bound to fail as well, since it basicly shuts off all media types for WMP among other things. I understand that WMP isn't the most interesting renderer, but I've come across several others software renderers that is recognized as WMP (same user agent etc.) that doesn't work because of it.

Your approach in #987 is much better than the current system imo, but there is so much legacy functionality and plugins built on the existing logic that I don't know if any such major changes will be accepted.

Nadahar:
I doubt this is up to date now, it's very old. That said, this won't make WMP work with UMS as those problems are on the DLNA level. All this will do is that it's possible to "push" content to WMP.

Nadahar:
I've updated the branch. I think there should be one renerer configuration for just BubbleUPnP and one for BubbleUPnP + MXPlayer, as what is supported is very different. That's also why I added icons for both options.

Also, the WMP profile need a cleanup in the User Agent/UPNP Search strings. Why don't you fix a little with this @Sami32 (in a second commit).

Sami32:
@Nadahar I post a PR already and you saw the result...so i'm not interested anymore by WMP renderer upgrade.

In contrary, users seem interested by BubbleUPnP + MXPlayer. This combination was already well tested by @ler0y  and @atamariya .
Adding one more for BubbleUPnP ? why not.

BTW, thank you very much to have upgraded this branch as i was in trouble ;-)

Sami32:
https://github.com/UniversalMediaServer/UniversalMediaServer/pull/987/commits/36ee01fc793019d6f6684bc7958f49c49e5935f5

https://www.bubblesoftapps.com/bubbleupnpserver/#changelog
https://www.bubblesoftapps.com/bubbleupnpserver/#android_configuration_transcoding

Nadahar:
@Sami32 I'm not sure what to look for in this link. Is this now ready and working?

Sami32:
@Nadahar It's informationnal links about BubbleUPnP formats support and Chromecast.

As i don't have any Android smartphone, don't rely on me for testing ;)
I though that you already did. Anyways, some people did already tested it.

Nadahar:
@Sami32 It's been soo long since I dealt with this, that I don't remember what I've tested and not. I'm just saying that I remember that we have a problem with UMS recognizing any Android device as "Chromecast". The Chromecast configuration catches too much. I haven't looked into why though.

Sami32:
@Nadahar Ok, i see.
Since BubbleUPnP Chromecast and our Chromecast library have been updated, it worst waiting for a new test.

Nadahar:
I'll merge this as I badly need BubbleUPnP support when testing #1287. Even though this isn't ideal as it is (doesn't separate between just BubbleUPnP or BubbleUPnP + MXPlayer), it's better than not having a profile at all. If anyone want to add more to this later, merging it as it is now should be no hinder.

ler0y:
@Nadahar 
>Even though this isn't ideal as it is (doesn't separate between just BubbleUPnP or BubbleUPnP + MXPlayer), it's better than not having a profile at all.

The BubbleUPnP player only support audio. The supported ones will be played with the BubbleUPnP audio player with this profile, so even without the MXPlayer installed this profile will cover the BubbleUPnP supported audio files.

Nadahar:
@ler0y Yes, but BubbleUPnP can be used with other players as well, with different format support. There's really no good way to solve this that I can see right now, so I guess this is ok for now.

PR #699, opened 2015-10-18
Use Splash Screen to avoid users to be confused when the UMS GUI start delays

valib:
For me sometimes the UMS GUI starts immediately but quite often there is almost the 10 seconds delay before the GUI appears. To avoid to be users starting UMS several times when there is that significant delay I offer to show the Splash Screen.


valib:
Some errors appears. I will make the PR again.

PR #700, opened 2015-10-18
Use Splash Screen to avoid users to be confused when the UMS GUI start delays

valib:
For me sometimes the UMS GUI starts immediately but quite often there is almost the 10 seconds delay before the GUI appears. To avoid to be users starting UMS several times when there is that significant delay I offer to show the Splash Screen.


Nadahar:
@valib Good idea. I have two comments:
* Why do you keep the image object in ```LooksFrame```? Can't that just be helt by ```Splash``` itself? Or did you think that it should be possible to load different splash images?
* Even though it can be good for newbies to see that something is happening, splash screens are typically annoying to experienced users. Therefore it's common practice to offer to disable splash screens, and I think we should too. There should be a configuration setting for it, and a checkbox in ```GeneralTab``` close to the other "startup options" like one instance, autostart etc.

valib:
@Nadahar thanks for the input. the first comment I accept and will change that. The same should be applied to the progress bar. We should decide whether or not to display it and it should be also hard-coded. I made it changeable only for testing.

I have a problem with your second comment. Yes to show the splash screen is obviously configurable but in this case the delay of the UMS GUI is unpredictable. Usually the GUI start instantly and the splash screen only blink and disappear. But as I wrote sometimes it takes more than 10 sec and in that case it is annoying that user doesn't know what is going on. I don't know if it is caused by the OS (WIN 10) or by yours changes in the language detection or implemented synchronizations.

Nadahar:
@valib 
> I have a problem with your second comment. Yes to show the splash screen is obviously configurable but in this case the delay of the UMS GUI is unpredictable. Usually the GUI start instantly and the splash screen only blink and disappear. But as I wrote sometimes it takes more than 10 sec and in that case it is annoying that user doesn't know what is going on. I don't know if it is caused by the OS (WIN 10) or by yours changes in the language detection or implemented synchronizations.

I haven't made any difference in language detection yet (those things are still in the "crowdin" branch"). What I've changed is the tags being used for languages, and more use of ```Locale``` instead of just a ```String``` with a tag. I doubt that has anything to with any added delays. 

When it comes to synchronization, I haven't done much of that, and as discussed over and over again that doesn't cost much at all except if you go though the locks millions of times or you end up having the threads wait for eachother. I don't have to complete overview over what's happening during startup, but to me it seems like the startup is mostly run by the "main" thread, and as such there shouldn't be much room for blocking.

But, if you've made the splash screen to compensate for a delay that's come recently, we should probably try to figure out what's causing the delay instead. Do you know at which version this added delay started?

Nadahar:
@valib I just checked out this branch and put some loggers in for timing. Here is what I see:

```logtalk
ERROR 22:19:02.750 [main] main start
ERROR 22:19:03.295 [main] Creating PMS instance
ERROR 22:19:03.296 [main] Starting Init
ERROR 22:19:03.326 [main] Creating splash
ERROR 22:19:03.487 [main] Done creating splash
ERROR 22:19:03.524 [main] Creating LooksFrame
ERROR 22:19:04.981 [main] Done creating LooksFrame
ERROR 22:19:05.046 [main] Disposing splash
```

valib:
@Nadahar I don't know when it started and as I wrote it could be the WIN version problem. I tried to figure out if it is related to your commits or not but it was not clear. Maybe it is only problem with my WIN setting or something like that. Actually it started when you made that changes but almost in the same time I changed the OS  from WIN 7 to WIN 10 so I can only guess where is the problem. :-)

Nadahar:
@valib I can push the test logs I've made now, and you can see what results you get. I'll remove that commit from the PR again later (without reversing, but by reseting so it won't be visible).

Nadahar:
Pull this PR now and filter the log on "error" (so that only these will show). It should give an indication of where the waiting is.

valib:
@UniversalMediaServer/developers if no objection I would merge this PR.

Nadahar:
@valib I haven't look at your latest changes, will check it out.

Nadahar:
@valib I still think there should be an option to disable the splash screen.

SubJunk:
@valib it's definitely necessary to be able to disable it, ideally by clicking a link on the splash screen like "click here to not see this again". Even then I'm not sure, splash screens are annoying IMO...

valib:
@SubJunk the delay caused by Splash is not significant and the time while it is shown is caused by the UMS GUI initialization. For me it usually takes one or two seconds and clicking on it is almost impossible but sometimes it takes almost 12 second and when showing nothing may make users nervous and like me try to start UMS again.

Nadahar:
@valib On slow computers loading the splash itself can take maybe 2-3 seconds extra, and I also have a bit of a hate for splash screens (simply because they do nothing useful). I do agree that they serve a purpose when loading times are long though, so that you know that something is happening.

But, you're saying that the loading time still is 12 seconds some times. Didn't we fix that? I'm pretty syre that whatever is causing that particular problem has to do with disk access on your computer. I have been running UMS testing on some pretty resource starved VM's though, so I've been waiting 30 seconds or more for it to launch. In such cases it's nice to know that something is happening, but I'm not sure it's worth 5 seconds extra waiting time for creating the splash :wink:

valib:
@Nadahar I don't know how slow is your PC but for me the Splash creation takes max 100 milliseconds.

Nadahar:
@valib My "main" computer has no problems, but I've tested it on virtual machines (VM's) that's been quite resource starved. In such cases I'm betting the creation can take a considerable amount of time, but such a slow computer would have no chance to do transcoding so I don't know how relevant that is.

I simply think that many, like me, "hate" splash screens and that to not alienate that crowd we should give them the option to skip it. I don't think turning it off should be at the splash screen though, because as you say, it's hard to click anything in the short time it's normally there.

valib:
@Nadahar, @SubJunk OK. it is not the problem to make it configurable like @Nadahar suggested. I will try to make it.

valib:
@SubJunk, @Nadahar what about this change?

src/main/java/net/pms/configuration/PmsConfiguration.java
```diff
@@ -3830,5 +3833,14 @@ public void setScreenSize(String value) {
 
 	public String getScreenSize() {
 		return getString(KEY_SCREEN_SIZE, "-1x-1");
-	}	
+	}
+
+	public boolean showSplashScreen() {
```
SubJunk: This should be isShowSplashScreen

SubJunk:
@valib I'll give it a try and will update the image before merging

valib:
@SubJunk feel free to change the image. I am not a designer to be able to make it good looking :)

SubJunk:
What do you think of this one? The lettering is pretty bold but I like the effect of having the text extend farther than the logo when there's a transparent background. I'm still not convinced a splash screen is a good addition to the program but I'm willing to go with the majority on this one.

Nadahar:
@SubJunk I think it looks ok, but I'm not sure about using "UMS". It's one thing that we say it all the time and can't be bothered to spell it out, but I'm not sure it will give that much meaning to everybody. Actually I think ```Loading``` and the logo would be enough, but then you don't get the "sticking out effect". Alternatively "Loading" on one line and "Universal Media Server" on the next is my two cents.

When it comes to having a splash or not, don't count me on either side. I'm kinda on the fence here, not really sure if I think it's a good idea. As I've said before, I think it's a good idea when software take a long time to start, so that the user knows that something is happening. On the other side, I'm not really sure that applies to UMS under normal circumstances and I find splash screens a little annoying in general. But, having tested this I can't really say it does any damage, you hardly get to see it. I guess what makes most sense to me is that it doesn't hurt much, and if UMS for some reason should take a long time to load under some circumstances those users won't start it 9 times in the meanwhile. The real problem lies elsewhere as I see it, so much work is being done in the constructor - including checking that the shared folders exist which can involve disc-spin ups, network connectivity issues etc.

ExSport:
Some programs have funny messages when some action takes unexpectedly long time to finish like "building atomic clock", "expanding universe", etc. So if it annoys someone when it is displayed every time what about to show it only for unexpectedly long occasions and avoid it for headless systems completely to avoid crashing😇

Nadahar:
@ExSport I don't know how we're supposed to predict how long it will take. It's not shown in headless mode.

ExSport:
@Nadahar Nice it is not initialised in headless mode.
I thought there can be code which monitors how long some action takes and if it is longer than expected, it will take action like generating splashscreen :-) I am not programmer so I give you sight from used side only. Btw. I am able to do something similar in Powershell so I am sure it will be easy way how to do the same in Java.

Nadahar:
@ExSport It would definitly be possible to create a timer, set a threshold and that check that timer from time to time. But, in the case where UMS is waiting for disc access for example, this is a blocking call and the waiting UMS code has no control. Then one would need to create a separate thread doing this, and some way for this thread to check if the other thread were finished or not. 

It's definitly doable, but I'm not sure it's worth the hazzle.

Nadahar:
@valib @SubJunk Maybe [this] (https://docs.oracle.com/javase/tutorial/uiswing/misc/splashscreen.html) should be used to create the splash screen instead? It will show the splash screen more or less immediately, but I don't know how we'd control if it should be displayed or not then given that it's spawned before UMS (in case it's turned off in user options or headless). I haven't studied it in detail though, it could be that it is easy to implement.

valib:
@Nadahar  I know about this options but as you wrote there is a problem how to control if it should be displayed or not. @SubJunk  what do think about merging of this PR?

Nadahar:
@valib @SubJunk I still don't think the plash "logo" should say "UMS". We should spell it out there at the very least.

@valib Why are there subtitle files in this PR, and why did you merge when it fails on tests?

SubJunk:
This should never have been merged regardless. I'm also on the fence about having a splash screen. I don't think it's a good idea to add things we are on the fence about so I don't think we should continue work on this feature.

Nadahar:
@SubJunk I think master is fixed now.

Regarding having a splash screen, it's now configurable but it defaults to on. We could simply change the default to off and there would be no change by default - and those that do think it's useful could active and use it.

PR #706, opened 2015-10-22
Show live subtitles folder configurable per DLNAResource

taconaut:
I'd like to be able to decide for each `DLNAResource` if the live subtitles folder should be displayed. I've added the same logic as for the transcode folder to do this.
This change will NOT modify the current behavior at all, but it allows to override `DLNAResource.isLiveSubtitleFolderAvailable()` in classes extending `DLNAResource` to decide if the live subtitles folder should be displayed.

### DIFF
--- a/src/main/java/net/pms/dlna/DLNAResource.java
+++ b/src/main/java/net/pms/dlna/DLNAResource.java
@@ -382,6 +382,15 @@ public abstract class DLNAResource extends HTTPResource implements Cloneable, Ru
 	}
 
 	/**
+	 * Checks if is live subtitle folder available.
+	 *
+	 * @return true, if the live subtitle folder should be shown
+	 */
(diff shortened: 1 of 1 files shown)


taconaut:
[This](https://github.com/taconaut/ums-mlx/blob/bf5e92735a68480d6213d3f06ef95f762866ce79/core/src/main/java/net/pms/medialibrary/dlna/MediaLibraryFolder.java#L92) is how I'm going to use it.
Please consider a merge.

Nadahar:
@taconaut @UniversalMediaServer/developers @SubJunk 
I don't know anything about live subtitles or how it works (which is why I haven't said anything here), but I think this looks like a very reasonable change that's can't hurt anything.

SubJunk:
Looks okay to me, thanks for the PR :)

Issue #709, opened 2015-10-25
Shared Folders control renders incorrectly on high-DPI display in Windows 10


johnddx:
On a 4k monitor, with DPI scaling at 200% to compensate, the Shared folder view compresses each item in the list to half-height.
![capture](https://cloud.githubusercontent.com/assets/15313197/10717222/31dd1796-7b0e-11e5-9736-047109bd0933.PNG)

Nadahar:
@johnddx I don't think UMS' GUI has been tested/adapted for DPI scaling at all. Is that the only issue you get?

johnddx:
There are others, but not as bad.  In several places the text is quite tiny, especially Logs and the Help tab.  

The Shared folders tab is the only place that's actually hard to use as a result (due to clipping).

Nadahar:
@johnddx I've made a suggested fix in #712. It won't make it for the next release though, as that's already in the pipeline.

Nadahar:
#712 is merged now, so this should be in the next release.

PR #712, opened 2015-10-28
Improve GUI font scaling to better support high-DPI displays (#709):

Nadahar:
In response to #709 I've made the following changes:
- Made font sizes and padding in help files CSS scale with OS font size
- Made the log view/traces scale with OS font size
- Made row height in tables in PluginTab and NavigationShareTab scale with OS font size

Please note that since I don't own a high-DPI display, it's a bit hard for me to test this, but I've tuned my Windows font size to 150% while testing and the results looks reasonable. While I don't say that this makes the GUI properly support font scaling, it is better than today and at least makes it functional. Maybe you can test it @valib, since you use font scaling above 100%.

### DIFF
--- a/src/main/external-resources/documentation/css/style.css
+++ b/src/main/external-resources/documentation/css/style.css
@@ -16,19 +16,21 @@
  * along with this program; if not, write to the Free Software
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
+ 
+ /*
+  * Please note:
+  * Any sizes that are to be relative to the OS base font size must be set
+  * in HelpTab.buildStyleSheet(). This is done to support OS font scaling
+  * used on example for high DPI monitors.
+  */
+  
 body {
 	font-family: Arial;
-	font-size: 14pt;
 	color: #000000;
-	padding: 10px;
 }
 
 dt {
 	font-weight: bold;
-}
-
-dd {
-	margin-bottom: 20px;
 }
 
 blockquote {
@@ -38,5 +40,4 @@ blockquote {
 }
 pre, tt {
 	font-family: Courier New, Courier;
-	font-size: 12pt;
 }
--- a/src/main/java/net/pms/newgui/HelpTab.java
+++ b/src/main/java/net/pms/newgui/HelpTab.java
@@ -25,9 +25,7 @@ import java.awt.Color;
 import java.awt.Desktop;
 import java.awt.Dimension;
 import java.io.File;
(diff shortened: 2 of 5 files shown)


valib:
@Nadahar tested and there are some issues. Font size scaling in the TracesTab works fine. But the Help Page font size is growing more than necessary. See:
ums1
I don't know which part of the code make it but it makes the space between lines much bigger like:
ums2

Nadahar:
@valib The font in the help tab is quite big without font scaling as well. I've just kept the same ratio. Without font scaling the "standard" font in Windows is 12 px - and the help page text is 14 px. When I tested at 150% the "standard" font was 18 px. To find the help text font I simply multiply wit 7 and divide with 6. 18*7/6=21. I think the help page font is too large in general, but I haven't touced this. I assumed that 14 was selected by design.

I haven't touched the ```About``` tab - I didn't know there were issues there. Do you use the default font size in UMS, or have you configured some special font size?

valib:
@Nadahar I only changed in the WIN 10 the text and application size to 150%

Nadahar:
@valib Without my changes, the help page doesn't scale (it stays at 14px font size). What I'm trying to say is that the help page scales as it's intended, but that the font size there is bigger than the rest of UMS altso for 100%. I don't think it looks very nice, but it didn't want to change anything, I just added scaling.

If there's consensus to make the help page font smaller, I'll do that. I'd recomment 12px in 100% - the same as the standard font. That will make the help page font 18px i 150% setting.

Nadahar:
@valib I've fixed it. I didn't know that, but it seems that all Swing components share a common ```StyleSheet``` (whether they are HTML or not), and when I manipulated the ```StyleSheet``` for the ```EditorPane``` in ```HelpTab``` I actually manipulated the shared ```StyleSheet```. The reason only ```AboutTab``` was affected was that it was the only one being built after ```HelpTab```. 

I'd also replaced ```pt``` (points) with ```px``` (pixels) in the ```StyleSheet``` since I at some point deducted that Swing didn't really support points and that is simply was an alias for pixels. However I drew that conclusion it was wrong - there definitly is a difference.

Will you test it again now? You must reset your local branch since I've amended the last commit.

valib:
@Nadahar tested and it looks good. Great job :+1:

Nadahar:
@valib Great, thank you for testing.

Any other objections to a merge?

Nadahar:
I've tested this quickly on Centos 7, Ubuntu 12 and OS X 10.8 as well and can't see any issues. Since there has been no objections I'm merging this.

PR #717, opened 2015-11-03
Replacing maven-assembly-plugin with maven-shade-plugin for building jar-with-dependencies

Nadahar:
I had to add a dependency in another branch I'm working on to access a few lines of code. I noticed that this added 1.8 MiB to the packaged JAR, and started reading. From what I can understand `maven-assembly-plugin` packs every dependency into the JAR, whether it is used or not.

Surprisingly it doesn't seem like there's a good solution for this in Java, and the closest I've come is to use the `maven-shade-plugin` instead. It has an option `minimizeJar` that excludes any unused dependencies from the built JAR. Trying this immediately shaved of 11 MiB from the JAR, but introduced some problems as well.

The manifest had to be built differently, but I think I figured that one out. But, `minimizeJar` removes too many classes - that is, classes that's not imported "the normal way". Trying to start this JAR leads to multiple errors `ClassNotFoundException`. It is possible to manually include these as special cases and still get the benefit of not including everything else we don't need in the finished JAR. More information on this can be found [here](http://stackoverflow.com/questions/8817257/minimize-an-uber-jar-correctly-using-shade-plugin).

As of now I've only fiddled with the one in the `windows` profile. It seems to me like all the profiles first build an identical `jar-with-dependencies` and then add profile-specific files. If I got things right, that means that the building of the `jar-with-dependencies` could be done in the `build` section instead of in all the profiles, and then the profiles could take over.

This needs more work, some classes must be manually included and the packing structure must be rewritten.

Before I do any more on this I'd like to know if you think it's a good idea, and if you see any pitfalls?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -925,8 +925,48 @@
 
 			<build>
 				<plugins>
-					<!-- Plugin to assemble a jar with dependencies -->
 					<plugin>
+						<groupId>org.apache.maven.plugins</groupId>
+						<artifactId>maven-shade-plugin</artifactId>
+						<version>2.4.2</version>
+						<configuration>
+							<createDependencyReducedPom>false</createDependencyReducedPom>
+							<minimizeJar>true</minimizeJar>
+							<promoteTransitiveDependencies>true</promoteTransitiveDependencies>
+							<transformers>
+								<transformer implementation="org.apache.maven.plugins.shade.resource.ManifestResourceTransformer">
+									<mainClass>net.pms.PMS</mainClass>
+									<!--
+										this is required to make the ImageIo image "plugins" provided by jai-imageio-core-standalone work.
+										without it, the following exception is thrown when PMS starts:
+
+										Configuration error: java.util.ServiceConfigurationError: javax.imageio.spi.ImageInputStreamSpi:
+											Provider com.sun.media.imageioimpl.stream.ChannelImageInputStreamSpi could not be instantiated:
+												java.lang.IllegalArgumentException: vendorName == null!
+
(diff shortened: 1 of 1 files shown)


Sami32:
@Nadahar Oh i didn't see that one yet :-)

Look a good idea too me, if your are able to do it, since now it's look like we need to keep an eye on UMS size with @atamariya PR.
I didn's even know about; i will check it later.

Sami32:
maven-shade-plugin (2.4.3)

  * Team upload.
  * New upstream release
    - Depend on libasm4-java instead of libasm3-java (Closes: #800855)
    - Depend on libmaven3-core-java instead of libmaven2-core-java
    - Depend on libplexus-utils2-java instead of libplexus-utils-java
    - New build dependency on libmockito-java
    - Removed descriptor.patch
  * Build with the DH sequencer instead of CDBS
  * Moved the package to Git
  * Use XZ compression for the upstream tarball
  * Standards-Version updated to 3.9.6 (no changes)
  * Removed the javadoc package

 18 Jan 2016 13:44:09 +0100

Sami32:
@nadahar I got compilation issue:

capture_09122016_233742

Nadahar:
> I got compilation issue:

@Sami32 That's not a compilation error, but the ant script that can't find a file it's supposed to copy. There probably was a compilation failure further up though. If you read my original text for this PR you will see that it is NOT complete and won't build. This is so long ago now that I don't remember the exact problem, but as I wrote some restructuring of ```pom.xml``` is needed. In addition there are some classes that must be added "manually" like parts of Swing.

SubJunk:
This seems like a good idea. Feel free to resurrect this if work resumes

PR #719, opened 2015-11-05
Auto-delete

taconaut:
From what we've discussed in #673 these seem to be the only relevant changes required to delete files to the recycle bin.

I haven't checked if files are part of folders_monitored as they will be anyway; otherwise it wouldn't have been possible to play them.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -336,6 +336,11 @@
 			<version>4.2.0</version>
 		</dependency>
 		<dependency>
+			<groupId>net.java.dev.jna</groupId>
+			<artifactId>jna-platform</artifactId>
+			<version>4.2.0</version>
+		</dependency>
+		<dependency>
 			<groupId>org.codehaus.plexus</groupId>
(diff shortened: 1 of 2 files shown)


Nadahar:
@taconaut If I haven't gotten something completely wrong, there's a misunderstanding about ```folders_monitored```. From what I understand shared folders are devided into ```monitored``` or not. Monitored shared folders are those who's media will be included in "new media", while the "normal" shared folder's media will not. @SubJunk plans to extend this ```folders_monitored``` logic to use for #673, so that only marked folders will have said logic.

What's not supported with this setup though is to have different such rules for different folders. You can have one folder where things are deleted and another where they are just marked. It might be fine, but I think I would have replaced the ```CheckBox``` for monitored with a ```ComboBox``` where you could choose between the different options. But, that comment stictly belongs in #673.

SubJunk:
Thanks @taconaut !

PR #722, opened 2015-11-08
The FreeDesktop.org Trash specification support

Nadahar:
While it as always was a lot more work than I imagined, it's done. This implements the trash bin specification that's supposedly used by the most popular Linux desktops, the [The FreeDesktop.org Trash specification v1.0] (http://standards.freedesktop.org/trash-spec/trashspec-1.0.html).

The standard have several weak points and it didn't feel well thought through once I started to dig into it. That said, I've followed the standard to the best of my ability with one exception. The standard defines that the root partition trash folder should be located at ```/home/user/.Trash```, but the KDE and Gnome have abandoned that location and instead use ```/home/user/.local/share/Trash```. This implementation checks both locations, but the problem is when none of them exist. According to the standard, the implementation then have to create the folder, but I can't create and use both and have to pick one. I've gone for the non-standard ```/home/user/.local/share/Trash``` location simply because this is what will make it work on most modern Linux installations. There's no way to remedy this without finding a way to detect which desktop environment (both type and version) is running, and I wouldn't take it that far.

Further, neither Centos 7 nor Ubuntu 12.04 that I tested supports the standard properly. 

On Ubuntu the problem arises if you delete two or more files or folders with the same filename and then later restores them. This implementation adds random characters to the filename in such situations until an unused (in the trash bin) name is found. Ubuntu will not rename them back to their original name, but will keep the random characters added. The user will at least not loose his/her file.

On Centos with KDE that part works, but it can't handle absolute paths in the ```trashinfo``` on non-root partitions, and suffixes them with the partition root making it an invalid path when trying to restore files. That means that restoring fails and that the user will have to manually pick up his file from the trash folder, but again - the file isn't lost at least.

On Centos with Gnome I don't think I encountered any problems, but that's probably just because I didn't test long enough. As always with Linux, nothing is ever done properly and standards are interpreted as informal guides that you just look at from time to time while coding.

In any case, it should mostly work under "normal" circumstances, that is when the files to be trashed are on the root partition and multiple versions of the same filename isn't to be held in the trash bin at the same time. This implementation won't under any circumstances delete the files, but will throw exceptions if the move can't be performed. The calling code will have to choose how to deal with that.

Because whoever wrote ```java.nio``` decided to make about everything package private (including almost all the classes), I've had to use reflection to do some of these things. I hate to have to do that, but with Java's flawed OO model it's hard to come around doing that. The alternative would be to have to implement my own calls to glibc, and I couldn't get myself to do that as long as the information I needed was available in the hidden classes. Shame on those that came up with the idiotic rules for ```protected``` in Java and those that coded all of this in ```package private```. It would have been very simple to do this with subclassing otherwise.

The changes with comments are:
* Updated JNA to 4.2.1 - no reason really, I started out with the approach to subclass ```java.net.jna.platform.FileUtils``` and updated and added a package. But, I soon realized that I couldn't use any of their logic and abandoned that approach. I still kept the version upgrade though.
* Added the following to ```FileUtil```:
  * class ```InvalidFileSystemException``` - an ```Exception``` class used by some of the new methods.
  * class ```UnixMountPoint``` - a struct/record to hold the data fetched by ```getMountPoint()```
  * method ```getMountPoint()``` - finds the mount point for any given file or folder residing on a Unix based file system.
  * method ```getUnixUID()``` - finds the integer used id on Unix systems for the user UMS is running under.
* Created class ```FreedesktopTrash``` with support for The FreeDesktop.org Trash specification v1.0 - this is where most of the work is done.

The only change to existing code is the bump from 4.2.0 to 4.2.1 for JNA. Everything else is new and does nothing unless called, so I'd say this should be safe to merge.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -333,7 +333,7 @@
 		<dependency>
 			<groupId>net.java.dev.jna</groupId>
 			<artifactId>jna</artifactId>
-			<version>4.2.0</version>
+			<version>4.2.1</version>
 		</dependency>
 		<dependency>
 			<groupId>org.codehaus.plexus</groupId>
--- a/src/main/java/net/pms/util/FileUtil.java
+++ b/src/main/java/net/pms/util/FileUtil.java
@@ -1,11 +1,16 @@
 package net.pms.util;
 
 import java.io.*;
+import java.lang.reflect.Field;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.nio.charset.Charset;
-import java.nio.charset.IllegalCharsetNameException;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.FileStore;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.attribute.PosixFileAttributes;
+import java.nio.charset.IllegalCharsetNameException;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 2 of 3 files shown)


SubJunk:
Cool, feel free to merge if no one else objects :)

Nadahar:
It's hard to see how anyone would have any objection, it doesn't touch the existing code, so I'm merging.

Issue #726, opened 2015-11-10
play audio file  for shorter   time  of total


anthraxs:
i tryed  multichannel  flac  and wav    L16 format 

multichannel  flac   total time 5:25   played   for only 1:48    

other   multichannel  flac   total time 2:21    played  for only 47 seconds 

DEBUG 2015-11-10 10:46:34.207 [main] 

TRACE 2015-11-10 10:48:03.813 [New I/O worker #8] net.pms.formats.FormatFactory Matched format FLAC to "E:\Bittorrent\1997-Depeche Mode - Ultra-flac\01 - Depeche Mode - Barrel Of A Gun-d2p-p2p.flac"
TRACE 2015-11-10 10:48:03.815 [New I/O worker #8] net.pms.dlna.DLNAResource Adding new child "01 - Depeche Mode - Barrel Of A Gun-d2p-p2p.flac" with class "net.pms.dlna.RealFile"
TRACE 2015-11-10 10:48:03.816 [New I/O worker #8] net.pms.formats.FormatFactory Getting player for resource "01 - Depeche Mode - Barrel Of A Gun-d2p-p2p.flac"
TRACE 2015-11-10 10:48:03.816 [New I/O worker #8] net.pms.formats.FormatFactory Player "FFmpeg" is incompatible
TRACE 2015-11-10 10:48:03.816 [New I/O worker #8] net.pms.formats.FormatFactory Player "MEncoder" is incompatible
TRACE 2015-11-10 10:48:03.816 [New I/O worker #8] net.pms.encoders.TsMuxeRVideo tsMuxeR cannot determine compatibility based on audio track for E:\Bittorrent\1997-Depeche Mode - Ultra-flac\01 - Depeche Mode - Barrel Of A Gun-d2p-p2p.flac
TRACE 2015-11-10 10:48:03.816 [New I/O worker #8] net.pms.formats.FormatFactory Player "tsMuxeR" is incompatible
TRACE 2015-11-10 10:48:03.816 [New I/O worker #8] net.pms.formats.FormatFactory Returning compatible player "FFmpeg Audio"
TRACE 2015-11-10 10:48:03.816 [New I/O worker #8] net.pms.configuration.FormatConfiguration Format "flac" failed to match supported line f:mpegps|mpegts   v:mpeg1|mpeg2|mp4|h264   a:ac3|lpcm|aac|mpa   m:video/mpeg
[...]

Issue #588, opened 2015-07-02
Transkoding audio files to LPCM gives wrong pitch (FFmpeg Audio) for some files


Nadahar:
Some audio files, when transcoded with FFmpeg Audio, plays with wrong pitch when I play them om both my available renderers (Onkyo TX-NR807 and foobar2000). There aren't many options for audio transcoding, toggeling resampling to 44,1/48 doesn't fix it. It's certain files, and they are always transcoded with wrong pitch (some to low, some to high). When I try to transcode the same file with the same options (only -f wav instead of -f sb16be and a filename instead of pipe), the files play normally in all the media players I've tested.

I've looked at bitrate, sampling rate, bit depth on the source files without being any smarter (it probably takes more). Do anyone have any knowledge of this, is it some kind of a bug, a DNLA/uPnP limitation or some configuring options I haven't found?

Examples:
- Format: OGG, Codec: Vorbis, VBR, 44.1 KHz (pitch too high)
- Format: M4A, Codec: AAC LC, VBR, 44.1 KHz (pitch too high)
- Format: M4A, Codec: AAC LC, VBR, 8 KHz, Mono (pitch smurf high)
- Format: 3GP, Codec: HE-AACv2, VBR, 32 KHz (pitch too high)
- Format: M4A, Codec: AAC LC, VBR, 22.05 KHz (pitch too high)
- Format: M4A, Codec: AAC LC, VBR, 44.1 KHz (pitch too high)
- Format: M4A, Codec: ALAC, VBR, 44.1 KHz, 16 bits (pitch too high)
- Format: AC3, Codec: AC3, CBR: 448 Kbps, 48 KHz, 16 bits, 5 channels (pitch too low)
- Format: WAV, Codec: DTS, CBR: 1411.2 Kbps, 44.1 KHz, 24 bits, 5 channels (pitch too low)
- Format: FLAC, Codec: FLAC, VBR, 44.1 KHz, 16 bits (pitch too high)

Nadahar:
I've found the problem. The problem only occurs when transcoding to LPCM. When transcoding to WAV or MP3 the samlpe rate is not specified probably because it's in the headers, but since LPCM is a raw stream there's no headers and it has to be specified. The specification of the sample rate is flawed and only "luck" will make it correct some times:

RendererConfiguration line 1288:
```java
					// Default audio transcoding mime type
					matchedMimeType = HTTPResource.AUDIO_LPCM_TYPEMIME;

[...]

PR #729, opened 2015-11-15
Fix missing or wrong sample frequency and audio channels when transcoding to LPCM (#588) (#726)

Nadahar:
As described in #588 there's a pitch/sample frequency bug when transcoding audio to LPCM. This fixes this, although because of the messy nature of the involved code I'm not sure if I've broken anything else. I don't know if these changes touches anything video at all, but if not I think it should be reasonably safe. Is LPCM used for video at all?

In my opinion this is merely a band-aid, there's so much mess here that should be cleaned up. I have a hard time to see what role the ```AudioProperties``` class plays, it seems to me like a half clone of ```DLNAMediaAudio```. It doesn't seem to be finished either, and defaulting to 48 KHz for audio sample frequency is simply wrong and begging for bugs. If something is unknown, treat it like an unknown imo.

I also feel that the whole "resample audio" function is outdated and wrong. It caters specifically to the PS3 (where you can choose between 44.1 and 48 KHz sample rates only), and doesn't make much sense for any other renderers. Being "universal" this logics should somehow be moved to the renderer configuration and not be a part of the base configuration.

If anyone can shed some more light on any of these issues, please do so, so I don't plan a future refactoring with vital missing pieces of the picture.

Back to this PR, please inspect and comment on the changes if you see anything wrong/dangerour. Also, please let me know if you think this is safe to merge.

### DIFF
--- a/src/main/java/net/pms/util/ProcessUtil.java
+++ b/src/main/java/net/pms/util/ProcessUtil.java
@@ -217,8 +217,13 @@ public class ProcessUtil {
 	// Whitewash any arguments not suitable to display in dbg messages
 	// and make one single printable string
 	public static String dbgWashCmds(String[] cmd) {
-		for(int i=0; i < cmd.length; i++) {
-			if(cmd[i].contains("headers")) {
+		for (int i=0; i < cmd.length; i++) {
+			// Wrap arguments with spaces in double quotes to make them runnable if copy-pasted
+			if (cmd[i].contains(" ")) {
+				cmd[i] = "\"" + cmd[i] + "\"";
+			}
+			// Hide sensitive information from the log
+			if (cmd[i].contains("headers")) {
 				cmd[i+1]= cmd[i+1].replaceAll("Authorization: [^\n]+\n", "Authorization: ****\n");
 				i++;
 				continue;
--- a/src/main/java/net/pms/configuration/RendererConfiguration.java
+++ b/src/main/java/net/pms/configuration/RendererConfiguration.java
@@ -17,6 +17,7 @@ import net.pms.PMS;
 import net.pms.dlna.*;
 import net.pms.encoders.Player;
 import net.pms.formats.Format;
+import net.pms.formats.v2.AudioProperties;
 import net.pms.io.OutputParams;
 import net.pms.network.HTTPResource;
 import net.pms.network.SpeedStats;
@@ -1251,7 +1252,7 @@ public class RendererConfiguration extends UPNPHelper.Renderer {
 	 *            configured for this renderer.
 	 * @return The mime type.
 	 */
-	public String getMimeType(String mimeType) {
+	public String getMimeType(String mimeType, DLNAMediaInfo media) {
 		if (mimeType == null) {
 			return null;
 		}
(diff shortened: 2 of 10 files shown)


Nadahar:
Oh, and there's one more "bug"/issue. In the renderer in use in #726 (```Vizio-SmartTV.conf```) there's not "supported" line for LPCM, yet still this is the default audio transcode format. This leads to the mimetype calculation falling back to V1/not ```mediainfo``` since it can't find a match. 

We should make sure that every renderer that uses mediainfo has "supported" lines for the transcoding formats. Any idea of how we go about doing this?

Nadahar:
While this PR fixes the issue in some situations, it doesn't seem to always do it. I still think this is better than nothing, and to fix the remaining issues I think we have to control output sampling frequency and channels according to the renderer's demands. That again requires some changes/expansion to the renderer configuration and the "supported" lines and is a bigger job. A separate issue should probably be created for this.

src/main/java/net/pms/util/ProcessUtil.java
```diff
@@ -217,8 +217,13 @@ public static String run(String... cmd) {
 	// Whitewash any arguments not suitable to display in dbg messages
 	// and make one single printable string
 	public static String dbgWashCmds(String[] cmd) {
-		for(int i=0; i < cmd.length; i++) {
-			if(cmd[i].contains("headers")) {
+		for (int i=0; i < cmd.length; i++) {
+			// Wrap arguments with spaces in double quotes to make them runnable if copy-pasted
+			if (cmd[i].contains(" ")) {
```
SubJunk: I think this should be tested on multiple operating systems before merging, since we made a change years ago that was similar and broke transcoding on some... I can't remember details.

Nadahar: @SubJunk This function is only for debug printing. It can't break anything as far as I can tell. The whole purpose was to make it so that one can copy-paste the command found in the log and it will actually work as intended. Processes run from Java send arguments as an array, not a space-separated string, but we turn it into a space separated string when logging it. I just made sure that if any of the arguments, one of the elements in the array, contains space, then that element is wrapped in double quotes before added to the debug string for logging. Has this created any problems?

SubJunk: @Nadahar oh yeah I didn't notice it was just for debugging, cool

Nadahar:
@UniversalMediaServer/developers @SubJunk Any objections to merging this?

SubJunk:
@Nadahar I'm not very knowledgeable about our LPCM support, it's a bit confusing. If you think this is an improvement we can merge it if you've tested it.

Nadahar:
@SubJunk I've tested it, and it seems to correct the problem in some circumstances. 

It's not very complicated. The thing about LPCM is that it is raw data only, it has no header. Because of that, any information like sample rate and channels have to be given together with the data stream, it's no "format" that can carry such information internally.

All I've done it to attach such information behind the mime type devided by semicolons, as is already done in some circumstances in existing code. This will only affect audio in any case. If the renderer doesn't understand the syntax, I'm guessing it's simply ignored and the results are as they are today.

SubJunk:
Sounds good to me

PR #733, opened 2015-11-17
StreamGobbler updates

Nadahar:

Following the discussions in db92c2a179fd260c7946a66b0b4cdcb798ab511e and 737936d3743f7f1bf555a42c80b1bea4d5f31670 and in particular the comment from @skeptical I decided to do a little optimization. Seeing that ```Runtime.getRuntime().exec()``` internally calls ```ProcessBuilder``` there's no performance loss in using ```ProcessBuilder``` directly instead, thus giving us the chance to merge the input(output?) and error streams for the process. This again means that we don't have to spawn threads to do the consuming, which is expensive.

Even though none of the current calls to ```StreamGobbler``` do this, I've kept the possibility to spawn threads in the event that the code don't want to wait/block until the stream is consumed, but I've added static methods to consume the stream in the callers thread. 

As a side note, I still haven't figured out why we read the input and not the output stream, but I see that it's the input stream that's used everywhere. Could it be so stupid that they named them from the Java POV, so that they exchanged the two? Any enlightenment is appreciated.

Here's the changelog:
* Expanded StreamGobbler to be able to do both blocking and non-blocking stream consumption
* Updated the callers to use blocking mode where it thread will wait for the stream to be consumed anyway.

What do you guys think? Is this a better way?

### DIFF
--- a/src/main/java/net/pms/dlna/RootFolder.java
+++ b/src/main/java/net/pms/dlna/RootFolder.java
@@ -1116,7 +1116,7 @@ public class RootFolder extends DLNAResource {
 										pb.redirectErrorStream(true);
 										Process pid = pb.start();
 										// consume the error and output process streams
-										new StreamGobbler(pid.getInputStream()).start();
+										StreamGobbler.consume(pid.getInputStream());
 										pid.waitFor();
 									} catch (IOException | InterruptedException e) {
 									}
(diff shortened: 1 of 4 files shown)


valib:
@Nadahar the```Process.inputStream``` is internally connected to the normal output of the subprocess. So we don't need to make e.g ```new InputStream(process.OutputStream())```. I see that it is not so logical but useful.

Nadahar:
I'm merging this since feedback has been positive.

Issue #714, opened 2015-10-30
WindowsNamedPipe creation problem with UMS 5.3.0 on Windows, Java 8 64bit


konrada:
When trying to transcode:
TRACE 2015-10-30 19:27:43.897 [New I/O worker #7] Looking for an audio track with lang: eng
TRACE 2015-10-30 19:27:43.897 [New I/O worker #7] Matched audio track: id: 0, lang: en, audio track title from metadata: Surround, audio codec: AC3, sample frequency:48000, number of channels: 6, bits per sample: 16
TRACE 2015-10-30 19:27:43.897 [New I/O worker #7] Searching for a match for: en with eng and off
TRACE 2015-10-30 19:27:43.897 [New I/O worker #7] Disabled the subtitles: id: 0, type: UNKNOWN, lang: off
DEBUG 2015-10-30 19:27:43.897 [New I/O worker #7] Reading engines: "ffmpegvideo,mencoder,ffmpegaudio,tsmuxeraudio,ffmpegwebvideo,vlcwebvideo,vlcvideo,mencoderwebvideo,vlcaudio,ffmpegdvrmsremux,rawthumbs" (default: "ffmpegvideo,mencoder,tsmuxer,ffmpegaudio,tsmuxeraudio,ffmpegwebvideo,vlcwebvideo,vlcvideo,mencoderwebvideo,vlcaudio,ffmpegdvrmsremux,rawthumbs")
DEBUG 2015-10-30 19:27:43.897 [New I/O worker #7] Reading maximum_bitrate: "0" (default: "90")
TRACE 2015-10-30 19:27:43.897 [New I/O worker #7] Using the video bitrate limit from the renderer config (25) which is lower than the one from the program settings (25)
TRACE 2015-10-30 19:27:43.897 [New I/O worker #7] Halving the video bitrate limit to 12500
DEBUG 2015-10-30 19:27:43.897 [New I/O worker #7] Reading DefaultVBVBufSize: true (default: false)
TRACE 2015-10-30 19:27:43.897 [New I/O worker #7] Adjusting the video bitrate limit to 11000 to make room for audio
DEBUG 2015-10-30 19:27:43.897 [New I/O worker #7] Creating pipe \\.\pipe\ffmpegvideo_58_1446229663897
DEBUG 2015-10-30 19:27:43.925 [New I/O worker #7] Caught exception
java.lang.IllegalArgumentException: Unrecognized calling convention: 3
[...]

Issue #710, opened 2015-10-28
Is Reviewable something for UMS?


Nadahar:
@UniversalMediaServer/developers @SubJunk 
I often feel that the code review process can be a bit messy, I don't really know how long I have to wait for response. I don't know if you've looked at it and have nothing to say about it or if it's unseen, and it's often somewhat unclear to me what's been agreed upon.

I looked quickly at [Reviewable](https://github.com/integrations/reviewable) and think it looks interesting. Maybe you guys think it's to much formalities, too much "bureaucracy", but I feel that more structure around this could be helpful. What do you think?

SubJunk:
It looks promising. I don't want it to create more work for us, but it looks like it will involve less work so that's good.

Nadahar:
It's hard to know before we've tried, but it's free for open source and we can just leave if it doesn't make things easier so I think it's worth a shot.

SubJunk:
It should be linked now

valib:
Agree it looks good.

Sami32:
Sure, but having the 2 together is perturbating, isn't it ?

@Nadahar Feel free to re-open it you see a purpose, that we all know about ;-)

Nadahar:
@Sami32 This was before GitHub introduced its own review tools - even though as we've seen they aren't as good.

PR #679, opened 2015-09-25
Crowdin integration

Nadahar:
@UniversalMediaServer/developers 
The POM is in a testing state here now, but I'm getting closer to finish the crowdin integration. I've had to fork [this Maven plugin] (https://github.com/glandais/crowdin-maven) and done some rather major modifications which you can see [here] (https://github.com/Nadahar/crowdin-maven).

Let me just start by saying: Aaaaaaargh :boom::facepunch::sweat_drops::punch::goberserk::skull::rage::collision: doesn't even begin to describe how much I hate Maven:exclamation: It's one undocumented piece of a complete mess.

But, with some hacking I've been able to get things to work more or less as planned. In the plugin I've this far implemented these goals:
* ```push``` - upload ```messages.properties``` to crowdin
* ```build``` - build or "compile" the translations database into downloadable files
* ```fetch``` - download the language files from crowdin into a temporary folder

What I have left is:
* ```apply``` - transfer (and rename, modify) the files from the temporary download folder into the ```resources/i10n``` folder
* some finishing touches to ```fetch``` where is also downloads the translation status of the files among others.
* ```pull``` where I'm thinking of running ```build```, ```fetch``` and ```apply``` in that order.

I see no point in integrating any of these into the normal Maven build process (even though we can if we will), so they will all be run like this:
```
mvn ums-crowdin:<goal>
```

All but ```apply``` will require a valid crowdin API key stored in ```~/.m2/settings.xml```. I've also built in some rather paranoid protection measures into ```push``` since any strings not included in the uploaded file will be deleted (with translations) from crowdin. ```Push``` therefore checks the crowdin project name vs the POM project name vs a plugin defined ```pushProjectName``` variable - which all has to match to do a ```push```. In addition, the variable ```confirm``` must be ```yes```, ```true``` or ```confirm```, like this:
```
mvn ums-crowdin:push -Dconfirm=yes
```

We'll need to figure out where to keep this plugin. If you can create another repository under ```UniversalMediaServer``` that would probably be the best @SubJunk ? We'll also need to store the JAR file somewhere that Maven can download it to avoid having to place it in the local Maven repo manually.

I see that there's different practices when it comes to properties files escaping. To get it right, I should know whether or not to escape: ```=```, ```:```, ```#``` and ```!``` in the strings. I don't know if it's needed or recommended, I've seen files both with and without these characters escaped which seem to work.

Any response on my questions is very welcome :wink:

PR #736, opened 2015-11-22
6.0.0

SubJunk:
This is not ready to merge until the following tasks are completed:

- [x] Add "SquareAudioThumbnails" and "SquareImageThumbnails" renderer options, to be false by default but will add true to PS3.conf
- [x] Let us know how many seconds are being asked for when byte-seeking is used by the renderer

This branch was created from the WatchedMediaIntegrationV3 branch, see the discussion at https://github.com/UniversalMediaServer/UniversalMediaServer/pull/673


SubJunk:
@UniversalMediaServer/developers does anyone know a good way to translate bytes into position of a video? Some renderers use byte-seeking where they request the video (or audio) from 2345678 bytes in, and Java makes it very easy for us to just split the input stream by that many bytes so we give the renderer what it wants, but how can we know what position that means for our GUI?

It's easy for CBR videos since we can just work it out from the bitrate, but most videos are VBR. For a rough estimate we can do the CBR way, which will still give us closer to the answer than what we have now which is nothing, but I wonder if there's an accurate way to do it.

SubJunk:
My todo for this branch:

- Add "SquareAudioThumbnails" and "SquareImageThumbnails" renderer options, to be false by default but will add true to PS3.conf
- Let us know how many seconds are being asked for when byte-seeking is used by the renderer
- Give the web interface some love

Nadahar:
@SubJunk About the byte seeking, I can't imagine that there can exist any more accurate method than basing it on the average bitrate for VBR without parsing the stream and reading the time codes. That's not a simple or quick feat, so I think we simply have to stick to the average bitrate.

Nadahar:
@SubJunk Is this the "new" WatchedMediaIntegrationV3, or should commits still be made to that branch?

skeptical:
@SubJunk ffmpeg logging gives us continuous feedback on size<-->time correspondence, which could be parsed using ```OutputTextLogger.filter()``` , see ```FFmpegWebVideo.parseMediaInfo()``` for a realtime playback parsing example.
```logtalk
DEBUG ... [ffmpeg-2-2] frame= 1958 fps=125 q=1.0 size=   24768kB time=00:01:05.33 bitrate=3105.5kbits/s    
```


@Nadahar :+1: on ```StreamGobbler```, it seems the next logical step might be to consolidate with ```OutputConsumer```, ```OutputTextLogger``` etc, i.e. transfer filtering capability and so on.

SubJunk:
@Nadahar yeah this is the new WatchedMediaIntegration so commit that here. If there are any other new features you want to commit to 6.0.0 but not 5.x, feel free to create a branch from this and a PR into this. Then when we release 6.0.0 final, we'll merge this into master. At least that's the way I imagine it happening but feel free to make suggestions.

@skeptical that's useful for when we use FFmpeg, but sometimes MEncoder or VLC are used, and other times (almost every time for devices like Panasonic and Samsung TVs) we don't use any transcoder, so it would be nicer to do it at a more fundamental level

skeptical:
@SubJunk yeah I know, but ffmpeg is possibly the only case where there's a practical solution.

NB You only need to do this sort of thing for the (tiny) subset of uninformative renderers that use a ```PlaybackTimer```, e.g. PS3.  Upnp renderers already send us accurate playback position via ```AVTransport``` info, and same is true for chromecast and web browsers.

I have an old unfinished implementation of this which has some similarity to yours vis ```PlaybackTimer``` and also addresses position retrieval for the typical case where we have direct position feedback. The basic idea is that position is part of playback info and it's the job of the renderer's ```BasicPlayer``` variant to track and provide it, and we should move away from ```DLNAResource.start/stopPlaying()``` schema. Here's the diff fyi:
```diff
commit bf56ec57f1904ae83e4a4a3944ceb1d351657a63
Author: infidel <aval57@yahoo.com>
Date:   Thu Feb 12 15:40:07 2015 -0500

    Track playback position accurately

diff --git a/src/main/java/net/pms/configuration/RendererConfiguration.java b/src/main/java/net/pms/configuration/RendererConfiguration.java
index f0de70a..9fcfbbc 100644
--- a/src/main/java/net/pms/configuration/RendererConfiguration.java
+++ b/src/main/java/net/pms/configuration/RendererConfiguration.java
@@ -2404,9 +2404,12 @@ public class RendererConfiguration extends UPNPHelper.Renderer {
 				@Override
 				public void run() {
 					state.playback = PLAYING;
+					long startTime = res.getStartTime();
 					while(res == renderer.getPlayingRes()) {
-						long elapsed = System.currentTimeMillis() - res.getStartTime();
-						state.position = DurationFormatUtils.formatDuration(elapsed, "HH:mm:ss");
+						long elapsed = System.currentTimeMillis() - startTime;
+						long streamed = res.getStreamPosition();
+//						state.position = res.setPosition((elapsed < streamed ? elapsed : streamed) / 1000);
+						state.position = DurationFormatUtils.formatDuration((elapsed < streamed ? elapsed : streamed) * 1000, "HH:mm:ss");
 						alert();
 						try {
 							Thread.sleep(1000);
diff --git a/src/main/java/net/pms/dlna/DLNAResource.java b/src/main/java/net/pms/dlna/DLNAResource.java
index dd6ecfd..c18eeb8 100644
--- a/src/main/java/net/pms/dlna/DLNAResource.java
+++ b/src/main/java/net/pms/dlna/DLNAResource.java
@@ -50,6 +50,7 @@ import net.pms.network.HTTPResource;
 import net.pms.util.*;
 import static net.pms.util.StringUtil.*;
 import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.time.DurationFormatUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -2298,6 +2299,8 @@ public abstract class DLNAResource extends HTTPResource implements Cloneable, Ru
 						}
 
 						startTime = System.currentTimeMillis();
+						position = 0;
+						streamPosition = 0;
 
 						for (final ExternalListener listener : ExternalFactory.getExternalListeners()) {
 							if (listener instanceof StartStopListener) {
@@ -2514,9 +2517,15 @@ public abstract class DLNAResource extends HTTPResource implements Cloneable, Ru
 
 			if (resume != null) {
 				if (range.isTimeRange()) {
-					resume.update((Range.Time) range, this);
+					if (range.isStartOffsetAvailable()) {
+						streamPosition = (long)(double)((Range.Time)range).getStart();
+					}
+//					resume.update((Range.Time) range, this);
+					resume.update(this);
 				}
 				params.timeseek = resume.getTimeOffset() / 1000;
+LOGGER.info("*** TIMESEEK: "+params.timeseek);
+//				params.timeseek = getPosition() / 1000;
 				if (player == null) {
 					player = new FFMpegVideo();
 				}
@@ -3431,7 +3440,8 @@ public abstract class DLNAResource extends HTTPResource implements Cloneable, Ru
 		notifyRefresh();
 
 		if (resume != null) {
-			resume.stop(startTime, (long) (media.getDurationInSeconds() * 1000));
+//			resume.stop(startTime, (long) (media.getDurationInSeconds() * 1000));
+			resume.stop(this);
 			if (resume.isDone()) {
 				parent.getChildren().remove(this);
 			} else if (getMedia() != null) {
@@ -3440,7 +3450,8 @@ public abstract class DLNAResource extends HTTPResource implements Cloneable, Ru
 		} else {
 			for (DLNAResource res : parent.getChildren()) {
 				if (res.isResume() && res.getName().equals(getName())) {
-					res.resume.stop(startTime, (long) (media.getDurationInSeconds() * 1000));
+//					res.resume.stop(startTime, (long) (media.getDurationInSeconds() * 1000));
+			res.resume.stop(this);
 					if (res.resume.isDone()) {
 						parent.getChildren().remove(res);
 						return null;
@@ -3763,4 +3774,37 @@ public abstract class DLNAResource extends HTTPResource implements Cloneable, Ru
 			dynamicPls.addChildInternal(vva);
 		}
 	}
+
+	private long position = 0, startPosition = 0, streamPosition = 0;
+
+	public long getPosition() {
+		return position;
+	}
+
+	public long getStreamPosition() {
+		return streamPosition;
+	}
+
+	public String getPositionString() {
+		return getPositionString(startPosition, position);
+	}
+
+	public static String getPositionString(long startPosition, long offset) {
+		return DurationFormatUtils.formatDuration((startPosition + offset) * 1000, "HH:mm:ss");
+	}
+
+	public String setPosition(String pos) {
+		position = convertStringToSeconds(pos);
+//LOGGER.info("*** POS: {} > {} ({}+{})",pos,getPositionString(),startPosition,position);
+		return startPosition == 0 ? pos : getPositionString();
+	}
+
+	public String setPosition(long pos) {
+		position = pos;
+		return getPositionString();
+	}
+
+	public void setStartPosition(long pos) {
+		startPosition = pos;
+	}
 }
diff --git a/src/main/java/net/pms/dlna/ResumeObj.java b/src/main/java/net/pms/dlna/ResumeObj.java
index 9237bbc..fe53e6e 100644
--- a/src/main/java/net/pms/dlna/ResumeObj.java
+++ b/src/main/java/net/pms/dlna/ResumeObj.java
@@ -8,6 +8,8 @@ import java.io.FileWriter;
 import java.io.IOException;
 import net.pms.PMS;
 import net.pms.configuration.PmsConfiguration;
+import static net.pms.util.StringUtil.*;
+import org.apache.commons.lang3.time.DurationFormatUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -62,6 +64,7 @@ public class ResumeObj {
 				r.getMedia().setDuration(res.resDuration / 1000.0);
 			}
 		}
+		r.setStartPosition(res.getTimeOffset() / 1000);
 		res.setMinDuration(r.minPlayTime());
 		return res;
 	}
@@ -70,7 +73,8 @@ public class ResumeObj {
 		File f = resumeFile(r);
 		ResumeObj obj = new ResumeObj(f);
 		obj.setMinDuration(r.minPlayTime());
-		obj.stop(startTime, (long) r.getMedia().getDurationInSeconds() * 1000);
+//		obj.stop(startTime, (long) r.getMedia().getDurationInSeconds() * 1000);
+		obj.stop(r);
 		if (obj.noResume()) {
 			return null;
 		}
@@ -96,7 +100,7 @@ public class ResumeObj {
 			String str;
 			while ((str = in.readLine()) != null) {
 				String[] tmp = str.split(",");
-				offsetTime = Long.parseLong(tmp[0]);
+				offsetTime = tmp[0].contains(":") ? convertStringToSeconds(tmp[0]) * 1000 : Long.parseLong(tmp[0]);
 				if (tmp.length > 1) {
 					resDuration = Long.parseLong(tmp[1]);
 				}
@@ -106,7 +110,8 @@ public class ResumeObj {
 		}
 	}
 
-	private static void write(long time, long duration, File f) {
+//	private static void write(long time, long duration, File f) {
+	private static void write(String time, long duration, File f) {
 		try {
 			try (BufferedWriter out = new BufferedWriter(new FileWriter(f))) {
 				out.write(time + "," + duration);
@@ -142,14 +147,28 @@ public class ResumeObj {
 
 	public void update(Range.Time range, DLNAResource r) {
 		if (range.isStartOffsetAvailable() && range.getStartOrZero() > 0.0) {
-			stop(System.currentTimeMillis() + getTimeOffset() - (long) (range.getStart() * 1000), (long) ((r.getMedia() != null ? r.getMedia().getDuration() : 0) * 1000));
+			stop(r);
 		}
 	}
 
-	public void stop(long startTime, long expDuration) {
-		long now = System.currentTimeMillis();
-		long thisPlay = now - startTime;
+	public void update(DLNAResource r) {
+		stop(r);
+	}
+
+//	public void update(Range.Time range, DLNAResource r) {
+//		if (range.isStartOffsetAvailable() && range.getStartOrZero() > 0.0) {
+//			stop(System.currentTimeMillis() + getTimeOffset() - (long) (range.getStart() * 1000), (long) ((r.getMedia() != null ? r.getMedia().getDuration() : 0) * 1000));
+//		}
+//	}
+
+	public void stop(DLNAResource r) {
+		long expDuration = (long) ((r.getMedia() != null ? r.getMedia().getDuration() : 0) * 1000);
+		long thisPlay = r.getPosition() * 1000;
 		long duration = thisPlay + offsetTime;
+//	public void stop(long startTime, long expDuration) {
+//		long now = System.currentTimeMillis();
+//		long thisPlay = now - startTime;
+//		long duration = thisPlay + offsetTime;
 
 		if (expDuration > minDur) {
 			if (duration >= (expDuration * configuration.getResumeBackFactor())) {
@@ -168,7 +187,10 @@ public class ResumeObj {
 
 		offsetTime = duration - configuration.getResumeRewind();
 		resDuration = expDuration;
-		LOGGER.debug("Resume stop. This segment " + thisPlay + " new time " + duration);
-		write(offsetTime, expDuration, file);
+		String offset = DurationFormatUtils.formatDuration(offsetTime, "HH:mm:ss");
+//		LOGGER.debug("Resume stop. This segment " + thisPlay + " new time " + duration);
+		LOGGER.debug("Resume stop: session={}ms position={}", thisPlay, offset);
+//		write(offsetTime, expDuration, file);
+		write(offset, expDuration, file);
 	}
 }
diff --git a/src/main/java/net/pms/util/BasicPlayer.java b/src/main/java/net/pms/util/BasicPlayer.java
index 6cbbb13..1e30378 100644
--- a/src/main/java/net/pms/util/BasicPlayer.java
+++ b/src/main/java/net/pms/util/BasicPlayer.java
@@ -91,6 +91,7 @@ public interface BasicPlayer extends ActionListener {
 		public DeviceConfiguration renderer;
 		protected State state;
 		protected LinkedHashSet<ActionListener> listeners;
+		protected String lastPosition;
 
 		public Minimal(DeviceConfiguration renderer) {
 			this.renderer = renderer;
@@ -133,6 +134,10 @@ public interface BasicPlayer extends ActionListener {
 
 		@Override
 		public void alert() {
+			if (state.position != null && ! state.position.equals(lastPosition) && ! StringUtil.isZeroTime(state.position) && renderer.getPlayingRes() != null) {
+				state.position = renderer.getPlayingRes().setPosition(state.position);
+			}
+			lastPosition = state.position;
 			for (ActionListener l : listeners) {
 				l.actionPerformed(new ActionEvent(this, 0, null));
 			}
@@ -330,6 +335,10 @@ public interface BasicPlayer extends ActionListener {
 			if (stopping && autoContinue && !forceStop) {
 				next();
 			}
+//			lastPosition = state.position;
+		}
+
+		public void stopping() {
 		}
 
 		@Override
diff --git a/src/main/java/net/pms/util/StringUtil.java b/src/main/java/net/pms/util/StringUtil.java
index d667b9f..e2fdbc7 100644
--- a/src/main/java/net/pms/util/StringUtil.java
+++ b/src/main/java/net/pms/util/StringUtil.java
@@ -24,6 +24,7 @@ import java.io.IOException;
 import java.io.Writer;
 import java.util.Formatter;
 import java.util.Locale;
+import org.apache.commons.lang3.StringUtils;
 import static org.apache.commons.lang3.StringUtils.isBlank;
 import org.apache.commons.lang3.text.translate.UnicodeUnescaper;
 
@@ -143,6 +144,20 @@ public class StringUtil {
 		}
 	}
 
+	public static long convertStringToSeconds(String time) {
+		long seconds = 0;
+		try {
+			String[] t = StringUtils.substringBefore(time, ".").split(":");
+			for (int i = t.length - 1; i > -1; i--) {
+				seconds += Integer.parseInt(t[i]) * MULTIPLIER[i];
+			}
+		} catch (NumberFormatException e) {
+//			LOGGER.debug("Error decoding time string: " + time);
+			return -1;
+		}
+		return seconds;
+	}
+
 	/**
 	 * Converts time to string.
 	 *
```

SubJunk:
@skeptical cool, thanks for the code. I'll see if I can integrate some of it into this branch. I don't have any renderers that use the UPnP stuff properly - my Panasonic TV doesn't, neither does the AirPlayer app for iOS, and using Chrome I can't seek anyway. So from my perspective it seems like the opposite - it's hard to find a device that actually sends the AVTransport info. Do you know if any iOS apps do?

SubJunk:
I've been working on this a bit more and there's just the "Let us know how many seconds are being asked for when byte-seeking is used by the renderer" task left from my todo list.

@skeptical don't suppose you have a POC for that too before I start writing it?

SubJunk:
@UniversalMediaServer/developers I'm fairly happy with the state of this feature now, can you guys please test it on your devices to see if it's working for you? Different renderers can have very different ways of doing things as we know. I suggest the following tests:

* Fast-forward through a file and observe whether the timer on the Status tab changes to match the position
* Rewind to make sure that updates the Status tab too
* Stop it near the end of the file (over 92%) and check that the fully played action is applied
* Maybe check more of the fully played actions too

It's passing these tests for me

SubJunk:
@UniversalMediaServer/developers I'm planning on doing another dual release over the next couple of days unless anyone thinks master is less stable than the last release?

@Optimusprimeums are you around? If not I have a mac I can try to get set up for the release, do you have any tips for the setup?

Nadahar:
@SubJunk I see that you have tagged 5.4.0 already. I thought I was going to make it this time, I've just finished merging #679 up to sync now. I guess you should check it out, but I think it's more or less ready for merging.

Optimusprimeums:
Hey @SubJunk, I've just got my net back on and i'm just getting things sorted. I don't think you'll need any tips to get your setup right for building UMS on mac but if you run into any issues i can run you through it. I think I got the POM right so it build like the others.

SubJunk:
@Optimusprimeums yeah it was pretty painless, thanks for setting it up so nicely :) I can probably do them now if you want, for at least another 6 months anyway

Optimusprimeums:
@SubJunk suits me mate, Let me know if/when you want me to do it again :)

SubJunk:
@UniversalMediaServer/developers are you guys okay with me merging this and making the next release 6.0.0? It doesn't seem like we have any bugs specific to 6.0.0 reported on the forums

Nadahar:
@SubJunk Aren't there still some unresolved issues?

Also, I think we need to figure out #714 first. Is seems like JNA 4.2.x causes the "Invalid calling convention" bug we have been struggeling with lately. As far as I remember, the reason for upgrading JNA was the trash handling of this branch. We need to figure out if we can "solve" the problems with 4.2.x or downgrade to 4.1 - and what that does to the trash handling.

SubJunk:
@Nadahar I'm not aware of any issues with it. @ExSport mentioned a possible one but I don't think that was confirmed. The code in this branch touches relatively few places so I don't think it's very dangerous.

Downgrading JNA to 4.1.0 should be fine for this branch, I don't see anything related in their changelog https://github.com/java-native-access/jna/blob/master/CHANGES.md

SubJunk:
I just confirmed moveToTrash is working the same with 4.1.0 on Windows 7

Nadahar:
@SubJunk Here's the thing that I think need to go on the list or that I'm unaware that is solved:

- [ ] Figure out what to do with the JNA version
- [ ] Support for Audio and Images
- [ ] Clarify file deletion: Is it always just move to trash/recycler, or does it actually delete files if that fails? Does the name of the function reflect this behavious?
- [ ] Specify "watched" behaviour on a per-share basis
- [ ] Solve #564 (Changes to shared folders doesn't take effect until UMS is restarted). I think it's very strange to add functionality to this area while leaving this major bug in place.
- [ ] I read either a forum thread or an issue lately that reported trouble with "played time" on the Status tab in 6.0.0 (only showing 18:18 or something of that kind).

SubJunk:
@Nadahar like the documentation in UMS.conf says, the files won't be deleted if there's no trash bin. Is there a succinct way you can think of to specify that in the option name? Given that it will be mostly Linux users affected by that and they're also the users most likely to read the conf file, I don't think it's too important.

JNA is rolled back.

Audio is supported, images don't seem to trigger a played event on my renderers so that doesn't work. I'm okay with releasing it without image support.

Specifying watched behaviour per folder will be really useful but this branch is already really useful so I see that as an enhancement rather than a necessity.

WRT #564 it will be good to fix that but it's not connected with this branch

I'll look for the forum thread about the played time bug

Nadahar:
@SubJunk Looking at the current 6.0.0, I have some comments to the string:

* Add an overlay to the thumbnail: Is this the option that will apply either a thumbnail overlay or a text depending on renderer setting? If so, should the text be different to reflect this? What about just "Tag media"?
* Hide the file: What about "Hide media"?
* Move the file to a different folder: What about "Move media to a different folder"?
* Move the file to the recycle/trash bin: What about "Move media to recycle/trash bin"?

Nadahar:
Looking at [this code] (https://github.com/UniversalMediaServer/UniversalMediaServer/blob/6.0.0/src/main/java/net/pms/dlna/DLNAMediaInfo.java#L1093), it seems that thumbnail generation is disabled if UMS is running headless. Why is that/is that correct?

SubJunk:
WRT headless comment, I think that part of code was failing in headless mode because it relied on a GUI being present, so that line was a bugfix

Nadahar:
@SubJunk Not the best bugfix I've seen, effectively disabling thumbnails for all headless installations..? Isn't there a better way to solve that?

SubJunk:
@Nadahar yeah I'm sure there's a better way. I wasn't involved in the bugfix and may be remembering it wrong, best to check the commit history for the file

Nadahar:
@SubJunk Here's a [new forum thread] (http://www.universalmediaserver.com/forum/viewtopic.php?f=9&t=5883) claiming that "mark all as played" doesn't work in neither 5.x or 6.0.

Nadahar:
I tried to track down disabling of thumbnail generation in headless mode, and I tracked it back to f8abfe9ef7118df75a4a26630fb20639d34bf675 commit from 28. March 2010. It's not mentioned in the commit description, so I assume that it falls under "Small changes", and I don't know where to look for more information about this.

I consider this to be a relatively big problem. The way I intend to use UMS once it gets stable enough is running headless as a daemon on a Linux box. To me that is the most ideal way to run things like UMS, running on a server that's always on so that it's there ready to serve clients at any time without slowing down or interfering with one of my desktop computers, and I can't believe I'm the only one that sees it this way.

I'd therefor like to try to track down the problem, but it's hard to know which circumstances caused this when I can't find any references to the original issue. Looking at the code, I can't find any call that would throw a ```HeadlessException```, and I'm assuming that's what has cause this condition in the first place.

Nadahar:
Maybe it would be an idea to find a new icon set for 6.0.0 to add to the "new and shiny" effect?

Nadahar:
Another thing I've been thinking of: Does all files have thumbnails? If not, shouldn't it "fall back" to marking the file name for those files that doesn't?

I'm trying to test thumbnail creation on a headless computer (Centos 7 with no GUI), I've disabled the check for headless so that it will try to create thumbnails anyway, but I'm not sure how to actually setup thumbnail generation. What options to I have set for them to be created? Also, UMS complains about not finding MPlayer on Centos - is that relevant?

src/main/java/net/pms/dlna/DLNAMediaInfo.java
```diff
@@ -958,22 +1092,60 @@ public void parse(InputFile inputFile, Format ext, int type, boolean thumbOnly,
 
 						if (sz > 0 && !net.pms.PMS.isHeadless()) {
 							BufferedImage image = ImageIO.read(new ByteArrayInputStream(thumb));
```
Nadahar: I've removed ```!net.pms.PMS.isHeadless()``` from this if, built and run it on a headless Centos server. Thumbnails are generated just fine, and the overlay work as well. Am I doing something wrong when testing? Do I have to do something special for this code to be used? If not, it seems like we can just remove that check and thumbnails will work for headless installations as well.

Nadahar:
Another thing I discovered when testing this: I was under the impression that the overlay would only be in one corner, but this one was covering the whole thumbnail making it impossible to see the original thumbnail. Is that intentional?

src/main/java/net/pms/dlna/DLNAMediaInfo.java
SubJunk: Cool, that's good news. Maybe it was added in error. The final months of PMS were littered with massive commits with little explanation of the changes contained, and we ended up reverting a lot of them, so this may be an example of one that slipped through the cracks.

SubJunk:
@Nadahar I'll try to reply in order of your posts.

1) That seems like a valid way to use UMS to me. It doesn't use enuogh resources on my computer for me to consider that factor, but I do have a fairly modern PC. We do have users who use it like that though, and have always done so. It seems you are commenting a lot on the stability lately, it's different for different devices, operating systems and uses, but I have had 99+% stability with PS3 and Panasonic TVs since version 1, much more than our competitors I've tried. It can always get better but I think it's good to put it in context, and I think reliability is part of our reputation, too.

2) I'm happy to look at new icon sets. I wasn't intending for 6.0.0 to be a big deal, since we have discussed being more liberal with our versioning (in keeping with what orgs like Mozilla and Google are doing), where basically a major version is for a large feature addition, but I do like the idea of updating icons if we can find nicer ones. I warn you though, I'm picky and I like our current set :)

3) I think all files should have thumbnails. If we don't find one specific to the file itself, we have generic ones to send, but they may not always send which is a bug.

4) It's intentional that it covers the whole thumbnail, since some renderers only display small thumbnails (my TV only gives about 90px width to each thumbnail) so any smaller and the text won't be readable. The overlay background is semi-transparent though so you should be able to see through it to the original thumbnail unless the original is very dark.

Nadahar:
@SubJunk 
1) I just keep getting the impression reinforced that particular ways to use UMS gets all the attention, and that the other ways is left to chance. This particular issue was from PMS though, so it's not relevant in that manner. But, the support for audio playback is riddled with bugs and shortcomings for example, the media library, which to me is a fundamental function in any media server, is renamed to "cache" (in English, it's still called media library in most translations) and scanning is not "recommended" - which goes a long way in making the media library take a back seat. Scanning itself is also very bugged, it keeps throwing all kind of Exceptions. Audio thumbnails has several flaws, the ```folder.jpg``` that overwrites file thumbnails, the automatic downloadng of covers doesn't seem to do anything, and the list goes on. The same with images, the fact that browsing large folders (with many elements) takes so long that it times out is ignored, which effectively renders it useless. There are many, many more issues, and trying to use these things make it quite obvious to me that it hasn't gotten any attention and that other "fixes" that has been to UMS probably has broken things there without being picked up or prioritized.

As to running it from a desktop, it's not that my desktop can't take it, it's more that I don't want my desktop doing server tasks. A server, in my view, should always be available and keep running without being corrupted/slowed down for months (the code doesn't close resources after use many places, which proves that running for a long time hasn't been on the agenda) . I want to be free to shut down my desktop, use it for gaming (don't tell me that running Arma3 at ultra settings isn't affected by UMS transcoding in the background), boot into another operating system or whatever.

When it comes to stability I guess we just have different standards. Earlier I had UMS running for weeks at a time on a "test server", and one I checked up on it it had always stopped working in some way or another. Also, playback from multiple renderers at the same time or scanning while browsing has to lead to corruption of state because of the lack of thread safety.

I don't mean to complain, I just wish those things would get more priority than features.

2) See if you find something you like, I won't as I'm pretty sure we completely disagree about what's nice, so there's no point in me looking. I haven't thought much about the UMS interface when it comes to "beauty", it's after all mostly a configuration interface that needs to be functional. It was just that when I tried the old versions the other days, I was surprised of how much better I thought it looked with the "old" icons, and thus these got my attention. 

I dislike anything remotely related to "fashion" or "being modern", I think that's concepts that has to do with social dominance and manipulation, not aesthetics. I actually believe people aren't able to judge what they really find beutiful until they've managed to completely rid themselves of any "soscial manipulation" from those deemed to be the "lead sheeps". So, we're probably on completely different planets as very few share my views on this.

3) Wouldn't it be nice to try to track down this bug then? Any idea of what trigger it?

4) Yes ok, I just thought I remembered something about putting it 45 degrees in one of the corners of the thumbnail from an early scetch or something, but my mind often plays tricks on me. It's hard to get it correct on all devices, I see now that on my TV via the PS3 it hides the background image pretty much if the thumbnail itself isn't very bright, while from firefox while browsing the web interface it looks good. 

I still think it would serve it's purpose if the text wasn't completely readable in all circumstances though, as it would be easy enough to recognize the "mark", and the font could be capitalized and bolded to make it readable at a smaller size, but by all means it works as it is.

Nadahar:
Another thing I discovered is that I think the renderer setting created is somewhat wrong. Some renderers show thumbnails once you play the item, but doesn't show them while browsing. What's relevant here is if thumbnails are shown when browsing, right? That's where you need to see what's played before and not, so we should probably change it to "Thumbnails while browsing" or something like that. It a renderer doesn't support that, thumbnail generation shouldn't be disabled, but using an overlay on "fully played" isn't sufficient, so text need to be added as well.

valib:
@Nadahar it seems to me like the renderer fault. Users can try to set ```ThumbnailAsResource = true``` but this is a maximum what we can do.

Nadahar:
I've just discovered a new bug or feature, I'm not sure which. If ```UMS.mon``` doesn't exist, it's never created, and thus neither new media nor fully played is working. Is this intentional?

SubJunk:
1) I'm happy to prioritize important bugs that I'm aware of. Most times I'm not aware of them.

I've never had problems running UMS for months, so I don't see those things, and I don't have trouble running several renderers at once either. I typically server a single instance of UMS to 2 Panasonic TVs, a PS3 and an iPhone with no issues that I can see.

We have made improvements to the speed of parsing large image folders over time and there may be more we can do, too.

I still think we can gain the greatest benefit from running more automated tests. Do any @UniversalMediaServer/developers have experience writing integration tests in Java instead of unit tests? One of my projects has great integration coverage, which is when the tests compile the program and use it as a user would, and tests for the correct output, which is something we don't have any of right now. Unit tests are less capable because they look at very specific parts of code instead of whole program features.

2) Okay

3) Sure, it would be good but it's an existing bug. In my opinion this feature is complete when it adds value without creating new bugs. It would be good to fix that issue but there is a lot of value in this branch already, I've been using this feature for months now and it's very stable, so there are a lot of users using a version of the program that is less useful and has the same level of stability, and every day we wait before releasing 6.0.0 is another day that those users have to notice that our competitors have a similar feature.

4) Cool, yeah there are a few different ways we could do it. Another approach would be to have a colored mark in a corner that symbolizes it was fully played. I wonder if that would be intuitive to a user without text. I'm sure they would notice an association with any mark on the thumbnail since they just played the file. It's probably a lower priority than other things though, and I'm keen to look into the bugs you are concerned about before then. I don't think they necessarily have to be timed with the 6.0.0 release, like I said before it seems like we have this big useful thing just sitting around now.

SubJunk:
@Nadahar I think that UMS.mon bug may be a permissions issue. I had it once but when I looked into it I couldn't reproduce it. I just verified that as I write this.

Nadahar:
@SubJunk I don't think so, I followed the code and couldn't find anywhere that is was actually created other than in dumpFile(). Any, if I recall correcly, it was never called when the file didn't already exist.

I did several tests playing things and they were not marked as played/watched - and I couldn't understand why. As a result of that I started tracing the code and found the fact that the file wasn't there as the cause. As soon as I created an empty file with that name, things started to get marked "played" and the file was populated. I have UAC turned off and am an administrator, so permissions are there. I have to look at the code again to remember exactly how it was, but as I read it I was wondering if it was intentional, that for some reason deleting UMS.mon was some "secret" way of disabling monitoring.

By the way, I've worked further on 6.0.0 and have fixed multiple bugs, but I haven't pushed it yet (still working on the same "6.0.0 misc" PR). There's still work in progress though, so I don't see any reason to push it before I'm done.

Nadahar:
@SubJunk The reason it behaves like it does is this:
```java
	private void parseMonitorFile() {
		File f = monitorFile();
		if (!f.exists()) {
			return;
		}
```
If ```f``` doesn't exist the rest is aborted, and thus ```dumpFile()``` with the fully played entries is never called.

Nadahar:
@SubJunk Right now I'm trying to figure out why you've used these static class variables like you have:
```java
	private static int thumbnailFontSizeVideo;
	private static int thumbnailTextHorizontalPositionVideo;
	private static int thumbnailTextVerticalPositionVideo;
	private static int thumbnailFontSizeAudio;
	private static int thumbnailTextHorizontalPositionAudio;
	private static int thumbnailTextVerticalPositionAudio;
	private static int thumbnailFontSizeImage;
	private static int thumbnailTextHorizontalPositionImage;
	private static int thumbnailTextVerticalPositionImage;
```
Many of these are never written, and thus will always be 0 from class initialization. You then use them to initialize the local variables in ```DLNAMediaInfo.parse()```, but you could just as well have set them to 0. I'm assuming that you forgot to set them, but what's the idea behind it? ```thumbnailFontSize<Type>``` is written if you have to reduce the font size. Is this simply to don't have to do the ```FontMetrics``` calculation over and over, so that it's only done once and then is correct after that? Was the idea to use the others in a similar way?

SubJunk:
@Nadahar like I said, if I delete UMS.mon and then watch something (to test, I fast-forwarded to the end of a video) UMS.mon is created. Maybe there's a better way to ensure the creation of that file.

Yes those variables are to prevent having to do the calculations each time and it looks like I forgot to finish their implementation, I've just done that. Nice catch :)

Nadahar:
@SubJunk Ok, I don't know what causes it then, but I could easily solve it by making sure ```parseMonitorFile()``` created the file if it didn't exist. I don't see how it could hurt.

Nadahar:
About the scaling of thumbnails. Do all renderers have a fixed thumbnail size? As it is now, the default is a "wide" format that is padded with black. I'd think it would be much nicer if the renderer could pad either with it's own background color or transparent. I realize that some renderers probably stretch instead, and in those cases it's better that UMS pads. Also, is thumbnails generally jpg or png? If png, we could pad with transparent instead of black.

When browsing with the web browser (web interface) now it doesn't look very nice.

SubJunk:
@UniversalMediaServer/developers I've just implemented a safety measure that will prevent us from merging any pull request into master if it either has failing tests or is out of date with master. I meant to do it ages ago but just found the setting :)

Nadahar:
@SubJunk Great ;)

SubJunk:
It also stops any commits from being pushed directly to master, so everything needs to go through Pull Requests now. I'm okay with that, but is it okay for you guys @UniversalMediaServer/developers ?

Nadahar:
I actually think it's a good idea, but I think we should test #710 then to see if that makes code review easier and more efficient. Among it's features, is show you a diff of only the code that's changed since the last time you reviewed that PR.

Nadahar:
I'm not so sure if the status checks are such a good idea anymore. Status checks in itself is great, but the fact that they force any small change to be a PR has consequences I didn't see. It means that we will have a LOT of merge commits in our history. Merge commits are close to pure evil in my opinion as they make history very hard to read and make rebasing more or less impossible.

When you have to make a PR for every version bump or changelog update, it will really impact history. I wish there was a way to force status checks on PR's but allow (small) commits directly, but it seems it isn't. I'm unsure what's the best option from those available at the moment.

SubJunk:
Yeah it is a shame it works like this. I"d like to still be able to push commits to master directly.

Nadahar:
Yes, personally I'd love for anything more than tiny changes to be PR's, but I think having PR's for every tiny little update of a comment or a version is way over the top. We're not that many, we could simply agree about how much should be a PR and what can go directly. Agreeing to always check for the green tick before merge shouldn't be hard at all, and it has only slipped once as far as I know. Maybe it's time to reconsider?

@UniversalMediaServer/developers Don't you guys have some views on this?

I still think protected branch is a good thing in itself, even without forced status checks.

SubJunk:
I've disabled the feature for now. If GitHub makes an update that allows us to use this feature for PRs but still be able to commit to master I'll re-enable it

Nadahar:
Agreed.

valib:
@Nadahar I agree with you. I think that PR's should be a new features or any changes which needs to be tested by us or community. The rest like change wording, renaming varibles to be more comprehensible (of course without influencing plugins), simple bug fixes etc. can go directly.

Nadahar:
@valib The problem is exactly where to draw the line. I'd say that any change that might end up in multiple commits absolutely should be a PR, and not only things that need to be tested. Things could also need to be reviewed (that is, that somebody else has some information relating to that change that the person making it doesn't have, and that need to be taken into consideration) without the need for extensive testing.

valib:
@Nadahar I agree and it is up to @SubJunk as an owner of this project to define that rules.

SubJunk:
Yeah I agree with you guys that only simple things should go into master. It's not the end of the world if something slips through the cracks since we can always revert, but we should strive for that. I agree that multiple commits should end up in a PR - master should only be single, complete commits with low impact and low chance of causing bugs.

PR #743, opened 2015-11-26
Updated README.md translators section

Nadahar:

In the short time we've been using [crowdin] (https://crowdin.com/project/universalmediaserver) some people have already delivered significant translations. We have actually gotten support for a new language (Slovak) that is already 100% translated in the last two weeks.

I figured it was time to update the translation credits based on this, and here is my suggestion.


SubJunk:
Nice one

PR #748, opened 2015-11-30
Added support for renderers AnyCast and Miracast M806

prescott66:
I have created two new conf files for two devices(AnyCast and Miracast M806). They are cheap DLNA renderers, but works well with my configs.
Special thanks to nadahar for his help!


Nadahar:
Regarding ```AnyCast.conf```, there's a problem with ```UserAgentSearch``` (in addition to having 2 entries). What is there now, is the whole agent sent out by the renderer. It looks very general, and it seems that the only copied some Firefox user agent. That's not a good idea imo, but what can be used when there is no "genuine" information in the string? Are there any other smart tricks to get the recognition right?

Nadahar:
@skeptical Aren't you the one to ask about recognition? What's the best / least worst way to recognize the AnyCast?

prescott66:
Yeah,i want to make UMS more usefull for more users, so feedback from devs will be appreciated.

skeptical:
@Nadahar if there's nothing truly unique in either User-Agent or dlna details (not sure if you've looked at it too?) there's no general solution.  The user can always manipulate loading priority, or 'select renderers', or customize the device conf (i.e. override for a specific uuid) using the 'customize' button, but you already know this.

Nadahar:
@skeptical The uPnP/DLNA detail are fine. It's the HTTP/user agent part that's problematic.

skeptical:
@Nadahar then the way it's supposed to work is that it may be temporarily misrecoginized if we see the ua first, but then re-recognized correctly when we receive the upnp details.

Nadahar:
@skeptical Ok, but what about ```UserAgentSearch``` then? Do we leave it blank, or fill int the Android dupe agent?

skeptical:
@Nadahar If the Android misrecognition is useful as a fallback, put it in, otherwise it will be an "unknown renderer" in its first reincarnation. Disclaimer, I'm going from memory here, so it may be that there's some additional wrinkles.

prescott66:
So i should revert useragentsearch to original,as it was before?

Nadahar:
@skeptical The Android user agent will work great for this device, my worry is that actual Android devices of the same version and language could get misrecognized as this device.

skeptical:
@Nadahar Setting ua in the way @prescott66 has done it is "possibly" safe, since it can only match itself, though as you say it looks so generic that there may be other android devices that would also send the same exact ua. In general setting a literal ua is a brittle approach, since the same device may use different uas in different contexts, and ua also often changes with firmware updates.

If you omit ```UserAgentSearch``` then Android.conf will be matched to the AnyCast based on ua (if we see ua first, the order of events is unpredictable).  In this case almost immediately afterwards we should get dlna details and switch to AnyCast.conf, since android is generic and low priority (see [UPNPHelper.rendererFound()](https://github.com/UniversalMediaServer/UniversalMediaServer/blob/master/src/main/java/net/pms/network/UPNPHelper.java#L635)).  Unless the user manages to press 'play' during this brief limbo period (which sometimes happens :-) there's no downside and I think this is preferable.  I would try this and restart a few or more times to see whether there are any issues.  Note that some renderers are a bit manic in their search behaviour (e.g. Panasonic if I remember right), and we always get an http request before dlna details.

Nadahar:
@skeptical @prescott66 The "good" thing with this device in this regard, is that it doesn't have an interface and has no browse functionality. It has to be controlled by a control point, so I'd think that it's unlikely that it will do anything http by itself. Once a control point is up, uPnP is probably estabilished unless ofcourse UMS is started at exactly the wrong time. I'd say we should go with a blank UA here.

prescott66:
@skeptical @Nadahar Guys so what to do now? For me works this devices well, with configs as you can see now(after all my commits).

Nadahar:
I think the right thing to do is to leave the user agent search blank, so that uPnP is the only way to recognize it after getting input from @skeptical and thinking some more about it as explained in my last comment.

You should probably just rebase 61de469cd70c37c71597c7c5e7564b7247b99117 away, to minimize commit history mess.

prescott66:
UA is blank now. i think this PR is ready to merge :smile:

Nadahar:
I think this is good to go now that the user agent stuff is sorted out, so I'll merge it now.

Issue #750, opened 2015-12-09
Live subtitles stopped working on standard def files since UMS 5.0.0


a1vdu:
Hi there, I'm using UMS with java 8 (64bit) on win7/64bit and using a PS3 as renderer.
I've made a clean install and I've tried ums 4.4.0, 5.0.0, 5.1.0, 5.2.0, 5.3.0
Since versione 5.0.0 Live subtitles won't show on standard def mp4 files (for example Arrow.S03E01.HDTV.x264-LOL.mp4 ) but they work just fine on mkv files ( The.Vampire.Diaries.S07E06.720p.HDTV.X264-DIMENSION.mkv ).

More in depth: If the standard def mp4 video in question already have an external subtitle downloaded, and I go on the Live subtitles option and I choose for example a different language, then the live subtitles works just fine.

Thanks in advance
vdu

(github won't make me upload the log file.. I had already a post with the log on the ums forum here http://www.universalmediaserver.com/forum/viewtopic.php?f=9&p=21621#p21621 )

Sami32:
https://github.com/UniversalMediaServer/UniversalMediaServer/issues/515

SubJunk:
Live subtitles have been removed because Open Subtitles went paid

PR #752, opened 2015-12-10
Make disable subtitles setting really working

valib:
This is a try to fix the option when subs are generally disabled. Curently are subs disabled but shown at the renderer.

[Review on Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/752)

### DIFF
--- a/src/main/java/net/pms/dlna/DLNAMediaInfo.java
+++ b/src/main/java/net/pms/dlna/DLNAMediaInfo.java
@@ -1394,7 +1394,7 @@ public class DLNAMediaInfo implements Cloneable {
 		}
 
 		// Check for external subs here
-		if (f.getFile() != null && type == Format.VIDEO && configuration.isAutoloadExternalSubtitles()) {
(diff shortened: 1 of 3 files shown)


Nadahar:
@valib I don't have a lot of movies, with or without subtitles, so I have limited testing abilities. When it comes to live subtitles, isn't that subtitles as well and should be disabled when subtitles are disabled?

That said, I don't know exactly why anyone would disable subtitles in general. If the audio/subtitle languages are configured correctly, shouldn't they appear only when really needed? Do you know what this is usually used for? If it's used to shut off subtitles because they cause trouble with a certain renderer (transcoding won't work etc.), shouldn't there be a renderer setting? I'm just guessing here, I can't quite imagine the situation you'd want to disable them.

@valib Did you look at #750 also when dealing with this?

valib:
@SubJunk I see a lot of PRs whose are waiting to be resolved. I think you should decide which of them should be merged and do it.

PR #755, opened 2015-12-14
3 dsubs merged with master

valib:


Nadahar:
@valib Why close?

valib:
@Nadahar because it was a wrong test of the merging. I still have a problem with the Eclipse setting when some of the testing subtitles files are not intentionaly made as changed and included to the commit. It would be nice to share the general Eclipse setting to avoid that unwanted changes.

Nadahar:
@valib Are you committing from Eclipse? I've never tried that, I'm using [TortoiseGit] (https://tortoisegit.org/) for most things git and it doesn't include untracked files if that's what happens. When I don't use TortoiseGit I use git command line.

You should try [TortoiseGit] (https://tortoisegit.org/) if Eclipse is giving you a hard time, I think it works really well under Windows adding overlay to file icons showing their current status and having all git operations available with rightclicking on files or folders.

valib:
Merged together with 3D subs changes

Issue #754, opened 2015-12-12
Samsung UN55ES6100 TV disconnects while streaming


Nadahar:
There's a [thread in the support forum](http://www.universalmediaserver.com/forum/viewtopic.php?f=9&t=5747&p=21645#p21645) about problems with disconnects after about 15 minutes when streaming to a Samsung TV on wired network. I wouldn't think there was much that could go wrong when streaming over wired network. UMS stops serving because the renderer disconnects, and that in itself looks normal.

The only thing I can find in [the log](http://www.universalmediaserver.com/forum/download/file.php?id=2869) that could possible be an issue is that UMS seems to switch renderer profile during playback. Could this be a problem, that we start out with a working profile and then switches to a profile that doesn't behave with the renderer in question? I'd like for you @skeptical to take a look and see if you understand what happens. Is the `UpnpDetailsSearch = [CD]S?\d{3}\d? , (Samsung|AllShare)` in the `Samsung C/D Series` profile simply casting to wide a net? It's matching `[TV]UN55ES6100 192.168.0.12 uuid:0d1cef00-00dc-1000-9bfe-1c5a3ed197c3 Samsung Electronics UN55ES6100 AllShare1.0 Samsung TV DMR http://www.samsung.com/sec http://www.samsung.com/sec`, but the `(Samsung|AllShare)` will match any device with `Samsung` **or** `AllShare` in it's uPnP details, won't it? That must be a mistake..?

skeptical:
@Nadahar the problem is case-insensitive match in ```RendererConfiguration.matchUPNPDetails()```, where the ```Samsung C/D``` regex:
```ini
UpnpDetailsSearch = [CD]S?\d{3}\d? , (Samsung|AllShare)
```
(after expanding '``` , ```' to '```.*```') gets a hit:
> [TV]UN55ES6100 192.168.0.12 uuid:0d1cef00-00dc-1000-9bfe-1c5a3e**d197c3 Samsung** Electronics UN55ES6100...

I think case-insensitive is obviously counter-productive here, will fix it in a bit when I get a chance (unless you do it first :-).

Nadahar:
[...]

PR #760, opened 2015-12-18
Updates to renderer search

Nadahar:
This PR started with #754 where UMS would switch renderer profile away from the correct one during playback. The switch ended up not being the problem, and #754 is still unresolved.

@skeptical quickly identified the problem, that parts of the UUID for the device matched the ``` UpnpDetailsSearch``` for the wrong profile, which was then given a positive match. The suggested resolution was to make the match case sensitive which would resolve the mismatch in this case.

But, me being me, I went in for a look and came up with some further suggestions. This PR is just that, suggestions. I'm not quite sure if everything here is a good idea, but I'd like to discuss it so I create the PR now. 

My suggested changes explained:

The first commit is uncontroversial and is mostly a bugfix from things I found while testing. If a renderer profile contained a image file name that couldn't be found, it would result in an NPE. I handled the NPE, and added a "image not found" image and a fallback to a Java icon if that file doesn't exist as well. This isn't a bug most user will meet, but it took a little time before I understood what was happening when I fiddled with my test renderer configuration, so I though it would be nice for others not to have to go through the same process. The commit consists of:
* Updates to renderer icon on StatusTab:
  * Handle a null icon without NPE
  * Created constructor in ImagePanel that takes an icon
  * Added new icon for missing icons "missing-icon.png"
  * Made logic do:
    * Try to add "missing-icon.png" if renderer icon isn't found
    * If the above failed, try to use the UIManager icon ```OptionPane.warningIcon```
    * If both the above fail, don't try to create a renderer icon action - no renderer panel will be available, but this should never happen

In the second commit I've made a new method ```UPNPControl.getDeviceMatchableDetailsString()``` that will build a space separated string with the UPnP details just like ```UPNPControl.getDeviceDetailsString()``` does, but which excludes ```address``` and ```udn``` (UUID). I can't see why we'd ever want to use those for matching a renderer profile, and the ```Device``` matching that uses IP address or UUID has it's own separate logic untouched by this. This should reduce false positives.

The third commit several things, and is probably the most controversal. I've made a new method ```StringUtil.matchConfigurationRegEx()```. My intention was to make a method that could be used universally for regex matches in configuration files to give the users some consistency. Today different rules apply to different fields, and I think it's unrealistic that the users will be able to keep track of these, I can hardly manage myself. I'm not sure if ```StringUtil``` is the correct location for this though, and I'm not very satisfied with the name but it's the best I've been able to come up with so far.

My idea was to use this for ```UserAgentSearch``` and ```UpnpDetailsSearch``` for now, and then migrate other regex fields later. I've only replaced ```UpnpDetailsSearch``` this far though, because I simply can't quite understand why things are done as they are for ```UserAgentSearch```. To me it seems like what's there is over complicated for what it does, so something must be eluding me. I don't understand why ```UserAgentAdditionalHeaderSearch``` is combined with ```UserAgentSearch``` either.

Not having ```UserAgentSearch``` use ```matchConfigurationRegEx()``` voids much of the purpose of ```matchConfigurationRegEx()``` as it does nothing to make things more consistant, so I'm currently not sure where to go with this from here. If I'm going to integrate it with ```UserAgentSearch``` I need to understand more of why things are done as they are.

In any case, the matching rules for ```matchConfigurationRegEx``` are:
* Matching is case sensitive unless the search string starts with ```(?i)```
* Normal regex rules apply, but dot matches all including newline
* If leading or trailing spaces are to be included, the regex must be double quoted
* Double quotes are stripped only if they exist in the first and last position of the regex, flags excluded

Examples of valid syntax:
* ```Case sensitive search term```
* ```(?i)caSE inSensitive search term```
* ```"Double quoted search term with trailing spaces   "```
* ```(?i)"caSE inSensitive double quoted search term"```

Initially I was going to use a simpler to case insensitivity flag, but in the end I ended up with the standard regex flag since there's no point in reinventing the wheel. That means that my code handling this is partly redundant, as regex would understand this by default. My code does however allow for the flag to be placed outside the regex string (if in double quotes), so I left it there for now as I figure it might be a little less confusing for users not having to mix it in the string.

I've also removed the ```' , '``` substitution for ```.*``` as discussed with @skeptical. But, I'm unsure if this is safe regarding backwards compatibility. I've updated the documentation and changed all the renderer configurations that used ```' , '```, so in that sense it's safe. But, any custom renderer configurations out there that use it would stop working. I honestly have no idea if that is a real-world problem or not. The reason for not keeping the substitution (but leave it undocumented) is my idea that this method should be universal. The substitution has only been a rule for the ```UpnpDetailsSearch```, so such a backwards compatibility could alter matching for other fields like ```UserAgentSearch``` that would use the method.

Some decisions have to be made. I strongly believe that it's important to standardize parsing, as the alternative is alienating to users. But, the situation is already at a point where I can't find any way of doing this without breaking anything. The easy solution is obviously just to leave everything as-is, but that will have penalties in the long run.

In a broader sense I think that is a major contributor to killing open source projects. If things get so specialized and intricate that only a few people understand it, only those few people will be able to maintain it. Increasingly they will wear down from feeling that they are the only ones "pulling the load", and in the end they will quite contributing, leaving the project in a state where noone is able to maintain or develop the project. I'm not saying that this little detail is that big a deal, but I think that general idea is something to keep in mind and always strive to make things as logical as possible *seen from the outside*. In the long run, that's the only thing that can keep the ship floating, keep new people coming in and sharing the load.

As you understand I'm a bit stuck with this, and I need some input to know where to go from here.

[Review on Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/760)


skeptical:
@Nadahar the basic logic of both upnp details search and ua search is actually consistent and not that complicated, the point is to avoid multiple match operations:
* sort both the regex fragments and the string values by 'key', i.e. by either http header name or upnp detail name as the case may be, so they're in the same order.
* join the regex fragments into one grand regex using ```.*``` as separator, and join the string values similarly into one glob with space as separator.
* perform just one match, after which you know conclusively if you have a hit or not.
Besides the obvious regex efficiency, it avoids having to decide how to proceed while in a partially-matched limbo state, which was the case previously when we used to go one header at a time in random/found order, and made things like sorting out renderer priority completely insane.

But you're right that it's a bit obscure in the case of ua, since the new logic is retrofitted onto the old ```UserAgent/AdditionalHeaderSearch``` paradigm, when in fact it's now effectively processed as just one combo ```"HTTPHeaderSearch"``` under the hood, same as ```UpnpDetailsSearch```.  So I'd be in favor of deprecating ```UserAgent/AdditionalHeaderSearch``` in favor of ```HTTPHeaderSearch``` or whatever, so it's clear that our matching logic is essentially the same.  A side point is that we only use ```AdditionalHeader``` for a few renderers anyway, so its presence as a separate setting tends to confuse rather than clarify things in the current scheme.

Regarding '``` , ```' as you know I don't mind discontinuing it, but it may be necessary to compromise the conceptual purity of a universal function since anyone who's pressed the 'Start a new configuration' button in the past year or so would end up with a broken conf otherwise (and btw ```RendererConfiguration.createNewFile()``` still needs updating in this regard).  This is why I suggested just undocumenting it and no more.  Or of course you could add code to repair and rewrite '``` , ```' confs on the fly too, since they would still be backward-usable and nothing would actually break.

I'm also good with ```matchConfigurationRegEx``` as a concept, i.e. why not have one flexible reusable function, but I don't think it's necessary for clarity or saves us from being killed or anything.  Clarity is external, i.e. in this case it means sticking with standard regex syntax in the confs (```(?i)``` good, '``` , ```' not-so-good in retrospect) and consistency in matching logic (which we already have underneath even though it isn't expressed well on the surface).  In terms of maintaining code regex is regex, it doesn't matter too much to me whether it happens in one unified function or not, it's understandable either way imo.  But I agree in principle that redundant code is bad, and we have way too much of that, and will still have massive amounts of that with or without ```matchConfigurationRegEx```.

Also small point, but why do you need to rename ```r``` to ```rendererConf``` in just one function, do you really think this is vital?  In our codebase we already use mostly ```renderer, render, mediaRenderer, and r``` to denote renderers, is it necessary to push yet another variant? Personally I like 'r' when I can't use 'renderer', since it's concise/readable/understandable when confined to local scope and I don't much like java-style long variable names anyway.  In things like this we should either all agree on standards, or otherwise try to live and let live.  There are endless cases of using 1-3 letter short var names in the codebase, and on the whole this is a good thing imo.

Nadahar:
@skeptical Thank you for replying.

About the user agent / additional header search logic, I still don't feel like I see the whole picture. I don't know what I'm missing, it just hasn't "clicked" for me yet. I understand the concept of user agent and headers, but I'm not quite on board with how it's done - and what's there for performance reasons, what's there for historical reasons and what's there for practical reasons. You gave me some useful input though, so I guess I just have to chew on it a bit more.

You make a couple of good points about ```' , '```. 

Regarding ```matchConfigurationRegEx``` as a concept, I never said that alone was critical, I merely tried to explain how I'm thinking around the subject. Every little drop counts as I see it, the total is the sum of all the details after all. My thoughts about maintainability wasn't about this particular case, but about "obfuscated" code in general.

My concern for clarity/consistency is on two levels: 
* From the view of the "advanced" users who try to build their own renderer profiles either for their own use or for contributing, but also for developers who lacks in depth knowledge of this or simply doesn't remember.
* From the view of the "external" developer trying to create a PR, find a bug or learn more about the project.

We might not agree on this point, I just think that the sooner a person unfamiliar with UMS understands the logic used, the better. This is my main concern, but I guess what is easy to understand/intuitive is somewhat subjective. 

I agree that code redundancy is bad for several reasons, but I also think that trying to make one method or object or concept do too many different things is bad if the things simply are to different and it ends up getting very complicated to cater to all needs. From that point of view alone, I don't think I'd advocate trying to generalize this, and I'm very aware that for it to have any value the amount of "special case" handling in the generalized code must be kept to a minimum. If that can't be achieved, I think keeping separate code is often better.

About renaming ```r``` to ```rendererConf``` I have to do things like this when I'm trying to read unfamiliary code. I simply need the names to be somewhat descriptive to be able to read the code. I don't pay that much attention to the exact name, it just needs to tell me what this object is without browsing further up again to find its declaration and initialization. I generally like for the object name to give me a hint about it's type. Most of the time I reverse those changes before committing, but in this case I had also changed something in there and couldn't just reverse it easily from diff view and figured that it couldn't hurt. I can change it back, it's in no way important to me that it stays like that.

I generally dislike short variable names though, even though I agree that locals is of less importance. In earlier days I used to use several such short names as a kind of "standard", but I've more or less quit doing that as I've experienced how painful it can be to read such code written by others. I think the only one that's stuck with me is ```i``` as a general integer counter/index, as I consider it almost a de facto standard. I think the extra time it takes to write is minimal, and it's worth it in terms of readability. Any decent IDE will have code completion anyway, so I rarely write it out more than the first time. Some times I write the method with "shorts" first, and then rename them when it's done, but I strive not to have "shorts" in the final code except ```i```.

Nadahar:
Undid variables renaming.

SubJunk:
I'm in favor of using standard but descriptive variable names. I think "renderer" is nicer than "r". If someone sees "r" they have to spend a few seconds finding out what it is, whereas renderer saves that time, and if we do that with all variables it adds up to significantly less time spent trying to figure out the code. I have renamed variables like that before.
I agree that we should strive to use the same variable name, maybe we could do a reformatting to standardize instances of RendererConfiguration to renderer, not that it's that important

Nadahar:
There's no reason this should stay open as it was never completed.

PR #761, opened 2015-12-20
Renderer hot-swapping fixes

skeptical:
This came up indirectly as a result of #754, where the renderer conf is switched during playback because it takes 15 minutes or so before we see the renderer via cling.  In this particular case there were no dire consequences specifically related to this hot-swapping afaict (due to dumb-luck mostly :-), but it wasn't a scenario I'd actually contemplated before and made me go back and reconsider.  What wasn't happening after the renderer switch was that the renderer's playback "controller" wasn't being upgraded from a passive playback timer to a upnp control, so this pr is a quick take on making that happen smoothly and avoid some possible pitfalls.

The main thing to be sure of is that there aren't any weird side effects in the normal case, i.e. for the vast majority of renderers where this doesn't happen, so if people would just run it for a while during normal use and confirm that all is well (or not, as the case may be) I'd appreciate it.

The second thing to test is the actual efficacy, which I've been simulating by hacking upnp service startup to be delayed until we're in the midst of playback, using the patch below and so far it seems to do the job vis internals and gui, but please try to trip it up if you feel the urge and bring any defects to light.

[upnp-hotswap-test.diff.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/67645/upnp-hotswap-test.diff.txt)

[Review on Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/761)

### DIFF
--- a/src/main/java/net/pms/configuration/DeviceConfiguration.java
+++ b/src/main/java/net/pms/configuration/DeviceConfiguration.java
@@ -87,7 +87,7 @@ public class DeviceConfiguration extends PmsConfiguration {
 		// Initialize our internal RendererConfiguration vars
 		renderCache = new HashMap<>();
 		sortedHeaderMatcher = ref.sortedHeaderMatcher;
-		// Note: intentionally omitting 'player = null' so as to preserve player state when reloading
+		// Note: intentionally omitting 'player = null' so as to preserve player state when reloading/replacing
 		loaded = true;
 		this.ref = ref;
 
--- a/src/main/java/net/pms/configuration/RendererConfiguration.java
+++ b/src/main/java/net/pms/configuration/RendererConfiguration.java
@@ -1534,6 +1534,18 @@ public class RendererConfiguration extends UPNPHelper.Renderer {
 		return player;
 	}
 
+	public void resetPlayer() {
+		// Replace the player if already existing and transfer its state
+		if (player != null) {
+			BasicPlayer old = player;
+			player = null;
+			getPlayer().replace(old);
+			LOGGER.debug("{}: swapped data from {} to {}", getRendererName(), StringUtils.substringAfterLast(old.getClass().getName(), "."), StringUtils.substringAfterLast(player.getClass().getName(), "."));
+		} else {
+			getPlayer();
+		}
+	}
+
 	/**
 	 * Sets the UPnP player.
 	 *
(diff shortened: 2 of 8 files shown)


skeptical:
@SubJunk regarding your [comment](https://github.com/UniversalMediaServer/UniversalMediaServer/pull/736#issuecomment-158855665) in #736 that we don't seem to get AVT PostionInfo from your Panasonic, try this branch with AVT position logging enabled:
```diff
+++ b/src/main/java/net/pms/network/UPNPControl.java
@@ -657,7 +657,8 @@ public class UPNPControl {
 		if (svc != null) {
 			Action x = svc.getAction(action);
 			String name = getFriendlyName(dev);
-			boolean log = !action.equals("GetPositionInfo");
+//			boolean log = !action.equals("GetPositionInfo");
+			boolean log = true;
 			if (x != null) {
 				ActionInvocation a = new ActionInvocation(x);
 				a.setInput("InstanceID", instanceID);
```
and maybe logs will shed some light.  Also, I seem to recall we already looked at this at some point, but couldn't find it via search.

SubJunk:
I finally did this but the renderer isn't sending useful information, just this continuously:

```
DEBUG 2016-04-10 18:06:38.679 [UPNP-VIERA VT60 Series] Received from VIERA VT60 Series[0]: Track=0
DEBUG 2016-04-10 18:06:38.679 [UPNP-VIERA VT60 Series] Received from VIERA VT60 Series[0]: TrackDuration=0:00:00
DEBUG 2016-04-10 18:06:38.679 [UPNP-VIERA VT60 Series] Received from VIERA VT60 Series[0]: TrackMetaData=
DEBUG 2016-04-10 18:06:38.679 [UPNP-VIERA VT60 Series] Received from VIERA VT60 Series[0]: TrackURI=
DEBUG 2016-04-10 18:06:38.679 [UPNP-VIERA VT60 Series] Received from VIERA VT60 Series[0]: RelTime=0:00:00
DEBUG 2016-04-10 18:06:38.679 [UPNP-VIERA VT60 Series] Received from VIERA VT60 Series[0]: AbsTime=0:00:00
DEBUG 2016-04-10 18:06:38.679 [UPNP-VIERA VT60 Series] Received from VIERA VT60 Series[0]: RelCount=0
DEBUG 2016-04-10 18:06:38.679 [UPNP-VIERA VT60 Series] Received from VIERA VT60 Series[0]: AbsCount=0
```

SubJunk:
Do we want to merge this?

skeptical:
@SubJunk hard to believe the Viera has a non-functional AVT service, maybe we're not supporting it right. If you want to test it independently try the cling [workbench](http://4thline.org/projects/cling/workbench/), which will let you monitor the service, set a UMS uri, etc.  Example with Kodi:
workbench
I'd try playing something on the Viera and seeing if workbench's AVT window is picking up any info at all, esp look at the 'instance id' tabs (0-7) across the top. UMS only looks at instance id 0.

Part 2: yes, imo it should be merged since it supports discovery/recognition events correctly, but the catch is that then your Viera's ```PlaybackTimer``` would automatically be upgraded to ```UPNPHelper.Player```, and there would have to be some renderer setting to undo this if in fact the Viera's AVT is useless.

SubJunk:
Feel free to resurrect this

PR #657, opened 2015-08-29
Htmlify JLabels + fix gui overlaps

skeptical:
A partial and mostly untested fix for #656 which
- implements automatic wrapping of long gui labels by ensuring text is wrapped in `<html>` tags.
- fixes `FormLayout`s and cell coordinates in transcoding tab and removes problematic `CellConstraints` intended to overlap labels leftwards.

Please test extensively. It still needs much prettifying which I'm hoping someone else jumps in to do because I've had enough of swing/jgoodies headaches for one day (correction: lifetime tbh :-).

[Review on Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/657)

PR #762, opened 2015-12-20
Add Farsi translation

skeptical:
@Nadahar mostly done (at least for this pass, hopefully crowdin volunteers will jump in).  Our current gui's support for rtl is really poor, so for best viewing results this should be cherry-picked into the htmlify branch where I've fixed a lot of the rtl issues. There's also an issue with [weird hovering behaviour in the jgoodies WindowsLookAndFeel RTL Table headers](https://github.com/UniversalMediaServer/UniversalMediaServer/pull/657#issuecomment-144875697) which I haven't looked into any further, but everything's ok in other LAFs afaict. All of the above applies to Arabic and Hebrew too, of course.

EDIT: there's one or two stubborn bi-di messages which split wrongly in the translation too, but this may be due to the limitations of gedit, which I was using, and they probably can be sorted out via the crowdin interface.


SubJunk:
@skeptical happy for this to be merged if you want

Nadahar:
@skeptical Great, I've actually already added Farsi (3d393d18881bf0762566d0fe224739a4fd6dc11b) in the crowdin branch (#679). As soon as I've figured out how to properly size the ```JOptionsPane``` the branch is more or less ready, so I think it will be easier if I simply import your translations to crowdin instead of merging this PR.

In #679 translation coverage is imported from crowdin into UMS when translations are pulled, and any language not reaching a certain threshold (20% at the moment, but we should probably raise that) will be hidden. Because of that, I could add Farsi and it wouldn't be visible anyway as long as it was untranslated.

Has the htmlify branch "stalled" on something, or is it basicly ready? Maybe you should "merge it up to date" and we should do some final tests on it? I seem to remember one issue I don't think is addressed, and that's the fact that several places in the GUI a space is added after label texts to create padding between the label and the component. Htmlifying will render the spaces useless. Using space like this is obviously terrible design, but we need to replace it with something, or have you handled this somehow?

Nadahar:
@skeptical I've imported [your translations] (https://crowdin.com/translate/universalmediaserver/1/enus-fa) to crowdin. Maybe you can take a look at the troublesome texts there? As soon as I have your crowdin username I'll give you edit permission.

skeptical:
@Nadahar htmlify branch is done more or less (but I haven't worked on it for a while so I should go back and review things).  I updated it last night when I pushed this pull request btw, so it's mergeable.

I logged in to crowdin using github id, and apparently it considers my username to be aval57.  Browsing through, I'm surprised to see some of the Microsoft translator suggestions aren't bad (better than google sometimes).  It all turns into garbage as soon as we move beyond simple phrases, tho.

Nadahar:
@skeptical It's hard for me to judge this, but it seems to me like the font in Farsi (and probably Arabic, Hebraic etc) is larger than the latin fonts. That obviously will make layout problems worse. Would it be possible to reduce it some without compromising readability?

skeptical:
@Nadahar I think there are 2 separate issues here.  First that for fa/ar the ascenders tend to go higher and descenders lower than latin fonts and this doesn't seem to be handled quite right, i.e. I see a lot of (slight) cropping at the top and bottom of characters, which of course annoys me but you may also be picking up on that since it just doesn't look quite right graphically.  Same thing happens in linux so it's not the font itself but swing's handling of it.

The second much worse issue is that Windows defaults to a terrible heavy blocky hard to read font which despite being too big, is also poorly delineated and reducing its size would make it even more unreadable I think. As I recall the Windows font choice used to be much better a few months ago when I was working on this, not sure if java changed or Windows, I did a (for me rare :-) win7 system update in the fall. In linux the font is quite nice, but would suffer if reduced in size.

Nadahar:
@skeptical I'm closing this as since I imported the translations to crowdin and #679 now being merged, this is effectively merged.

PR #770, opened 2015-12-26
Rewritten 3D subs conversion fixed fot merging

valib:


Nadahar:
@valib I haven't gotten around to test this yet as I'm trying to finish the crowdin PR, but I have planned to test it as best as I can without 3D renderers available.

valib:
@UniversalMediaServer/developers this PR should be merged because when the ffmpeg new implementation was introduced the font size for 3D subs are incorrectly calculated. This PR fix it.

SubJunk:
I'm okay with merging it

Nadahar:
@valib I don't have an opinion on the actual changes, but I spotted some things in the diff that I'd like to change. One thing is the generalization, I doubt that ```<Object>``` is the correct type to use and I'd like to verify that. The second is that it's no point in updating the translations since we started using crowdin (only change ```message.properties```, the others are auto generated).

If you want I could do a quick commit to address these things.

valib:
@Nadahar the ```<Object>``` type is suggested by Eclipse so I use it otherwise there is a warning. And about the changes of the ```message.properties``` I deleted not needed mesages and maybe the czech translation is not needed becuse of the crowdin.

Nadahar:
@valib I know why Eclipse warns, it's because the type should be specified. But, I doubt ```Object``` is the correct type. Give me some minutes to look at it...

About the translations, it doesn't really matter if you change them but it serves no purpose. Translations have to be added at crowdin, the translation files are generated by crowdin. The reason why I don't want them in the PR is because changes like that can create merge conflicts (for other branches) for no reason.

Nadahar:
I'm okay with merging this too now

Nadahar:
@valib I think you can merge this now - I'll make some further changes to the checkboxes once this is merged.

PR #773, opened 2015-12-30
Updates to  #770

Nadahar:
@valib I made this PR to your PR. If you want to import my changes, just merge this PR and your PR will be updated.

I fixed the ```ComboBox``` models somewhat, but several of them should have ```KeyedComboBoxModel``` instead. That would resolve current issues with missing translation of parts of the values and the fact that the current value isn't selected when you enter the page (instead, a "copy" of the current value on the top is). But, I've done some major updates to ```KeyedComboBoxModel``` in #679, so I can fix that after #679 is merged.

I also reversed the translation changes simply to reduce potential merge conflicts.


valib:
@Nadahar I like the changes you made but reverting the translation changes made them a little bit confusing. The 3D depth is no longer in % but in pixels and the range is not -2 to 2 but -5 to 5.

Nadahar:
@valib I don't understand. I didn't change ```messages.properties```, the only one that matters. The other's will be overwritten by the crowdin import anyway.

The range of that checkbox was -5 to 5, it's easy to change it to -2 to 2 if you want, but the things I did didn't change any functionality...?

valib:
@Nadahar I see your point. I am not so much familiar with crowdin and if it will be ovewritten I have no objection to merge this PR as it is.

Nadahar:
@valib The whole point with crowdin is that it manages all the translations. We don't edit the translation files anymore, we edit ```messages.properties``` and those changes are made visible to the translators that then change their respective translations. Finally, new files are generated at crowdin and integrated into UMS.

I'm going to merge this now then.

PR #776, opened 2016-01-01
Implemented workaround for Java bug JDK-8034057

Nadahar:
Based on [this forum thread] (http://www.universalmediaserver.com/forum/viewtopic.php?f=9&t=5854).

[JDK-8034057] (https://bugs.openjdk.java.net/browse/JDK-8034057) is a rare bug, and I'm not sure if the exact bug that cased this issue, but it any case it's very closely related. The bug is fixed, but not for any released Java version so it seems we have to handle it for quite a while.

I found out after some testing that I got a particular exception, ```FileSystemException``` when this occurred if I made the call "one step deeper" into NIO. I've then brought back and slightly modified the "old" write tests that actually try to write to a file to determine write permission, and when this particular exception is thrown I fall back to trying to write to a file or folder.

It will make UMS run for anyone who's using the open source ImDisk RAM drive for Windows as their temp folder.

It shouldn't need much testing since the old, tested code is used. I have tested that it resolves the issue with ImDisk.

### DIFF
--- a/src/main/java/net/pms/util/FilePermissions.java
+++ b/src/main/java/net/pms/util/FilePermissions.java
@@ -21,8 +21,15 @@ package net.pms.util;
 
 import java.io.File;
 import java.io.FileNotFoundException;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.nio.file.AccessDeniedException;
+import java.nio.file.AccessMode;
+import java.nio.file.FileSystemException;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import com.sun.jna.Platform;
 
 /**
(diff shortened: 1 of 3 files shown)


Nadahar:
@UniversalMediaServer/developers This seems to fix it, tested by both me and the reporting user. This is a small change, merge?

src/main/java/net/pms/util/FilePermissions.java
```diff
@@ -176,4 +251,73 @@ public synchronized String toString() {
...
+	/**
+	 * Must always be called in a synchronized context
+	 */
+	private boolean testFolderWritable() {
+		if (!folder) {
+			throw new IllegalStateException("Can only be called on a folder");
+		}
+		boolean isWritable = false;
+
+			File file = new File(
```
SubJunk: Indentation?

SubJunk:
Looks good to me

src/main/java/net/pms/util/FilePermissions.java
Nadahar: Copied the old code and reorganized some - a slip - will fix.

Nadahar:
I don't think this will introduce any problems, and it solves one so I'll merge this now.

PR #777, opened 2016-01-02
Proposed changes to 6.0.0

Nadahar:
@SubJunk I've had a quick run through the code, and suggested some changes. Use what you want of this, I'll comment in the commit where I think that's needed.


src/main/external-resources/UMS.conf
```diff
@@ -436,7 +436,7 @@ hide_new_media_folder =
 #   2 = Hide the file
 #   3 = Move the file to a different folder
 #   4 = Move the file to the recycle/trash bin. It will not do anything if
-#       there is no bin on the operating system
+#       this isn't possible
```
Nadahar: Changed because it isn't true - it depends on the filesystem and the device, not only the OS. Windows for example have no recycler support in many cases.

src/main/java/net/pms/dlna/DLNAMediaInfo.java
```diff
@@ -807,7 +807,7 @@ public void parse(InputFile inputFile, Format ext, int type, boolean thumbOnly,
 							int thumbnailHeight = renderer.getThumbnailHeight();
 
 							// Make sure the image fits in the renderer's bounds
-							boolean isWatchedThumbnail = configuration.getFullyPlayedAction() == 1 && MediaMonitor.isWatched(file.getAbsolutePath());
+							boolean isWatchedThumbnail = configuration.getFullyPlayedAction() == 1 && MediaMonitor.isFullyPlayed(file.getAbsolutePath());
 							thumb = UMSUtils.scaleImage(thumb, thumbnailWidth, thumbnailHeight, isWatchedThumbnail);
```
Nadahar: Changed to be more general - watched only applying to videos

src/main/java/net/pms/dlna/DLNAResource.java
```diff
@@ -567,8 +567,8 @@ public void addChild(DLNAResource child, boolean isNew) {
 					}
 
 					// Hide watched videos depending user preference
-					if (configuration.getFullyPlayedAction() == 2 && child.format.isVideo() && MediaMonitor.isWatched(child.getSystemName())) {
```
Nadahar: Format is unreliable to determine content type, several (container) formats can either audio or video. That was the idea behind the "new" methods on media instead of format, they actually look at the content not the file extension.

src/main/java/net/pms/dlna/DLNAResource.java
```diff
@@ -1536,9 +1536,17 @@ private String getDisplayName(RendererConfiguration mediaRenderer, boolean withS
 				mediaRenderer != null &&
 				!mediaRenderer.isThumbnails() &&
 				configuration.getFullyPlayedAction() == 1 &&
-				MediaMonitor.isWatched(file.getAbsolutePath())
+				MediaMonitor.isFullyPlayed(file.getAbsolutePath())
 			) {
-				displayName = Messages.getString("DLNAResource.4") + ": " + displayName;
+				if (media != null) {
```
Nadahar: I take it this was forgotten since you had created the strings.

src/main/java/net/pms/dlna/MediaMonitor.java
```diff
@@ -20,24 +20,22 @@
 import org.slf4j.LoggerFactory;
 
 public class MediaMonitor extends VirtualFolder {
-	private static Set<String> watchedEntries;
+	private static Set<String> fullyPlayedEntries;
 	private File[] dirs;
 	private PmsConfiguration config;
-	private static final FileUtils FILE_UTILS = FileUtils.getInstance();
```
Nadahar: I'm not sure if MediaMonitor is always just one thread (even with multiple renderers playing back), but having shared static instances is a pain synchronization wise. I recon move to trash/recycler isn't that frequent an operation and that the avoided synchronization issues is worth the extra construction.

src/main/java/net/pms/dlna/MediaMonitor.java
```diff
@@ -156,13 +154,15 @@ public void stopped(DLNAResource res) {
...
-		LOGGER.trace("   minimum play time needed: " + (fileDuration * configuration.getResumeBackFactor()));
+		if (fullyPlayedAction > 0) {
+			LOGGER.trace("Fully Played feature logging:");
+			LOGGER.trace("   duration: " + fileDuration);
+			LOGGER.trace("   getLastStartPosition: " + res.getLastStartPosition());
+			LOGGER.trace("   getStartTime: " + res.getStartTime());
+			LOGGER.trace("   getLastStartSystemTime: " + res.getLastStartSystemTime());
+			LOGGER.trace("   elapsed: " + elapsed);
+			LOGGER.trace("   minimum play time needed: " + (fileDuration * configuration.getResumeBackFactor()));
+		}
```
Nadahar: I assume this logging only is relevant if fully played action is enabled

src/main/java/net/pms/dlna/MediaMonitor.java
```diff
@@ -206,49 +206,44 @@ public void stopped(DLNAResource res) {
...
-					File watchedFile = new File(rf.getFile().getAbsolutePath());
+					File playedFile = new File(rf.getFile().getAbsolutePath());
 
 					if (fullyPlayedAction == 3) {
 						// Move the video to a different folder
-						String newDirectory = configuration.getFullyPlayedOutputDirectory();
-						if (!newDirectory.endsWith("\\")) {
-							newDirectory += "\\\\";
-						}
+						String newDirectory = FileUtil.appendPathSeparator(configuration.getFullyPlayedOutputDirectory());
```
Nadahar: I made this method just for cases like this, and it takes into account different OS'es. What you made here would only work on Windows.

src/main/java/net/pms/dlna/MediaMonitor.java
```diff
@@ -206,49 +206,44 @@ public void stopped(DLNAResource res) {
...
+									LOGGER.info("Failed to move {}", playedFile.getName());
 								}
+							} catch (InterruptedException e) {
+								LOGGER.warn("Abandoning moving of {} because the thread was interrupted, probably due to UMS shutdown", e.getMessage());
+								LOGGER.trace("", e);
+								Thread.currentThread().interrupt();
 							}
-						} catch (Exception e) {
-							LOGGER.info("Failed to move {} because {}", watchedFile.getName(), e.getMessage());
 						}
```
Nadahar: I removed this catch as there was nothing that could throw and Exception that got here.

src/main/java/net/pms/dlna/MediaMonitor.java
```diff
@@ -206,49 +206,44 @@ public void stopped(DLNAResource res) {
...
 
-								if (watchedFile.renameTo(new File(newDirectory + watchedFile.getName()))) {
-									LOGGER.debug("Moved {} because it has been fully played", watchedFile.getName());
+								if (playedFile.renameTo(new File(newDirectory + playedFile.getName()))) {
+									LOGGER.debug("Moved {} because it has been fully played", playedFile.getName());
 								} else {
-									LOGGER.info("Failed to move {}", watchedFile.getName());
+									LOGGER.info("Failed to move {}", playedFile.getName());
 								}
+							} catch (InterruptedException e) {
```
Nadahar: I changed the InterruptedException behaviour as that is not an error but a normal occurrence when UMS is shut down (all threads get interrupt signals then, which means they should hurry up and finish what they're doing). Any caught interrupt should be reenabled so that parent objects also knows it's time to quit.

src/main/java/net/pms/dlna/MediaMonitor.java
```diff
@@ -206,49 +206,44 @@ public void stopped(DLNAResource res) {
...
 							if (Platform.isLinux()) {
-								FreedesktopTrash.moveToTrash(watchedFile);
+								FreedesktopTrash.moveToTrash(playedFile);
 							} else {
-								FILE_UTILS.moveToTrash(Arrays.array(watchedFile));
+								FileUtils.getInstance().moveToTrash(Arrays.array(playedFile));
 							}
 						} catch (IOException | FileUtil.InvalidFileSystemException e) {
-							LOGGER.error(String.format("Failed to send the file %s to the trash after it has been fully played. Please enable the functionality on your operating system.", watchedFile.getAbsoluteFile()));
-							LOGGER.trace("The error was", e);
```
Nadahar: As previously described, this is not only determined by the OS and the message was misleading.

src/main/resources/i18n/messages.properties
```diff
@@ -119,10 +119,10 @@ FoldTab.64=<html><strong>Default:</strong> Enabled if filename prettifying is en
 FoldTab.65=Monitor played status of files
 FoldTab.66=<html><strong>Default:</strong> Disabled<br><strong>Notes:</strong> Folders must have "Monitor played status of files" enabled to have their contents included in the New Media folder.</html>
 FoldTab.67=Do nothing
-FoldTab.68=Add an overlay to the thumbnail
-FoldTab.69=Hide the file
-FoldTab.70=Move the file to a different folder
-FoldTab.71=Move the file to the recycle/trash bin
+FoldTab.68=Tag media
```
SubJunk: The word "file" seems better to me than "media". Also "tag" has another common meaning that isn't what we mean, so I don't like that word either.

Nadahar: No, it's just suggestions. I like media better, since it's a "media" server. Tag was just an attempt to find a common expression for both overlay and added text if thumbnails are not supported. It should express both imo.

Nadahar: @SubJunk If you want the rest of the PR, I can change the strings.

SubJunk: @Nadahar "mark" seems good, although it seems to me that the text is a fallback rather than a feature on its own, since most renderers support thumbnails. It might make more sense to mention just the overlay here, and note the text fallback in the config and a tooltip.

If you think "media" is a better choice feel free to use it. I personally like "file" especially when talking about file operations like moving to folders and trash bins, but as usual I'm not insistent on it.

The other improvements in this PR look good to me.

Nadahar: @SubJunk I don't mean to "force my will" in any way. To me "media" made more sense, but that might just be for me. Maybe mediafile is even better?

Regarding option 1, I don't know which renderers shows thumbnails when you browse but from those I have used with UMS yet, only the PS3 does. In that list is my Onkyo Receiver, Foobar2000, MediaMonkey, WMP, VLC, various Android DLNA applications, a Panasonic 50" smart TV (don't remember model) and a Bluray player (Sony I think). My collection of renderers might not be representative, but I don't think it's correct to disregard it as only a fallback. It's not that I they the two are equally fancy, I just think it should be obvious to the users what the option do.

Nadahar:
@SubJunk I did a push now so you can see what I'm working with. This is by no means complete, I just commited the last commit where I am now.

src/main/resources/i18n/messages.properties
```diff
@@ -119,10 +119,10 @@ FoldTab.64=<html><strong>Default:</strong> Enabled if filename prettifying is en
...
 FoldTab.67=Do nothing
-FoldTab.68=Add an overlay to the thumbnail
-FoldTab.69=Hide the file
-FoldTab.70=Move the file to a different folder
-FoldTab.71=Move the file to the recycle/trash bin
+FoldTab.68=Mark media
+FoldTab.69=Hide video
+FoldTab.70=Move file to a different folder
+FoldTab.71=Move file to recycle/trash bin
 FoldTab.72=Fully played action:
```
Nadahar: @SubJunk I've changed the strings - are these better?

SubJunk: Now we are using 3 different words - media, video and file. Wouldn't it be better to be consistent?

Nadahar: I think we're talking about slightly different things here, so I don't think so. I thought about "file" and found out that to be exact, that's what we handle in the two latter cases (we move or recycle the files themselves). In the first case, we mark any media type inside UMS (but we don't do anything to the files), and in the second case we only hide video media. I didn't think it was necessary to spell out "video media", but if you feel that is more consistant it's fine for me..? But I think the separation between what happens inside UMS (media) and what happens on a file system level (file) is relevant here.

Nadahar:
@UniversalMediaServer/developers While testing I've come to the conclusion that thumbnails isn't generated for images when ```MediaInfo = false```. I'm trying to figure out why, and it seems to come from [this constructor] (https://github.com/UniversalMediaServer/UniversalMediaServer/blob/master/src/main/java/net/pms/dlna/DLNAMediaInfo.java#L491). 

Can anyone shed some light on why it's done this way?.I don't even understand the comment ```this class manages thumbnails by default with the parser_v1 method```. The exact lines it's attached to makes sure that thumbnails isn't generated with parser_v1, so what's the logic?

Nadahar:
Another thing on the "to do list": In d2d210c071cf540ab2afd325f4ac19fb03bf3a5b a "lazy method" ```DLNAResource.resumeName``` was made instead of calling ```DLNAResource.resumeStr(getDisplayName()``` The problem is that calling ```getDisplayName()``` results in a call to ```getDisplayName(null, true)``` where ```null``` is the renderer. This results in no renderer specific options being applied, and things like fully played actions doesn't work. There are multiple calls to this method, and the web interface use it. All calls where renderer information is available need to be rewritten.

Nadahar:
@SubJunk I'm finally done with the refactoring. I fixed a couple of bugs while I were at it, but there's multiple issues left for me to look at. 

I'm afraid this PR is getting too complex as it touches many different things, so I'm wondering if I should do the rest in another PR building on this. I haven't really changed anything, I've refactored/reorganized and fixed bugs only. It would be nice if you'd look through this and we could come to a final conclution about the function names.

Remaing issues that I'm aware of:
- [ ] Image thumbnails not working for ```MediaInfo = false```
- [ ] ```DLNAResource.resumeName()```
- [ ] RendererConfiguration Thumbnails setting should be changed to thumbnailedBrowsing
- [ ] I have an alternative suggestion for RendererConfiguration isSquare
- [ ] I think scaling should be implemented slightly different

SubJunk:
@Nadahar was refactoring and reorganising really necessary? This pull request now has 912 additions and 309 deletions, so there is nearly 3x the amount of code and now I have to go through it and relearn the code that I just wrote very recently, for very little gain in terms of what users experience.

This is open source and as such you can work on, or not work on, whatever you want, but I wonder if this kind of thing is the best use of our time when there is so much we both want to do to this program.

The bugfixes are great, and some can go into master since they don't seem in the scope of 6.0.0

Nadahar:
@SubJunk It's not the refactoring that's cost me time, its testing and tracking down all kind of issues that's often not related to 6.0.0. It's really a jungle to try to figure out all the different ways a single thing can be done many times, and where the involved code is scattered around UMS. My goal is not to add to this, but instead slowly try to structure things so that it's humanly possible to get an overview.

You don't really have to relearn the code, it's just moved around a bit. I do things like this to make things easier to understand and to get better overview, and as I see it even though it doesn't benefit the users short term it will benefit them long term in that the code is easier to maintain and fewer bugs are generated. The bugfixes have all been as a consequence of my testing. The restructuring have also been with done with the "per share" fully played action in mind, so it will be easier to implement. I'd bet that the bulk of the added lines are from the two type classes I've created (```FullyPlayedAction``` and ```MediaType```). Those didn't take me much more than 5 minutes each to create, even though they add up to many lines it's very straight forward and I just copy another class of the same type and modify it. I've also added some Javadocs here and there, but if you exclude those two things the code should actually be shorter now as I've reduced the almost identical code used 3 places to one.

But, if you just want to get it out with the remaining issues, just release 6.0.0 as it is. There's nothing that prevents us resolving the remaining issues after release.

SubJunk:
@Nadahar fair points. Well what I propose is that we merge your code into the 6.0.0 branch, and then merge 6.0.0 into master, and then the remaining things you want to address can be working off master instead of 6.0.0

The changes you listed above that have yet to be done will be great but I think 6.0.0+this is already better and more stable than master is, so that tells me it's a good time to merge.

What do you think?

Nadahar:
@SubJunk It's fine by me.

SubJunk:
Cool, do you want to resolve the conflicts in this branch and then merge it? If not I can

Nadahar:
@SubJunk Done and done.

Issue #779, opened 2016-01-03
Tooltip margin


SubJunk:
Some tooltips seem to have double margins/paddings now. For example, on the Navigation/Share Settings tab, hover over the buttons down the bottom and there's a huge padding, but it's different on the checkboxes

Nadahar:
@SubJunk I tried reverting the tooltip border change I merged in, but it makes no difference. The tooltip rendering is horribly bugged, and should be looked at.

I suspect it could have to do with changes I did in #679 related to the ```LanguageSelection``` dialog, but I find it strange since this happens without the dialog ever being launched. I'll look closer at any changes there.

The extremely hopeless thing is that all the HTML enabled Swing components share one instance of the CSS. So, if any component make any changes to the CSS formatting, it has implications for everything else. I suspect this must be what's going on here.

Nadahar:
What's apparant is that it's our "custom" tooltips that have the huge margin, where the standard tooltip is used the margin is normal, and the shadow is drawn.

I've seen this before, the bug isn't new I think, but it has been relatively rare. For some reason it seems to happen consistantly now.

Nadahar:
I've confirmed that it is the stylesheet that's changed. Replacing it with an empty stylesheet removes ALL padding... :(

SubJunk:
@Nadahar which stylesheet?

Nadahar:
The shared stylesheet between all components that use ```HTMLEditorKit```, like ```JTextPane``` and ```JEditorPane```. I have fixed the issue locally here, but I'm considering how to best implement the changes.

Nadahar:
@SubJunk Have you tested this?

SubJunk:
Fixed in #780

PR #780, opened 2016-01-03
Further tooltip tweaks (#779)

Nadahar:
What I've done: 
* Detached ```HyperLinkToolTip``` from the default Swing shared ```StyleSheet```. This means that the tooltip is no longer affected by changes to the default shared ```StyleSheet``` used in components using ```EditorKit```. This restores tooltip padding to normal.
* Changed ```ToolTipUI``` back to the inherited to preserve UI style. The ```ToolTipUI``` was overridden in this class probably because the default ```ToolTipUI``` wouldn't render the custom content. Overriding ```getPreferredSize()``` fixes the rendering, and thus using the inherited ```ToolTipUI``` better preserves LAF (like drawing a drop-shadow).

What I haven't done is address to two remaining issues I'm aware of regarding tooltips:
* ```ToolTipUI``` isn't properly rendering ```JComboBox```. This seems to be a Swing bug, but still looks ugly.
* Solving "clickability" on tooltip links. This should be done by either having a delay before the tooltip close (only if there's a link in it preferrably), or by positioning the tooltip so that it overlaps the component and it's possible to move the mouse there without it closing.

This should solve #779.

### DIFF
--- a/src/main/java/net/pms/newgui/LooksFrame.java
+++ b/src/main/java/net/pms/newgui/LooksFrame.java
@@ -351,7 +351,7 @@ public class LooksFrame extends JFrame implements IFrame, Observer {
 
 		// Customize the colors used in tooltips
 		UIManager.put("ToolTip.background", new ColorUIResource(PMS.getConfiguration().getToolTipBackgroundColor()));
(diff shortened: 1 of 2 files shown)


SubJunk:
Just tested this, I'm happy for it to be merged

PR #785, opened 2016-01-05
Updated headless handling and Transcode tab comboboxes

Nadahar:
These changes are small and shouldn't require much testing. I've tested it without issues. This is a combined PR consisting of two unrelated changes.

There has been an inconsistancy in headless handling. The way it has been, ```isHeadless()``` only tested for the ability to run a graphical environment while system properties ```CONSOLE``` and ```NOCONSOLE``` was used to decide if headless mode was chosen or forbidden. From what I could see, only the code in ```main()``` respected these rules, and calls other places in the system simply referred to ```isHeadless()``` - which would constitute a bug if UMS was run in an environment capable of running a GUI but intentionally was running without.This might partly be my fault due to earlier headless changes, but in any case it's confusing and bug prone.

What I've done is made ```isHeadless()``` take both factors into consideration by creating a new method ```forceHeadless()``` that will set headless mode regardless of system capability, and calling that if startup arguments dictate it. It should remove several bugs if running headless in a graphical environment.

The other change is the comboboxes on the transcoding tab. They were made in a "hackish" way where contents were parsed and anything after ```/*``` where considered to be a comment and that before it the actual value to be used. It didn't allow translation for all options, and the options looked somewhat cryptic. I've replaced those with ```KeyedComboBoxModel``` which allows for the key (the value read and written to configuration) and the value (the option shown) to be independent. This means that the options can now be fully translated, be less cryptic, and the comboboxes will be initialized at their current selection. 

There is one drawback with this change that can be mended, but I didn't want to go through a lot of work to do that if it isn't actually needed. When selecting codepage for non-unicode subtitles, the combobox used to be editable so that users could enter codepage values not found in the list. That isn't compatible with ```KeyedComboBoxModel``` since the key and value are different - and thus the user would only specify a value, not a key, when editing. I can make it work if needed, but I figure that most "rare" character sets use unicode these days and we have a relatively broad selection in the existing list. I'd say we just see if a complaint ever come about this - I doubt it, and if it does it's probably better to just add that or those codepages to the list. Codepages not in the list can still be specified in the configuration file, just not configured in the GUI.

My intention was to get this into the upcoming release, what do you think?

### DIFF
--- a/src/main/java/net/pms/PMS.java
+++ b/src/main/java/net/pms/PMS.java
@@ -543,10 +543,10 @@ public class PMS {
 
 		registry = createSystemUtils();
 
-		if (System.getProperty(CONSOLE) == null) {
+		if (!isHeadless()) {
 			frame = new LooksFrame(autoUpdater, configuration);
 		} else {
-			LOGGER.info("GUI environment not available");
+			LOGGER.info("Graphics environment not available or headless mode is forced");
 			LOGGER.info("Switching to console mode");
 			frame = new DummyFrame();
 		}
@@ -708,9 +708,9 @@ public class PMS {
 		// Any plugin-defined players are now registered, create the gui view.
 		frame.addEngines();
 
-		// To make the cred stuff work cross plugins
-		// read cred file AFTER plugins are started
-		if (System.getProperty(CONSOLE) == null) {
+		// To make the credentials stuff work cross plugins read credentials
+		// file AFTER plugins are started
+		if (!isHeadless()) {
 			// but only if we got a GUI of course
 			((LooksFrame)frame).getPt().init();
 		}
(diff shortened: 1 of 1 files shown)


SubJunk:
@Nadahar it looks good, but I'm not sure about removing the ability to input custom transcoding qualities. There are a few valid scenarios I can think of for doing that

Nadahar:
I wasn't aware that I did that. I thought I only limited custom codepages - and that the others were already limited. If that's the case, I agree. I have to dive in and make a logic for adding custom choices to the ```KeyedComboBoxModel``` then.

SubJunk:
@Nadahar if I enter a custom value, it's removed when I go out of the box. Do you want to quickly merge the headless changes and leave the other things for the next release? If you're around now I can wait 10 mins before I tag. I don't have much time so I can't wait longer than that

Nadahar:
@SubJunk I'll split it into two PR's, hang on

Nadahar:
Done, merging.

SubJunk:
Nice one

valib:
@SubJunk not related to this PR but there is a new version 0.7.81 of the MediaInfo which could be updated to UMS.

SubJunk:
@valib cool, it looks good. I'll update it after the release

PR #786, opened 2016-01-05
Updated comboboxes in TranscodingTab

Nadahar:
This has been separated out from #785 since there further work needs to be done on this change. I'm pasting the text regarding this change from #785 below:

I've changed the comboboxes on the transcoding tab. They were made in a "hackish" way where contents were parsed and anything after ```/*``` were considered to be a comment. The part before ```/*``` were used as the actual value It didn't allow translation for all options, and the options looked somewhat cryptic.

I've replaced those with ```KeyedComboBoxModel``` which allows for the key (the value read and written to configuration) and the value (the option shown) to be independent. This means that the options can now be fully translated, be less cryptic, and the comboboxes will be initialized at their current selection. 

There is one drawback with this change that can be mended, but I didn't want to go through a lot of work to do that if it isn't actually needed. When selecting codepage for non-unicode subtitles, the combobox used to be editable so that users could enter codepage values not found in the list. That isn't compatible with ```KeyedComboBoxModel``` since the key and value are different - and thus the user would only specify a value, not a key, when editing. I can make it work if needed, but I figure that most "rare" character sets use unicode these days and we have a relatively broad selection in the existing list. I'd say we just see if a complaint ever come about this - I doubt it, and if it does it's probably better to just add that or those codepages to the list. Codepages not in the list can still be specified in the configuration file, just not configured in the GUI.

Remaining tasks:
- [x] Regain the ability to specify custom values for some of the comboboxes.


Nadahar:
I've created a new subclass of ```KeyedComboBoxModel``` called ```KeyedStringComboBoxModel```. It is not generic as it's anchestor and is constrained to only using strings, but I found it to be the only practical solution since there's no "clean" way to find the type of a generic object. Object types must be known because this class needs to convert key to value and vice versa.

@SubJunk Does this work better?

Nadahar:
@SubJunk Could you check if you think this way works?

SubJunk:
Now it says `Custom value "14"` if I enter `14`, so if that's how it's supposed to work then it seems okay. It also stores the value correctly in the config file.

Nadahar:
@SubJunk That was how I meant for it to work yes, the question is if you think it's understandable/a good way.

SubJunk:
It seems fine to me, if you prefer it. We don't lose any existing functionality

Nadahar:
@SubJunk I've made a generic combobox model that can be applied to any combobox that uses strings that does this. That's the reason for the generic "Custom value". I take it you don't like it, do you think it's preferrable without the added text? Only showing the value like "cp738"?

SubJunk:
@Nadahar I really don't mind, either way is fine with me. Does anyone else have any opinion @UniversalMediaServer/developers ?

valib:
@Nadahar I prefere not to use "Custom value" and leave it empty. For general use the implementation of  the user setting is OK but especially for the Codepage to allow users to set any value doesn't make sense because only supported CodePages are relevant. The wrong value will make an error in transcoder. I think that ```setEditable``` should be set false

Nadahar:
@valib So you think it's better to just show a "cryptic" code when a custom value is configured?

Regarding the codepages, as it is today any value can be entered - valid or not. I've kept that behaviour. If we turn off editable (like it was earlier in this PR), we'll exclude all codepages not in the list. Is that really a good idea?

valib:
@Nadahar I mentioned it in the EDITs of my comment.

Nadahar:
Ok, I thought that was a feature not a bug. There are other codepages than those listed - are you sure that's not the reason it is editable? What if a user want to use such a codepage?

I'm also wondering why a blank option is needed. What does blank mean? Is that some kind of "auto detect"? If so, we should make an endtry with a blank key and "Auto detect" as a value. Also, isn't this only used if the subtitle isn't unicoder? From that perspective, it's a bit odd that "UTF-8" is in the list.

valib:
@Nadahar the "Auto detect" is a wery good idea. But it should be set as a key with blank value if possible. Otherwise the logic should be change too. Now the blank value is used to determine the auto detection. The auto detection recognises all codepages including the "UTF-8" but I agree that  "UTF-8" should be removed from the list because our goal is to convert subs from unicode to  "UTF-8" and force  that as default will produce an error from e.g. FFmpeg when subs are in unicode.

Nadahar:
@valib Blank key should be no problem, when you say blank you mean ```""```, not ```null``` right? What about other codepages? Do we simply ignore those? The most common ones are in our list, and if some needed codepage is missing, we could add it (given we're told about it). Also, it would still be possible to enter any value in ```UMS.conf```, just no to select it from the GUI...

valib:
@Nadahar the ```isNotBlank(configuration.getSubtitlesCodepage()) ``` is used to check if the codepage is set or not so  it could be both  ```""``` or  ```null```. The reason why is sometimes necessary to set codepage manually is because the automatic detection is not 100% accurate. For example before the automatic detection was improved I needed to set ISO-8859-2 for the czech language because subs were false detected as cp1251 and wrong characters were shown. The supported codepages for automatic detection are listed here http://userguide.icu-project.org/conversion/detection#TOC-Detected-Encodings so we should consider to update our list. Unfortunately I failed to find whose codepages are supported in FFmpeg.

valib:
@Nadahar to be more precise. The key could be ```Auto detect``` not blank but the corresponding value stored in ```configuration.getSubtitlesCodepage()```   should be ```""```

Nadahar:
@valib I've added "Auto detect" with "" as key, changed to non editable and added some more character sets. As long as there's no information as to which character sets are available (from FFMpeg), all we can do is guess. 

I think it's likely that we've covered almost everything that's actually used, and I also think it's likely that languages with non-latin characters use unicode a lot today, since non-unicode character sets must have been a real pain for them.

What do you think, does this look ok?

valib:
@Nadahar great job. Tested and it could be merged.

PR #790, opened 2016-01-07
Decrease the delay before the video starts playing

valib:
@UniversalMediaServer/developers can you test it if it is worth to merge?

[Review on Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/790)

### DIFF
--- a/src/main/java/net/pms/configuration/WebRender.java
+++ b/src/main/java/net/pms/configuration/WebRender.java
@@ -115,6 +115,7 @@ public class WebRender extends DeviceConfiguration implements RendererConfigurat
 		configuration.addProperty(SUPPORTED, "f:gif m:image/gif");
 		configuration.addProperty(SUPPORTED, "f:tiff m:image/tiff");
 		configuration.addProperty(TRANSCODE_AUDIO, MP3);
+		configuration.addProperty(TRANSCODE_FAST_START, true);
 		return true;
 	}
 


Nadahar:
@valib What exactly does it do? How do I test it? I see that it sets faststart, but for what under what circumstances and what is the advantage/disadvantage of doing this?

valib:
@Nadahar there is a standard delay before the video starts to stream to to the renderer. This PR changed this delay to 0 for the web interface and the video should start faster. For me it starts faster but at the beginning the video is a little bit choppy. So I would like others to test it. You can put on your PC running UMS to the web browser (Firefox or Chrome) the address 172.0.0.1:9001 to accsess the UMS web interface and test it.

Nadahar:
@valib In know how to use the web interface, but I think the address you intended was ```127.0.0.1:9001```? It's the same as ```localhost:9001```.

What I meant was enable faststart (which effectively drops the startup buffer) for web players? Is there a special reason why web play doesn't need buffering?

As I see it, the startup buffering logic should be changed in general. How much buffering is needed before playback is started is probably dependent on the "power" of the computer running UMS, not the renderer. It should be possible to adjust this instead of just turning it on or off. Some people might have to live with a long buffering pause due to weak hardware, while others can get away with a much shorter delay. Some buffering however, even if it's just 100ms or something like that, is probably a good idea in any case. But, the file being transcoded probably also matters, the heavier it is to transcode, the more buffering time is needed.

valib:
Seems to me to be unacceptable so I close it.

PR #793, opened 2016-01-10
Change the DLNAList.clear() logic

valib:


skeptical:
@valib this would break our id scheme. The reason for subclassing ```DLNAList``` was so that add/remove resulted in corresponding changes to ```GlobalIdRepo```, so ```DLNAList.clear()``` should remove only its own children from ```GlobalIdRepo```, not delete all ids globally.

valib:
@skeptical ahh I see. Thanks for the explanation.

skeptical:
@valib there is a flaw in the simplistic 1:1 correspondence logic of ```DLNAList```` and ```GlobalIdRepo``` though, which pops up if we try to move a child, e.g.
```java
folder2.add(folder1.remove(dlna))
```
after the move the resource still exists but its id is gone.

PR #800, opened 2016-01-17
Improve prettifying

SubJunk:
This branch is mostly to use IMDB information more in filename prettifying.

TODO:
- [x] Check that the title from IMDB is not very different to the one in the filename, since sometimes incorrect results are returned from the search
- [x] Use movie titles from IMDB
- [x] Fix commas in strings breaking the variables
- [x] Fix variables accessed in wrong order

[Review on Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/800)


Nadahar:
I'd consider [omdbapi.com] (http://www.omdbapi.com/) instead or in addition. IMDB is doing what it can to hinder scrapers by changing things and generally being difficult. OMDB has an actual API and no scraping is needed. 

[themovedb.org] (https://www.themoviedb.org/documentation/api) is another possibility.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/800#-:-K8G41G-gITuTF3KCQrW:2103417670)*

SubJunk:
@Nadahar we use the opensubtitles API which seems very similar to the OMDB one

Nadahar:
@SubJunk Do you mean that the content of the database or the API interface itself is similar? The important thing is the quality of the database and the search. I'm just saying that relying on IMDB maybe isn't the smartest since it's against their TOS (you have to pay to use their data) and they will try to sabotage any "freeloaders" as best they can. That means that any scraper is likely to break in the future requiring constant rewrites.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/800#-:-K8G8p7lXEUY-cy0KMl0:-1049882473)*

SubJunk:
@Nadahar we have been using opensubtitles since this feature was introduced 3 years ago with no problems. It seems like the content and API are similar, OMDB even uses IMDB IDs so it's not entirely separate from IMDB

Nadahar:
@SubJunk I think we're talking slightly passed each other. IMDB ID has been a kind of defacto standard ID for movie content, so most such databases I've looked at has IMDB ID. That does not mean that the database is that of IMDB. But, if I understand it correctly, when you talk about IMDB you're really talking about opensubtitles? If so, we don't have the scraping problem as that's an open API.

I have no idea which database is "best" of opensubtitles and OMDB, if opensubtitles generally seem to have correct and sufficient information I guess that's fine. What I was warning against was using IMDB.

SubJunk:
Ah cool :)

ExSport:
Is it possible to configure scraper (IMDb in this case) with language option?
Like in KODI and other addons where scraper returns localized posters (CS/DE/EN/FR/...) and movie names including plot where available.
The best way to configure it (also) separately as I am using GUI in EN but it will be good to use scraper in CS :+1: 
Thx!

Nadahar:
> The best way to configure it (also) separately as I am using GUI in EN but it will be good to use scraper in CS :+1:

The rest of what you see on the renderer is that of your UMS language, so I think it would be natural if that too was in the same language. Why don't you run UMS in CS?

ExSport:
I don't understand to translated technical words in Czech😉
Any program I saw in Czech I always found that I don't understand to the settings words what they mean. 
Czech translation is e.g. 3x longer than in English so when translator have to shrink descriptive text I never understood what is going on:) In English I know for what the settings was intended but in other language it can be misleading😇

Nadahar:
@ExSport There's a difference between translations and BAD translations. @valib is translating to Czech, so I'm sure that language is pretty much spotless in UMS. If you find any faults, you can submit corrections at [crowdin.com] (https://crowdin.com/project/universalmediaserver) :wink:

ExSport:
An example: I am an admin of Windows technologies like Windows servers. I understand anything there (settings, logs, error messages...) but only in English.
When I must work with Czech edition it is nightmare. Professional translation but anyway I don't understand to many. Many tech words in my opinion should not be translated at all and other ones I don't understand what they mean or where this settings to find but in English no problem. 
Sometimes for tech guy there is no option in non-English wording as translation will loose simplicity and original meaning.
Nothing against Valib😎
E.g. equivalent for Transcode folder in Czech? Can't imagine the understandable short word right now😈

Nadahar:
@ExSport ```Hide the "Transcode" folder``` is translated to ```Skrýt složku "Transkódování"```. It doesn't seem much more complicated than the English version to me, even though I don't read Czech. 

There are lots of bad translations in Windows, I've worked a lot with different Windows server products as well, and always used English not Norwegian. But, the main reason for this isn't that I don't understand the Norwegian interface, but that you also get Norwegian errors and event log entries. Those are hopeless to search for on the net, so I prefer using English.

UMS on the other hand has the log in English nomatter what interface language you use, so that's not a problem.

ExSport:
It is an English word converted into the shape used in Czech☺
Anyway  for whatever reason, Czech is complicated for transcoding, MEncoder,FFmpeg, audio, video,mediaserver,... subjects. At least for me😎

valib:
@Nadahar agree with @ExSport that the translation of some technical phrases is almost impossible and could not be translated. But in that case they could be understandable only for us a "technical" people. :wink: 
@ExSport I would appreciate if you suggest some better wordings at the [crowdin.com] (https://crowdin.com/project/universalmediaserver).

Nadahar:
@valib @ExSport The problem isn't foreign to me in any way, I'm translating to Norwegian remember?

I did find a few difficult cases when I translated, but most things are solvable if you just think about it. Things like ```transcode``` can actually just be translated to ```konvertovat``` if a similar word is missing. But, constructing a word is also possible. In Norwegian we use the prefix ```trans``` in the same way as in English, and ```code``` is ```kode```, so it's easy to construct ```transkode```.

Other things are names (like FFmpeg, MEncoder) that doesn't need translating. A few things can be impossible to find a reasonable translation for, and also be commonly known as the English term, and in these cases I keep the English word instead of contructing something nobody would understand.

If there's a will there's a way. The problem with much translated software is that the translators seem to be people with language, not computer, experience. They often have a tendency to want to construct a native work regardless of if it will make sense to the average computer user. But, we are free to choose a different path.

ExSport:
We get little bit OT. My initial post wasn't about translation but possibility to make prettifying language-wide.
It means if GUI is set to Norwegian, UMS will checks original filename and returns localized one.
Or better when this can be also overridden in conf file so English GUI but Czech names. 
I expect it is possible as I saw few IMDb scrapers which gathered localized info including posters.
And now make it configurable :))))

valib:
@Nadahar this discussion leads me to to the conclusion that also the English wording should be corrected too before it is translated. For example the TRANSCODE folder name seems not to be correct. This folder not only offers the possibility to transcode videos but also offers to stream the video which is definitely not a transcoding. I personally dont't know what the name of it would be bether but maybe @SubJunk  can help us.

Nadahar:
@valib I agree, it's very important that we get any English strings correct before they are send to the translators, which is why I'm trying to keep an eye on new strings being commited. But, when it comes to existing strings I'm sure many things could be improved. Translation is like most things, shit in means shit out.

valib:
@Nadahar in case of the TRANSCODE folder I see two possibilities. The forlder name will be changed or we should rename files to be streamed to "Not transcoded" instead

valib:
Sorry guys I tried to make some comment but I realised that it is not relevant and when I tried to clear it I push the wrong button.

SubJunk:
Yeah that's a good point about the word "transcode", I had the same discussion years ago on the PMS project. I don't know of a word that works better.

I've also been wondering if we can use a better convention than `#--WORD--#`, because that looks ugly. Like maybe `UMS: Word` or `:Word` or even just `Word` but make sure they appear above the other folders. I'm not sure what I prefer, does anyone have ideas?

SubJunk:
@UniversalMediaServer/developers I'd like to merge this before the 6.0.0 release, since in addition to the improvements to prettifying it also fixes bugs currently affecting people. Any objections?

Nadahar:
I agree that ```#--word--#``` is very ugly indeed. Making them appear on top is no problem for renderers that doesn't sort the content, but some gotta be smart an sort themselves. I don't know what would be the best replacement though, they should stand out in some way.

---

Reviewed 5 of 5 files at r2, 1 of 1 files at r3, 3 of 3 files at r4.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[pom.xml, line 392 \[r4\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/800#-K92Gjr-yd_t8TyT08Wy:-K92Gjr-yd_t8TyT08Wz:-110742484) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/70e17d4fe3d111fee2377d0fdfc85b25e94ea98f/pom.xml#L392)):*
I don't think there's any need to add a lib only to get around the comma issue. You could simply do a "replaceall" on the way into and out of storage. We already use ```&comma;``` for this other places, so that would be the most logical substitute for comma.

---

*[src/main/java/net/pms/util/FileDb.java, line 77 \[r4\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/800#-K92I35TKdnGF3DhRXlf:-K92I35TKdnGF3DhRXlg:-1723946955) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/503d73e32b8a93ffd7a0582e4ed4507b41dd5889/src/main/java/net/pms/util/FileDb.java#L77)):*
```java
for (String s : tmp) {
    s = s.replaceAll("&comma;", ",")
}
```

---

*[src/main/java/net/pms/util/FileDb.java, line 128 \[r4\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/800#-K92I__LSjrghgc27auL:-K92I__LSjrghgc27auM:-1723458937) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/70e17d4fe3d111fee2377d0fdfc85b25e94ea98f/src/main/java/net/pms/util/FileDb.java#L128)):*
```java
for (String s : data1) {
     s = s.replaceAll(",", "&comma;")
}
```

---

*[src/main/java/net/pms/util/FileUtil.java, line 565 \[r4\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/800#-K92JIHPcVqd_Gy1JATd:-K92JIHQqkdNawKogpcF:2124967703) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/70e17d4fe3d111fee2377d0fdfc85b25e94ea98f/src/main/java/net/pms/util/FileUtil.java#L565)):*
Wasn't it opensubtitles, no IMDB?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/800#-:-K92KJ09vPuuC2UFPZI6:699730889)*

SubJunk:
Review status: 3 of 5 files reviewed at latest revision, 4 unresolved discussions.

---

*[pom.xml, line 392 \[r4\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/800#-K92Gjr-yd_t8TyT08Wy:-K9v7DYv6ZsckwnxuJCX:-718090005) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/70e17d4fe3d111fee2377d0fdfc85b25e94ea98f/pom.xml#L392)):*
I disagree. I like using libs. I've been influenced by my current workplace in that way I guess. Not a big deal either way.

---

*[src/main/java/net/pms/util/FileUtil.java, line 565 \[r4\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/800#-K92JIHPcVqd_Gy1JATd:-K9v7c_wbAWxLpDCR4_u:-1069899034) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/70e17d4fe3d111fee2377d0fdfc85b25e94ea98f/src/main/java/net/pms/util/FileUtil.java#L565)):*
IMDb via OpenSubtitles

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/800)*

valib:
Reviewed 1 of 5 files at r2, 2 of 3 files at r4, 2 of 2 files at r5.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/800)*

Nadahar:
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[pom.xml, line 392 \[r4\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/800#-K92Gjr-yd_t8TyT08Wy:-K9xlswfQsQDR9wsbNnZ:-136045748) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/70e17d4fe3d111fee2377d0fdfc85b25e94ea98f/pom.xml#L392)):*
I'm not against libs in general, but including a whole lib for doing a small and easily performed task. It adds build time, it adds to the file size, it opens the possibility for bugs out of our control (from the external library) that can suddenly appear, it is one more library maintain the version for, and in this case I'm pretty sure it will be many times slower (even though it very well be negligible) since a lot of superfluous parsing it probably performed.

---

*[src/main/java/net/pms/util/FileUtil.java, line 565 \[r4\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/800#-K92JIHPcVqd_Gy1JATd:-K9vQw-QmWdKhXM75i79:-738982990) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/70e17d4fe3d111fee2377d0fdfc85b25e94ea98f/src/main/java/net/pms/util/FileUtil.java#L565)):*
How can that be? To you mean that OpenSubtitles have a scraper in place doing real-time scraping of IMDb? IMDb only allows licenced access to it's data as far as I know.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/800)*

SubJunk:
I think this is good to merge now. If you have a different way you'd like to format it @Nadahar feel free to change it, as always. I don't think there's any reason to hold up this merge though, especially considering it fixes bugs in master

Nadahar:
I decided to rid it of the CSV dependency, and I't glad I did. I hadn't looked closely at the class, but it is made for using any separator, and simply defaults to comma. The CSV methods completely broke this. I've created ```encodedSeparator``` - the value the separator should be replaced with if it is in any of the strings. This defaults to ```&comma;```. The replacement will take place both ways, when reading and writing.

I've also fixed or improved some other small things while I were at it.

I have *not* tested this as I'm not sure how it's used. I ended up rewriting much more than I intended, so basic testing is needed.

---

Reviewed 3 of 3 files at r6.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/800#-:-KA8v9HJqQ_vGtdrNprF:-1185226048)*

SubJunk:
I'm very not in favor of taking a PR that's about to be merged and doing reformatting that puts it back into a testing phase. It's a waste of time and I have very little time to start with.

Nadahar:
@SubJunk Then I don't know what you meant by this:
> If you have a different way you'd like to format it @Nadahar feel free to change it

It's easy to remove the last commit if you want to merge with the bugs.

SubJunk:
@Nadahar I take back my previous comment, your improvements are welcome and are worth putting it back into the testing phase. I'm frustrated to have already spent the time testing this PR before that commit, but in the end it is better to add it now and retest. Thanks

Nadahar:
@SubJunk There's no need to retest everything, only the storing in the filedb

valib:
Reviewed 2 of 3 files at r6.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/800)*

SubJunk:
@Nadahar it no longer supports commas after the last commit - it's just storing them as regular commas

Nadahar:
@SubJunk What settings and actions do I need to test this? Is enabling prettifying and browsing some movies enough to populate filedb?

SubJunk:
Prettifying and IMDb lookup, yes

Nadahar:
@SubJunk I've fixed the bug (the replacement wasn't applied to the array), and a couple of other small things.

Testing this lead me to look more into this though, and there's several "major" bugs here. First of all, it fails to look up around 70% of my sample files - I haven't checked this enough to claim it, but from the random samples I traced through only those found by hash were actually found. ```formattedName``` was often blank when sent to the lookup routine, which seemed to be a common cause for failure.

The whole "file database" concept is also very strange to me. It's no database at all, it's a flat text file that is rewritten entirely for every change that is stored. It will perform horribly for large "databases". It's also full of flaws, I didn't dig in to find out why, but a lot of the lookups I saw was being done when tracing was never written to file. There is no thread safety here at all (and the lookups are done from different threads), so I'm assuming that a lot of the information keep getting overwritten by other threads. I've grown tired of constantly stumbling over huge issues that requires a lot of rewrite/redesign that noone else than myself see as a problem, so I'm just leaving it like it is.

---

Review status: 3 of 5 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/main/java/net/pms/dlna/DLNAMediaDatabase.java, line 166 \[r7\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/800#-KADYgQcAUKXWD23s_H2:-KADYgQcAUKXWD23s_H3:68062053) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/a9a1b9dfc4fd5874175f0dc5bd8d154f1df97fa9/src/main/java/net/pms/dlna/DLNAMediaDatabase.java#L166)):*
This lead to a NPE if default renderer was null (as null was returned from getRootFolder)

---

*[src/main/java/net/pms/util/FileDb.java, line 93 \[r6\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/800#-KADZD8gM_MpowjROhA2:-KADZD8gM_MpowjROhA3:-414938977) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/7598cd6fcd6895ef105f5ab91e84d022322ca550/src/main/java/net/pms/util/FileDb.java#L93)):*
This was the reason it didn't work - stupid of me, but results weren't stored in the array when looped that way, the change was only applied to the loop local ```element```string.

---

*[src/main/java/net/pms/util/FileDb.java, line 156 \[r6\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/800#-KADZHeum2btka9AHQ8h:-KADZHevd6q-Otpsg_ZI:-1559709536) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/7598cd6fcd6895ef105f5ab91e84d022322ca550/src/main/java/net/pms/util/FileDb.java#L156)):*
The same issue applied here

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/800#-:-KADZQfH7DbTyciWjUbA:899002469)*

SubJunk:
@Nadahar I agree it would've ideally been written differently. I think it would've been better to use our h2database implementation for it. However it was developed as it is and I find it very useful as a user, so I'm grateful to the developer who made the feature. It can and will be improved, especially as we start to rely on it more.
I get about 90% success with my files and the ones that didn't match are rare, often they don't exist on IMDb so it's not us failing, so it seems there's an issue you're running into that I'm not.

SubJunk:
@Nadahar I've confirmed that the fix worked, are we good to merge now?

SubJunk:
Remember that just because something merges doesn't mean development stops. We are free to have as many branches for the same feature as we want, with merges along the way when we determine the branches offer enough of an improvement vs. master at the time. We can immediately start working on an ImprovePrettifyingV2 branch if we want to.

Nadahar:
@SubJunk Good to merge as far as I'm concerned

PR #802, opened 2016-01-18
Revive cover art

Nadahar:
I was planning to make some more progress until I created the PR, but it's not easy to elude @valib's hawk eye. This isn't finished by a long shot. 

I ended up completely replacing the existing CoverUtil since none of the code in it actually work anymore. This is due to changes at Amazon and DiscoDogs, where authentication (and a lot of handling with tokens and whatnot) is needed today. Amazon only allows use of their covers if you are selling their music, and DiscoDogs requires an "account" to allow access and from the little I read there were lots of problems with their current authentication scheme. As I see it neither of those services is particularly interesting anymore, so I went for Cover Art Archive instead.

Tasks:
- [x] Create a new class structure with an abstract parent class and subclasses for actual implementation
- [x] Create a type class for CoverSupplier
- [x] Create Cover Art Archive subclass
- [x] Create database caching for MBID by tag
- [x] Create database caching for cover by MBID
- [x] Refine search logic to get more hits

[Review on Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/802)


Nadahar:
@valib I'm not done, which is why I hadn't made the PR yet. The fetching of audio covers from CoverArtArchive is working, but it takes some time (and bandwidth) to get them so I've started making a caching solution for them. 

I'm planning to use the database already in UMS (the h2 database), but not surprisingly the database handling is somewhat of a mess. I'm trying to figure out how to use what's there, even though I must say it's tempting to just write it from scratch. Thread safety seems to be foreign in the exists code, and the organization is strange to put it mildly.

Since looking up covers on Cover Art Archive requires two lookups (unless the tag contains MBId in which case the first lookup isn't done) I'm planning to cache both separatly. I know that thumbnails themselves supposedly are cached (though this seems partly broken to me), but that's only if the cache is enabled and it's on a per-track basis. A cover is typically per album or single, so a lot of tracks will have the same cover. When caching this, it means that when one track has been cached for a given album, no more lookup is needed for that album, which will hopefully speed things up a bit. These caches will also be unaffected by clearing of the media library.
To retrieve a cover I first have to find the musicbrainz id for the given release (single or album), and then I need to query CoverArtArchive for the cover for the given MBId. Both lookups take time, so I'm planning on caching both.

valib:
@Nadahar nice :-) I know that is not finished yet. It was only a remark that I am surprised how bad is the current implementation.

Nadahar:
This still isn't finished, but it's getting closer. All the "heavy lifting" creating and maintaining database tables is done, and I think I've figured out how to work with Java connections, statements and resultsets. 

Database caching is in place to the tag search, to any MBID that needs looking up is first checked in the cache before going on the internet. Implementing the same for the binary covers should be relatively straight forward now.

---

Review status: 0 of 16 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/802#-:-K8_KNrsCZ6PddvgJJlp:-199445483)*

Nadahar:
Oops, it seems I forgot to commit the classes handling the database.

---

Review status: 0 of 16 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/802#-:-K8_LuEEPIFwWlInY_WO:-1599257732)*

SubJunk:
Nice work!

Nadahar:
@UniversalMediaServer/developers This PR has gotten very large for doing such a small thing. The reason is that I chose to create a new "framwork" for working with the database. The old and this new can coexist, but my goal is that all the tables will be moved to this in the future (given that you guys approve the way I've organized it).

The main differences between the "old" and the "new" is:
* This keeps tabs on versions on table level instead of database level. If one table is modified, only that table need to be upgraded - the other tables can be used as is. The "old" logic simply drops all tables (and thereby all information in them) if the database version doesn't match. I consider that a waste for people upgrading.
* This actually upgrades tables (will, not yet since the only table in it is in version 1, but the logic is made for doing it) instead of dumping them and recreating. If changes to a table is made in such a way that an ```ALTER TABLE``` is impossible or would currupt data, the upgrade code can still choose do dump and recreate the table, but as a last resort instead of by default. Most table changes are usually adding fields, and by using a sane default the upgrade code can easily handle that without deleting existing data.
* This takes multithreading into account, trying to avoid duplication of data and partial overwrites while allowing concurrent read access.
* This gathers all operations for a given table in one class, so that it's reasonably easy to get an overview of that table for understanding operations or making changes.

valib:
@Nadahar when testing I have got a ```org.h2.jdbc.JdbcSQLException: Table "MUSIC_BRAINZ_RELEASES" not found;``` and when I checked the code the method ```checkTable()``` is not used anywhere so the table is not created.

Nadahar:
@valib It's supposed to be called from line 69 in ```Tables.java```:

```java
TableMusicBrainzReleases.checkTable(connection);
```

---

Review status: 0 of 19 files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/802#-:-K8a3BXrPEbdZNU4_PNt:721835226)*

Nadahar:
Review status: 0 of 19 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/main/java/net/pms/database/Tables.java, line 69 \[r7\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/802#-K8a3n5GrKRC-KmAtQXj:-K8a3n5GrKRC-KmAtQXk:-411964460) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/56b943fbf001a423e29bb71b7b337c6ac430da5c/src/main/java/net/pms/database/Tables.java#L69)):*
@valib Here is the call

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/802)*

valib:
@Nadahar thanks for explanation but the ```checkTables()``` is not called either.

Nadahar:
@valib I forgot that, I had some temporary test code that did that which has later been removed. I made a new commit fixing that. Thanks

---

Review status: 0 of 19 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/802#-:-K8a6XiZmiYLGlFRNKSs:-1217292463)*

Nadahar:
Caching of covers is now done and it seems to be working. I've discovered that the search for MBID is too narrow though, the tag information must match exactly to find a MBID. I'll have to see if I can do something about that, because as it is now it doesn't find very many of my sample files.

Nadahar:
I've expanded the search quite a lot, and it now find most of my sample music that is tagged. It's all down to the quality of the tag though, nothing to do about that. I've seen a few false positives when the tag only contain title, but there's not much I can do about that - it's all down to guessing then.

Excluding yet not discovered bugs, this PR is ready to merge. The search has gotten a bit complex, and I suspect I could spend an endless amount of time refining it, so I think we'd rather take on cases of misidentification as they arise (IF there's sufficient information available to make it better).

Most of this code is outside existing code, so I don't think potential bugs in this should have too serious consequences. The only thing that worries me is the changes to the database initialization, but I haven't seen any issues with that during my testing.

SubJunk:
Cool. Do you want it included in 6.0.0?

Nadahar:
@SubJunk Sure, but it might need a little bit of testing?

Nadahar:
Btw, I've discovered a "serious" bug in the fullyplayed/thumbnail/rescale logic. Thumbnails are cached this way, so when viewed from another renderer (with different thumbnail ratio) they look completely weird. I also suspect that the overlay sometimes is applied to the same thumbnail several times, which would explain why they sometimes are extremely dark.

I'm working on it as we speek, but it won't be a 5 minute fix.

SubJunk:
@Nadahar yeah I have commented on that before too, it's on a list of my things to fix. Happy for you to do it.
My fix was going to be to use different databases per renderer, which would have other benefits like allowing us to cache file display names.

Nadahar:
@SubJunk I've been thinking of a per-renderer cache too, but that's a bigger job. My current idea is to measure the time it takes to do scaling/overlay, and if it's negligible simply to do it on the fly - on a per browse request. There are other problems with caching it too, like if the user turns the function on and off, deletes UMS.mon etc.

Nadahar:
@SubJunk If we're going to release things "as is", we could include this too as it should be "low impact" as well. The only thing that has any potential to negatively affect anything is the changes to the database initialization - but it seems to be solid from all the testing I've done on this. 

The rest of this PR either replace already broken code (```CoverUtil```) or is completely new and unrelated to existing functionality.

SubJunk:
Cool, I'm happy to merge it

---

Reviewed 7 of 7 files at r1.
Review status: 4 of 23 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/802#-:-K93_4X3-OXQILoau5rk:-1114197115)*

Nadahar:
@UniversalMediaServer/developers Anyone else have any objections/views?

Nadahar:
No unresolved issues at Reviewable and no further comments - merging.

PR #805, opened 2016-01-23
Updates to RemoteWeb

Nadahar:
I wanted to find a better solution to the NPE avoided in ec6131aca7152ae69b7fbc0e4e66ff4860df7747. I ended up changing quite a lot of the ```Exception``` handling and fixing all FindBugs bugs and thread unsafe access to a ```HashMap```:

* Reorganized ```Exception``` handling
- Fixed synchronization for ```roots``` map
- Made ```ResourceManager``` creation run inside ```doPrivileged()```
- Made sure ```UMS.jks``` is closed after read
- Updated unused method ```associate()``` to avoid potential ```ClassCastException``` should it ever be used
- Made reading of credentials use ```UTF-8``` and closing the file in case of an ```Exception```
- Removed unused variable ```p``` in ```handle()```

The ```Exception``` handling is the only thing that's actually changed, everything else should be bug fixes with no behaviour change. The contructor now throws ```IOException``` into ```PMS.init()``` where it's called, and I believe that is the correct way to do it since ```PMS.init()``` need to handle if the creation of the web interface fails. It doesn't handle it very gracefully at the moment (but no worse than several other ```Exceptions```), but I'm working on another branch improving ```Exception``` handling during startup that will improve this if I ever find the time to finish it.

What do you think?

[Review on Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/805)

### DIFF
--- a/src/main/java/net/pms/remote/RemoteWeb.java
+++ b/src/main/java/net/pms/remote/RemoteWeb.java
@@ -4,9 +4,12 @@ import com.sun.net.httpserver.*;
 import java.io.*;
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
+import java.nio.charset.StandardCharsets;
+import java.security.AccessController;
+import java.security.GeneralSecurityException;
 import java.security.KeyStore;
+import java.security.PrivilegedAction;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
@@ -25,6 +28,7 @@ import org.apache.commons.lang.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+@SuppressWarnings("restriction")
 public class RemoteWeb {
 	private static final Logger LOGGER = LoggerFactory.getLogger(RemoteWeb.class);
 	private KeyStore ks;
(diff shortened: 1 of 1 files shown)


Nadahar:
@UniversalMediaServer/developers Any objections to merge this?

valib:
@Nadahar no objections from me.

valib:
Reviewed 1 of 1 files at r2.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/805)*

Nadahar:
It seems noone else has views on this, so I'm merging it.

SubJunk:
This seems fine to me but we should have waited longer before merging.

Nadahar:
@SubJunk I waited a while and there were no response from anyone but @valib. This is some minor adjustments to Exception handling and bugfixes really, nothing to controversial.

We can revert it if there's a problem.

SubJunk:
I don't have a problem with it being merged or staying merged, but waiting under 2 days for a response isn't long enough. I came into the PR the first chance I had to comment and look at it, and saw it had been merged before I could.

SubJunk:
Good work on the code by the way, it looks like a good change

Issue #698, opened 2015-10-17
New Media feature not working


SubJunk:
It seems that the "New Media" folder feature isn't working for me - on my PS3 if I watch a video to the end, it's not marked as watched and the file "UMS.mon" still doesn't exist.
The ".resume" files are created if I watch halfway through, and not created if I watch the whole video, so I think the overall logic is still working but it's just not recording the watched status in the file. That is how it used to work, didn't it? @SharkHunter ?

SubJunk:
The check in MediaMonitor.java:130 (if (monitorClass(tmp)) is failing when the video has finished playing

SubJunk:
Playing the "Clear All" video makes the PS3 display an error but correctly adds all files to UMS.mon

SubJunk:
Oh I had forgotten that it only works if you played the file via the New Media folder, duh. Closing this ticket now.

Nadahar:
@SubJunk This seems relevant to #806. I can't get this to work - what do you mean that it has to be played from the New Media folder? It this special folder found anywhere else?

PR #806, opened 2016-01-23
Mark all as played

Nadahar:
The `VirtualVideoAction` "Mark all as played" does not work as reported [here](http://www.universalmediaserver.com/forum/viewtopic.php?f=9&t=5883). With the new fully played functionality this has become even more of an issue, every upgrade of UMS will end up erasing all fully played information. Even though it's far from perfect, using this would atleast allow marking all media currently shared as fully played, so only new media won't be. 

I actually think this should be completely redone in such a way that a version upgrade doesn't affect what's considered fully played as it almost renders the whole fully played function useless.

Even so, "Mark all as played" is better than nothing, and it still should work if it's there. I've not been able to solve this. I've spent some time on it, and I don't know what broke this but I can see that there's issues with other `VirtualVideoAction`s as well, so I think the problem might lie outside this specific implementation. From what I can see, the code is never actually called.

I've done some changes in while trying to solve it, so I'm posting these here:
- Made MediaMonitor make one instead of multiple (One per shared folder) "Mark all as played" folder(s)
- Multiple bugfixes in MediaMonitor
- A bugfix in DLNAResource

I have to many other things to tend to, so I can't dig any deeper into this now. I'd really wish some of you would see if you could find out why the code isn't called and push to this PR.

[Review on Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/806)

### DIFF
--- a/src/main/java/net/pms/dlna/DLNAResource.java
+++ b/src/main/java/net/pms/dlna/DLNAResource.java
@@ -2813,7 +2813,7 @@ public abstract class DLNAResource extends HTTPResource implements Cloneable, Ru
 	 *
 	 * @return The system time when the resource was last (re)started
 	 */
-	public double getLastStartSystemTime() {
+	public long getLastStartSystemTime() {
 		return lastStartSystemTime;
 	}
 
--- a/src/main/java/net/pms/dlna/MediaMonitor.java
+++ b/src/main/java/net/pms/dlna/MediaMonitor.java
@@ -16,12 +16,11 @@ import net.pms.util.FileUtil;
 import net.pms.util.FreedesktopTrash;
 import net.pms.util.FullyPlayedAction;
 import org.apache.commons.lang.StringUtils;
-import org.fest.util.Arrays;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class MediaMonitor extends VirtualFolder {
-	private static Set<String> fullyPlayedEntries;
+	private static Set<String> fullyPlayedEntries = new HashSet<>();
 	private File[] dirs;
 	private PmsConfiguration config;
 
@@ -29,8 +28,12 @@ public class MediaMonitor extends VirtualFolder {
 
 	public MediaMonitor(File[] dirs) {
 		super(Messages.getString("VirtualFolder.2"), "images/thumbnail-folder-256.png");
-		this.dirs = dirs;
-		fullyPlayedEntries = new HashSet<>();
+		if (dirs == null) {
+			this.dirs = null;
+		} else {
+			this.dirs = new File[dirs.length];
+			System.arraycopy(dirs, 0, this.dirs, 0, dirs.length);
+		}
 		config = PMS.getConfiguration();
 		parseMonitorFile();
 	}
@@ -74,35 +77,41 @@ public class MediaMonitor extends VirtualFolder {
 		}
 	}
 
-	public void scanDir(File[] files, final DLNAResource res) {
+	public void scanDir(File[] files, final DLNAResource resource) {
 		if (files != null) {
-			final DLNAResource mm = this;
-			res.addChild(new VirtualVideoAction(Messages.getString("PMS.150"), true) {
-				@Override
-				public boolean enable() {
-					for (DLNAResource r : res.getChildren()) {
-						if (!(r instanceof RealFile)) {
-							continue;
+			if (resource.getChildren().isEmpty()) {
+				// Only add "Mark all as played" it there are no children here
+				// already to prevent one entry per monitored folder
+				final MediaMonitor mediaMonitor = this;
+				resource.addChild(new VirtualVideoAction(Messages.getString("PMS.150"), true) {
+					@Override
+					public boolean enable() {
+						for (DLNAResource r : resource.getChildren()) {
+							if (!(r instanceof RealFile)) {
+								continue;
+							}
+							RealFile rf = (RealFile) r;
+							fullyPlayedEntries.add(rf.getFile().getAbsolutePath());
 						}
-						RealFile rf = (RealFile) r;
-						fullyPlayedEntries.add(rf.getFile().getAbsolutePath());
+						mediaMonitor.setDiscovered(false);
+						mediaMonitor.getChildren().clear();
+						try {
+							dumpFile();
+						} catch (IOException e) {
+							LOGGER.error("Error dumping media monitor: {}", e.getMessage());
+							LOGGER.trace("", e);
+						}
+						return true;
 					}
-					mm.setDiscovered(false);
-					mm.getChildren().clear();
-					try {
-						dumpFile();
-					} catch (IOException e) {
-					}
-					return true;
-				}
-			});
+				});
+			}
 
 			for (File f : files) {
 				if (f.isFile()) {
 					if (isFullyPlayed(f.getAbsolutePath())) {
 						continue;
 					}
-					res.addChild(new RealFile(f));
+					resource.addChild(new RealFile(f));
 				}
 				if (f.isDirectory()) {
 					boolean add = true;
@@ -110,7 +119,7 @@ public class MediaMonitor extends VirtualFolder {
 						add = FileUtil.isFolderRelevant(f, config, fullyPlayedEntries);
 					}
 					if (add) {
-						res.addChild(new MonitorEntry(f, this));
+						resource.addChild(new MonitorEntry(f, this));
 					}
 				}
 			}
@@ -149,7 +158,7 @@ public class MediaMonitor extends VirtualFolder {
 		 * This is not a great way to get this value because if the
 		 * video is paused, it will no longer be accurate.
 		 */
-		double elapsed;
+		long elapsed;
 		if (res.getLastStartPosition() == 0) {
 			elapsed = (System.currentTimeMillis() - res.getStartTime()) / 1000;
 		} else {
@@ -241,7 +250,7 @@ public class MediaMonitor extends VirtualFolder {
 							if (Platform.isLinux()) {
 								FreedesktopTrash.moveToTrash(playedFile);
 							} else {
-								FileUtils.getInstance().moveToTrash(Arrays.array(playedFile));
+								FileUtils.getInstance().moveToTrash(new File[] {playedFile});
 							}
 						} catch (IOException | FileUtil.InvalidFileSystemException e) {
 							LOGGER.warn("Failed to move file \"{}\" to recycler/trash after it has been fully played: {}", playedFile.getAbsoluteFile(), e.getMessage());


Nadahar:
Review status: 0 of 2 files reviewed at latest revision, 5 unresolved discussions.

---

*[src/main/java/net/pms/dlna/DLNAResource.java, line 2816 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/806#-K8kAfpbKBpcdw2qPp1a:-K8kAfpcIk0XTKYNmxRV:1821676518) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/fcffd793c047a8b77f350b7e7da5dc04ba5386ee/src/main/java/net/pms/dlna/DLNAResource.java#L2816)):*
This must be a bug, ```lastStartSystemTime``` is of type long. Casting it to double in this way has no meaning that I can see.

---

*[src/main/java/net/pms/dlna/MediaMonitor.java, line 36 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/806#-K8kBHnbf8w57-B-DiGj:-K8kBHnc35Vef7SU2FtL:-1305669440) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/0de6ecab1f9e80737b8d02f5d81f979a66a1d4d7/src/main/java/net/pms/dlna/MediaMonitor.java#L36)):*
I did this to make sure an external object wasn't used by this class object. If the caller of the constructor later changes the array, it would be changed inside MediaMonitor as well.

---

*[src/main/java/net/pms/dlna/MediaMonitor.java, line 97 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/806#-K8kBiNUJCaipwSNn__l:-K8kBiNUJCaipwSNn__m:475215165) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/fcffd793c047a8b77f350b7e7da5dc04ba5386ee/src/main/java/net/pms/dlna/MediaMonitor.java#L97)):*
I moved this to ```discoverChildren()``` as it would otherwise be called once for every shared folder. I can't see that this has ever worked correctly, which means that any testing done on this must have been done with only one shared folder. The drawback is that the "Mark all as played" item now appears below all the other items, so it should maybe be moved so that it's inserted before the other children. It's hard to know what will work and not as long as I can't get the code to be called.

---

*[src/main/java/net/pms/dlna/MediaMonitor.java, line 107 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/806#-K8kCsDjhYybFc3JKSIq:-K8kCsDjhYybFc3JKSIr:428429760) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/0de6ecab1f9e80737b8d02f5d81f979a66a1d4d7/src/main/java/net/pms/dlna/MediaMonitor.java#L107)):*
This fixes and NPE if listFiles() return ```null```

---

*[src/main/java/net/pms/dlna/MediaMonitor.java, line 162 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/806#-K8kD2eIzVeGqimtAA4v:-K8kD2eIzVeGqimtAA4w:1070981455) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/0de6ecab1f9e80737b8d02f5d81f979a66a1d4d7/src/main/java/net/pms/dlna/MediaMonitor.java#L162)):*
I see no reason to keep fractional seconds here, and the way the calculation was done (casting to double after calculating) it would never have a fractional value anyway.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/806)*

valib:
@Nadahar this PR makes an error ```ERROR 2016-01-23 23:15:21.187 [net.pms.PMS.main()] A serious error occurred during Universal Media Server initialization: null```

Nadahar:
@valib Yes, it's not complete. I don't get that error though so a full stack trace would be helpful.

valib:
@Nadahar here is the log 
[debug.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/102073/debug.txt)

Nadahar:
@valib I hadn't read the documentation for ```System.arraycopy()``` good enough. I didn't catch that the destination array also had to be initialized, but since this probably is a low level C method it makes sense. Sorry about that, I don't know how it doesn't fail at my computer.. :blush: 

I've amended and force pushed the change (no need for a new commit for this), so you probably have to fetch and reset your branch to update it.

valib:
@Nadahar now it work good.

Nadahar:
Review status: 0 of 2 files reviewed at latest revision, 6 unresolved discussions.

---

*[src/main/java/net/pms/dlna/MediaMonitor.java, line 97 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/806#-K8kBiNUJCaipwSNn__l:-K8lHodExC0ke1ir-q-A:-118274665) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/fcffd793c047a8b77f350b7e7da5dc04ba5386ee/src/main/java/net/pms/dlna/MediaMonitor.java#L97)):*
I moved this section so it's before the actual files and foldes are added, to make "Mark all as played" appear on top of the list again.

---

*[src/main/java/net/pms/dlna/MediaMonitor.java, line 120 \[r2\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/806#-K8lHfalLJdNUJikKFo-:-K8lHfalLJdNUJikKFo0:751031107) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/5f539e2d5797db5661fa158c889809cb31732c01/src/main/java/net/pms/dlna/MediaMonitor.java#L120)):*
I just cleaned up some debug code I left behind here

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/806)*

valib:
Review status: 0 of 2 files reviewed at latest revision, 6 unresolved discussions.

---

*[src/main/java/net/pms/dlna/MediaMonitor.java, line 97 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/806#-K8kBiNUJCaipwSNn__l:-K8p_OZJJdn-dK7iLZci:-1882218909) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/fcffd793c047a8b77f350b7e7da5dc04ba5386ee/src/main/java/net/pms/dlna/MediaMonitor.java#L97)):*
The "Mark all as played" is missing in subfolders.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/806)*

SubJunk:
Why do upgrades need to erase all fully played information? I've never lost any of that kind of info

Nadahar:
@SubJunk I don't know why, I've just seen users claim that it does. I'm not sure on which platform though.

SubJunk:
On Windows it should persist unless they choose the "Clean install" option, which warns that they will lose all UMS preferences and related files. On OS X it seems to keep its preferences but I should verify that. On Linux it will save them if they extract over the top of their existing folder, not otherwise.

Nadahar:
@SubJunk That might be, I didn't check that further. The important thing for me was that it doesn't work, which I've confirmed when I've tested it. That's why I noticed the comment you had in #698 where you said that it only works from the "New Media" folder.

That made me believe that there's something major I'm missing, because as far as I can tell, it only exists in the "New media" folder - and it doesn't do anything when I test it.

Nadahar:
@SubJunk By the way you should take a look at [this] (http://www.universalmediaserver.com/forum/viewtopic.php?f=9&t=6062). Reportedly UMS stutters when he disables subtitles in the configuration for files with subtitles but play just fine with subtitles. I can't quite understand how that can be, is it a case where MEncoder is used if subtitles are active and FFmpeg if not, and that the problem really is with FFmpeg? In any case, you know much more about the transcoding engines and I really don't know where to start on that one.

Nadahar:
@SubJunk By the way you should take a look at [this] (http://www.universalmediaserver.com/forum/viewtopic.php?f=9&t=6062). Reportedly UMS stutters when he disables subtitles in the configuration for files with subtitles but play just fine with subtitles. I can't quite understand how that can be, is it a case where MEncoder is used if subtitles are active and FFmpeg if not, and that the problem really is with FFmpeg? In any case, you know much more about the transcoding engines and I really don't know where to start on that one.

valib:
Reviewed 1 of 2 files at r1.
Review status: 0 of 2 files reviewed at latest revision, 6 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/806)*

valib:
Reviewed 1 of 2 files at r1, 1 of 1 files at r2.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/806)*

Nadahar:
Review status: 0 of 2 files reviewed at latest revision, 5 unresolved discussions.

---

*[src/main/java/net/pms/dlna/MediaMonitor.java, line 97 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/806#-K8kBiNUJCaipwSNn__l:-K9ZMJZ8Zb-8u76KVosN:1632648925) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/fcffd793c047a8b77f350b7e7da5dc04ba5386ee/src/main/java/net/pms/dlna/MediaMonitor.java#L97)):*
Thanks, I hadn't thought this well enough through. It's a bit tricky since this method is run several times (one per monitored folder) for the top level "New Media" folder and then once for every child folder. I moved it back where it was and instead made a requirement that the resource could have no children already if "Mark all as played" are to be added. That should mean that the top level "New Media" folder only gets one entry both when there are multiple monitored folders and when "new Media" is entered several times. Test please.

---

*[src/main/java/net/pms/dlna/MediaMonitor.java, line 107 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/806#-K8kCsDjhYybFc3JKSIq:-K9ZM1AYX6o2OBHVzT2l:739280585) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/0de6ecab1f9e80737b8d02f5d81f979a66a1d4d7/src/main/java/net/pms/dlna/MediaMonitor.java#L107)):*
This fix only worked for the top level, null check moved to scanDir() instead

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/806)*

valib:
Reviewed 2 of 2 files at r3.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/806)*

Nadahar:
I must emphasize that even though this PR fixes some minor issues, it doesn't fix the main issue - making "Mark all as played" work. I consider this to be a start for anyone to continue rather than a PR up for merging.

---

Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/806#-:-K9gv5tatFtnN6Bk5-Zj:-239007165)*

SubJunk:
It works for me with this branch when the button is visible - which is in the root of the New Media folder

Nadahar:
It doesn't for me, which makes the bug even stranger. I haven't tested with the PS3 as it's placed far from the computer, so I used Kodi to test. Maybe that's the thing, that it only works on some renderers, like the PS3?

SubJunk:
I tested it on the web interface. I went into the New Media folder and clicked the button, then I got a notice in the upper right corner saying "Mark All as Played enabled" or something like that, which we should change the wording for but it's a cool style of notification anyway. Does it work for you on the web interface?

Nadahar:
I have an idea - the Kodi/XMBC renderer profile is completely fubar - that is, supported media isn't configured so it comes up with [FFmpeg] - transcoding. Maybe that's what disables it. There are users on the forum complaining about the same, so the problem might be just that.

Nadahar:
Yes, it works from me from the web interface. I'm betting that the transcoding is the problem - the question is how to fix it. Ideally the action should play whether the video is transcoded or not.

SubJunk:
I just checked and it also works in master via the web interface

Nadahar:
Yes, as I said I never found the problem, so I've only dealt with some other bugs and created a new one (that it only appears in the root folder).

SubJunk:
I think the New Media folder used to display all media files in the root, meaning there were no subfolders. There may have been a regression with that in 6.0.0

Nadahar:
It doesn't look like it was only one folder on the code, but there's still something odd when I test. Ideally, when you run the "mark all as played" on the top level, there should be nothing in the new media folder when you revisit it..? It is for me, in the subfolders.

---

Review status: 1 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/main/java/net/pms/dlna/MediaMonitor.java, line 87 \[r4\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/806#-K9o9xKHj-5UeAMWt93z:-K9o9xKHj-5UeAMWt94-:-1091272736) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/44e8b4a43a418a0e90f1e91efbaba0b38fb7713e/src/main/java/net/pms/dlna/MediaMonitor.java#L87)):*
This was just a little refactoring bug by me that meant it wasn't created in subfolders. It does now, but I force pushed to the branch must be reset.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/806#-:-K9oAD7JXuDo95RN8FDm:1746351176)*

Nadahar:
Looking at the code ran that's no surprise. It doesn't mark as played recursively. It also shows all subfolders whether they have "new" items or not. 

I was looking at not adding the folder in question unless it has children (when building the new media folder), but ```DLNAResource.addChild``` prevents adding children anywhere but last. That could be rewritten, but ```DLNAResource.addChild```is a mess so I abandoned that. 

---

Review status: 1 of 2 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/806#-:-K9oAlU5ADg3yFYBh7cQ:-1641585039)*

SubJunk:
@Nadahar I wonder if it's more intuitive that it marks folders as fully played recursively or not... I imagine there would be times when users wish it did that, and times when they wish it didn't, so it may be safer to have it as it is currently so that it doesn't mark things as played that the user didn't want to mark, since we have no easy way for them to unmark those files.

Nadahar:
@SubJunk I see arguments for both solutions, but if you consider a library with a lot of folders, going through them all recursively manually isn't really viable. We could make an extra option for recursive, I honestly don't know what is the best solution. It's easy to implement in any case, the question is how we think it's best. We could have "Mark folder as played" and "Mark all as played" for example. 

---

Review status: 1 of 2 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/806#-:-KA1soeWb25MhwEfmsIl:2061800654)*

Nadahar:
I guess this has been made obsolute by #942.

PR #807, opened 2016-01-24
Web interface tweaks

SubJunk:
This is a work in progress and will probably be a small PR with minor tweaks to improve the web interface.

[Review on Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/807)


SubJunk:
@UniversalMediaServer/developers any objections to these tweaks being merged? They are relatively minor, mostly CSS to make the web interface fit together better

valib:
@SubJunk tested and it looks good.

Nadahar:
I can't seem to find a way to "activate" your changes. I've tried building, cleaning and building, building with Maven but I still get "Browse the media on <name>" when entering the web page. What am I missing?

---

Reviewed 1 of 5 files at r1.
Review status: 1 of 5 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/main/external-resources/web/web.css, line 141 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/807#-K8nPEBOQzgc3MMybZdc:-K8nPEBOQzgc3MMybZdd:1401088925) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/88dcef0d589bcd6f5f478b7bcdd3368d7c31856f/src/main/external-resources/web/web.css#L141)):*
When you say fit well at 1080p, do you mean if someone is crazy enough to run the browser maximized?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/807#-:-K8nVvmylc9R2BjZC5Lg:-81821830)*

valib:
@Nadahar to make it working in Eclipse I had to copy the ```Workspace\UniversalMediaServer\src\main\external-resources\web``` folder to the root folder ```Workspace\UniversalMediaServer```

Nadahar:
@valib Ah, thanks

Nadahar:
It generally looks better, even though I'm a bit skeptical about the whole concept of the "start" page as long as there's no choices there.

I have a couple of comments, see the in-code comments. I don't really see the change to the default layout, so I'm guessing that the dynamic scaling take over very quickly? 

Generally I'm not too fond of the idea to assume that people run their web browser in full screen. Even though I know that many old people that's confused by multiple Windows do, I expect the general user to be more advanced than that. I have 1920 x 1080 + 1920 x 1200 screens, but I rarely run the browser bigger than.. 1600 x 900 or something there about. If I wanted to reduce my screen to just one application, I would have installed Window (8) instead of Windows.

---

Reviewed 4 of 5 files at r1.

---

*[src/main/external-resources/web/start.html, line 16 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/807#-K8nnxOfMIUN2A_YryhY:-K8nnxOg3RT6SDrVYUIY:-579425989) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/88dcef0d589bcd6f5f478b7bcdd3368d7c31856f/src/main/external-resources/web/start.html#L16)):*
Maybe we should use the same rules as is used for DLNA so it's consistent?

```java
				String profileName = "";
				if (configuration.isAppendProfileName()) {
					profileName = " [" + configuration.getProfileName() + "]";
				}

...
					s = s.replace("Universal Media Server", serverName + profileName);
```

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/807#-:-K8nojF6dcd6Xgbo8buH:18980532)*

SubJunk:
@Nadahar it would be interesting to get analytics for it so that we know what the most common screen sizes are. I think most non-PC renderers will display web browsers in full screen which would be 1920px wide. That's my guess. I don't know which percentage of our web interface users are using computers to play content vs. TV or console apps. I think 1920px is a good assumption to make for our users' screen resolutions for their PCs and renderers, and I think most people generally use browsers in full screen mode.

Our website analytics back that up, but that doesn't help us to know the renderer resolution.

SubJunk:

---

*[src/main/external-resources/web/start.html, line 16 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/807#-K8nnxOfMIUN2A_YryhY:-K8pck3xCD7-fP-r_KhK:421126156) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/88dcef0d589bcd6f5f478b7bcdd3368d7c31856f/src/main/external-resources/web/start.html#L16)):*
Yeah that would be good

---

*[src/main/external-resources/web/web.css, line 141 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/807#-K8nPEBOQzgc3MMybZdc:-K8pcuYcHbwX6u1O-XCt:1024363472) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/88dcef0d589bcd6f5f478b7bcdd3368d7c31856f/src/main/external-resources/web/web.css#L141)):*
I suspect most of our users do, especially considering that most non-PC devices don't give you a choice in that

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/807)*

valib:
Reviewed 1 of 6 files at r2.
Review status: 6 of 11 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/807)*

valib:
Reviewed 5 of 6 files at r2.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/807)*

Nadahar:
I don't think we need any analytics to figure out that 1920x1080 is the most common resolution, atleast excluding mobile devices. But I strongly disagree that most people will use web browsers maximized. I have been working as an It Manager in the past and have seen a lot of help desk cases, and in my experience only 60+ generally run applications maximized.

---

---

*[src/main/external-resources/web/start.html, line 16 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/807#-K8nnxOfMIUN2A_YryhY:-K8pfgDSFCysh4cSD0ft:2040597324) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/88dcef0d589bcd6f5f478b7bcdd3368d7c31856f/src/main/external-resources/web/start.html#L16)):*
#808 took care of that

---

*[src/main/external-resources/web/web.css, line 141 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/807#-K8nPEBOQzgc3MMybZdc:-K8pfvx_ozyUC1LoWdKn:685792068) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/88dcef0d589bcd6f5f478b7bcdd3368d7c31856f/src/main/external-resources/web/web.css#L141)):*
But don't you expect that the web interface is first and foremost for PC use? If you use a DLNA device, why use the web interface? If you're on mobile, I'd rather use something like BubbleUPnP

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/807#-:-K8pehsyjraDQz9aZSP5:-2145152794)*

SubJunk:
I'm not sure. Either way it only affects the pre-loaded display, and I'm not confident that we could come up with an estimate of the most common size of a windowed browser, so unless there's a way to achieve the effect we want with pure CSS I think this is a good default.

Anyway it still looks good on any resolution, the only difference is that there will be a small right margin on other resolutions until the page loads.

---

Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/807#-:-K8pupTKLx32n4NwoA8d:1123272530)*

Nadahar:
Yes, as I said - I couldn't really notice. I just noted it because I was scared of this assumption being used to "ruin" other resolutions in the future.

---

Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/807#-:-K8pvyPKWKFkEdOaXswN:703109686)*

WolfganP:
Another interesting tweak to the web interface is allowing local players (ie mpc, vlc, ..) to take over the rendering instead of the web player, displaying an extra icon with the direct file link as if being served (UMS Web becomes a remote media browser then).

I.e: in a PC configured with VLC as default player, if another icon can be added to the entries linked to the mediafile directly as served by UMS (ie http://ums_server:5001/get/523/test.mp4), then the web UI can be used to navigate the media files and folders, and the default player process the media link once that link is clicked (I tested with VLC and mpc-hc on a PC and works beautifully)

Idea originally discussed at http://www.universalmediaserver.com/forum/viewtopic.php?f=13&t=2947#p18377

Nadahar:
@WolfganP Can't you already do that with bump?

skeptical:
@WolfganP does pressing the 'download' button on a browser play page trigger external playback for you?
image

Nadahar:
Reviewed 1 of 2 files at r3.
Review status: 10 of 11 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/main/external-resources/web/jquery.ums.js, line 14 \[r3\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/807#-K8tZ0mdMWAgxITns6wy:-K8tZ0mdMWAgxITns6wz:-321975989) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/ab2e74d48270e2baa7b710d7f23cee73235bd0ea/src/main/external-resources/web/jquery.ums.js#L14)):*
This sounds like a bad idea. Why would we want to do that? IMO it much better that the thumbnails are shown in their native aspect and not padded.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/807)*

WolfganP:
@skeptical I didn't try the download button, I assume it *may* work. But it feels like a workaround to me (by the time the external player -ie VLC- launches, you already jumped into another page, loaded and started the web player and probably a full transcode on UMS side).

It will be cleaner to launch the external app from the "browsing folder" page (it only requires to link to the source file via http, ie 'http://ums_server:5001/get/523/test.mp4' like UMS does when serving the file, no need to reinvent the wheel).
As there's no definite DLNA client for windows, this improved UMS Web UI may serve well that purpose with just the install of a capable local player (NOT wmp :-), and when considering the parallel efforts in getting metadata/coverart from imdb/pother sources this may be easy to expand and look pretty, plus is eye catchy for uninitiated users.
Just my 2 cents...

skeptical:
@WolfganP well, why don't you try it first to confirm that this is what you're asking for: similar download button, but under the browse page items (and isn't that what I wrote in my comment? :-).

WolfganP:
Sorry @skeptical , my bad (I understood you was talking about a browser functionality, not the page itself). 
The Download button exports a ref 'http://ums_server:9001/raw/144' which doesn't trigger a download (just opens another tab and then errors, VLC also claims it can't open that MRL).
Looking at the logs, the way UMS responds when serving a renderer (URL pointing to a file like 'http://ums_server:5001/get/523/test.mp4') works better for native integration between the browser and a native player (at least in windows).
I hope it helps.

skeptical:
@WolfganP that's why I asked, since the download url doesn't include a literal 'extension' but otoh the data is sent with correct mimetype in its Content-Type header. I was curious to know if this was sufficient to activate an external media plugin.

Also, I tried installing the vlc browser plugin to see what happens, but it isn't detected by chrome at all (discontinued plugin api support) and in firefox it never seems to activate for me, even though I enabled it and configured it to handle the relevant mimetypes. I thought it was a linux thing but I get the same behaviour in Windows. I didn't try WMP, but if you have any quick tips please pass them along :-).

Of course all this isn't strictly speaking our concern, from ums point of view our client is the browser itself and as long as we provide a 'normal' download link then the browser is free to use it in whatever cool way it wants.  Will look at it some more.

Nadahar:
Just to mention it here, I have posted #694 a long time ago, which makes it possible to remote control WMP via BUMP for any installed codec. The problem is how to configure what's supported as that depends on installed codecs.

skeptical:
@WolfganP try #809.  I put it in a separate branch instead of right here since it conflicts a little with the thumbnail padding feature in this thread, which I agree with @Nadahar isn't a great idea.

valib:
I support this idea but the implementation will not be so easy as @skeptical made in #809. The UMS still responds like the renderer which is asking for the file is a browser and starts transcoding for not supported media. For that purpose a new ```DlnaResouce``` must be set like it is in the TRANSCODE folder to stream the media. What is worse the subs support should be updated too.

skeptical:
@valib Thx, (my excuse is)  I wanted to first confirm whether it's working for him at all :P.  Btw all the transcode folder items are available for download the same way.

valib:
@skeptical not all possible options are in the TRANSCODE folder especially the subtitles support. Anyway I tested it and it works. The only problem is that it is launching a new browser page not the external player. If you have installed e.g. a DivX or VLC module in the browser it could be OK.

skeptical:
@valib for some reason vlc-browser-plugin+firefox isn't working for me (as noted above) so someone should actually verify that it results in an external player launch.  We can easily send a true untranscoded file with a better url via slight changes to ```RemoteRawHandler``` I think, but that won't help with subtitles (but the place to fix that is in the transcode folder I suppose).

Nadahar:
Wouldn't it be better to take this discussion in #809? As s side note I though VLC abandoned their support for the browser plugin in 1.0. For me at least (with dreambox who's web interface use it), only pre 1.0 versions work.

skeptical:
@Nadahar agreed, onwards to #809.

valib:
@Nadahar agree that we can go to the #809. @skeptical I tested it and when I disabled the DivX module in the Firefox the OS (Windows 10) offers to me to use the  external player (the VLC) and there is a problem. The VLC waits for the downloading of the whole file and after that starts to play it. When I set to the VLC to play a stream ```http://ums_server:5001/get/523/test.mp4``` like @WolfganP mentioned above  it works.

SubJunk:
@UniversalMediaServer/developers I'm planning to merge this before releasing 6.0.0 since it seems fairly low-impact, unless there are objections

Nadahar:
@SubJunk There are objections if you look at what's already posted (has been there for a day or so). @skeptical has also said that he's with me on this. Aspect ratio lock for thumbnails is a terrible idea in my view, and I don't see why you would do that. If your focus it that it looks "out of order" when there's a mix of icon aspect rations, you should rather put them in some kind of grid than manipulate the thumbnails themselves.

---

Review status: 10 of 11 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/807#-:-K92FMWbDS1jGUx_YuZX:1713243096)*

SubJunk:
@Nadahar "terrible" is quite a strong opinion. Any reason for that? I don't see a comment from @skeptical and the only reason I see from you is that it's "better" the other way.

---

Review status: 10 of 11 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/main/external-resources/web/jquery.ums.js, line 14 \[r3\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/807#-K8tZ0mdMWAgxITns6wy:-K93T09jp0l21FeuqvcW:-2099560700) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/ab2e74d48270e2baa7b710d7f23cee73235bd0ea/src/main/external-resources/web/jquery.ums.js#L14)):*
The reason I did it is because I went into the transcode folder and could hardly see any of the text on the square icons because they were so short, and that's a time when we really need to see the text because that's what decides which version we want to watch.

Another reason is that it can result in a jarring experience if you are scrolling down and the widths change per row - it happens for me because some of my videos have custom covers that are a poster aspect ratio - which again leads to barely any space for titles and breaks up the flow of browsing.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/807#-:-K93TtewrmLoT_uAa7Zk:134209157)*

Nadahar:
A little further up in this very text you will find:
> @WolfganP try #809. I put it in a separate branch instead of right here since it conflicts a little with the thumbnail padding feature in this PR, which I agree with @Nadahar isn't a great idea.

It's hard to explain exactly what I think is so terrible about it because it's so obvious to me it's like an instinct. It simply looks like a bug, like rendering artifacts, it's very hard to explain it otherwise. If I said we should have pink circles around the icons instead, how would you explain that in any other way then that it just looks... bad? I simply can't see how filling the screen with black rectangles can bring anything good, and the only thing I can think as a possible reason to want to do it is to get the icons "in line", to follow a grid. But that must be easily achievable without adding black bars, using a table or something like that? 

---

Review status: 10 of 11 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/main/external-resources/web/jquery.ums.js, line 14 \[r3\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/807#-K8tZ0mdMWAgxITns6wy:-K93m_yPwpnUv1Dnm5XU:-226760703) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/ab2e74d48270e2baa7b710d7f23cee73235bd0ea/src/main/external-resources/web/jquery.ums.js#L14)):*
It seems to me like both those issues could be solved by placing the items in a grid without doing the padding.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/807#-:-K93kXYo8YgvuKUyGH4l:496798930)*

SubJunk:
@Nadahar maybe the effect you are describing is the same as removing line 144 in web.css, since we aren't adding real padding, just giving a black background to the container of the thumbnail.
Then it's just a question of whether we like the black background or not. I like it because it matches the way we already do thumbnails - it makes no sense to me to have black backgrounds for generated thumbnails but not for others - so I think we should be consistent either way.

Since some renderers need thumbnails to be a consistent aspect ratio, or else they stretch them, we would need to make a renderer setting to avoid a regression for those renderers, especially since I have several of those renderers myself.

---

Review status: 10 of 11 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/807#-:-K95cH0lUQRhLRzOR7g-:-1373404830)*

SubJunk:
Here's a comparison of this so far:

Current master:
old

New without background:
new-nobg

New with background:
new-bg

Nadahar:
@SubJunk In my view the middle one is preferrable, it's much easier to actually see the thumbnail/cover and it doesn't look bugged. I agree that arranging them in a grid makes it less messy in addition to having the capacity to show more of the string.

skeptical:
@SubJunk I prefer current master, the irregular dynamic close-packing effect is much more visually lively/organic than forcing a grid.  Of course I'm biased, having had a hand in creating it :-).

Nadahar:
@skeptical High/narrow thumbnails leave little room for the title text, that is a problem I've seen multiple times that the grid layout partly solves. I do however that basing the grid on the first element is way too random.

skeptical:
@Nadahar, I agree vis narrow images (but there's always the tooltip, or we could allow 2 line captions :-).  On the whole the dynamic "fill" is very visually pleasing to me, reminiscent of [perfect squares](https://en.wikipedia.org/wiki/Squaring_the_square)  as opposed to plain grids:

Nadahar:
I've browsed music where every entry has the same text (artist or album info), and using the tooltip to differentiate the entries is less than ideal.

When it comes to visually pleasing I don't know, what I do know is that it's easier for my brain to get the overview when they are organized in a grid. That might just be because of internal limitations on my part, but that's the way it works :wink:

skeptical:
@Nadahar as you always point out this stuff is very subjective. I love it when I see cover art, photos, tracks and liner notes all scattered together in one of our web audio folders, and the "same text" problem could be solved by multi-line captions.

Nadahar:
@skeptical It could, but it wouldn't help my brain getting any overview.

It's a bit impressive though. We are three people discussing this and we have three different views. No wonder peace treaties are so difficult to land :wink:

skeptical:
@Nadahar I give your brain much more credit than you seem to :-). I would also point out that google images uses the exact same dynamic fill strategy as ours, so I think there's an argument for it being humanly-intuitive.

Nadahar:
@skeptical I think the google case is somewhat different, since that's a unorganized display of images that doesn't represent any structure. You're supposed browse the images themselves instead of grasping an underlying structure represented by images. Google don't have texts with different width requirements than the images that should be presented with the images either, nor do they have the "tricky" buttons that's supposed to belong to to an element in an intuitive way.

I could also list a lot of examples of the opposite, in fact most thumbnailed listings I can think of are in some form of a grid layout. But, I'm open to the fact that what's intuitive vary from person to person, I guess it depends both on the brain's inner workings and what the viewer is used to.

But, we either have to pick one way, or make some kind of "view selection" (which would probably be the best solution). We could have "packed", "gridded" and "list" for example (represented by unintuitive icons if we want to follow the latest trends). I tend to prefer lists to grids in general, and a list view would easilty show the full text although it would have to restrict the thumbnail size to be meaningful.

skeptical:
@Nadahar I was just thinking the same thing, it would be good to have a web settings page for web 'experience' stuff, and store it in a cookie (and hopefully not ums.conf, since no one but us devs has the stamina for grokking all those settings :-).

valib:
Agree with @Nadahar. I also prefer the view without background and the "view selection" is a good idea to give users a choice.

WolfganP:
I also prefer the current master, which mimics the trend of "more natural" lists (aka google images).
My only concern is that small/weird AR images doesn't make a mess of the item title and icons down below the image (like overlapping), so a minimum width should be contemplated.

WolfganP:
Another helpful tweak may be to display a clickabe current "path" in the top bar to allow easier back navigation, ie: Media / TV-Series / Batman / Season 4 / ##TRANSCODE##

Besides, I have mixed feelings with the fixed right bar (I always push for "content is king" initiatives) when most of the times it will be empty or with just a couple entries. Maybe an improvement is to make it auto collapsible, or display the folder entries as boxes in the top bar?

skeptical:
Sorry guys, our new kitten added the 'invalid' label when he decided to take a nap on my keyboard :-).

@WolfganP :+1: on your suggestions, would be good to make these user-configurable. Btw as you may already know, narrowing the browser window will switch it to 'mobile' layout, and the sidebar moves to the top, etc.  There's a related minor bug here too, where the header sometimes spontaneously switches to narrow vs normal height and back again for a split second when we touch the scrollbar.

WolfganP:
@skeptical thx for the tip, I did noticed it while resizing the browser when testing the web UI a few days ago, excelent work BTW. 
The issue I see is that the folders list still arrange vertically, so the top bar may become a huuuuuuuuge section with tons of empty space (ie I have a Movies folder where most of the items inside are folders, one per movie and the mobile layout of top bar takes like two browser pages long).

Nadahar:
@SubJunk This padding has shown to be more complicated than I imagined. As stated before I don't think the black padding is very nice but I appreciate the grid pattern although that doesn't go for everyone else. 

There's one point I didn't see in the beginning, and that's the need to indicate what area "belongs" to what thumbnail to make the text and the "buttons" clearly linked to the item they belong. I guess I would have preferred a slightly darker version of the background already there to make that connection over black, but I guess you opted for black because the generated video icons already pad with black. I don't fancy that too much either, but that's another discussion. All in all I'd say it's acceptable with black from my view as of now. I would prefer having multiple "views", but if noone is willing to make it that's not really relevant.

What I'm more worried about with this solution is the fact that it forces all thumbnail sizes to be like the first one listed. I think that assumes too much about what is presented there. I can see the logic if most thumbnails are generated "double width" thumbnails and you have some covers in between, but I don't like tailoring something for just one case. I prefer general solutions that's as good as possible in all imaginable situations. What if the first thumbnail happens to have the "cover" shape? It won't play out quite that nice then. Also, what happens if the first item doesn't have a thumbnail?

Also, if you're listing audio or images that typically have square or "sqarish" thumbnails, the outcome can be quite random depending on the first item. In addition the space for the titles will still be way to small to give any meaning in many circumstances, especially for audio that have a tendency to start with a fixed prefix of album and/or artist. I'd prefer a solution that tried to take all this things into consideration, maybe with a minimum width to ensure readability, maybe with multiline text or maybe both - and where the size of the thumbnails were based on all and not just the first item.

That said, I don't want this to stop this PR, you have some nice tweaks in here and if you can't or won't adjust the thumbnail scaling/padding arrangement, it could always be done later.

---

Reviewed 1 of 2 files at r3.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/807#-:-K9feArLPBxF4Bpkv6kf:-1878032074)*

SubJunk:
@Nadahar I agree with you. I've been thinking of just setting the aspect ratio to 1.78 for the thumbnail containers. I also agree about having different view options, and I'd like to tie that in with the IMDb info so we pull more fields than we currently do and display things like actors, IMDb ratings, etc. in the other view, which I imagine to be like a "detailed" view that contains one or two media items per row.

I think I'll make the 1.78 change and then merge this, and work on the detailed view in another branch.

valib:
Reviewed 6 of 6 files at r2.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/807)*

SubJunk:
I've addressed the aspect ratio issue now. Am I good to merge?

skeptical:
> I also agree about having different view options

@SubJunk I took this to mean you'll provide a way to choose between 'grid view' and 'dynamic fit' before merging?

valib:
Reviewed 3 of 5 files at r4.
Review status: 10 of 12 files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/807)*

SubJunk:
@skeptical sorry I should've been more clear. My views on the dynamic fit are that it doesn't suit this program. I really like that kind of layout on image gallery websites, and website owners love them because they make people take longer to look through lists and prevents advertisement-fatigue by making their eyes scan all over, instead of being able to expect where content is.
In my opinion our goal is the opposite - we want to minimize the time the user spends looking for their media in the menu, so they can play it as quickly as they want.

If you think it's useful though, beyond just looking cool (which it does, I really like it from a pure design perspective as opposed to UX perspective) we can keep both options with a toggle, like you said.

The extra option I want to add is different, which would be a detailed view which has one video per row, where the thumbnail/cover is on the left of the row, and the details from IMDb like director, actors, rating, etc. are on the right of the row. It's a view that Plex do well and I've been wanting to add for a long time.

SubJunk:
@UniversalMediaServer/developers there are now toggle buttons on the upper right corner of the screen on browse pages, hopefully now we are good to merge?

Nadahar:
:lgtm: 

---

Reviewed 1 of 5 files at r4, 6 of 7 files at r5.
Review status: 15 of 16 files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/807#-:-K9xkpmuTD-0isqJwukH:980077423)*

valib:
@SubJunk how it works? I see that buttons are still the same and for me thumbnails are only make in grid. I think It would be better to see the chosen option by changing the button look maybe make them in the different color.

SubJunk:
@valib good idea, that's done now. Good to merge?

valib:
@SubJunk good job. The buttons are working but  all pictures and videos are always padded to 16/9 aspect so the grid view is only working for them. It is only working for audio thumbnails.

Nadahar:
Reviewed 2 of 2 files at r6.
Review status: 15 of 16 files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/807)*

Nadahar:
Reviewed 1 of 7 files at r5.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/807)*

SubJunk:
@valib we recently reverted the thumbnail padding by default so if you reset your cache that should be different. I think black is the best choice for the background in grid mode, but yeah we could make it transparent in dynamic mode because then the text and buttons will be aligned with the image anyway.

SubJunk:
How's this?

Nadahar:
Reviewed 3 of 3 files at r7.

---

*[src/main/resources/images/thumbnail-video-120.jpg, line 0 \[r7\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/807#-KA4HYP6e9A7KaKlQMT3:-KA4HYP7GJmT-cxxpU6u:1323003043) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/23a25e2b957ce200a2eaa90ff7f9e5e2ff5ca436/src/main/resources/images/thumbnail-video-120.jpg#L0)):*
I think it's very hard to understand that the new icon is a video. Is it supposed to indicate a video?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/807)*

SubJunk:
No, it's the icon that's used for audio files too (as in valib's screenshot) so I chose something more generic

Nadahar:
Ah ok, it's only the name that's wrong then. There should be no problem to see what is video,  audio and image from DLNAMediaInfo.is<type>, but you might not think it's worth it?

---

Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/807#-:-KA4Mfwzi_nz_8s6fK_z:-53971209)*

skeptical:
@SubJunk how's this? It still has a few css glitches (dropdown menu still sometimes underlaps despite fiddling with ```z-index```), but I thought the header menu was starting to look too busy, especially since view/pad settings are only used rarely.

SubJunk:
@Nadahar I'm not sure if it would be better to have different default icons depending on media type. I guess it would depend how nice the icons are

@skeptical good idea, yeah it was looking busy. We can tweak it a bit and fix the overlapping. Also we should probably make the search box into a realtime filter, there's no need to have a page load before displaying the results since it's not searching recursively - or make it search recursively

valib:
Reviewed 5 of 8 files at r8, 1 of 1 files at r9, 2 of 2 files at r10.
Review status: 21 of 24 files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/807)*

Nadahar:
Reviewed 3 of 8 files at r8.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/807)*

SubJunk:
@skeptical are you happy with the changes to the dropdown?

valib:
Reviewed 5 of 10 files at r11.
Review status: 22 of 27 files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/807)*

skeptical:
@SubJunk what does a guy have to do to get retired around here :-)?

Switching to horizontal orientation is a very creative dodge :+1:, I like it. It still drops behind the bump player, tho.  Didn't fix it, but added a few notes to make clear the thinking behind bump position/z-index.

valib:
Reviewed 1 of 1 files at r12.
Review status: 23 of 28 files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/807)*

valib:
Reviewed 2 of 5 files at r1, 1 of 5 files at r4, 3 of 7 files at r5, 2 of 3 files at r7, 3 of 8 files at r8.
Review status: 23 of 28 files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/807)*

valib:
Reviewed 5 of 10 files at r11.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/807)*

SubJunk:
@skeptical it's not a dodge - the initial issue where it was falling behind the layer is also fixed, I just thought it looked better this way too :) I didn't test the bump player so I'll do that

SubJunk:
@skeptical how's this?

SubJunk:
@skeptical how's this?

skeptical:
@SubJunk :+1:  nice: makes it look like we know what we're doing + visually amusing + minimal coding.

PS Thx for updating jumpy. Another wishlist TODO: in dynamic layout the 'WATCHED' overlay is forcing padded aspect. I know you've mentioned it as a possibility yourself,  but imo it's worth considering dropping text altogether and just using a smaller symbol instead, e.g.:
watched48

SubJunk:
@skeptical great, thanks for all the reviewing. It's always good to give the appearance of knowing what we're doing ;)
Yeah I think a symbol would be better than text, and that one is nice. The green tick here on GitHub is nice too. I can make that feature or if you get a chance you could, it's not like you were looking to retire from the project or anything

@UniversalMediaServer/developers are we good to merge this branch now?

valib:
Reviewed 4 of 4 files at r14.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/807)*

SubJunk:
@skeptical the start of an implementation of that image overlay feature is at https://github.com/UniversalMediaServer/UniversalMediaServer/pull/845
It's working but could use some safety around sizing, and could use a bigger overlay icon since 48x48 seems a bit small. Where did you get the icon from?

skeptical:
@SubJunk it's a gnome icon ([gnome-stencils.svg](http://git.gnome.org/browse/gnome-icon-theme-symbolic/tree/src/gnome-stencils.svg), we should probably credit it if we use it), here it is by itself in svg: [watched.svg.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/144683/watched.svg.txt), and larger pngs:
watched72watched96watched72 f gold swatched96 f gold s

SubJunk:
Moving the conversation about the image overlays to #845 

Are you happy if I merge this branch @skeptical ?

skeptical:
@SubJunk :+1:

WolfganP:
Please don't forget to merge https://github.com/UniversalMediaServer/UniversalMediaServer/pull/809

skeptical:
@SubJunk  if you don't mind: newer _jumpy-0.3.9b_ update for the plugins tab: [jumpy-info](http://skeptical.github.io/jumpy/jumpy-info), [jumpy-inst](http://skeptical.github.io/jumpy/jumpy-inst), thx :-).

SubJunk:
@skeptical done

PR #809, opened 2016-01-25
Web: Add download button to browse page items

skeptical:
Per @WolfganP's suggestion in #807, for testing.


skeptical:
> As s side note I though VLC abandoned their support for the browser plugin in 1.0.
@Nadahar their current installer offers to install the plugin, and apparently @WolfganP is using it.

Nadahar:
@skeptical I know that they offer it, but I've never seen it work since v1.0. VLC being open source I don't expect that one hand knows what the other is doing, so I've just assumed that the information that they should remove it from the installer never got there. But, I haven't studied it, just stuck to pre 1.0 versions.

skeptical:
@Nadahar touche, on the whole open source guys are pretty clueless I guess :-)

Nadahar:
@skeptical I didn't mean that, but when it comes to organization and structure there's some issues in that nobody can "order" anyone to do anything, so boring/tedious tasks (like having an overview and making sure things fit together/are consistent) have a tendency to be nobody's responsibility.

To pollute this PR a little too, I'm a bit... confused at the moment. I'm trying to figure thumbnailing out properly, and thus I'm reading the DLNA spec. Did you guys know that the way we do transcoding (and almost everything else) is completely opposite of how it's designed?

From what I can see from the specs, you can offer any number of alternative "sources" for any given object (say a video). A media server is required to offer any media in a set of very restrictive, low quality formats and is then free to offer as many better alternatives as it wants. The idea is that the renderer should then choose the best quality format that it supports. That means that, if I read this correctly, we could serve the file as is as "best quality" and one or more reduced quality alternatives that's transcoded, and the renderer would pick the best one. If that actually work as described, the our whole system with supported formats isn't needed, it's just to offer alternative URL's for transcoded alternatives and start transcoding when the request for one of them is received. 

Do any of you know if this approach has been tested and abandoned because it doesn't work, or do we simply not do this because of ignorance?

skeptical:
> I tested it and when I disabled the DivX module in the Firefox the OS (Windows 10) offers to me to use the external player (the VLC) and there is a problem. The VLC waits for the downloading of the whole file and after that starts to play it. When I set to the VLC to play a stream http://ums_server:5001/get/523/test.mp4 like @WolfganP mentioned above it works.

@valib does the same thing happen if you try the links on [this page](http://download.blender.org/peach/bigbuckbunny_movies/)?

@Nadahar just kidding re clueless.  I say it's ignorance (and to take a page out of your own book, shouldn't you open a separate issue for this :-)

valib:
@Nadahar what do you mean? How can I test it? If I set one of the files as a stream to the VLC it plays. Or I should add it to the web.conf and use the UMS?

valib:
So the web interface should recognize the default OS player and if possible to set them to play a stream otherwise it wouldn't work.

Nadahar:
@skeptical I don't know enough about it to create an issue, I just wanted to mention it - I just read something about it by coincidence while looking for the thumbnail "rules", and assumed there was a whole story to this that I don't know. I'm sorry for going off topic, it was just very much on my mind at the moment I wrote that.

@valib I'm not sure I understand what you asked me, but to clear up any misunderstanding I was only relaying my experience with the VLC browser plugin - I haven't looked into what you're really trying to do here, I just thought I should mention that so you don't waste time trying to get a potential none-working browser plugin to work.

valib:
@Nadahar firstly I didn't understand well what are you asking me. When I realize what it is about I tested the standard VLC player because the Firefox VLC plugin doesn't work for me.

skeptical:
@valib well it seems you sorted it out. I only asked to test the other links so as to confirm that there's nothing wrong or unusual with our own download link per se, but so far it looks like the issue is with the vlc plugin. I think this feature is only worth pursuing in the scenario where the browser itself launches the external player,  maybe @WolfganP can point us in the right direction.

Nadahar:
I'm not sure I understand the need for this, but I guess more always more. Why would anyone use DLNA to access files from a PC in the first place? There are many good media players out there capable of maintaining a libraray of files available ona file system, Kodi for example. Kodi can also do a reasonable job via DLNA. Transcoding, in many ways the main purpose of software like UMS, isn't needed as "all" codecs can be supported by installing them.

I'd much rather use a system like Kodi on a "media center PC" than a webbrowser + VLC, but unless I'm accessing the PC via a remote controller I'd much rather just use a media player. In addition, WMP can be used as a push-to renderer with #694 via either BUMP or UMS' renderer panel.

I'm not saying "don't do it", I'm just saying think about if it's actually a need for this before digging into it for real.

skeptical:
@Nadahar Also vlc itself can supposedly act as a DLNA client too.  If all that's required for browser media plugins is a download link then of course it's worth doing, but not beyond that imo.

Nadahar:
@skeptical The first thing I tried was to "push" to VLC. But with VLC 2.08 (that I tested with), UMS won't even recognize it as a renderer. VLC isn't able to browse UMS, and I suspect that the reason is the same as for WMP - they try to enumerate the whole tree first so that they can organize the content themselves. That obviously fails with UMS because of all the virtual folders, virtual video action++. 

I don't know why it fails now, I've been able to detect it before, but I know as much from earlier that VLC's DLNA implementation isn't very solid. I tried quite a lot to get it to work with UMS earlier, as a test DLNA client, but after reading some forums and having no success I got the impression that it's not something they really care about and is just something someone has added without paying much more attention to it. 

I have the same result now with VLC 2.2.1. Remembering how much I struggled with it earlier I just concluded that VLC probably just implements DMP and not DMR. I might be wrong, lots of assumptions here, but that was the first thing I checked when this discussion started, because then BUMP/Renderer panel could be used to control VLC. 

Do I understand you correctly if you're talking about a "somewhere in the middle" solution where VLC can't receive control commands but somehow receive URL's that it then can use if it were just another renderer?

> DMP
Digital Media Player
DLNA Device Class having home network environmental characteristics, with the role of finding content exposed by a DMS or M-DMS and rendering the content locally

> DMR
Digital Media Renderer
DLNA Device Class having home network environmental characteristics, with the role of rendering content it receives after being setup by another network entity

WolfganP:
Thx for opening an specific issue on this topic. I don't know why the discussion fixed on VLC plugin and FireFox, as I'm not using it nor I considered VLC as a requirement, it was just an example (user may have MPC as his/her preferred/default player configured and it just have to work OK, same if using Chrome or Edge)

My idea relies on the browser using the internal "open with" mechanism (ie FireFox about:preferences#applications) and default system media player, that's why I suggested the icon to export links pointing to a file type URL (as in http://ums_server:5001/get/523/test.mp4) as if UMS is just serving the file as natively supported, no renderers or transcode engines analysis needed (it's on the client side to proceed to play or fail -ie mkv with wmp on win7-)

Regarding the Kodi comments, probably you're thinking on a dedicated media PC, I was just thinking in eventual connections from any PC/Mac (even mobile phones) where a browser and some media player is surely installed by default, and UMS acts as browseable direct media server (no need to map network folders or the like).

Does it make sense?

skeptical:
@WolfganP thx for the pointer, so I uninstalled the vlc plugin and now the browse page download links successfully open an external player (I tried vlc app and ffplay), but the catch is that this happens only after download is complete which negates its usefulness quite a bit.

However the good news is instant external playback is triggered if the link is wrapped inside a playlist, e.g. m3u
```m3u
#EXTM3U
#EXTINF:-1,ums-test
http://192.168.1.5:5001/get/984/%2Fg%2Fvid%2Ftest%2Fvideo_for_everybody%2Fbig_buck_bunny.mp4
```
so I'll give it a shot :-).

@Nadahar I could have sworn I had a dlna-capable vlc once, but you're right it isn't listed in the supported services now (2.2.1).

ExSport:
Nice catch! Hopefully m3u trick will work also for MPC HC (Home Cinema builds) as it also supports all codecs internally and with hw acceleration. I prefer it instead of VLC as it had sometimes problems with some of my files which were played  incorrectly (up-bottom)

skeptical:
Seems to work, but not extensively tested.  Should stream unless it's a transcode folder item.

Nadahar:
Wrapping in M3U has been used by the dreambox webinterface for years (as a fallback if the VLC browser plugin isn't installed or doesn't work). I've assigned .m3u to VLC in my Windows file association and have used that without problems for at least ten years. I think that's pretty safe with VLC, but it does require you to bind .m3u to VLC, and other players tend to "steal" that association when installed.

skeptical:
@Nadahar right, what happens to the playlist is ultimately up to the user, not us.  Btw I don't claim this to be a "trick" (@ExSport was just being kind :-), it's very standard for streaming links.

valib:
@skeptical it doesn't work for me. I tried the Firefox, Chrome and Microsoft Internet Explorer and it always open the new page with 
result and not to open the default player. Before the change to use the ```m3u8``` it was working and VLC was called. The same happened when I set the default player to MPC. Tested on the Windows 10. Maybe I should set something to make it working. What's more as I stated some time ago the subtitles support should be updated too. When testing the video with external subtitles when i forced the VLC to play the m3u8 playlist I made from the UMS respond e.g. ```#EXTM3U
#EXTINF:65,1avataR part
 [FFmpeg] {ExternĂ­ titulky}
http://10.0.0.5:5001/get/42/_NOTRANSCODE_E%3A%5CFilmy%5C1avataR+part.avi``` it makes an error ``` Caught exception: java.lang.ClassCastException: net.pms.configuration.RendererConfiguration cannot be cast to net.pms.configuration.DeviceConfiguration```

skeptical:
@valib I only tested in firefox (chrome doesn't seem to have a setting to launch external apps).  Try disabling any media plugins you have and then pick various apps from the "open with" dialog that pops up when you press the link.  In linux it's working for me with vlc (app, not plugin), Movie Player (system default), and vlc.

Can you post a log for the exception?

src/main/java/net/pms/network/RequestHandler.java
```diff
@@ -94,8 +94,17 @@ public void run() {
 			// header matches are attempted and if those fail as well we're stuck with the
 			// default renderer.
 
-			// Attempt 1: try to recognize the renderer by its socket address from previous requests
-			renderer = RendererConfiguration.getRendererConfigurationBySocketAddress(ia);
+			// Attempt 1: If the reguested url contains the no-transcode tag, force
+			// the default streaming-only conf.
+			if (request.getArgument().contains(RendererConfiguration.NOTRANSCODE)) {
+				renderer = RendererConfiguration.getStreamingConf();
+				LOGGER.info("Forcing streaming.");
```
valib: @skeptical I think it should be set to```debug``` or better to ```trace``` not to the ```info```. It spams the info whenever the chunk is asked to stream. The same should be applied for ```RequestHandlerV2```

skeptical: :+1:

valib:
@skeptical I am aware of it and I disabled all media plugins before testing but without success.

skeptical:
@valib must be a Windows thing, will look at it later when I get a chance.

valib:
@skeptical here is the log 
[debug.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/105932/debug.txt)

skeptical:
@valib whew, I hate windows :D.  As @Nadahar commented above mimetype/extension support in firefox is mostly driven by registry associations, so
* there was no 'm3u' item in about:preferences#applications for me.
* there's no easy way to add mimetype associations inside firefox, so instead I purged vlc and reinstalled it (including the browser plugin), which then thru its registry mods produced the desired m3u application association (to vlc plugin) in firefox, along with a dozen or so other media types.
* I went through and set all vlc plugin items to 'Always ask', and voila found that I got the 'open with' dialog for m3u, so far so good.
* but m3u8 still downloaded as text in a new tab, even though in explorer (file mgr) m3u8 was plainly associated with vlc (did I mention I hate Windows?).
* after patching ums to produce m3u instead of m3u8 links, all is well, ums links seem to open up in vlc and ffplay.

Will revert to m3u here for now, though the proper solution would probably involve looking into why the m3u8 registry entries (same as m3u, confirmed via regedit) didn't propogate to firefox. Probably the vlc plugin's fault I would guess.

Nadahar:
@skeptical Each time I have to recreate the m3u <-> VLC association because some stupid installer have hijacked it, I just create an empty file on my desktop, renames it to .m3u, rightclick -> open with -> Select default application -> VLC.

I m3u8 wont work I have no idea though, but m3u kind of suck being limited to latin-1.

WolfganP:
Thx a lot for the prompt implementation! I'll try building this branch, test this feature and report back.

WolfganP:
Just built webdownload branch with UMS-builder to test this functionality, but I can't browse any folder below root (http://fserver:9001/browse/0), Chrome & Firefox errors in:
"The fserver page isn’t working
fserver didn’t send any data.
ERR_EMPTY_RESPONSE" 

Log attached.
[ums_dbg-webdownload.zip](https://github.com/UniversalMediaServer/UniversalMediaServer/files/106953/ums_dbg-webdownload.zip)

skeptical:
@WolfganP how are you deploying webdownload, with full nsis installer or just copying the jar (or maybe using 'quick compile' setup)? The web resources have changed, so web/browse.html won't work unless it's the new version.  If you're using the jar only, try renaming the web folder to web.bak and copying/symlinking the web folder from this branch into your program folder.

skeptical:
Well for some reason when I booted into Windows today firefox now (magically :-) shows a 'm3u8' item in about:preferences#applications, so I restored m3u8 use here, and am getting (reasonably) good luck with external apps in Windows.

I should note that once in a while the m3u/8 files produced here fail in ffmpeg-based apps (vlc, mpc, ffplay) due to apparently very picky implementation of m3u in ffmpeg, so I'm trying to see if I can satisfy that.  We could also try pls or asx I suppose.

@Nadahar yeah, I still have an old set of .reg files from my Windows days for undoing hijacked file associations.

WolfganP:
@skeptical Thx, that was the issue. I figured it out when analyzing the files changed by this PR.
I tested Chrome and Firefox on Win7, with both VLC and MPC-HC players.
The only issue I have is with MPC-HC, it freezes trying to open the stream via the .m3u link, but if add the mediafile link directly (ie http://fserver:5001/get/178/_NOTRANSCODE_E%3A%5C%21TV+various%5Ctestfile.mkv as exported by the m3u) it works perfectly, so I assume is a MPC issue and nothing UMS related.
Again, thanks all for the quick implementation.

skeptical:
@WolfganP so far so good, what's the setting to enable external apps in chrome?

valib:
@skeptical when I tried to play video with subtitles I have got ```DEBUG 2016-01-27 21:29:31.323 [New I/O worker #6] Caught exception: java.lang.ClassCastException: net.pms.configuration.RendererConfiguration cannot be cast to net.pms.configuration.DeviceConfiguration```

skeptical:
@valib working on it :-).

valib:
@skeptical I am not forcing you :smile:

skeptical:
@valib I'm digging my own grave, I know, but you're lending a hand it seems :-).

skeptical:
@valib try it now.  It seems to me that default streaming should be 'as-is', and subtitles available via transcode folder.  What do you think?

valib:
@skeptical it works but I think that it would be better to transcode the video when there are external subtitles. I know that m3u doesn't support subtitles so I think this is the better option how to show them then to go to the TRANSCODE folder. This implementation still will have a lot of advantages when all other videos will be streamed.

WolfganP:
I assume this is ready to merge but it's waiting to ride along the changes at https://github.com/UniversalMediaServer/UniversalMediaServer/pull/807
Is there any additional test (Windows platform) I can run to help debugging?

skeptical:
@WolfganP this branch is unrelated to #807 and so it's not waiting for that.

* I refactored everything so the webdownload branch needs to be deleted and checked-out again.
* The m3u8 files are now generated from a template ```web/play.m3u8```, so it's possible to edit this file to change the m3u formatting.  I removed the ```EXT-X-TARGETDURATION``` line, since this conflicted with multi-item playlists, but it can be put back into the template if need be like so:
```
#EXT-X-TARGETDURATION:{{targetDuration}}
```
* There's a new 'downoad as folder' button to get a playlist of all items in the folder.

skeptical:
> There's a new 'download folder' button

Oops, forgot to add the icon folderdown16 - fixed now :-).

valib:
*[src/main/java/net/pms/configuration/RendererConfiguration.java, line 243 \[r4\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/809#-K9P39Yon-227s9H2zpR:-K9P39Yp1apTiRARiYk4:13511669) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/f6177312247e880d69319ebc7a350df96cd31c00/src/main/java/net/pms/configuration/RendererConfiguration.java#L243)):*
I think that this should be deleted. When it is not used the video with subtitles is transcoded accordingly to the users setting. To play video without subtitles users can use TRANSCODE folder.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/809)*

valib:
Reviewed 2 of 12 files at r1, 6 of 9 files at r2, 10 of 10 files at r4.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/809)*

Nadahar:

---

*[src/main/java/net/pms/configuration/RendererConfiguration.java, line 243 \[r4\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/809#-K9P39Yon-227s9H2zpR:-K9PgA2hoZ0CMJ2a1Uhf:-1603751099) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/f6177312247e880d69319ebc7a350df96cd31c00/src/main/java/net/pms/configuration/RendererConfiguration.java#L243)):*
That sounds like the right thing to do. The same rules should apply here as elsewhere.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/809)*

skeptical:
@valib @Nadahar you're mostly right, of course.  The reason I was holding back is that here we're dealing with "super-renderers" that can basically handle anything, so ideally we shouldn't be transcoding at all but just passing everything along.  Internal subs travel with the file but I was looking for a way to package/transfer external subs similarly, since vlc/mpc/etc would easily handle them if  available.  Supposedly there's such a thing as a [streamable mkv](https://matroska.org/technical/streaming/index.html), which might allow for bundling srt with the video file, but so far I can't get ffmpeg to mux one on the fly (I'd tried/failed at it a few years back too, but hope springs eternal :-). Possibly there's a playlist-related way to do it to, but no luck finding anything so far.  Suggestions are welcome.

But on the whole the m3u approach is arguably better than html5/flash, since there are almost no format constraints and most everything can be instantly streamed, :+1:  to @WolfganP for his persistence :-).

WolfganP:
@skeptical thanks for the thumbs up, x10 to you guys for implementing this so promptly and improve on the original idea :-)

Regarding streamable (aka soft) subtitles, I seem to recall tsMuxer allow that somehow, so wouldn't the tsMuxer entry in ##TRANSCODE## folder do the trick if any subtitle file (ie .srt) exists? (http://www.videohelp.com/software/tsMuxeR)

Nadahar:
@skeptical That (about the subtitles) sounds reasonable, the important thing is that this doesn't end up in this state if a good solution for streaming them isn't found.

---

Reviewed 2 of 12 files at r1, 4 of 9 files at r2, 10 of 10 files at r4, 1 of 1 files at r5.
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[src/main/java/net/pms/configuration/RendererConfiguration.java, line 226 \[r5\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/809#-K9TCMaxQt5kiM-vb2JH:-K9TCMaxQt5kiM-vb2JI:438342791) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/c9e0f689bd05ef0c3486065c7f3752d69a7ecfb0/src/main/java/net/pms/configuration/RendererConfiguration.java#L226)):*
I'd really love if we could just write things out when introducing new stuff, it makes the code so much more readable that it's worth the extra second it takes to write imo.

---

*[src/main/java/net/pms/configuration/RendererConfiguration.java, line 243 \[r5\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/809#-K9TCmR5prnJqv8Wt6aD:-K9TCmR5prnJqv8Wt6aE:1653865) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/c9e0f689bd05ef0c3486065c7f3752d69a7ecfb0/src/main/java/net/pms/configuration/RendererConfiguration.java#L243)):*
Same here...

---

*[src/main/java/net/pms/network/RequestHandlerV2.java, line 127 \[r5\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/809#-K9TDG95kYmPH26cFq2T:-K9TDG95kYmPH26cFq2U:-510466975) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/c9e0f689bd05ef0c3486065c7f3752d69a7ecfb0/src/main/java/net/pms/network/RequestHandlerV2.java#L127)):*
:+1: 

---

*[src/main/java/net/pms/remote/RemoteBrowseHandler.java, line 162 \[r5\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/809#-K9TERiYDM3BtAtmkUu5:-K9TERiYDM3BtAtmkUu6:-628683193) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/c9e0f689bd05ef0c3486065c7f3752d69a7ecfb0/src/main/java/net/pms/remote/RemoteBrowseHandler.java#L162)):*
This is probably just due to my lack of overview of this whole area of code, but are you replacing the bump functionality instead of adding this?

---

*[src/main/java/net/pms/remote/RemoteUtil.java, line 124 \[r5\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/809#-K9TEjKxLnlzyJaljbEx:-K9TEjKyxyLkWzwBvgRp:-510466975) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/c9e0f689bd05ef0c3486065c7f3752d69a7ecfb0/src/main/java/net/pms/remote/RemoteUtil.java#L124)):*
:+1: 

---

*[src/test/java/net/pms/configuration/RendererConfigurationTest.java, line 216 \[r5\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/809#-K9TF6r6RqoY-tPvNswN:-K9TF6r7SvvglX-AiLuI:-1029163598) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/c9e0f689bd05ef0c3486065c7f3752d69a7ecfb0/src/test/java/net/pms/configuration/RendererConfigurationTest.java#L216)):*
and again..

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/809#-:-K9TFQXa_hT4UgqlBcbE:-1917022962)*

valib:
Reviewed 1 of 1 files at r5.
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/809)*

SubJunk:
This looks good to me

SubJunk:
@skeptical do you have any interest in resolving the conflicts in this branch so it can be merged or is it abandoned now?

skeptical:
@SubJunk abandoned? What would I say to @WolfganP then :-)? Please double-check, since you've made a lot of later css changes in #807. Also, since you've changed the bump icon, for consistency could you please generate a matching large version to replace _web/bump/bump32.png_ (for the big bookmarklet button on _bump.html_).

@WolfganP it'd be great if you give this a workout to make sure everything's still ok :-).

WolfganP:
@skeptical sure I'll test it. I switched machines so I need to reinstall my UMSbuilder environment, and will report back after testing. Thanks a lot!

SubJunk:
@skeptical it mostly looks great, the download button fits in really well with that set. I'm not sure about the placement of the folder download button though, it looks out of place to me. I'll generate the 32px image soon, nice spotting

SubJunk:
@skeptical is this what you meant? I can make the PSD available if you want to make edits or feel free to totally change it

skeptical:
@SubJunk I personally prefer my original icon bump16-2x i.e. it's meant to convey 'casting' like cast_x5f_icon_x5f_off_x5f_light_2_16, but I'm guessing you don't like giving up the extra caption real estate and prefer the smaller generic 'play' button.  I can see the logic of that too, so it's not a big deal :-).

SubJunk:
@skeptical I'll try to make it more similar to the original while still fitting there

SubJunk:
@skeptical I've updated the icons in master. Can you please come up with a different placement of the folder download button before we merge?

skeptical:
@SubJunk believe it or not I wrestled with its placement quite a bit before settling on that spot :-).  It 'naturally' wants to be within the media area so it's visually associated with that set of items, and it seems silly to open up a dedicated space for an incidental feature like this.  I tried the other 3 corners + top-center, and all those were more awkward to my eye than the top inside corner.

SubJunk:
@skeptical how about to the right of the Back button, inside the left panel?

valib:
@SubJunk what about to merge this PR?

SubJunk:
@valib I don't think the button fits there, do you? I'm happy to be outvoted

SubJunk:
The little "download" icons under the thumbnails are confusing for me because as a user I think I would expect that to download the video. I might be confused when I get a file I don't know what to do with. Maybe we can use an icon that is like that but with some lines around the arrow, to make it more clear you are downloading a playlist not the video itself?

SubJunk:
This will need to be reworked since we merged the remote control branch

SubJunk:
@UniversalMediaServer/developers what do you think of this approach? It gives us more room for the title of the media, and adds the new download button

image

SubJunk (approved):

valib:
@SubJunk this PR doesn't work for me. In the produced m3u8 file there is the URI with the `_NOTRANSCODE_` inside the file name but this must be removed before the UMS starts streaming this file otherwise renderers are waiting for the stream which the UMS cannot produce because the file with `_NOTRANSCODE_`doesn't exits.

SubJunk:
@valib maybe I broke it when I synchronized with `master`. I didn't intentionally change any of that logic. Any ideas for a fix?

PR #813, opened 2016-01-29
Misc bug fixes to FileUtil

Nadahar:
@UniversalMediaServer/developers This started out as a simple bugfix for [this bug] (http://www.universalmediaserver.com/forum/viewtopic.php?f=9&t=6011). But, after fixing that bug I ran FindBugs and Threadsafe, and this lead to much more (and changed this from being a commit to a PR).

I hope we can expedite this once quickly, as there's no (intended) behaviour change here. I have multiple questions and comments though, so this needs to be reviewed. I'll make the comments in Reviewable.

[Review on Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/813)

### DIFF
--- a/src/main/java/net/pms/util/FileUtil.java
+++ b/src/main/java/net/pms/util/FileUtil.java
@@ -1238,13 +1238,16 @@ public class FileUtil {
 				}
 			}
 
-			// Replace periods with spaces
-			filename = filename.replaceAll("\\.", " ");
+			// Replace periods and underscores with spaces
+			filename = filename.replaceAll("\\.|_", " ");
 		}
 
 		if (PMS.getConfiguration().isIgnoreTheWordThe()) {
-			// Remove "The" from the beginning of files
-			filename = filename.replaceAll("^(?i)The[ .]", "");
+			// Remove "the" from filename
+			filename = filename.replaceAll("(?i)\\bThe\\b", "");
+
+			// Replace multiple whitespaces with space
+			filename = filename.replaceAll("\\s{2,}"," ");
 		}
 
 		return filename;
--- a/src/main/java/net/pms/util/FileUtil.java
+++ b/src/main/java/net/pms/util/FileUtil.java
@@ -30,14 +30,16 @@ import org.slf4j.LoggerFactory;
 import com.sun.jna.Platform;
 import com.ibm.icu.text.CharsetDetector;
 import com.ibm.icu.text.CharsetMatch;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import org.apache.commons.lang.WordUtils;
 
 public class FileUtil {
 	private static final Logger LOGGER = LoggerFactory.getLogger(FileUtil.class);
-	private static Map<File, File[]> cache;
-	private static int S_ISVTX = 512; // Unix sticky bit mask
+	private static final ReentrantLock subtitleCacheLock = new ReentrantLock();
+	private static final Map<File, File[]> subtitleCache = new HashMap<>();
+	private static final int S_ISVTX = 512; // Unix sticky bit mask
 
 	// Signal an invalid parameter in getFileLocation() without raising an exception or returning null
 	private static final String DEFAULT_BASENAME = "NO_DEFAULT_BASENAME_SUPPLIED.conf";
@@ -244,13 +246,21 @@ public class FileUtil {
 		return f.substring(0, point);
 	}
 
-	private static String edition;
-	private static String formattedName;
+	private static final class FormattedNameAndEdition {
+		public String formattedName;
+		public String edition;
+
+		public FormattedNameAndEdition(String formattedName, String edition) {
+			this.formattedName = formattedName;
+			this.edition = edition;
+		}
+	}
 
 	/**
 	 * Remove and save edition information to be added later
 	 */
-	private static void removeAndSaveEditionToBeAddedLater() {
+	private static FormattedNameAndEdition removeAndSaveEditionToBeAddedLater(String formattedName) {
+		String edition = null;
 		Matcher m = COMMON_FILE_EDITIONS_PATTERN.matcher(formattedName);
 		if (m.find()) {
 			edition = m.group().replaceAll("\\.", " ");
@@ -258,35 +268,35 @@ public class FileUtil {
 			formattedName = formattedName.replaceAll(" - " + COMMON_FILE_EDITIONS, "");
 			formattedName = formattedName.replaceAll(COMMON_FILE_EDITIONS, "");
 		}
+		return new FormattedNameAndEdition(formattedName, edition);
 	}
 
 	/**
 	 * Capitalize the first letter of each word if the string contains no capital letters
 	 *
-	 * @param loopedOnce
 	 */
-	private static void convertFormattedNameToTitleCase(boolean loopedOnce) {
+	private static String convertFormattedNameToTitleCaseParts(String formattedName) {
 		if (formattedName.equals(formattedName.toLowerCase())) {
-			String formattedNameTemp = "";
+			StringBuilder formattedNameBuilder = new StringBuilder();
 			for (String part : formattedName.split(" - ")) {
-				if (loopedOnce) {
-					formattedNameTemp += " - " + convertLowerCaseStringToTitleCase(part);
-				} else {
-					formattedNameTemp += convertLowerCaseStringToTitleCase(part);
+				if (formattedNameBuilder.length() > 0) {
+					formattedNameBuilder.append(" - ");
 				}
-				loopedOnce = true;
+				formattedNameBuilder.append(convertLowerCaseStringToTitleCase(part));
 			}
-			formattedName = formattedNameTemp;
+			formattedName = formattedNameBuilder.toString();
 		}
+		return formattedName;
 	}
 
 	/**
 	 * Capitalize the first letter of each word if the string contains no capital letters
 	 */
-	private static void convertFormattedNameToTitleCase() {
+	private static String convertFormattedNameToTitleCase(String formattedName) {
 		if (formattedName.equals(formattedName.toLowerCase())) {
 			formattedName = convertLowerCaseStringToTitleCase(formattedName);
 		}
+		return formattedName;
 	}
 
 	/**
@@ -294,7 +304,7 @@ public class FileUtil {
 	 *
 	 * @param fileNameWithoutExtension
 	 */
-	private static void removeGroupNameFromBeginning(String fileNameWithoutExtension) {
+	private static String removeGroupNameFromBeginning(String formattedName, String fileNameWithoutExtension) {
 		if (!"".equals(formattedName)) {
 			if (formattedName.substring(0, 1).matches("\\[")) {
 				int closingBracketIndex = formattedName.indexOf(']');
@@ -309,14 +319,16 @@ public class FileUtil {
 		} else {
 			formattedName = fileNameWithoutExtension;
 		}
+		return formattedName;
 	}
 
 	/**
 	 * Remove stuff at the end of the filename like release group, quality, source, etc.
 	 */
-	private static void removeFilenameEndMetadata() {
+	private static String removeFilenameEndMetadata(String formattedName) {
 		formattedName = formattedName.replaceAll(COMMON_FILE_ENDS_CASE_SENSITIVE, "");
 		formattedName = formattedName.replaceAll("(?i)" + COMMON_FILE_ENDS, "");
+		return formattedName;
 	}
 
 	/**
@@ -338,7 +350,7 @@ public class FileUtil {
 	 */
 	private static final String COMMON_FILE_EDITIONS = "(?i)(?!\\()(Special[\\s\\.]Edition|Unrated|Final[\\s\\.]Cut|Remastered|Extended[\\s\\.]Cut|IMAX[\\s\\.]Edition|Uncensored|Directors[\\s\\.]Cut|Uncut)(?!\\))";
 	private static final Pattern COMMON_FILE_EDITIONS_PATTERN = Pattern.compile(COMMON_FILE_EDITIONS);
-	
+
 	/**
 	 * Returns the filename after being "prettified", which involves
 	 * attempting to strip away certain things like information about the
@@ -353,11 +365,10 @@ public class FileUtil {
 	 */
 	public static String getFileNamePrettified(String f, File file) {
 		String fileNameWithoutExtension;
-		formattedName = "";
+		String formattedName = "";
 		String formattedNameTemp;
 		String searchFormattedName;
-		edition = "";
-		boolean loopedOnce = false;
+		String edition = "";
 
 		// These are false unless we recognize that we could use some info on the video from IMDb
 		boolean isEpisodeToLookup = false;
@@ -375,7 +386,11 @@ public class FileUtil {
 			// Then strip the end of the episode if it does not have the episode name in the title
 			formattedName = formattedName.replaceAll("(?i)[\\s\\.]S0(\\d)E(\\d)(\\d)([eE]|-[eE])(\\d)(\\d)(" + COMMON_FILE_ENDS + ")", " - $1$2$3-$5$6");
 			formattedName = formattedName.replaceAll("[\\s\\.]S0(\\d)E(\\d)(\\d)([eE]|-[eE])(\\d)(\\d)(" + COMMON_FILE_ENDS_CASE_SENSITIVE + ")", " - $1$2$3-$5$6");
-			removeAndSaveEditionToBeAddedLater();
+			FormattedNameAndEdition result = removeAndSaveEditionToBeAddedLater(formattedName);
+			formattedName = result.formattedName;
+			if (result.edition != null) {
+				edition = result.edition;
+			}
 
 			// If it matches this then it didn't match the previous one, which means there is probably an episode title in the filename
 			formattedNameTemp = formattedName.replaceAll("(?i)[\\s\\.]S0(\\d)E(\\d)(\\d)([eE]|-[eE])(\\d)(\\d)[\\s\\.]", " - $1$2$3-$5$6 - ");
@@ -384,19 +399,23 @@ public class FileUtil {
 			}
 
 			formattedName = formattedNameTemp;
-			removeFilenameEndMetadata();
+			formattedName = removeFilenameEndMetadata(formattedName);
 
 			// Replace periods with spaces
 			formattedName = formattedName.replaceAll("\\.", " ");
 
-			convertFormattedNameToTitleCase(loopedOnce);
+			formattedName = convertFormattedNameToTitleCaseParts(formattedName);
 		} else if (formattedName.matches(".*[sS][1-9]\\d[eE]\\d\\d([eE]|-[eE])\\d\\d.*")) {
 			// This matches scene and most p2p TV episodes after their first 9 seasons that are more than one episode
 
 			// Rename the season/episode numbers. For example, "S11E01" changes to " - 1101"
 			formattedName = formattedName.replaceAll("(?i)[\\s\\.]S([1-9]\\d)E(\\d)(\\d)([eE]|-[eE])(\\d)(\\d)(" + COMMON_FILE_ENDS + ")", " - $1$2$3-$5$6");
 			formattedName = formattedName.replaceAll("[\\s\\.]S([1-9]\\d)E(\\d)(\\d)([eE]|-[eE])(\\d)(\\d)(" + COMMON_FILE_ENDS_CASE_SENSITIVE + ")", " - $1$2$3-$5$6");
-			removeAndSaveEditionToBeAddedLater();
+			FormattedNameAndEdition result = removeAndSaveEditionToBeAddedLater(formattedName);
+			formattedName = result.formattedName;
+			if (result.edition != null) {
+				edition = result.edition;
+			}
 
 			// If it matches this then it didn't match the previous one, which means there is probably an episode title in the filename
 			formattedNameTemp = formattedName.replaceAll("(?i)[\\s\\.]S([1-9]\\d)E(\\d)(\\d)([eE]|-[eE])(\\d)(\\d)[\\s\\.]", " - $1$2$3-$5$6 - ");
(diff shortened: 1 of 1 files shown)


Nadahar:
There were some use of unprotected static class variables here. That is "single threaded" mindset, and something we have to stop doing. While storing things between methods in class variables can be convenient, it's very error prone if multiple threads invoke any of the involved methods even without corruption due to unsynchronized access. I've replaced those that didn't have to be class variables with locals, and protected those that had to be class global.

---

Review status: 0 of 1 files reviewed at latest revision, 8 unresolved discussions.

---

*[src/main/java/net/pms/util/FileUtil.java, line 248 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813#-K9E6QK9Bd5xRPDqOkan:-K9E6QK9Bd5xRPDqOkao:1635805869) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/280eb4be029a7e7687fb6adf2f5b48b1c47ea771/src/main/java/net/pms/util/FileUtil.java#L248)):*
These are a big no-no, and there were no need for them either.

---

*[src/main/java/net/pms/util/FileUtil.java, line 253 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813#-K9E6fBc7ggaV90HdglC:-K9E6fBc7ggaV90HdglD:-1186322174) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/280eb4be029a7e7687fb6adf2f5b48b1c47ea771/src/main/java/net/pms/util/FileUtil.java#L253)):*
Generally void methods should ring a bell, for them to do anything useful they have to work on some shared object. It static as well, the alarm should really go off it no protection is in place. I've made all these return their results instead of storing them in shared variables.

---

*[src/main/java/net/pms/util/FileUtil.java, line 266 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813#-K9E8Q2A3vdfhh7ukIM2:-K9E8Q2A3vdfhh7ukIM3:-2013137891) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/280eb4be029a7e7687fb6adf2f5b48b1c47ea771/src/main/java/net/pms/util/FileUtil.java#L266)):*
@valib I removed loopedOnce as I couldn't see that it was used for anything. It was always false when any of these methods were invoked, and primitives doesn't return their state so if the intention was that changes to this should change the same named variable in the calling method, it won't work. If that is needed for this to work, this must be further rewritten.

---

*[src/main/java/net/pms/util/FileUtil.java, line 268 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813#-K9E7TizDVIHq91Q0kWQ:-K9E7TizDVIHq91Q0kWR:-97065336) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/280eb4be029a7e7687fb6adf2f5b48b1c47ea771/src/main/java/net/pms/util/FileUtil.java#L268)):*
@valib I can't see what's the difference in result between this once and the overloaded version of ```convertFormattedNameToTitleCase()```. I had to rename this one to ```Parts``` because after removing the inactive loopedOnce variable there was nothing separating the two. 

I don't understand why it's necessary to split them on ```-``` and capitalize them separatly - is that really necessary?

---

*[src/main/java/net/pms/util/FileUtil.java, line 275 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813#-K9E86B4gfKgiJA_0QVX:-K9E86B4gfKgiJA_0QVY:-229269423) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/280eb4be029a7e7687fb6adf2f5b48b1c47ea771/src/main/java/net/pms/util/FileUtil.java#L275)):*
Looping and modifying a String is very inefficient, using a StringBuilder instead.

---

*[src/main/java/net/pms/util/FileUtil.java, line 720 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813#-K9EA6Oa9Ad1g7vXpNJ0:-K9EA6ObGYao9-ZsLKPP:1643606801) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/280eb4be029a7e7687fb6adf2f5b48b1c47ea771/src/main/java/net/pms/util/FileUtil.java#L720)):*
This code must have been flawed. What's the point of having a cache if it's cleared as soon as one call where ```useCache == false``` is made. The whole method was synchronized, which I removed, since only a small part actually needed to be protected. There's no need to hold a lock longer than necessary. 

The implementation is still very inefficient (thus my comment) as it will lock the whole cache while browsing one folder. It should be locked on a per-folder basis so multiple folders could be populated at once and then inserted into the cache, but that requires more rewriting than I was up to now - and I'm unsure when/how often this is called, although I assume it's whenever a renderer browses a folder, which will make multiple renderers have to wait for eachother.

---

*[src/main/java/net/pms/util/FileUtil.java, line 1193 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813#-K9EBeubLiJObu5VsJtc:-K9EBeubLiJObu5VsJtd:-1051010857) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/280eb4be029a7e7687fb6adf2f5b48b1c47ea771/src/main/java/net/pms/util/FileUtil.java#L1193)):*
This was the original bug I was fixing, it assumes to much about the string (it seems to have been written only thinking about the case at hand, not how it would work in general). It would crash out in a ```StringIndexOutOfBoundsException``` whenever a file or folder name ended with a ```]```, and it could also remove much more than intended depending on where the square braces was placed. I've redone it using regex instead, imposing stricter rules for when it will act. I don't know how long these anime "tags" can be, but I've currently restricted the length between the square braces to 10 characters (that has to be at the start of the String) for this to act.

---

*[src/main/java/net/pms/util/FileUtil.java, line 1358 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813#-K9EClGGWlbfEW1WGa-s:-K9EClGHQUOusw_M6zr4:1646114412) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/a1728a5fe5d8b6d2daeffcbffbbfa58fd9dfc3a8/src/main/java/net/pms/util/FileUtil.java#L1358)):*
Fixed a missing stream close.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813#-:-K9E5OGgG3xJEVweIPTI:-183707127)*

Nadahar:
@SubJunk It's a shame you tagged without this one, the bug is on the forum and is "fixed" there, so people will expect it to be fixed in the next version. I've seen this error in many logfiles, but never knew what it was.

SubJunk:
Review status: 0 of 1 files reviewed at latest revision, 8 unresolved discussions.

---

*[src/main/java/net/pms/util/FileUtil.java, line 268 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813#-K9E7TizDVIHq91Q0kWQ:-K9Gf-5TtqKMbkBW147l:2023745848) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/280eb4be029a7e7687fb6adf2f5b48b1c47ea771/src/main/java/net/pms/util/FileUtil.java#L268)):*
I probably should've written comments for that part of the code, since I don't know why it does it at a glance. I do remember having a good reason for that and the loopedOnce variable.

---

*[src/main/java/net/pms/util/FileUtil.java, line 1193 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813#-K9EBeubLiJObu5VsJtc:-K9GdrqrMm_Yf5Ruzyh_:741476736) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/280eb4be029a7e7687fb6adf2f5b48b1c47ea771/src/main/java/net/pms/util/FileUtil.java#L1193)):*
10 characters won't be enough for popular release groups. 20 would be safer.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813)*

valib:
*[src/main/java/net/pms/util/FileUtil.java, line 266 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813#-K9E8Q2A3vdfhh7ukIM2:-K9GsmRqTI03FL9bQbyW:-985468683) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/280eb4be029a7e7687fb6adf2f5b48b1c47ea771/src/main/java/net/pms/util/FileUtil.java#L266)):*
@Nadahar I only a little bit rewrote the original code so and the loopedOnce was there.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813)*

valib:
Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, 8 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813)*

Nadahar:
Review status: 0 of 1 files reviewed at latest revision, 8 unresolved discussions.

---

*[src/main/java/net/pms/util/FileUtil.java, line 266 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813#-K9E8Q2A3vdfhh7ukIM2:-K9HLlaWCqupykGuQHz7:1852856968) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/280eb4be029a7e7687fb6adf2f5b48b1c47ea771/src/main/java/net/pms/util/FileUtil.java#L266)):*
Yes, but my guess is that in the original code loopedOnce's value was remembered between each run - then it would make some sense. The only way to figure this out properly is to go back in git history and look at earlier code.

---

*[src/main/java/net/pms/util/FileUtil.java, line 268 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813#-K9E7TizDVIHq91Q0kWQ:-K9HMI3b4E3sQYHoH9U-:1485437710) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/280eb4be029a7e7687fb6adf2f5b48b1c47ea771/src/main/java/net/pms/util/FileUtil.java#L268)):*
Yes, but the question is if that still applies the way the code was now. I suspect that the loopedOnce function has since been lost, for it to have any meaning it must have been retained between runs.

Splitting on ```-``` is very hard for me to see how could have any effect the way the code works now. The method that's called on the separate strings (or the total string in the other variant) splits on space and then loops and replaces the first letter of each word with a capital letter except for certain words. It's very hard to see how the results of this can possible differ when treating a long string versus several shorter ones.

To get to the bottom of this I think the original code must be studied, and the changes done since that must be identified and taken into consideration.

---

*[src/main/java/net/pms/util/FileUtil.java, line 1193 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813#-K9EBeubLiJObu5VsJtc:-K9HOLoPRjDyR7e8v8l4:5174689) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/280eb4be029a7e7687fb6adf2f5b48b1c47ea771/src/main/java/net/pms/util/FileUtil.java#L1193)):*
ok

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813)*

SubJunk:
Review status: 0 of 1 files reviewed at latest revision, 7 unresolved discussions.

---

*[src/main/java/net/pms/util/FileUtil.java, line 268 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813#-K9E7TizDVIHq91Q0kWQ:-K9Kf4-XNRbJLQUybqaG:-817946757) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/280eb4be029a7e7687fb6adf2f5b48b1c47ea771/src/main/java/net/pms/util/FileUtil.java#L268)):*
It looks like it was just a different way of doing exactly what you've rewritten - making sure that the " - " is added to all iterations except the first one. Doing the length check achieves the same thing so this is fine for reformatting.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813)*

Nadahar:
Review status: 0 of 1 files reviewed at latest revision, 7 unresolved discussions.

---

*[src/main/java/net/pms/util/FileUtil.java, line 268 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813#-K9E7TizDVIHq91Q0kWQ:-K9KlhAo4efGmDAhnFh2:558531262) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/280eb4be029a7e7687fb6adf2f5b48b1c47ea771/src/main/java/net/pms/util/FileUtil.java#L268)):*
Yes, that part is obvious to me. The part I didn't understand was that the variable was passed as an argument to the method, and was always false. A logical way to do that would have been to use a local variable that was initialized to false. Because of the way it was done, it seems to me that it was somehow supposed to "remember" if it was true from one method call to another, but since this is a primitive (not a reference), that won't work and the value will never be "returned" to the caller.

What I'm afraid of is that this was the functionality at one stage, and that it has been "lost in refactoring". If not, the passing as argument was meaningless and this implementation should work well.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813)*

Nadahar:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/main/java/net/pms/util/FileUtil.java, line 1193 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813#-K9EBeubLiJObu5VsJtc:-K9LIoQULHNDdN4UonLn:1479510348) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/280eb4be029a7e7687fb6adf2f5b48b1c47ea771/src/main/java/net/pms/util/FileUtil.java#L1193)):*
I've changed it to 20 characters.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813)*

SubJunk:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/main/java/net/pms/util/FileUtil.java, line 268 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813#-K9E7TizDVIHq91Q0kWQ:-K9Lxv0JOghkYSBdJm2O:-552311755) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/280eb4be029a7e7687fb6adf2f5b48b1c47ea771/src/main/java/net/pms/util/FileUtil.java#L268)):*
The loopedOnce variable doesn't need to be passed as a parameter to the method.

---

*[src/main/java/net/pms/util/FileUtil.java, line 1193 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813#-K9EBeubLiJObu5VsJtc:-K9LyENJH78osqHkJK-m:232232202) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/280eb4be029a7e7687fb6adf2f5b48b1c47ea771/src/main/java/net/pms/util/FileUtil.java#L1193)):*
Cool. I'll test it this week if I get a chance.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813)*

SubJunk:
Reviewed 1 of 1 files at r2.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813)*

valib:
Reviewed 1 of 1 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813)*

Nadahar:

---

*[src/main/java/net/pms/util/FileUtil.java, line 1193 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813#-K9EBeubLiJObu5VsJtc:-K9fciU8Gc-DQQ3c4zOG:610963288) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/280eb4be029a7e7687fb6adf2f5b48b1c47ea771/src/main/java/net/pms/util/FileUtil.java#L1193)):*
Did you have a change to test this?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813)*

SubJunk:
@Nadahar I just tested it and it seems to have changed the file ordering

SubJunk:

---

*[src/main/java/net/pms/util/FileUtil.java, line 1193 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813#-K9EBeubLiJObu5VsJtc:-K9jnNzIjYdhhyi_rWlk:-1758849584) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/280eb4be029a7e7687fb6adf2f5b48b1c47ea771/src/main/java/net/pms/util/FileUtil.java#L1193)):*
I just tested it and it seems to have broken file ordering for files that use underscores for spaces. This is the order of some files after this change:

[DeadFish]_Natsume_Yuujinchou_-_01 [BD][720p][MP4][AAC].mp4
The.100.S02E02.1080p.BluRay.x264-SHORTBREHD.mkv
Banshee.S03E08.1080p.HDTV.x264-BATV.mkv

Whereas it should be the other way around - the correct order being "1, B, N" instead of "N, 1, B"

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813)*

Nadahar:

---

*[src/main/java/net/pms/util/FileUtil.java, line 1193 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813#-K9EBeubLiJObu5VsJtc:-K9mFXYx_tgXLrD7EM4u:-1078990417) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/280eb4be029a7e7687fb6adf2f5b48b1c47ea771/src/main/java/net/pms/util/FileUtil.java#L1193)):*
@SubJunk I just tested in current master, and it lists them in the exact same order. What settings do you use (Prettifying and sorting)? Have you compared to master and found a regression in this PR?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813)*

Nadahar:

---

*[src/main/java/net/pms/util/FileUtil.java, line 1193 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813#-K9EBeubLiJObu5VsJtc:-K9mGudOjijGdDAsX91I:-1575311002) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/280eb4be029a7e7687fb6adf2f5b48b1c47ea771/src/main/java/net/pms/util/FileUtil.java#L1193)):*
When I turn on prettifying and set sorting to alphabetic, I get the exact same sort order in this PR and in master, "1, B, N". Are you sure there's actually a bug here?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813)*

Nadahar:
Review status: 0 of 1 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/main/java/net/pms/util/FileUtil.java, line 1243 \[r3\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813#-K9mNQTQpPCeUlTCDsE8:-K9mNQTQpPCeUlTCDsE9:1772976257) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/76aa35751a27e98c7a74052713fffeef29f18e46/src/main/java/net/pms/util/FileUtil.java#L1243)):*
I added underscores here, not that I can understand how this PR have affected that. I thought prettifying already had treated the filename before it came here, so I thought both dots and underscores were already gone at this stage..?

---

*[src/main/java/net/pms/util/FileUtil.java, line 1251 \[r3\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813#-K9mNgGv0BkuFGERlwdY:-K9mNgGv0BkuFGERlwdZ:602946448) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/76aa35751a27e98c7a74052713fffeef29f18e46/src/main/java/net/pms/util/FileUtil.java#L1251)):*
I fixed ignoring "the" while I were at it. The option says ignore "the", not ignore "the" in the beginning of words only. Also, it would only ignore "the" followed by space or dot as it were, not all "non word" character on either side will match. This also have unicode support, even though that's not relevant as this is relevant for English titles only.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813)*

SubJunk:
Reviewed 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[src/main/java/net/pms/util/FileUtil.java, line 1193 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813#-K9EBeubLiJObu5VsJtc:-K9nJ_llZxB1LRwCNgNB:-83140487) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/280eb4be029a7e7687fb6adf2f5b48b1c47ea771/src/main/java/net/pms/util/FileUtil.java#L1193)):*
Yes the issue isn't in master using the default sorting, which is alphanumeric. It's possible there is a bug in master with alphabetic sorting. It's working the same as master after 76aa357 with alphanumeric sorting.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813)*

Nadahar:
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[src/main/java/net/pms/util/FileUtil.java, line 1193 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813#-K9EBeubLiJObu5VsJtc:-K9nNTl1O5Uycx54QW4j:1384170043) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/280eb4be029a7e7687fb6adf2f5b48b1c47ea771/src/main/java/net/pms/util/FileUtil.java#L1193)):*
I'm not sure I follow you. Are you talking about with prettifying on or off? Where is the difference, with alphabetic or alphanumeric sort?

I need to find where the difference is so I can find the cause.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813)*

SubJunk:
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[src/main/java/net/pms/util/FileUtil.java, line 1193 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813#-K9EBeubLiJObu5VsJtc:-K9nssMroqcTCVSLbjXn:-1975280700) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/280eb4be029a7e7687fb6adf2f5b48b1c47ea771/src/main/java/net/pms/util/FileUtil.java#L1193)):*
The issue seems to be resolved after 76aa357

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813)*

valib:
Reviewed 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813)*

Nadahar:

---

*[src/main/java/net/pms/util/FileUtil.java, line 1193 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813#-K9EBeubLiJObu5VsJtc:-K9o13XTr8K8sabl94tG:-1295626228) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/280eb4be029a7e7687fb6adf2f5b48b1c47ea771/src/main/java/net/pms/util/FileUtil.java#L1193)):*
Ok, I see it now. It was only valid for anime folders with prettify on.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/813)*

SubJunk:
So this seems ready to merge now in my opinion - I don't notice any behavior change between this and master

Nadahar:
Great.

@UniversalMediaServer/developers Anyone else has any objections?

valib:
No objection.

PR #817, opened 2016-01-31
Fix not to be subtitles parsing twice

valib:
It needs testing but it works for me. It works when the MediaInfo is used but it needs some improvement when FFmpeg is used for the media parsing. Any help is appreciated. :smile:

[Review on Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/817)

### DIFF
--- a/src/main/java/net/pms/dlna/DLNAMediaInfo.java
+++ b/src/main/java/net/pms/dlna/DLNAMediaInfo.java
@@ -331,9 +331,8 @@ public class DLNAMediaInfo implements Cloneable {
 	}
 
 	/**
-	 * @return the number of subtitle tracks embedded in the media file.
+	 * @return true when there are subtitle tracks embedded in the media file.
 	 */
-
 	public boolean hasSubtitles() {
 		return subtitleTracks.size() > 0;
 	}
@@ -1402,8 +1401,17 @@ public class DLNAMediaInfo implements Cloneable {
 
 		// Check for external subs here
 		if (f.getFile() != null && type == Format.VIDEO && configuration.isAutoloadExternalSubtitles()) {
-			FileUtil.isSubtitlesExists(f.getFile(), this);
+			hasExternalSubs = FileUtil.isSubtitlesExists(f.getFile(), this);
 		}
+	}
+
+	private boolean hasExternalSubs;
+
+	/**
+	 * @return true when there are external subtitle tracks for the video file.
+	 */
+	public boolean hasExternalSubs() {
+		return hasExternalSubs;
 	}
 
 	/**
--- a/src/main/java/net/pms/dlna/DLNAMediaInfo.java
+++ b/src/main/java/net/pms/dlna/DLNAMediaInfo.java
@@ -1408,7 +1408,7 @@ public class DLNAMediaInfo implements Cloneable {
 	private boolean hasExternalSubs;
 
 	/**
-	 * @return true when there are external subtitle tracks for the video file.
+	 * @return true when there is an external subtitle file matching the renderer subtitles setting for the video file.
 	 */
 	public boolean hasExternalSubs() {
 		return hasExternalSubs;
--- a/src/main/java/net/pms/dlna/DLNAMediaSubtitle.java
+++ b/src/main/java/net/pms/dlna/DLNAMediaSubtitle.java
@@ -188,6 +188,7 @@ public class DLNAMediaSubtitle extends DLNAMediaLang implements Cloneable {
 					LOGGER.debug("Set detected charset \"{}\" and language \"{}\" for {}", match.getName(), lang, externalFile.getAbsolutePath());
 				} else {
 					subsCharacterSet = null;
+					LOGGER.debug("No charset detected for {}", externalFile.getAbsolutePath());
 				}
 
 			} catch (IOException ex) {
--- a/src/main/java/net/pms/dlna/DLNAMediaSubtitle.java
+++ b/src/main/java/net/pms/dlna/DLNAMediaSubtitle.java
@@ -185,12 +185,14 @@ public class DLNAMediaSubtitle extends DLNAMediaLang implements Cloneable {
 				if (match != null) {
 					subsCharacterSet = match.getName().toUpperCase(PMS.getLocale());
 					lang = match.getLanguage();
+					LOGGER.debug("Set detected charset \"{}\" and language \"{}\" for {}", match.getName(), lang, externalFile.getAbsolutePath());
 				} else {
 					subsCharacterSet = null;
 				}
+
 			} catch (IOException ex) {
 				subsCharacterSet = null;
-				LOGGER.warn("Exception during external file charset detection.", ex);
+				LOGGER.warn("Exception during external file charset detection: ", ex.getMessage());
 			}
 		}
 	}
--- a/src/main/java/net/pms/dlna/RealFile.java
+++ b/src/main/java/net/pms/dlna/RealFile.java
@@ -53,12 +53,8 @@ public class RealFile extends MapFile {
 	public boolean isValid() {
 		File file = this.getFile();
 		resolveFormat();
-		if (getType() == Format.VIDEO && file.exists() && configuration.isAutoloadExternalSubtitles() && file.getName().length() > 4) {
-			setHasExternalSubtitles(FileUtil.isSubtitlesExists(file, null));
-		}
-
+		boolean checkSubs = getType() == Format.VIDEO && configuration.isAutoloadExternalSubtitles() && file.getName().length() > 4; // TODO why should be file name longer than 4 chars?
 		boolean valid = file.exists() && (getFormat() != null || file.isDirectory());
-
 		if (valid && getParent().getDefaultRenderer() != null && getParent().getDefaultRenderer().isUseMediaInfo()) {
 			// we need to resolve the DLNA resource now
 			run();
@@ -83,6 +79,12 @@ public class RealFile extends MapFile {
 			if (getParent().getDefaultRenderer().isMediaInfoThumbnailGeneration()) {
 				checkThumbnail();
 			}
+
+			if (checkSubs) {
+				setHasExternalSubtitles(getMedia() != null && getMedia().hasExternalSubs());
+			}
+		} else if (checkSubs && valid && getParent().getDefaultRenderer() != null && !getParent().getDefaultRenderer().isUseMediaInfo()) {
+			setHasExternalSubtitles(FileUtil.isSubtitlesExists(file, null)); // TODO it is still parsing subs twice when FFmpeg is used for parsing the file
 		}
 
 		return valid;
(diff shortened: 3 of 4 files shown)


Nadahar:
It's hard for me to see exectly what you are doing without digging a lot into the code. It would be easier if you explain what the current situation is and what you are trying to do about it.

---

Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/main/java/net/pms/dlna/RealFile.java, line 56 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/817#-K9TOeXFevC6-c8sgef0:-K9TOeXGzRiaM43xb3EY:-1212773928) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/bc3a653787a82b68e2bad805a753b053e56d2be4/src/main/java/net/pms/dlna/RealFile.java#L56)):*
I don't know, but it's probably to make sure it has an extensions. ".sub" is 4 characters

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/817#-:-K9TP-GkdYyHmnXRlLCm:-1843415737)*

valib:
@nadahar the current code is parsing subtitles and sets the charset and language twice. It is first called from the RealFile when the ```FileUtil.isSubtitlesExists()``` is called and the second from LibMediaInfoParser when is called ```media.finalize(type, inputFile)```. This is the situation when the MediaInfo is used. When the MediaInfo is not used I couldn't find the reason why it is doing.

---







*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/817#-:-K9TtqC_KyOGyHXQIWJU:453955332)*

valib:
*[src/main/java/net/pms/dlna/RealFile.java, line 56 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/817#-K9TOeXFevC6-c8sgef0:-K9U-6cDLfgFRFQ-HSoK:-1038617727) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/bc3a653787a82b68e2bad805a753b053e56d2be4/src/main/java/net/pms/dlna/RealFile.java#L56)):*
I am not sure because this is about the video file name not subtitles. Maybe it has something to do with the web streamed videos. I don't know.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/817)*

Nadahar:
It's called many more places than that. If we're going to clean this up, it must be cleaned up properly. From what I can find, it's called from:

* ```DLNAMediaInfo.finalize(int, InputFile)```
* ```RarredEntry.isValid()```
* ```RealFile.isValid()```
* ```SevenZipEntry.isValid()```
* ```ZippedEntry.isValid()```
* ```FileUtil.doesSubtitlesExists(File, DLNAMediaInfo)```
* ```DLNAResource.getDlnaOrgPnFlags(RendererConfiguration, int)```
* ```Player.setSubtitleOutputParameters(String, DLNAMediaInfo, OutputParams)```

It all looks like a complete mess to me.

---

Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/main/java/net/pms/dlna/RealFile.java, line 56 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/817#-K9TOeXFevC6-c8sgef0:-K9U1AoAJR09QMUPFCZs:-537119470) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/bc3a653787a82b68e2bad805a753b053e56d2be4/src/main/java/net/pms/dlna/RealFile.java#L56)):*
Maybe it's from before we used MediaInfo? Then extenstions were used exclusively to determine file "format", and as such if could be a requirement that the videofile had an extension.

I investigated this, and that code is from the very first commit of PMS (079805f534da336ef84f4c5ba7ba998a8a1705b6). The rest looked completely different then, look [here] (https://github.com/UniversalMediaServer/UniversalMediaServer/blob/079805f534da336ef84f4c5ba7ba998a8a1705b6/ps3mediaserver/net/pms/dlna/RealFile.java#L41)

The code at that stage removed the last 4 characters and replaced them with ```.IFO```, ```.srt``` or ```.sub```. 

The test is not relevant any more and should be removed. This logic is now handled by ```FileUtil.browseFolderForSubtitles()``` which strips the video file extension and then tries supported extensions.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/817#-:-K9U9SMyfs6mbuvlCGgS:385784603)*

valib:
@Nadahar thanks for the investigation.

valib:
Reviewed 3 of 4 files at r1, 1 of 1 files at r2.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/817)*

valib:
@UniversalMediaServer/developers this PR solved the problem when the MediaInfo is used which is now the basic setting for the majority of renderers. I suggest to merge it and the rest of problems which @Nadahar mentioned we can investigate and solve in the future.

SubJunk:
No objections from me - as far as I can see this is an improvement on the existing code

Nadahar:
@valib This was a bad idea, please revert it.

I had made several comments in the code and was writing about it - but you merged it while I was writing.

Here is the biggest of my comments, but several others were lost:

I've looked at the code, and found a couple of problems that I've commented on. I also think merging partial solutions is a bad idea in general, as I've repeatedly said. I honestly don't know what the rush is to merge things before they are complete, and I'm not sure it's doing the product any favours. If things are going to be done "later" to fix this, why not do it before merge instead of merging something we know doesn't take properly care of a given issue?

I also have a more general problem with this approach. As I see it, the motivation for not fixing it "properly" is to not have to gain the necessary overview. Manipulating something that's not fully understood is a sure way to create bugs. Also, for every such change getting the full overview to properly take care of the issue is made harder. The bugs created by not having the overview will probably also have to be "patched" at one stage making the situation even less logical and more messy. After a while everything is so messed up that it's impossible to understand what happens and why, and not least what the intention of the code is.

I think we have plenty of such "traps" already, and only patience and hard work can untie these knots. I'm constantly trying to do just this various places in the code, and I hate to see more of the same being applied. I'm not in a situation to say that we have to stop doing this now, but I must say that it makes it much more difficult to find motivation for trying to figure things out and get them organized.

The reason why it's hard to get the overview as it is today is very likely because this mechanism has already gone through some iterations. The longer one continue fixing band-aids with new band-aids the harder it will be to fix it properly, and in the end it will become impossible without breaking a lot of things.

Nadahar:
I've looked at the code, and found a couple of problems that I've commented on. I also think merging partial solutions is a bad idea in general, as I've repeatedly said. I honestly don't know what the rush is to merge things before they are complete, and I'm not sure it's doing the product any favours. If things are going to be done "later" to fix this, why not do it before merge instead of merging something we know doesn't take properly care of a given issue?

I also have a more general problem with this approach. As I see it, the motivation for not fixing it "properly" is to not have to gain the necessary overview. Manipulating something that's not fully understood is a sure way to create bugs. Also, for every such change getting the full overview to properly take care of the issue is made harder. The bugs created by not having the overview will probably also have to be "patched" at one stage making the situation even less logical and more messy. After a while everything is so messed up that it's impossible to understand what happens and why, and not least what the intention of the code is.

I think we have plenty of such "traps" already, and only patience and hard work can untie these knots. I'm constantly trying to do just this various places in the code, and I hate to see more of the same being applied. I'm not in a situation to say that we have to stop doing this now, but I must say that it makes it much more difficult to find motivation for trying to figure things out and get them organized.

The reason why it's hard to get the overview as it is today is very likely because this mechanism has already gone through some iterations. The longer one continue fixing band-aids with new band-aids the harder it will be to fix it properly, and in the end it will become impossible without breaking a lot of things.

---

Review status: all files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[src/main/java/net/pms/dlna/DLNAMediaInfo.java, line 1409 \[r2\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/817#-K9lOnpd0XFyQLqI7Os7:-K9lOnpe6o11ZKe98VT3:-285576338) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/325d06ae53baaa49b7d7a8d18ddfa5e07261ddd9/src/main/java/net/pms/dlna/DLNAMediaInfo.java#L1409)):*
When introducing class global "holding variables" like this, multithreading must always be taken into consideration. Basicly, if it's not going to be protected it must be ensured that only one thread will ever access the value. It's very hard to see that that's the case here (two renderers could access the same DLNAMediaInfo object), and thus synchronization is needed. 

I seem to remember that the subtitle scan is being cached, so this lookup shouldn't be that expensive. I just dealt with all synchronization issues in this class, I'd hate for a new one to be introduced as soon as it's fixed.

---

*[src/main/java/net/pms/dlna/RealFile.java, line 56 \[r2\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/817#-K9lQk2YfTDhP0wz5wWP:-K9lQk2ZEETGox_O3HNJ:1920515218) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/325d06ae53baaa49b7d7a8d18ddfa5e07261ddd9/src/main/java/net/pms/dlna/RealFile.java#L56)):*
We should check getMedia.isVideo() instead. Format isn't reliable for this, as it will identify several audio files as video for several formats that have the same extension for both.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/817#-:-K9lRq6-yJhkotW5x0Ww:-745222075)*

Nadahar:
@valib You only waited 12 hours from posting this before merging, most of which I was sleeping, and as said I was answering this as you merged it. I really don't get the rush.

I don't remember the details in my other comments

Nadahar:
@valib Also, I'm not sure this is really needed. ```FileUtil.browseFolderForSubtitles()``` is cached, something I've fixed in #813. The cache solution was there, but someone had broken it by deleting the cache often making it probably quite ineffective.

I'm not saying that a cleanup of this isn't needed, but as soon as #813 is merged, the actual folders won't be double scanned anyway.

valib:
@Nadahar sorry that I was so quick. Unfortunately the problem is somewhere else than in ```FileUtil```.  It should be a bug in the ```DLNAResource``` or their subclasses. Your changes in #813 don't fix anything for this issue. I partially reverted  the commit because it at least reduces logspam.

Nadahar:
@valib I didn't mean that #813 made fewer calls to ```isSubtitlesExist()```, but if you look further in, you'll see that ```isSubtitlesExist()``` calls ```browseFolderForSubtitles()```. #813 fixes that so that the folders aren't actually scanned, but a cache is used if they have already been scanned. That cancels out most of the expense of calling ```isSubtitlesExist()```. As I said, I'm sure a cleanup of the calls/the call structure to ```isSubtitlesExist()``` would be good, but with #813 the actual disk folders aren't scanned twice.

PR #820, opened 2016-02-03
DLNAMediaInfo bugfixes

Nadahar:
I've fixed the bugs pointed out by FindBugs and Threadsafe plus a couple I found myself in ```DLNAMediaInfo```:
* Fixed many synchronization bugs in ```DLNAMediaInfo```
* Removed access to external mutable objects in ```DLNAMediaInfo```
* Some minor ```Exception```/logging fixes/tweaks in ```DLNAMediaInfo```

I'll give some comments in the code, please review.

[Review on Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/820)

### DIFF
--- a/src/main/java/net/pms/dlna/DLNAMediaInfo.java
+++ b/src/main/java/net/pms/dlna/DLNAMediaInfo.java
@@ -18,11 +18,10 @@
  */
 package net.pms.dlna;
 
-import java.awt.Color;
 import java.io.*;
 import java.nio.file.Files;
 import java.util.*;
-import net.pms.Messages;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
 import net.pms.PMS;
 import net.pms.configuration.PmsConfiguration;
 import net.pms.configuration.RendererConfiguration;
@@ -37,7 +36,6 @@ import net.pms.util.CoverUtil;
 import net.pms.util.FileUtil;
 import net.pms.util.FullyPlayed;
 import net.pms.util.MpegUtil;
-import net.pms.util.CoverArtArchiveUtil;
 import net.pms.util.ProcessUtil;
 import static net.pms.util.StringUtil.*;
 import net.pms.util.UMSUtils;
@@ -86,7 +84,8 @@ public class DLNAMediaInfo implements Cloneable {
 	 */
 	public static final long TRANS_SIZE = Long.MAX_VALUE - Integer.MAX_VALUE - 1;
 
-	private boolean h264_parsed;
+	private final Object videoWithinH264LevelLimitsLock = new Object();
+	private Boolean videoWithinH264LevelLimits = null;
 
 	// Stored in database
 	private Double durationSec;
@@ -139,11 +138,7 @@ public class DLNAMediaInfo implements Cloneable {
 	public String aspectRatioContainer;
 	public String aspectRatioVideoTrack;
 
-	/**
-	 * @deprecated Use standard getter and setter to access this variable.
-	 */
-	@Deprecated
-	public byte thumb[];
+	private byte thumb[];
 
 	/**
 	 * @deprecated Use standard getter and setter to access this variable.
@@ -157,8 +152,13 @@ public class DLNAMediaInfo implements Cloneable {
 	@Deprecated
 	public int bitsPerPixel;
 
+	private final ReentrantReadWriteLock referenceFrameCountLock = new ReentrantReadWriteLock();
 	private byte referenceFrameCount = -1;
+
+	private final ReentrantReadWriteLock avcLevelLock = new ReentrantReadWriteLock();
 	private String avcLevel = null;
+
+	private final Object h264ProfileLock = new Object();
 	private String h264Profile = null;
 
 	private List<DLNAMediaAudio> audioTracks = new ArrayList<>();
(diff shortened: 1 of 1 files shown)


Nadahar:
Review status: 0 of 1 files reviewed at latest revision, 6 unresolved discussions.

---

*[src/main/java/net/pms/dlna/DLNAMediaInfo.java, line 146 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/820#-K9caI43ufnisquKcIfn:-K9caI445q8vkM8bro8f:1169633290) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/ad629ef7aec7f7503389f8578084f47c7390e020/src/main/java/net/pms/dlna/DLNAMediaInfo.java#L146)):*
I've had to make a few objects private to prevent unsynchronized or externally mutable access. I've removed the deprecated note on these.

---

*[src/main/java/net/pms/dlna/DLNAMediaInfo.java, line 155 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/820#-K9cabaOCUta04KbDE4y:-K9cabaOCUta04KbDE4z:-266650063) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/d7bc69e7489bca76efb50246d38172edb04f0930/src/main/java/net/pms/dlna/DLNAMediaInfo.java#L155)):*
I've replaced synchronization on the class instance itself (this is what happens when ```synchronized``` is declared in the method declaration) with explicit synchronization objects. While it's very easy to simply synchronize on the class instance itself, it ends up with a very bad liveliness because it will hold locks much longer than necessary and worse it will lock when it doesn't need to (it will lock access to variable B while variable A is being accessed). It can also easily lead to deadlocks when two or more objects call eachother for no other reason than "lazy synchronization".

I've used ```Object``` for synchronization when there are only a few reads and writes, or where access seems balanced (both read and write are locked if either is performed). I've used ReentrantReadWriteLock for objects that are normally read. When separating between reads and writes, concurrent reads are allowed as no long as no write lock is held.

---

*[src/main/java/net/pms/dlna/DLNAMediaInfo.java, line 758 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/820#-K9cckRE0mjvLXlSooAn:-K9cckRE0mjvLXlSooAo:-1711588705) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/d7bc69e7489bca76efb50246d38172edb04f0930/src/main/java/net/pms/dlna/DLNAMediaInfo.java#L758)):*
There's no reason to abort the whole parsing if a NumberFormatException occurs.

---

*[src/main/java/net/pms/dlna/DLNAMediaInfo.java, line 1433 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/820#-K9cd4OKnSoe_wTTFmMP:-K9cd4OLnVPbsdHR1_jF:-1204540593) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/d7bc69e7489bca76efb50246d38172edb04f0930/src/main/java/net/pms/dlna/DLNAMediaInfo.java#L1433)):*
There seemed to be a major bug here. This method kept track of whether it had been run before (```h264_parsed```) and only run once for any given instance of ```DLNAMediaInfo```, but it didn't keep the result. If it were queried a second time it always returned true.

I don't know if it's actually called several times, but I wouldn't be surprised, but in any case this can't be the intended behaviour. The result from the first check is now kept and simply returned on subsequent queries.

---

*[src/main/java/net/pms/dlna/DLNAMediaInfo.java, line 1776 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/820#-K9cdkh8hiD6rVfESq17:-K9cdkh8hiD6rVfESq18:548272765) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/d7bc69e7489bca76efb50246d38172edb04f0930/src/main/java/net/pms/dlna/DLNAMediaInfo.java#L1776)):*
These "failsafe" threads are all unsynchronized time bombs just waiting to create bugs. The state handling when a timout occurs is also questionable (they just abort the task without setting some kind of results), but I didn't do anything with that.

---

*[src/main/java/net/pms/dlna/DLNAMediaInfo.java, line 2088 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/820#-K9ceNPBYSC0GC-XHSPC:-K9ceNPCxhE6Zf4ddnws:-1842188680) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/d7bc69e7489bca76efb50246d38172edb04f0930/src/main/java/net/pms/dlna/DLNAMediaInfo.java#L2088)):*
thumb (being an array) was externally mutable, so it had to be made private and copies made when setting and getting.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/820)*

SubJunk:
Reviewed 1 of 1 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/820)*

SubJunk:
I'm fairly ignorant about this type of thing but reviewed it to the best of my ability and it looks good. I'm okay to merge if everyone else is.

valib:
Reviewed 1 of 1 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/820)*

PR #828, opened 2016-02-07
Support external Blu-ray subtitles (.sup)

SubJunk:
MEncoder and FFmpeg don't seem to support transcoding this format when it's external (they both do it when it's internal, though) so we should convert the PGS to a supported format like VOBSUB before we pass it to the transcoder. It should be very fast since no extraction is needed.

[Review on Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/828)


Sami32:
@SubJunk Cool ;-)
Until FFmpeg support, i discover some ways to play external PGS (.sup)
http://www.videohelp.com/software/XySubFilter/version-history#changelog
http://www.zachsaw.com/?pg=suptitle_pgs_avisynth_plugin

Nadahar:
I'd really wish there was another way. Converting them to vobsub removes a lot of quality and can in some cases make them almost unreadable. I thought FFmpeg had a SUP reader..?

Sami32:
Yes, but it's a little weaky...
Since i saw these last few days that they worked on subtitles, i should post a SUP issue that i never take time to report...

Issue #831, opened 2016-02-11
opensubtitles.org abuse


Nadahar:
After looking through some code, I discovered that we're abusing opensubtitles.org in about as many ways as possible. First of all, we're using a user agent specifically reserved for TESTING only, we're calling it IMDb when we're actually using opensubtitles and we're generally not crediting them in any way.

Here are the [rules for using opensubtitles.org] (http://trac.opensubtitles.org/projects/opensubtitles/wiki/DevReadFirst). I'd say we should stop using it until these things are fixed.

SubJunk:
Good find. I've emailed them to get a user agent and will make the changes necessary to give credit to them.

SubJunk:
Start this at https://github.com/UniversalMediaServer/UniversalMediaServer/pull/832 and I added links to opensubtitles.org on our About and Comparison pages

Nadahar:
:+1:

PR #832, opened 2016-02-12
Give credit to OpenSubtitles (#831)

SubJunk:
[Review on Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/832)


SubJunk:
I merged this but if it's not enough we can add more credit

Nadahar:
This looks sufficient to me, but I think they wanted a little something on the web page as well.

SubJunk:
If it's important that it's a secret we can use our domain as a proxy for the requests but that should be a last resort since it's inefficient and slower

Nadahar:
@SubJunk That doesn't really help. Anyone that can read the source can also mimic the call made to the domain (or just trace the code and read the value in the debugger). 

However you twist and turn it, you can't make a secret available for an open source software that whoever reads the code can't get. You have no place to "hide" the key needed to release the secret however you put it.

This is a general challenge regarding all such matters, and any solution would surprise me. The question is if it's any point doing easily reversable things like BASE64, encryption with a known key, downloading it etc. It's really just an insult to anyone trying to find it.

Nadahar:
@SubJunk OpenSubtitles do support user login though and do encourage using it. As long as we only use a user agent each IP will be limited to 200 request per day. Registered doners however has no such restriction, so giving the possibility to enter user credentials would be the right thing to do. For normal UMS users 200 request should be no problem, especially since we cache the results, but for users that have to use a proxy or VPN for some reason, it can be a big problem. Anyone using a given proxy/VPN server has the same IP address, and thus such addresses tend to be "saturated" on any per IP solution.

SubJunk:
@Nadahar we can easily make the user agent added via our proxy so that users are never exposed to it

Nadahar:
@SubJunk Yes, but what does it accomplish?

SubJunk:
@Nadahar why would it be easy for anyone to get that way?

Nadahar:
@SubJunk Because whatever you do, UMS must be able to "decode" it back to the string it's supposed to be when sending it during the call to opensubtitles. Since this is open source, anyone that reads the code (it's only revealed to those reading the code anyway) can either:
* Look in the code how to "decode" it 
* Put a breakpoint somewhere in the code after decoding and read the value from the debugger

Nadahar:
@SubJunk Sorry, I didn't get the proxy transform bit. Yes, that would work for keeping it secret, but that would require any opensubtitle traffic to go via the proxy and would quickly saturate the number of requests for the proxy IP address. That would mean that all UMS users combined would be limited to 200 requests a day.

SubJunk:
Yeah that's true. I guess we can just see how it goes as it is. I wonder if any similar services have higher limits. Most people won't be using proxies anyway so it should be okay.

Nadahar:
I think we're talking slightly passed eachother. A lot of people are using VPN these days for various reasons, as more and more laws are being passed in favor of big business more and more people have to protect their identity to remain free.

Just to be sure that everything is clear: OpenSubtitles accepts both user credentials and a user agent. The user agent is to indentify the application, this is what we have to put in our code somewhere and that I was talking about it being impossible to keep secret (unless all calls are to go through some proxy where we inject the actual code). Not that I think that's a big deal.

But, in addition there's the user credentials. If you leave them blank or just send a regular registered OpenSubtitle users credentials the combination of that user agent and your public IP has a 200/day limit. But, if the user either is a donor or a VIP at OpenSubtitles, there are other rules. Our user agent should be the same in any case (it's there to identify the application), but if we have users the possibility to specify credentials as well (maybe via our existing credentials system) and they had an "elevated" user account at OpenSubtitles, that limit wouldn't apply. That means that a user going on the internet via VPN or proxy that is either a donor or a VIP can have their own number of lookups despite being on a "shared IP". Get it?

SubJunk:
I think the number of users who have credentials for that site will be so low that it's not worth considering that as a convenience overall. It would be nice to add support for the credentials for the few users who have it but if we have enough users using VPNs then it's a real issue and we should look to use another service. I have no idea what percentage of our users use VPNs but they are certainly the demographic that would be most likely to.

Nadahar:
For looking up information there's other services, but for live subtitles I'm not sure there are so many options. Isn't adding the possibility for credentials very easy if we use the existing credentials system? Just define some "name" that applies to opensubtitles and check if a username and password is entered for that "name"..?

SubJunk:
Yes, it's possible but I don't think it's very useful since I don't think it will be common for users to have those accounts

ExSport:
I am not sure if opensubtitles is right provider if it is limited to 200 hits daily. 
I suppose general user will make more hits daily. Every db re-initialization, scanning library, miss hits, all is counted. 
I have maybe one folder with more than 200 documents in folder which will overcome the limit by only browsing the folder once. Caching? I think most of them will not be cached as nothing will be found but I suppose after some time  UMS will try it again. Again without success but again it will block opensubs for whole day.
Not sure how it works so maybe also my folders with hundreds of files from video camera will be retrieved=not-cached miss-hits? I know parsing imdb directly is less efficient and slower...but without limits. 
I like the idea how MovieInfo plugin does it. It must parse whole page but it works 😊
Also using Google search seems has better success than searching imdb directly. I also use it for csfd.cz site and using Google with search string "movie name year site:m.imdb.com/title/" has almost 100% hit ratio.
Nothing is perfect so I know that any change on site will break page parsing😇
But the limit of 200 seems to be low for me.  Also is UMS understanding these limits so over request is not taken as movie name not found so next time it should be retrieved again. 
As I said I know nothing about how this feature works ☺
Only my two cents to the pool...

Nadahar:
I did a lot on the MovieInfo IMDB and Google lookup in https://github.com/SharkHunter/MovieInfo/commit/42d2e7a58118df99784400c59002e244c4529cf4. If I don't recall it completely wrong, IMDB search wasn't too successful and Google must was the way. But, Google also has protection in place and quickly puts up a captcha if too many searches comes too fast from the same IP. It also analyzes the type of search for signs of automated searches.

The bottom line is that neither Google nor IMDB allows such use, thus what MovieInfo does must be considered abuse. In addition we can expect that they will keep coming up with evolved methods of protection against such use in the future.

There are issues with using OpenSubtitles too, especially that they "forbid" using automated lookups and demands that the user must make an action for each lookup. Ideally OpenSubtitles should be used for live subtitles only and not for general movie info imo. The problem is what source to use for information. themoviedb.org might be the best solution, but I haven't looked into the details there.

PR #833, opened 2016-02-13
Misc prettifying related bug fixes

Nadahar:
This isn't a complete bugfix in any way, I did some testing to find why so few of my sample videos were looked up, and found some bugs on the way. This PR is these bugfixes. The one in ```UPNPControl``` isn't related to prettifying, but it's here as I discovered it while debugging this.

None of these bugs makes a big difference on the "hit rate" although a few more is looked up. I found that the biggest problem is that the search term is not properly cleaned up/prettified before the search is being done, so many searches have no hits. Any file found by hash will work though. As a side note the hash value should be kept instead of being calculated each time a lookup is done.

I started to look into opensubtitles itself to look for some kind of "fuzzy" search, but didn't continue on once I discovered #831. I'm posting the bugfixes already made though as it must be better to get them in than not even if they doesn't fix the main problem.

[Review on Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/833)

### DIFF
--- a/src/main/java/net/pms/network/UPNPControl.java
+++ b/src/main/java/net/pms/network/UPNPControl.java
@@ -379,7 +379,7 @@ public class UPNPControl {
 	}
 
 	public static String getFriendlyName(Device d) {
-		return d.getDetails().getFriendlyName();
+		return d != null ? d.getDetails().getFriendlyName() : null;
 	}
 
 	public static String getUUID(Device d) {
--- a/src/main/java/net/pms/util/FileUtil.java
+++ b/src/main/java/net/pms/util/FileUtil.java
@@ -536,6 +536,7 @@ public class FileUtil {
 
 			// Rename the year. For example, "2013" changes to " (2013)"
 			formattedName = formattedName.replaceAll("(?i)\\[(19|20)(\\d\\d)\\].*", " ($1$2)");
+			formattedName = removeFilenameEndMetadata(formattedName);
 
 			// Replace periods with spaces
 			formattedName = formattedName.replaceAll("\\.", " ");
@@ -545,6 +546,10 @@ public class FileUtil {
 			// This matches rarer types of movies
 			isMovieToLookup = true;
 			formattedName = removeFilenameEndMetadata(formattedName);
+
+			// Replace periods with spaces
+			formattedName = formattedName.replaceAll("\\.", " ");
+
 			formattedName = convertFormattedNameToTitleCase(formattedName);
 		} else if (formattedName.matches(COMMON_FILE_ENDS_MATCH)) {
 			// This is probably a movie that doesn't specify a year
@@ -598,7 +603,7 @@ public class FileUtil {
 		}
 
 		// Remove extra spaces
-		formattedName = formattedName.replaceAll("  ", " ");
+		formattedName = formattedName.replaceAll("\\s+", " ");
 
 		/**
 		 * Add info from IMDb
@@ -613,7 +618,7 @@ public class FileUtil {
 		if (file != null && (isTVSeriesToLookup || isMovieToLookup)) {
 			InfoDb.InfoDbData info = PMS.get().infoDb().get(file);
 			if (info == null) {
-				PMS.get().infoDbAdd(file, searchFormattedName);
+				PMS.get().infoDbAdd(file, StringUtil.hasValue(searchFormattedName) ?  searchFormattedName : formattedName);
 			} else if (isTVSeriesToLookup) {
 				int showNameIndex = indexOf(Pattern.compile("(?i) - \\d\\d\\d.*"), formattedName);
 				if (StringUtils.isNotEmpty(info.title) && showNameIndex != -1) {
@@ -640,7 +645,7 @@ public class FileUtil {
 					}
 					LOGGER.trace("The similarity between '" + info.title + "' and '" + formattedName + "' is " + similarity);
 				} else {
-					int yearIndex = indexOf(Pattern.compile("\\s\\(\\d\\d\\d\\d\\)"), formattedName);
+					int yearIndex = indexOf(Pattern.compile("\\s\\(\\d{4}\\)"), formattedName);
 					String titleFromFilename = formattedName.substring(0, yearIndex);
 					double similarity = org.apache.commons.lang3.StringUtils.getJaroWinklerDistance(titleFromFilename, info.title);
 					if (similarity > 0.9) {
@@ -705,7 +710,7 @@ public class FileUtil {
 
 		return convertedValue;
 	}
-	
+
 	public static int indexOf(Pattern pattern, String s) {
 		Matcher matcher = pattern.matcher(s);
 		return matcher.find() ? matcher.start() : -1;
--- a/src/main/java/net/pms/util/ImdbUtil.java
+++ b/src/main/java/net/pms/util/ImdbUtil.java
@@ -19,8 +19,9 @@ public class ImdbUtil {
 
 	public static String extractImdb(File file) {
 		String ret = extract(file, IMDB_REG);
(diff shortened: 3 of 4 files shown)


Nadahar:
I've made comments in the code where needed, please review the comments.

---

Review status: 0 of 4 files reviewed at latest revision, 6 unresolved discussions.

---

*[src/main/java/net/pms/network/UPNPControl.java, line 383 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/833#-KANKZewBJKNoc2zbKw2:-KANKZexDbV0KsdHHWbg:1145055391) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/4f22d55095cf98ff74ed4d9e014cddaa4ad17bec/src/main/java/net/pms/network/UPNPControl.java#L383)):*
This is to prevent a NPE when ```d == null```

---

*[src/main/java/net/pms/util/FileUtil.java, line 540 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/833#-KANKldBlWb7gsC5O9Dn:-KANKldBlWb7gsC5O9Do:74598164) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/4f22d55095cf98ff74ed4d9e014cddaa4ad17bec/src/main/java/net/pms/util/FileUtil.java#L540)):*
@SubJunk You'll have to verify these changes, but it seemed to me like this and the one below must have been forgotten for "this variant".

---

*[src/main/java/net/pms/util/FileUtil.java, line 607 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/833#-KANLB4Crmh1BO-nbcnJ:-KANLB4Crmh1BO-nbcnK:1501235094) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/4f22d55095cf98ff74ed4d9e014cddaa4ad17bec/src/main/java/net/pms/util/FileUtil.java#L607)):*
Replace any number of whitespaces with one space instead of just two

---

*[src/main/java/net/pms/util/FileUtil.java, line 621 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/833#-KANLIr379Wkd7a7Cce0:-KANLIr379Wkd7a7Cce1:-924656660) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/4f22d55095cf98ff74ed4d9e014cddaa4ad17bec/src/main/java/net/pms/util/FileUtil.java#L621)):*
Send ```formattedName``` when ``` searchFormattedName``` is blank (which is in most cases). This is the term that end up being sent to opensubtitles as the search phrase, and if this is blank the filename will be used instead. This made matching possible for a few of my samples.

---

*[src/main/java/net/pms/util/ImdbUtil.java, line 25 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/833#-KANM1azMKJx6yZr90Vw:-KANM1azMKJx6yZr90Vx:-696189031) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/4f22d55095cf98ff74ed4d9e014cddaa4ad17bec/src/main/java/net/pms/util/ImdbUtil.java#L25)):*
The old code was wrong for use with opensubtitles (and that is the only place this is used form what I can see). Opensubtitles specify that only the numbers should be included (not ```tt```).

---

*[src/main/java/net/pms/util/OpenSubtitle.java, line 57 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/833#-KANMPAppkD0UamUV9p_:-KANMPAppkD0UamUV9pa:-344975813) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/4f22d55095cf98ff74ed4d9e014cddaa4ad17bec/src/main/java/net/pms/util/OpenSubtitle.java#L57)):*
Only writes were synchronized while reads were unsynchronized for ```token```, so I rewrote synchronization. I'm not 100% sure if this is the optimal way to synchronize this, but after reading for a while about it without being any wiser I ended up using this approach. This way is thread-safe, I'm just unsure if some nanoseconds could have been shaved off by doing it differently.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/833#-:-KANN7Ya3-LqgDrDatWt:1516075007)*

SubJunk:
Reviewed 4 of 4 files at r1.

---

*[src/main/java/net/pms/util/FileUtil.java, line 540 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/833#-KANKldBlWb7gsC5O9Dn:-KAOF-ctnZozBcMJLbgC:1347088457) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/4f22d55095cf98ff74ed4d9e014cddaa4ad17bec/src/main/java/net/pms/util/FileUtil.java#L540)):*
I think these things should be left as they are until we have specific examples that could be improved. When I wrote this feature each variant was based on a common naming convention I had at the time which included a lot of sample files from forum users. So if something is omitted like this was, it was on purpose at the time.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/833)*

Nadahar:

---

*[src/main/java/net/pms/util/FileUtil.java, line 540 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/833#-KANKldBlWb7gsC5O9Dn:-KAOGy-cnIboXubTPO-P:-497442240) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/4f22d55095cf98ff74ed4d9e014cddaa4ad17bec/src/main/java/net/pms/util/FileUtil.java#L540)):*
I have a hard time to see the case where this would be beneficial. The first change is simply to remove those standard 720p, 1080p etc tags for movies where the year is in square brackets in the same way it is done for normal brackets. Im which situation do you want to keep those tags when prettifying or before searching for a match on opensubtitles? The reason I changed it is because I have files where this didn't work because it were omitted. 

The same applies to the second, why would you want to keep the "dots for space" for titles with the year in "normal" brackets?

I think it's more likely that your samples didn't have a lot of movies with the these variants (you refer to them as rare), so these options didn't "evolve" much. 

I understand the "don't touch what's working argument, but as stated, it doesn't work for most of my stuff - so it probably depends on the source. Also, for most "mainstream"/scene releases my bet is that most will match by hash, in which case all of this isn't used and the lookup will be correct in any case.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/833)*

valib:
Reviewed 4 of 4 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/833)*

SubJunk:

---

*[src/main/java/net/pms/util/FileUtil.java, line 540 \[r1\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/833#-KANKldBlWb7gsC5O9Dn:-KAOw4WfHZsPkQOlITgy:-1977776365) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/4f22d55095cf98ff74ed4d9e014cddaa4ad17bec/src/main/java/net/pms/util/FileUtil.java#L540)):*
If it works better for you this way it's fine, I don't anticipate any issues with it. I suspect I left it out originally for efficiency since my files didn't need it but if yours do then it's a good improvement.

We should keep in mind it's important that it works without Open Subtitles too for those users who don't enable external network - and that's the way the feature was written in the first place. It's an offline feature that is enhanced (soon to be greatly enhanced) by online information.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/833)*

SubJunk:
I'm good to merge this

Issue #739, opened 2015-11-23
UMS doesn't free DLNA ports after closing


porst17:
While trying to debug #738, I wanted to switch back and forth between UMS and minidlnad/ReadyMedia. However, after stopping the UMS server and application, I am still unable to start minidlnad again. From the minidlnad log:
```
[2015/11/23 10:39:25] minissdp.c:647: debug: SSDP M-SEARCH from 192.168.178.1:48569 ST: urn:schemas-upnp-org:device:avm-aha:1, MX: 5, MAN: "ssdp:discover"
[2015/11/23 10:39:26] minissdp.c:647: debug: SSDP M-SEARCH from 192.168.178.35:40398 ST: urn:lge-com:service:webos-second-screen:1, MX: 5, MAN: "ssdp:discover"
[2015/11/23 10:39:27] minissdp.c:647: debug: SSDP M-SEARCH from 192.168.178.35:40398 ST: urn:lge-com:service:webos-second-screen:1, MX: 5, MAN: "ssdp:discover"
[2015/11/23 10:39:28] minissdp.c:647: debug: SSDP M-SEARCH from 192.168.178.35:54067 ST: urn:lge-com:service:webos-second-screen:1, MX: 5, MAN: "ssdp:discover"
[2015/11/23 10:39:28] minissdp.c:647: debug: SSDP M-SEARCH from 192.168.178.35:40398 ST: urn:lge-com:service:webos-second-screen:1, MX: 5, MAN: "ssdp:discover"
[2015/11/23 10:39:29] minidlna.c:154: warn: received signal 15, good-bye
[2015/11/23 10:39:29] upnpevents.c:185: debug: removeSubscriber(uuid:ad9f1fac-91af-11e5-9e45-460401000000)
[2015/11/23 10:39:29] upnpevents.c:185: debug: removeSubscriber(uuid:ad9e3dd0-91af-11e5-9e45-460401000000)
[2015/11/23 10:39:29] upnpevents.c:185: debug: removeSubscriber(uuid:ad9b0f84-91af-11e5-9e45-460401000000)
[2015/11/23 12:10:41] minidlna.c:1026: warn: Starting MiniDLNA version 1.1.4.
[2015/11/23 12:10:41] minissdp.c:114: error: bind(udp): Address already in use
[2015/11/23 12:10:41] minidlna.c:1057: info: Failed to open socket for receiving SSDP. Trying to use MiniSSDPd
[...]

PR #489, opened 2015-02-23
Visibility debug

skeptical:
@SubJunk Just a temp branch to help debug the ums visibility issue.  If you could create a binary for affected users to try we'll at least get some more useful logs :-).

[Review on Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/489)

Issue #236, opened 2014-02-06
Often netwok port results blocked, therefore UMS cannot work


Simone98RC:
UMS reports free network ports as blocked, rarely it works without error messages.
I already checked if these ports were really blocked, but it was not.
Please fix it, in future releases you could set a casual port if default (5001) isn't accessible.

Arglanir:
Using Windows 10's resource monitor, I can see after UMS shutdown, that the port 5001 is still used, but attached to no process.
So if I restart UMS, it cannot bind to the port 5001 and thus says it cannot work.
![image](https://cloud.githubusercontent.com/assets/5308010/9833649/555c1810-599f-11e5-84a5-f98ada1bb1cb.png)
Is it the same bug or should I start another?

Nadahar:
@Arglanir It's probably the same bug. We need to figure out what causes this to be able to solve it. This might be a stupid question, but you do close UMS on the ```Quit``` button and not by clicking ```x``` on the window? I'm also wondering if run UMS from a administrator account, and if you have [UAC] (https://en.wikipedia.org/wiki/User_Account_Control) enabled or not?

It would also be great if you could create trace logs of this happening (closing down without releasing the port) and then post the log [in this thread] (http://www.universalmediaserver.com/forum/viewtopic.php?t=1656).

Arglanir:
Yes I quit it by using the "Quit" button (not the X button of course, as it only minimizes it)
I was not running it as administrator to it was not possible for UMS to change it's own files.
I tried to make the error appear today by restarting UMS a lot of times (with port 5002) but I did not succeed in bugging UMS.

But even when it's shutdown and not running, in the Resource monitoring window I still have 5 weird local connections to UMS configured port 5002 (but not attached to a process). But they disappear after some time.

Using port 5001, UMS did not succeed in starting. (I cannot log to the UMS web server so [here are the debug files](https://dl.dropboxusercontent.com/u/1889097/ums_dbg_pb_port5001.zip))
But there are a lot of connections for the local port 5001 that have been created (but still not attached to a process).

I restarted using port 5002, and it worked, but a lot a

Nadahar:
@Arglanir 
> I was not running it as administrator to it was not possible for UMS to change it's own files.

Could you please rephrase this sentence as I'm not sure what you mean.

Are you running UMS as a service?

skeptical:
[...]

PR #835, opened 2016-02-14
IFrame removal and consequences

Nadahar:
This PR is somewhat of a mess in the sense that it's digressed into a lot of changes that has a very weak or no relation to the original issue. It does solve some fundamental problems though, like thread safety for the PMS and LooksFrame singleton instances. This has been a mess, especially during tests.

I also think it cleans up parts of the code so it's easier to understand. The whole IFrame - LooksFrame - DummyFrame construct seems to me like an attempt at a lazy handling of headless mode (where you could just code "pretending" you had a GUI) that caused more confusion and bugs than solutions, so I simply removed it.

The original issue was discovered when I tried to look at the lingering UMS issue (#236, #489, #739), and I got the following stack trace:

``` logtalk
INFO  2016-02-06 09:42:54.795 [Thread-800] Exception in thread "Thread-800" java.lang.Error: Interrupted attempt to aquire write lock
INFO  2016-02-06 09:42:54.797 [Thread-800]  at javax.swing.text.AbstractDocument.writeLock(AbstractDocument.java:1349)
INFO  2016-02-06 09:42:54.799 [Thread-800]  at javax.swing.text.AbstractDocument.insertString(AbstractDocument.java:702)
INFO  2016-02-06 09:42:54.800 [Thread-800]  at javax.swing.text.PlainDocument.insertString(PlainDocument.java:130)
INFO  2016-02-06 09:42:54.800 [Thread-800]  at javax.swing.JTextArea.append(JTextArea.java:477)
INFO  2016-02-06 09:42:54.801 [Thread-800]  at net.pms.newgui.TracesTab.append(TracesTab.java:186)
INFO  2016-02-06 09:42:54.801 [Thread-800]  at net.pms.newgui.LooksFrame.append(LooksFrame.java:523)
INFO  2016-02-06 09:42:54.801 [Thread-800]  at net.pms.logging.FrameAppender.append(FrameAppender.java:109)
INFO  2016-02-06 09:42:54.805 [Thread-800]  at ch.qos.logback.core.UnsynchronizedAppenderBase.doAppend(UnsynchronizedAppenderBase.java:88)
INFO  2016-02-06 09:42:54.806 [Thread-800]  at ch.qos.logback.core.spi.AppenderAttachableImpl.appendLoopOnAppenders(AppenderAttachableImpl.java:48)
INFO  2016-02-06 09:42:54.806 [Thread-800]  at ch.qos.logback.classic.Logger.appendLoopOnAppenders(Logger.java:273)
INFO  2016-02-06 09:42:54.806 [Thread-800]  at ch.qos.logback.classic.Logger.callAppenders(Logger.java:260)
INFO  2016-02-06 09:42:54.807 [Thread-800]  at ch.qos.logback.classic.Logger.buildLoggingEventAndAppend(Logger.java:442)
INFO  2016-02-06 09:42:54.807 [Thread-800]  at ch.qos.logback.classic.Logger.filterAndLog_0_Or3Plus(Logger.java:396)
INFO  2016-02-06 09:42:54.808 [Thread-800]  at ch.qos.logback.classic.Logger.debug(Logger.java:503)
INFO  2016-02-06 09:42:54.808 [Thread-800]  at net.pms.io.WindowsNamedPipe.run(WindowsNamedPipe.java:348)
```

I started to try to figure out how this could happen, and several fundamental design issues dawned upon me. There are many aspects to this, first of all UMS doesn't keep track of the threads it spawns so it can't close them down when closing. Secondly `InterruptedException` is generally not respected and is often just caught and ignored. The JVM expects the application to terminate its threads before it takes charge of the shutdown process (initiated by `System.exit()` which is in itself not the way to terminate). Any remaining threads are given an interrupt signal. Those that still doesn't shut down will have to be "killed" by the JVM, and when some of these might be holding locks and/or blocking in some way there really isn't much of a surprise to me that shutdown doesn't always complete successfully.

Changing all of this is a huge task that's bound to have side effects, so I've not attempted to do this here. What I've done is that I've attacked one of what I consider to be the design flaws that creates this situation and try to correct that including all side effects. I drew the line here for now, this has already a huge potential to break something, so I think this needs to be tested and deemed stable before another one of the big issues is attacked.

It seems that you're not all aware of this (@skeptical told me eariler), but there's a special "awt" dispatcher thread that runs all GUI initiated code. All changes to the GUI must be run in this thread since the GUI code is generally not thread-safe. Tasks can be scheduled for this thread by calling `SwingUtilities.invokeLater(Runnable)`. This is the only safe way to do it. I found calls from all over the code modifying GUI components, so I tried to come up with a general solution to the problem. It took the form of the static method `LooksFrame.updateGUI(LooksFrameUpdater)` and a static inner class `LooksFrameUpdater`. `LooksFrameUpdater` is just an abstract implementation of the `Runnable` interface with some repetitive code already implemented. If you look at it it should be pretty easy to understand. You implement the actual code in `doRun()` in an anonymous inner type/subclass of `LooksFrameUpdater`. Since I've removed the `IFrame` interface I've implemented all calls to it from the outside to use this method - so the GUI code should be thread-safe with this.

I already have a half-done branch that attacks some of the same issues but primarily looks at the startup sequence and error handling during startup. When this is deemed safe and merged, I can sync merge that branch and then continue that to make step two. I don't know how many steps it will take to get this all under control, but I figured I might as well start. 

I'll make some comments in the code later, but right now I'm just posting it here for you to review or test.

The commit message would have been gigantic if I should have made it detailed, so I ended up with the following fuzzy description:
- Removed `IFrame` interface
- Made all `IFrame` calls go to LooksFrame
- Made `LooksFrame.updateGUI()` and class `LooksFrameUpdater` for easy GUI updates from other threads
- Made a lot of synchronization bugfixes
- Made general bugfixes where found
- Changed `FrameAppender` "callback" logic and made sure `FrameAppender` is terminated before the GUI is closed

[Review on Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/835)


skeptical:
@Nadahar this one reminds me of your old 'stream-of-consciousness' style of committing where we mostly see a log of your work session as it happened. My comments

* Removing ```Iframe``` is a breaking change. "_**seems to me** like an attempt at a lazy handling... it caused more confusion and bugs than solutions, so I simply removed it_" is itself a bit lazy and not a valid/sufficient reason for introducing a breaking change (hint: plugins sometimes call ```PMS.getFrame()``` too).

* With something as fundamental as ```SwingUtilities.invokeLater()```, imo it's better to not to hide it behind a spoonfeeding wrapper but just fix it in code so it's clear at a glance what's going on. This is more informative/instructive/self-explanatory to your teammates, whatever 'it seems that [they]'re not all aware of'.

* Some more breakage occurs in the somewhat vague "_Made general bugfixes where found_" part (possibly another 'seems to me' improvement): the reason for 2 configuration vars in ```Player``` and subclasses (static var```_configuration``` plus instance var ```configuration```) is that ```DeviceConfiguration``` is allowed to override ```PmsConfiguration``` settings too (for reference see 5 commits starting at cb39545, esp 7c26b1c), so:
 * at the beginning of ```launchTranscode()``` we temporarily point the Player instance's ```configuration``` var to the given ```DeviceConfiguration```.
 * consequently during ```launchTranscode()``` while the output command is being built, the Player's helper functions are actually looking up PmsConfiguration settings in the ```DeviceConfiguration``` when they make calls like ```configuration.getSubsColor()``` and so on.  If we have a custom device conf, these settings could be different from the static ```_configuration``` derived solely from *UMS.conf*.
 * at the end of ```launchTranscode()``` once the command is built we restore ```configuration``` to its previous value (which is presumably same as ```_configuration```).
 * needless to say, this feature is undone by this proposed fix. I agree it may not be immediately obvious, but custom device confs were retrofitted on top of the older conf logic and my idea was to avoid a lot of disruptive refactoring since everyone else is always working on this area of the code too.

* 'Seems to me' when one finds oneself writing things like '_The commit message would have been gigantic..._' it would be a good thing to then consider '_so maybe I should try dividing this commit into coherent logical parts, that way everyone else can more easily see exactly what I'm doing, and who knows, I may also spot something I didn't see before_'.

Nadahar:
@skeptical I said I was going to write more specifics, I was just too tired of it all to do that yet. 

The main reason I removed IFrame was because it complicated thread safety, which has been my guiding line through all of this. Since the hang after exit that I managed to get logged (I'm not saying it's the only one, but that's clearly one of them) clearly shows that there's something seriously wrong with the locking scheme there I started to look at how this was tied together. I found that the reference to IFrame was copied here and there which made it impossible to "dereference" it. To shut down in a proper manner you have to dereference objects (either explicit or by scope) so that they are garbage collected. I took a look at the whole construct and couldn't see that it made much sense, the only thing DummyFrame actually did was to capture all logging event in memory in headless mode so that you're guaranteed to run out of memory in the end. That is gone with this change, so the web interface log feature probably needs a check too if it ever were to be used.

The reason the commit message is like it is is simply because when I make a change like this there's a lot of compile errors that I have to go through and sort out. I committed as soon as that was done, but by that time the PR had this size. As I've said before I don't commit "in the middle" of something where something is half implemented, I don't see any use of doing that.

Honestly I don't really care so much if some of the plugins must be updated. Replacing ```PMS.getFrame()``` with ```LooksFrame.get()``` isn't exactly a huge job, it would take me about 5 minutes per plugin. There is will to spend considerably more time and energy arguing against changing anything in any plugin than it is to change plugin code.

I'm sorry that the wording was to vague for you, honestly I don't remember all the reasons for everything I did right now, this has been a lot of hours. That was why I wanted to comment in the code, because I would hopefully better remember what and why when I went through it.

The reason for making the wrapper wasn't to hide anything but to avoid having to repeat the same code over and over again. It's important to have a capture all exception clause in the ```run()``` method to avoid crashing the swing/awt thread, and doing that with logging is what I wanted to avoid repeating. It's not complicated in any way, so anyone that look up what it does will understand it instantly.

About the ```DeviceConfiguration``` I don't understand why it's undone. To me it seems like the change was only used in that method and then restored, so I replaced it with a local one. Is the problem that you want calls made from elsewhere too to use the overridden configuration? Doing it this way creates synchronization issues since you write to the configuration reference. To properly synchronize it there would have to be created locks that were inherited and used properly in all the subclasses, and I didn't see why a local variable wouldn't achieve the same.

That said, I haven't yet understood the reason for doing it this way at all. If there are a few of the ```PmsConfiguration``` settings that's wanted on a per renderer basis, why not just make options for them in the ```RendererConfiguration``` that defaults to the value in ```PmsConfiguration```? I've assumes that there's a bigger picture I haven't seen, but when you talk about subtitle colors it seems to me that that would have been the best solution.

That said, I have no intention of forcing this on anyone. If there's no will to do the painful changes needed to solve the more fundamental design problems that keep making fixing things difficult, I'll just continue in my own branch or fork. My patience has grown thin on this never ending discussion that all comes down to that we can't fix the major stuff because it might break some minor stuff.

SharkHunter:
What major stuff are you talking about? What is major? Rendering a plugin useless i a major impact and should be avoided if possible. Which is what I think @skeptical is after. This is not a major enough issue to motivate a breakable change. One that can be avoided. 
If we should go down this path I have a more pressing concern why is the main class called PMS?  Change that instead! (note this might contain some sarcasm).

Nadahar:
@SharkHunter What I consider major is all the bugs that's caused by fundamental design flaws that is reported constantly by the users and is ignored completely. Trying to tackle such things is to me more important than avoiding to have to change a couple of calls in some plugins.

SharkHunter:
@Nadahar - obviously we have a completely different mindset about what is a major bug and what is acceptable to fix that bug. If the fix of a bug affects more users than benefit from the fix then the fix is bad. Before breaking backwards compability  you must estimate how big an impact this is  etc. Something you simply ignore and just change. How many users are affected by this? 1? 10? 100? 
And pls explain why this is so important to change but not the name of the main class? To me it is on the same level. Very little gain. But I don't have any objections. I'm happy with my system and I'm not about to update any of my plugins just due to some name changes occurs. Now I haven't updated stuff here for ages (I've been occupied with other stuff) but this path is not one I'm going to follow.

Nadahar:
@SharkHunter I think we have estabilished already that we have a completely different view on coding in general. I have tried to explain why I think it's important, but you just ignore my explanations and repeat the same allegations (this pattern repeats itself in any discussions we've had). So, I see it as a waste of time to yet again try to explain my views to you, I simply don't believe that you're interested in seing any viewpoints that doesn't match your own.

Frankly I was expecting this kind of response, but was hoping I was wrong. I figured I'd give it one more chance though. What I don't understand is if you want everything to be as it is, why can't you just stick with an existing version?

SharkHunter:
@Nadahar "simply don't believe that you're interested in seing any viewpoints that doesn't match your own" and neither are you. You change and don't do any sort of impact analysis. 
"What I don't understand is if you want everything to be as it is, why can't you just stick with an existing version?" 1st I don't want everything to be as it is. But if you break stuff the gain should be huge (read new cool feature or bug that prevent videos to be played). And I don't understand why you don't can make it backwards compatible? 
Who will update all plugins? You? Do you have write permissions in all plugin repos? Or does someone else has to be involved in the updating?

Nadahar:
@SharkHunter 

> You change and don't do any sort of impact analysis. 

Let's get one thing clear: You know absolutely nothing about what I think. 

I only do things I believe to be to the better, but I don't share the idea that everything that a plugin ever has used must be frozen in all eternity. Normally, plugins break in version changes. Have you tried to make a Firefox plugin for example? I have, and it was just a few months before it was broken, and that wasn't easy changes like this but major refactoring where I really had to figure out how to do some things all over again. I've actually never heard of a product other than UMS that can't be improved if it has the possibility to affect a plugin, and I think it's a failed strategy. I know that you and @skeptical this it's great, but I don't.

You know very well that I don't have write access to "all plugins" whatever that is. I don't think I have write access to a single plugin in fact. As you very well know I've done this before by making pull requests for those plugins that's there for automatic downloading, but not all PR's were accepted because of the same ill-will to change that drives this. I'm not going to waste my time with that again. 

The fact is that this plugin argument is a lot of bullshit as I see it. Plugins can use any part of the code, so if we were to follow that logic to the full, nothing can be changed because we don't know what every potential plugin out there does. The plugins you're concerned about however, are the plugins that's controlled by us. And those can very easily be updated. The ones we don't know about are more problematic, but at the same time it's self explanatory that you can't freeze all the code because some to us unknown plugins might use that particular piece of code. It really feel this is just about prestige, ill-will and defiance.

SharkHunter:
@Nadahar - the problem with the plugins are simply one issue I have with rewriting. Since you still can't present a good enough reason why this is so important or how it impacts the rest of the system implies that it probably isn't that important. I'm asking again why is this so important? Why is it impossible to make it backwards compatible? The only answer you given so far is that we are ill-willing 
reactionary grumpy old men. There is no ill-will to change. But the change should be good.

skeptical:
@Nadahar trying to get to the root of the 'lingering ums instance' bug in Windows is vital, I agree.  The best way I know of to do that is to learn how to reproduce it first.  If it's not possible to reproduce it (which I wasn't able to do) then one falls back on generic coding fixes, but in that case it should at least be testable/verifiable whether the generic coding fixes had any discernible effect on the bug. The answer shouldn't to embark on a wholesale purge where suspects are lined up and shot on the street (I'm only kidding, this is just for rhetorical effect). As of now we know for sure that 2 things have been broken here, but we don't know for a fact that anything significant has been fixed.  This is because the effort has lost its focus imo.

This codebase, for better or worse (both I would argue), has evolved/devolved into something of a hacker's playground where the left hand often doesn't know what the right hand is doing, but it still manages to do some nice things.

Regarding moving settings back and forth between 'app' and 'renderer' categories, I would say that these are to some extent false categories, and some features (including, but not limited to, all-encompassing ```DeviceConfiguration``` overriding) are a recognition of this. @SharkHunter has often pointed out that since dlna has no working concept of a 'user', then in practice 'renderer' becomes something of a substitute for that.  Why shouldn't a renderer have different shared folders, etc?  I only used ```getSubsColor()``` as an example because it's an easy demo of the principle.  It's not my concern whether ```subtitles_color``` is intrinsically an app or renderer setting, I'm just letting users determine that for themselves.

Nadahar:
@skeptical This PR fixes lots of bugs, but it's only a first step as I explained. The reason is that if I should do it all in one go it would be to difficult to pick up and correct what's broken along the way. My goal is to solve as many synchronization/concurrency issues as possible (where this already clears up all the errors given by Threadsafe in the PMS class, and many in many of the other classes plus that it takes care of getting the GUI updates over to the event dispatcher. Once the bulk of the synchronization issues is tackled (I don't know where to start with DLNAResource - it's the class with most concurrency bugs that I'm aware of and it seems like a real nightmare to try to sort out, so I was thinking of leaving that for later), I am planning to get control over the startup and shutdown sequence, make UMS take responsibility over it's spawned threads, make all ```run()``` methods have a "last line of defence" exception catcher/logger to avoid crashing the thread, go through all ignored ```InterruptedException``` handling and make sure that all threads actually terminates when one is signalled and then create graceful shutdown of all threads and the GUI (and thus the event dispatcher thread) when quitting. This should allow the removal of ```System.exit()``` (which is really a "kill" command that should only be used in "emergencies"). I am convinced that once this is done, the "lingering instance" will be gone, but I can't "prove" it until it's actually implemented.

When the requirement is that this must be proven before any changes can be made, even though all the changes make sense in themselves to make sure thing functions in an organized way, I see that as an unreasonable demand that effectively blocks major issues like this from ever being resolved. 

It's easy to prove cause and effect for simple issues, it's not the same for complicated issues. That doesn't mean that complicated issues don't exist, and I guess that's my bottom line here.

When it comes to broken things, the only thing I know of that is probably broken in this PR is the web interface log retrieval in headless mode. I don't consider having to change tiny bits of code in some plugins as breaking something, and I still don't see how the ```DeviceConfiguration``` is broken. Is it because during this periode of time, other calls to ```configuration``` (from outside the method in question) should be rerouted as well? If so, that's a hackish way that obviously needs to be thread safe - the very construct I was aiming to avoid by using a local variable instead.

It was never my intent to break ```DeviceConfiguration``` even though I struggle to see that the complications it makes is worth it. UMS do have support for multiple profiles where you can block/allow renderers as needed, and IMO that is the better solution to such situations. That allows complete separation between the two for the database, all configuration files etc. It's not that I don't appreciate freedom of choice, but I think UMS is to some extent crippeled by all the non coherent solutions that have to work dispite of eachother instead of in harmony. It makes evolving very difficult.

SharkHunter:
synchronization, catching all errors and terminate threads "nicely" costs. In implementation time, in performance etc. And we'll get new (other) bugs. If the synchronization was a disaster then nothing would work. Now it works pretty good indicating that it is not a big problem regardless of what any tool says. It's an academic problem. Last time we had a big rewrite of the code base it took at least 6 months before all bugs were ironed out. Call me stupid but I at least find it more fun do do new stuff than rewriting old to solve some small bugs (and yes these bugs are small).

skeptical:
@Nadahar no one is saying that "complicated issues don't exist".  Just "even with complicated issues, it should be possible to determine whether a given fix has (or seems to have) any effect".  If you can't show the efficacy of a fix, then it shouldn't be implemented at the cost of losing existing capabilities.

SharkHunter:
@skeptical - is as always much better at summarizing what I mean.

Nadahar:
@SharkHunter It's like we're on different planets. Synchronization IS a disaster. The tool didn't point that out, I did. The tool is to help origanize/find the bugs and to prioritize them by severity. I get a strong feeling that you've never bothered to look into multithreading in any detail at all, your understanding seems to be based on fobia rather than knowledge. I found the tool as a consequence of seeing that there are a lot or really fundamental synchronization flaws here. You don't have to take my word for it, or that of the tool. Just read about the matter - I'd say pretty much any source you can find on the matter, and it will confirm what I'm saying. I don't see the reason for me to spend my time trying to explain to you the basics of how this works, especially when you're only interesting in "defending" your stance that it's an "academic problem", the concepts and consequences are simple and straight forward even though the implementation might not always be.

To repeat myself into oblivion: Bugs ARE NOT IRONED OUT as it is. UMS suffers from all the sympthoms of flawed multithreading - we simply see completely different realities. The cost in implemention time is small compared to the cost in discussion time, and the performance cost is negligible unless you do it wrong (so that threads spend a lot of time waiting). The situation as it is now is much worse also performance wise, where e.g all DLNAResources are resolved multiple times for some unknown reason nobody seems to be able to figure out. That's cost on a completely different scale, yet you seem to have no rush to address this despite being extremely worried about the nanoseconds lost to locks.

Again, the bugs in question aren't small and nobody have asked you to solve it. You are trying to block me from attempting to solve it, that's what happening, so I find it hard to see how it's relevant what you think is more fun to do.

@skeptical So what you're saying is that doing things in multiple steps is impossible because the effect doesn't show on step one? Do you think it's better to complete it all and then never being able to merge because there are too many things changed for anyone to be capable of having an overview? I know what the result of that would be, the message then would be: This should have been split into multiple PR's. Making any progress in this area is impossible if one is to follow the rules set by you.

And please stop referring to breaking things. This was never meant to break anything. The reason for reviewing a PR is to see if it breaks things and to solve that. Causing minor changes in plugin code is not breaking things by any definition I know of. It would be easy to keep ```getFrame()``` and just forward that to LooksFrame, but the point is that the plugins need to check ```isHeadless()``` before doing the GUI stuff. It's not exactly rocket science or a lot of work to implement, but since that change is needed there's no reason to keep ```getFrame()```.

When it comes to the ```DeviceConfiguration``` and the web interface log I'm completely open to solving this, but I need factual information about how to go about solving it instead of general complaining.

SharkHunter:
@Nadahar - now it is my turn "You know absolutely nothing about what I know about synchronization". I have only worked with multicore systems for 10-15 years and even teached students so I probably know nothing. 
Synchronization is a tricky thing. Rule one is that you should know what you're doing. Rule two is make sure you really need it. If synchronization were gods gift to programming everything would be synchronized but everything isn't right? Synchronization costs. It needs locks and threads has to be stalled. 

So question one is the synchronization of UMS a total piece of crap? I'm not convinced. If it were a problem then it would show up more often. So I would say that most cases work. If you now go in and start adding sync points then you'll screw up a working timing which could cause other interesting problems. 
I'm wondering what you mean with synchronization? What are you going to synchronize?

Nadahar:
@SharkHunter I don't know what you know, but I know that you don't want to adhere to the most basic principles. I find it very strange to have to debate whether thread safety is necessary. 

I agree that it can be and is tricky, and I think Java is making it worse because of it's "helper" functions (synchronized, volatile) which abstract the programmer from what's really happening. As I see it, synchronized methods should almost never be used, as they lock much more than they have to.

But, even though there are challenges there's not an option to just say "we pretend it isn't there", or "we like the excitement". I'm not a gambler, I don't understand what draws people to gambling - to me that is completely pointless. I guess that might have to do with it, but I see ignoring thread safety as pretty much the same as gambling - except that there's no jackpot to be won.

I keep hearing that you say that not everything is synchronized. I don't know what code you refer to, I'm sure there are a lot of code out there that doesn't deal with thread safety properly but there's wast amount of buggy software as well. To me it's as simple as multithreading equals thread safety. I'm using synchronization in a broad sense, I'm not talking about the Java keyword, I'm talking about making sure that reads and writes or writes and writes are exclusive. There's many ways to do that, using locks is to most obvious - but sometimes that can lead to unsolvable situations where you either get blocking or starvation. Then you have to consider different strategies. 

As I see it, AWT/Swings approach to do everything from one thread is also a way to address it, albeit a bad one. But there might be reasons why it's the only logical approach, there could be OS limitations and whatnot, I don't know. But they still respect thread safety.

Another way that I've used before is to create "message objects" that's used for all inter thread communication. You then only use a locking mechanizm to control which thread currently "owns" that object and the rest is without locks. Not all code can be "fixed" just by applying locks, sometimes the code must be structured so that access to shared objects are done only a few places, so that most of the code can run without locks. Data exchange and actual "work" can often be separated. Some times it might be ok to just make a copy of shared data and then work with that knowing that it might not be relevant in the end. It all depends - but just hoping that timing and coincidence saves the day is not acceptable. Timing won't be the same from computer to computer or installation to installation either, so even if it's running without any noticable issues on once computer it doesn't mean the same is true on other computers.

valib:
Guys I am neither an expert in the thread safety nor in the synchronization issues  and honestly I didn't take care of them when I made some code changes but the @Nadahar approach and changes he made looks reasonable to me. Agree that making all that changes in one PR is not good and should be divided into some small steps but generally I support this.

SubJunk:
These changes and the reasons for making them look good to me. I really appreciate the work you're doing here @Nadahar 

I agree about splitting it into smaller chunks. I know it creates more work for you but it makes it more possible for us to review and understand the changes when they're in smaller and more specific groupings.

You guys might be interested in this link which makes some good points https://secure.phabricator.com/book/phabflavor/article/writing_reviewable_code/

SharkHunter:
So this is how it feels to be a grumpy old man, cool. I still haven't seen any good answer to why we need to fix something that isn't broken other than it's bad design and it might fix some bug. That the fix destroy some other (working) capabilities is just simply ignored. 

Now I haven't been that productive lately and the fact that a new clone of UMS doesn't even compiles on my computer should provide a hint to my reptile brain that it's time to move on. I'm not interested in having to look for bugs that occurs due to a complete rewrite. Past experience shows that no one except me tests with web based content and I'm not interested in testing old stuff that used to work but suddenly doesn't work, but this other bug I never seen before is now solved.

SubJunk:
@SharkHunter I agree with you about breaking existing functionality. I won't knowingly merge anything that does that. We have the ability to update most of the plugins for UMS, and we should keep deprecated methods around when possible for those we can't update.

Nadahar:
@valib @SubJunk Thank you for your support, but there's a fundamental problem here that need to be resolved somehow: We seem to want to do completely different things that directly contradict eachother. I can't have this discussion every step of the way, I loose all motivation when this happens, and it's basicly due to the fact that we have different priorities and different goals.

I don't know the exact goals for the rest of you, but my goal has slowly morphed from wanting to improve everything audio (which seem to have been neglected for a while) to try to fix as many bugs and design flaws as possible. There is new stuff I want to make, but I've come to the conclusion that as it is now with hacks on hacks on hacks and where the bugs seem to be considered pets is a too hostile environment for me to make new things. The probability that it will affect something seemingly unrelated or break some special hack is simply too big.

You guys seemed to be much more in line before I came along, so the best solution might be if I go my own direction. Maybe it's a dead end and I won't manage to push through it alle, maybe it turns out more stable and easy to handle - I honestly don't know. I just know that I don't want to break up the existing team because I'm not aligned with the rest of you.

@SharkHunter I get that you don't share my views, but I'm getting tired of you just repeating the same claims that I'm just out to do some "academic" or general design "improvement" with no practical advantage, at the cost of breaking things that used to work without caring. It is NOT true and it won't be true even if you repeat it 100 times. All it does to me is give me an impression that you don't bother to actually read and try to understand what I write, but only look for an angle where you can ridicule how completely "off" I am.

SharkHunter:
@Nadahar - So if you're so sure this is the right path then it wouldn't be hard to convince me right? Your line of argument is still "adhere to the most basic principles. I find it very strange to have to debate whether thread safety is necessary" this is not saying "bug X is due to bad thread saftey in files Y,Z etc.". I understand what you say I just don't see the gain since you just use general arguments like the one above. 
But like said I'm not going to be part of this road forward. Rewrite whatever you want.

skeptical:
Well I'm also a proud member of the grumpy old man club, and I don't see that having endless mudfights over intangibles is a productive way forward, plus I've been more or less emeritus (or demeritus if you like) lately in terms of my own participation here, so I think this may be the time for me to move on too.

SharkHunter:
@skeptical - meet you over in the grumpy old corner just behind Waldorf and Statler, there might be beer there  :)

skeptical:
@SharkHunter we can shake our canes at the passersby :-).

SharkHunter:
@skeptical - and complain about how everything what better before and the youth of today! Can't even tie their shoe laces!!!!

Nadahar:
@SharkHunter @skeptical Why should the two of you leave if I go my own way? Isn't your problems first and foremost with my wishes/ideas?

ExSport:
Like a young boys guys.
The path @Nadahar is going I like because I many times struggled with problems which changed in time so sometimes something worked but other time not so it was very hard to catch the line and root cause. Thread-safty can be one of the correct path where you don't fix anything until you make more right fixes which globally will make an improvement. 
I understand also SharkHunter but also don't see the problem here as there are so small nr. of plugins without source code available if any (talking about functioning plugins in latest UMS versions) so don't see problem here. And not talking about that Nadar offered rewriting plugins by own after some breaking change done as he also did in past. 
As my parents have quite multimedia home I saw many times undeterminant problems there. Maybe for others UMS is rock stable in their environment but if you will have TVs,PS3,NAS,AndroidTV,Mobiles all together so UMS sees 15 network connected devices at time...then it is much easier to get instabilIity problems....who knows if thread safty can help here or not but I like the idea Nadar goes as I know there is lot of hidden bugs already from time of PMS...
Thanks to all of you guys and hopefully you will continue participating to this project!

SharkHunter:
I'm not saying a rewrite is not in place. I'm just saying that it during that time will be a very bumpy road where the system will be worse. I can't speak for my grumpy fellow but I think our reaction was mostly that stuff is added a bit here and a bit there where each bit breaks something and doesn't (obviously) solve anything. 
I have no interest in this rewrite since the upsides are small (it MIGHT work) but the downsides are huge (there is a rather substantial risk that you have new/same bugs after and just spend a lot of time). This plus the fact like @skeptical says I (we) have been in an emeritus status in the project for a long time now so it is probably time to move on. 

@skeptical - if we ever get a chance to meet you have a grumpy beer on me :)

skeptical:
@SharkHunter deal! Last time I was in Sweden was 1977, but who knows? Let me know if you're ever anywhere near Philadelphia.

SubJunk:
anchorman-well-that-escalated-quickly

Nadahar:
@SubJunk True, and I really don't know how to interpret it. I really don't know what's the cause and effect or the current situation here now.

valib:
As I said some time ago "guys take it easy" when the the discussion went to blaming each other I think that now is the time to repeat that. I am also a grumpy old man (almost 60 years old) but I believe that the @Nadahar approach can make some changes which can improve the basic  shortages of the original PMS code. If not we can simply revert that changes but I think we should give them a change.

Nadahar:
@SubJunk Regarding "reviewable code" what's recommended isn't far from what I'm doing. This commit built on one change, replacing the ```IFrame``` interface. What I doesn't do is separate out bugfixes I do as I stumble upon them during the primary change. If I were to do that I would never get it done, there are so many small bugfixes in here. I think those recommendation would be good when the number of needed bugfixes were fewer. The bugs are mostly the same, failure to consider ```null```, failure to close resources, failure to consider character sets for IO, failure to handle unexpected situations and bad exception handling (catching much wider than needed and/or failure to do something sensible when an exception occur). I think most are selv explanatory though except the thread-safety changes which might not be so obvious.

I'm well aware that the commit message was vague in this case, I wrote so from the start. The reason is that I had to touch so many places in the code to do the primary task (and fixing the bugs I happened to stumble over), that I didn't have the overview when committing. It's a consequence of the share number of places I had to do changes. As I wrote initially, I was planning to comment in code to make up for this, but the reaction to the idea itself took away all motivation at the time, and since feeling that a merge was unlikely I didn't see any reason to do so either.

@skeptical @SharkHunter I'm still unsure what to do here. I offered to go my own way so that I wouldn't have to break your rules about not risking breaking or having to rewrite anything. If you're leaving the project and the others are positive to this, there might not be a reason for me to do so. At the same time, there's no need for you to leave because of this, and I'd hate to be the reason why you leave.

What is your current stance, is this a "If this is done, we're gone" or have you for some unknown reason decided that this time is the time to exit anyway even if I do this on the sideline? I don't know what to think, and it really feels like an over reaction to me. Is the problem that @SubJunk and @valib showed some support for the idea at all, is that somehow a "no turning back point"? You already must have known that I had this position, it's not the first time we've discussed these things - so that can't be what is different.

Nadahar:
I've made the necessary changes for MovieInfo to work with this PR here: https://github.com/Nadahar/MovieInfo/commit/dbd73ce2fe0552f0cdbfd1d04cad93e54ddc52e8

valib:
Review status: 0 of 30 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/main/java/net/pms/newgui/StatusTab.java, line 329 \[r2\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/835#-KAqOSPf4gkZjud9HiUj:-KAqOSPf4gkZjud9HiUk:1864389238) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/d68fc357e7331ea0fc5234827b2e595fd3214ba0/src/main/java/net/pms/newgui/StatusTab.java#L329)):*
Is it necessary to use SwingUtilities.invokeLater() when this method addRenderer() is called also from SwingUtilities.invokeLater() ?

---

*[src/main/java/net/pms/newgui/StatusTab.java, line 356 \[r2\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/835#-KAqIEWwB8ggRO73lccm:-KAqIEWwB8ggRO73lccn:1725695920) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/d68fc357e7331ea0fc5234827b2e595fd3214ba0/src/main/java/net/pms/newgui/StatusTab.java#L356)):*
Is it necessary to use SwingUtilities.invokeLater() when this method updateRenderer() is called also from SwingUtilities.invokeLater() ?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/835)*

Nadahar:
Review status: 0 of 30 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/main/java/net/pms/newgui/StatusTab.java, line 329 \[r2\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/835#-KAqOSPf4gkZjud9HiUj:-KAtoR3lmaU0P5g1nDLy:1192974412) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/d68fc357e7331ea0fc5234827b2e595fd3214ba0/src/main/java/net/pms/newgui/StatusTab.java#L329)):*
No, great catch. I didn't look that far in, I just checked that addRenderer() made changes to GUI components that were not running in the event dispatcher thread. This inner one can go.

---

*[src/main/java/net/pms/newgui/StatusTab.java, line 356 \[r2\]](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/835#-KAqIEWwB8ggRO73lccm:-KAtozf-A5Gt53BttHHU:-1810779133) ([raw file](https://github.com/universalmediaserver/universalmediaserver/blob/d68fc357e7331ea0fc5234827b2e595fd3214ba0/src/main/java/net/pms/newgui/StatusTab.java#L356)):*
No, you are correct. This one I should have discovered as it were completely redundant. I do think that the most logical place to put the runnable is outside the GUI classes though, so I think keeping the outer one is "most correct" (even though that's a completely invalid argument :wink: )

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/835)*

Nadahar:
There are still unresolved issues in this PR (device configuration, web interface log). Also, in the light of recent developments I see no reason why this would ever be merged. It also has little value on it's own, it was only meant as a first step in a multi-step rewrite, and since the rest of this won't happen this should just be forgotten. I'm therefore closing this.

---

Review status: 0 of 30 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/835#-:-KAtrdFnh2NiHZjCm9Q0:-2030136774)*

valib:
@Nadahar sorry i deleted this branch. I tried to clean up branches already merged or closed. Maybe we can consider if we should continue with this PR.

Nadahar:
@valib No big problem for me, I have it locally as well - but I don't remember my plans and logic for this now anyway. All I remember is that I had hardly scratched the surface of a lot of interconnected issues that in my view should be straightened out when this discussion clearly stated that such a cleanup wasn't appreciated.

Sami32:
@Nadahar I don't have any problem with that PR.
If it's work fine we keep it, if we get issues, big or too many, we just revert it.
Is it look fair to you ? This work is just too up of my Java knowledge anyway.

Nadahar:
@Sami32 This PR has little use by itself, it's just the start of a larger work.

SubJunk:
Feel free to resurrect

PR #845, opened 2016-02-24
Use image instead of text for fully played overlay

SubJunk:
[Review on Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/845)


SubJunk:
@skeptical Another angle to take is something like this, indicating that the item can be re-played
ic_refresh_48px-128

Keen to get opinions on icons and colors @UniversalMediaServer/developers

skeptical:
@SubJunk Good idea: better visual 'semantics' imo, i.e. doesn't connote approval like a checkmark.

valib:
@SubJunk I don't agree this is a good idea. This mark is obviously used to play video again. Why not to use a checkmark? We have it already in the UMS so why not use it?

valib:
What about this change  12a6fbe ?

SubJunk:
@skeptical thanks

@valib like you said, the mark is used to play the media again, and that's what clicking or selecting the thumbnail does. So it seems to be appropriate. The tick is nice too and you're right it fits the style. I think I like the replay icon concept more though, like @skeptical said a tick can mean different things

SubJunk:
@UniversalMediaServer/developers I tried a new icon, it's one I made so we don't have to worry about rights, and means we can modify it however we want. I also made the code safer. Thoughts?

It seems a bit slow though so there's probably some speed improvements to be done.

SubJunk:
@UniversalMediaServer/developers the new icon looks cool in my opinion. I'll do some benchmarking to find out what the difference is in performance, and if it's noticeably slower (which it seems to be to me) I'll see what optimizations can be made to reduce that difference. Maybe using another image library would be faster? Any comments or tips would be welcome

SubJunk:
Hmm I did some benchmarking and it seems to take 0-3ms per hundred overlays so that's not the speed issue. I'll merge tomorrow unless anyone objects.

Issue #686, opened 2015-10-05
Handling firewalls


Nadahar:
@UniversalMediaServer/developers @SubJunk 
It seems to me like UMS isn't taking care of firewalls and simply assumes that everybody will turn them off. It can be quite tricky to deal with, as there are a number of different software firewalls available and it's hard to test incoming access without cooperating with an external host. Most OS'es these days also comes with firewalls enabled by default, and not-so-technical users will be scared to turn them off even if they are protected by a router and NAT.

I do think that more could be done though. The very least we should do is to inform the users of what incoming ports need to be open for UMS to work. HTTP wise that's pretty simple, just two TCP ports (one for the DLNA server and one for the web interface)? On the uPnP/UDP side I really don't know what is needed, uPnP is a messy standard network wise with it's broadcasting and whatnot. It shouldn't be to hard for us to figure out what was needed though (isn't UDP port 5000 involved?), and we should probably show this information in the GUI dynamicly so that it reflects the current configuration.

What would be really nice is if it would be possible to contact a renderer and request for it to contact UMS back (instead of replying in the estabilished channel) and in that way be able to test that it could reach UMS. I don't know if there are any such possibilities in DLNA or uPnP though.

We should also offer to open the necessary ports at the very least on Windows firewall, from the installer. I don't know if the install processes on the other OS'es would allow something like that, Linux is hopeless in that there's no standard for anything, and OS X is just foreign to me.

What do you guys think?

Nadahar:
Here's some [more information] (http://universalmediaserver.com/forum/viewtopic.php?f=9&t=5778&p=21813#p21813) about the subject. Firewall issues are by far the most common support thread topic, and I'd be willing to bet that a lot of people simply abandon UMS before going to any forums because they can't get it to work for this reason.

WolfganP:
I agree with @Nahadar, most of the "UMS doesn't detect my renderer" or viceversa is one of the most repeated subjects at the forums (mostly in Windows) and firewall/AV related. 
Isn't there any way for UMS to include a firewall blocking test routine at first start that may alert users? And/or a settings button that may open the Firewall Rules window for the users to check against a FAQ article or specific forum post?

Nadahar:
[...]

PR #849, opened 2016-02-28
UMS TCP/UDP Ports and Firewall Rules

sainokawara:
I have completed the code supporting Issue #686 and hope this contributes to creating the Windows Firewall "rules" necessary to run UMS as a Windows Service.  I was told that "SubJunk" would write the associated documentation.  I am more than willing to attempt this as long as it is not too complex a task for me.

Three routines were changed for the code part:
PmsConfiguration.java - "validity checks" the UMS Port number: Valid range ports 1024-65534.
UPNPControl.java - forces Cling's TCP Port to the UMS Port number + 1.
setup.nsi has been restructured to use a "Components Page":

   The UMS Base & Windows Firewall Rules are listed as "Required".
   AviSynth install selection is "Recommended" and "Checked".
   The UMS Windows Start Menu is "Recommended" and "Checked".
   A Documentation Shortcut is "Optional" and "Checked".
   Creation of a Desktop Shortcut is "Optional" and "Not Checked".

The install script was restructured to exploit NSIS functions thus simplifying the amount & complexity of the code necessary to add this support and, hopefully, future changes AND to make the UMS install process appear more consistent with other products/projects.  Several "deferred deletions" were eliminated so that an "uninstall" IMMEDIATELY deletes the highly visible UMS Shortcuts in the Windows Start Menu.  Although I dislike postponing any deletions until a Windows restart because it can result in user confusion, I did not change the rest of these.  Also, I added a "Documentation" Shortcut to the UMS Window's Start Menu because, in my opinion, this function should be available before UMS is installed and when it is not running AND the UMS "Help" function is inferior to the support provided by web browsers.  And, I cannot see wasting code on enhancing the UMS "Help" function because UMS is not and should not be a web browser.  It is a media server!

In my opinion, if getting UMS to run as a Windows Service were "trivial", that is the way an average user should install/configure it.  To them, it is just a "service" that should be available anytime their PC is running without them having to do anything such as starting it!

[Review on Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/849)


Sami32:
@SubJunk Is it fine for you ?
[debug.log_PR849.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/410302/debug.log_PR849.txt)

Sami32:
IMO, users running it as a service should give feedback first.

---

Reviewed 4 of 4 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/849#-:-KSHF2fmG19Tkl4T3yKu:b-mao5vj)*

Nadahar:
@Sami32 Reviewing is a bit confusing at the moment since we're running reviewable while GitHub has just introduced its own review functionality. They overlap a lot, and it's hard to know what belongs to what some times. 

The Windows installer is only used for Windows, so OS X and Linux is irrelevant here. I doubt very many users run UMS as a service at the moment because it's complicated to get it right. Configuring the Windows Firewall would without a doubt reduce the number of people that can't get their renderer to find UMS, but the last time I checked this also included other changes to the installer that should be carefully reviewed before merging. I disagree with several things like making the firewall rules required, including the "documentation" link that simply links to the same outdated documentation contained within UMS' "help" tab. The check if AviSynth is already installed is also removed AFAICR.

It looks like the files are commited with "Windows" newlines as well (CR+LF), making the diffs unusable. I'm not sure if this would have any consequences, but it should be cleaned up before thinking about merging both to be sure and to get useful diffs.

ErdnussFlipS:
@Sami32 could you improve the comments, what @Nadahar said? It would be nice, if this can be merged. 😀

Sami32:
@ErdnussFlipS To be honest, i don't think that @Nadahar will agree to merge it.

He planned to do some work on it: https://github.com/UniversalMediaServer/UniversalMediaServer/issues/686

Nadahar:
#686 got kindof stuck since we couldn't agree on the details, and non of the other team members bothered to be involved. However, there's some good things this/#686. But, #1195 will change how ports are allocated and this should be done once #1195 is complete.

It should be noted that this only applies to UMS under Windows, and also that it's not really needed if UMS is run as an application, Java is installed as normal and no the standard Windows firewall is used with no custom configuration. The reason is that the Java installation by default opens all incoming ports for all Java programs.

If running as a service, ports need to be opened since the service is running under a wrapper for which the Java installation generated firewall rules doesn't apply.

In the end, it's all a bit of a mess because to make such rules you have to make a lot of assumptions about the system you're installing on. This approach will make it work if everything in Windows is left to its defaults, but IMO anyone running a firewall should be responsible for configuring it as wanted. The currect practice of applications altering the firewall rules is fundamentally bad, because it means that the computer "protected" by this firewall isn't supervised by anyone - and application can and do "what they want" with it. This can compromise or completely disable a firewall, and in the end it all simply results in a false sense of security for the user. In the end, you can't both be protected and not bother to understand the protection. It's like you'd want your home to be burglar proof, but you didn't want to have anything to do with keys, alarms or codes. Everybody understands that a lock is only any good if you control who can unlock it (has a key) - it's the same with firewalls, except that most people allow strangers to hand out keys to other strangers.

I also believe that running a firewall on each computer is an unnecessary complication with little benefit. Most people have a router between them and the internet, and all firewalling should be done there. That will protect any devices behind the router, not only single devices. You should generally trust devices on your local network, if you don't you should address that (i.e properly encrypt your Wi-Fi network). The only instance that I can think of where regular users benefit from a firewall is when using a laptop to connect to foreign Wi-Fi nets either free or paid for. In those cases you will be unprotected by a router, and need to protect the individual computer.

SubJunk:
Feel free to resurrect if the problems here are solved. It seems there are probably line break changes caused massive diffs which make it impossible to review

Nadahar:
@SubJunk GitHub finally offers to turn off whitespace comparison without editing the URL manually. If you use the little "Diff settings" button on the right when looking at the "Files changed" tab, you can disable it which will ignore the line ending difference and any other whitespace change.

A little warning: Although comparing without whitespace changes can make seeing the real changes much easier, one must remember to always turn it back on again before actually approving or merging something. The whitespace changes are still a part of the commit(s) in question, and might pose problems themselves.

SubJunk:
Oh cool, thanks for the tip

PR #875, opened 2016-04-10
Make upnp hello delay configurable

thomashoell:
I found it annoying that you have to wait up to three minutes for UMS to become visible to the client.
So I made the delay configurable via the general tab.

While the minimum configurable delay is 10s, it still defaults to 180s when untouched.

---
This change is [Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/875)


SubJunk:
Thanks for the PR. I agree it's something we should refine.
Maybe instead of this, we could just make it happen every 10 seconds on unmatched network interfaces, then use 180 seconds only when there is an active renderer on that network interface.

SubJunk:
I've done this instead, I'm happy for feedback on it but will close this issue in the meantime https://github.com/UniversalMediaServer/UniversalMediaServer/commit/eab608c9aad4410228cd0c71615ad26a10a565be

SubJunk:
This has been implemented in a different way now, you can use ALIVE_delay

PR #896, opened 2016-05-22
Tweaks to Sony Bravia NX800 config

prescott66:
Added config to reduce high video bitrate, which causes video playback problems.

### DIFF
--- a/src/main/external-resources/renderers/Sony-BraviaNX800.conf
+++ b/src/main/external-resources/renderers/Sony-BraviaNX800.conf
@@ -32,6 +32,10 @@ UserAgentAdditionalHeaderSearch =
 UpnpDetailsSearch = BRAVIA KDL-\\d{2}NX800
 TranscodeFastStart = true
 KeepAspectRatio = true
+TranscodeVideo = MPEGTS-MPEG2-AC3
+MaxVideoBitrateMbps = 35
+
+
 
 # Supported video formats:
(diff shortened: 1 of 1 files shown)


SubJunk:
@prescott66 thanks for this. Did you test it on a wireless or wired network?

prescott66:
@SubJunk now i am using wired network using power-lines...before i have wireless network and there was a problem, but in that times, i dont knew about this configuration abilities :smile: this configuration is best for now.

prescott66:
@SubJunk i am using this config everyday.without problems. please merge it

SubJunk:
@prescott66 so this is needed even on a wired connection?

prescott66:
yes.all is perfect

ExSport:
@prescott66
I am on the boat with SubJunk as it doesn't seem to be optimal for wired connection.
Example: Panasonic TV
Whatever network is used, TV starts stutter with bandwidth higher than 70Mbit.
It is too much for WIFI but OK for Ethernet.
Using 35 also for Ethernet should lead to:
- higher CPU usage due to converting full HD videos to so small bitrate
- buffer underflow/overflow as converting full HD high bitrate to so small mpeg2 bitrate is almost no way - no real-time transcoding due to CPU or quality set with so small bitrate
- stuttered video, pixelated video, poor quality video
It means it can work for you but not for others with different files, codecs, resolutions, subtitles requirements, poor CPUs etc.
I am sure Sony will not be limited to 35Mbits only but it is limitation for your wifi environment. Why limit others to use better video quality by forcing so low bitrate also for cable?
There should be also other "problems/limitations"

Issue #880, opened 2016-04-18
Renderer's gui panel not working after UMS 5.5.0


Sami32:
Hello,

sorry to first posting this report to the wrong place in the "Support General Help and Support" of your Website.
So here are the photos and logs :
![capture-5 5 0](https://cloud.githubusercontent.com/assets/18471069/14613993/fcc58c0a-0596-11e6-82dd-f0953b84c9c7.jpg)
![capture-6 2 0](https://cloud.githubusercontent.com/assets/18471069/14613991/fbd60c02-0596-11e6-9f00-f79033c88f93.jpg)

For the UMS log i am not able to upload them as ZIP as i get "Unfortunately, we don't support that file type" ???
Whatever, i post them on your website support part. I hope that will help you.

Sami32:
I tested with latest JAVA 7u101 and JAVA 8u92, and with UMS 6.0.0a1 and i get same broken "Player Control Panel".

Sami32:
![capture-1](https://cloud.githubusercontent.com/assets/18471069/15193874/ec549d24-17b7-11e6-9126-1fe51de22b8e.gif)

The _Unknow renderer_ is a family laptop downstair, but the _VIERA E6 Series_ and  _Panasonic TX-L32E6E_ are the same TV !
As you can see, the _Panasonic TX-L32E6E_ give a one line _Player Control panel_ when the _VIERA E6 Series_ give Java error as mentioned before.
Sometimes this kind of display can happen, so i am not sure it could help to debug the problem but just in case...i leave the debug file as well :
[http://www.universalmediaserver.com/forum/download/file.php?id=3222](url)

valib:
@Sami32 recognition of the Panasonic TVs is a little bit tricky. They use basicly the same User-Agent header so simply disable the Panasonic TX-L32V10E if it is not in your net.

Sami32:
@valib OK, thank you. When i don't make tests, normally disable all other drivers...but your UMS and programming skill are much better than mine so i appreciate your notice, advice :-)

About a new test run with UMS 6.3.0 and Java 1.8u92, here are the log :

14:11:14 INFO  Exception in thread "AWT-EventQueue-0" java.lang.ClassCastException: net.pms.configuration.RendererConfiguration$PlaybackTimer cannot be cast to net.pms.util.BasicPlayer$Logical
[...]

Issue #818, opened 2016-02-02
Issues playing files from "Renderer panel"


Nadahar:
I've had issues with the same previously, but when I'm testing right now it works for me. I've been trying to search previous issues and pull requests for anything related, but haven't found it (although I think I remember seeing something about this once).

I'm creating this issue hoping it's not a duplicate. There's a user having this issue now, reproducable (for him). I'm guessing @skeptical is the man for the job. Debug files can be found in [this post](http://www.universalmediaserver.com/forum/viewtopic.php?p=22855#p22855).

Here are what seems to be the essence:

``` logtalk
DEBUG 2016-02-01 16:58:27.752 [AWT-EventQueue-0] net.pms.network.UPNPControl Action failed: Error: Illegal MIME-type (HTTP response was: 500 Internal Server Error)
DEBUG 2016-02-01 16:58:27.752 [AWT-EventQueue-0] net.pms.network.UPNPControl Sending upnp AVTransport.Play [Speed, 1] to NICK-TV[0]
DEBUG 2016-02-01 16:58:27.762 [AWT-EventQueue-0] net.pms.network.UPNPControl Action failed: Error: Transition not available (HTTP response was: 500 Internal Server Error)
```

Nadahar:
About the DLNA guidelines - these are free for anyone to download but not share. Downloading them requires registration. You can download them by [filling out the form here] (http://www.dlna.org/guidelines/).

skeptical:
@Nadahar I could have sworn I still have a copy of those, but apparently not.  Not being a big fan of headaches, I don't think I ever when beyond skimming it over.

Just added 5b2c67fb, not as a solution, but just for the capability of turning metadata off if absolutely necessary.  It's hard to imagine that there's anything actually wrong on our side here, or what we could change to satisfy the viera. Maybe the problem will go away when firmware gets updated, who knows.

Nadahar:
@skeptical I'm surprised that it even works without metadata.

I'm not surprised that it won't accept those we send, they break several rules and if the parsing of those are written thinking that they can follow the rules and not deal with a lot of variations beyond the standard, I can easily see how the parsing can fail. What I am surprised about is the fact that these same "broken" metadata seems to be accepted when delivered as an answer to a browse request. I would think, given that they are supposed to follow the same rules, that they would use the same parsing for both.

Are you sure that these metadata are supposed to be equal to those sent as a response to a browse request?

[...]

PR #899, opened 2016-05-25
Viera AVT and related fixes

skeptical:
Parking this here in case @UniversalMediaServer/developers or others with Vieras want to jump in :-). The story so far:
- Contrary to previous reports the Viera's AVT service appears to work with UMS, at least over ethernet, as shown by @Sami32's results with the patches in #880 using `PushMetadata=false`.
- With `PushMetadata=true` he gets mixed results, possibly due to unrelated issues from running on XP. It would be good to pin this down by testing on newer Windows or other platforms.
- Per @valib's experiences  lack of UMS AVT functionality probably has to do with network topology rather than the Viera itself, so I've disabled @SubJunk's hard-coded Viera hack in 296a941, and substituted a generally equivalent renderer setting:

``` sh
ForcePlaybackTimer = true
```
- To enable extra upnp debug logging, add this to `UMS.conf`:

``` sh
upnp_debug = true
```

---

This change is [Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/899)

### DIFF
--- a/src/main/java/net/pms/configuration/PmsConfiguration.java
+++ b/src/main/java/net/pms/configuration/PmsConfiguration.java
@@ -283,6 +283,7 @@ public class PmsConfiguration extends RendererConfiguration {
 	protected static final String KEY_TRANSCODE_FOLDER_NAME = "transcode_folder_name";
 	protected static final String KEY_TRANSCODE_KEEP_FIRST_CONNECTION = "transcode_keep_first_connection";
 	protected static final String KEY_TSMUXER_FORCEFPS = "tsmuxer_forcefps";
+	protected static final String KEY_UPNP_DEBUG = "upnp_debug";
 	protected static final String KEY_UPNP_ENABLED = "upnp_enable";
 	protected static final String KEY_UPNP_PORT = "upnp_port";
 	protected static final String KEY_USE_CACHE = "use_cache";
@@ -3810,6 +3811,10 @@ public class PmsConfiguration extends RendererConfiguration {
 	 */
 	public void setAutoSave() {
 		((PropertiesConfiguration) configuration).setAutoSave(true);
+	}
+
+	public boolean isUpnpDebug() {
+		return getBoolean(KEY_UPNP_DEBUG, false);
 	}
 
 	public boolean isUpnpEnabled() {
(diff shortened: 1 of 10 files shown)


SubJunk:
@skeptical that's exciting, I'll try this out later

SubJunk:
@skeptical with this branch, I see no playback timer when playing videos. I tried with PushMetadata true and false. Happy to provide logs if they would help, let me know if there are any actions to perform for the logs

skeptical:
@SubJunk Not sure what you mean exactly. Unless you set ```ForcePlaybackTimer=true``` you should get a regular ```UPNPHelper.Player```, which is what we're testing. POC here would be to:
* Connect UMS and Viera via ethernet.
* Set ```ForcePlaybackTimer=false```
* Set ```upnp_debug=true```
* Push media from the player control panel or bump something from the web interface, testing both ```PushMetadata=true/false```.

If you have any success at all, then it's a matter of debugging whatever still doesn't work (e.g. maybe metadata, connecting via wifi, etc). If you read through #880 you'll see some suggestions on how to use cling workbench and twonky for comparison. First step would be to post logs, naturally :-).

Sami32:
https://github.com/UniversalMediaServer/UniversalMediaServer/issues/818

SubJunk:
@skeptical unfortunately I'm on wireless now since AC gives me enough bandwidth finally, and I don't have an easy way to change that

Sami32:
Since i'm at reading this nice literature ;-) i post here some extract that could be useful:

> Requirement [7.3.99.6]: In conjunction with 7.3.99.5, a MediaRenderer may progressively 
> update the values of NrTracks and AVT.NumberOfTracks, such that the total count is 
> not provided immediately.
> The definition of progressively update is that the MediaRenderer updates the state 
> variables such that a control point is able to receive GENA events at a rate that does 
> not exceed one GENA event every 0.2 seconds and the control point is able to acquire 
> the values using AVT:GetPositionInfo.

> Requirement [7.3.99.7]: In conjunction with 7.3.99.6, a MediaRenderer that progressively 
> updates the values of NrTracks and AVT.NumberOfTracks should provide the total 
> count within 10 seconds of the previous call to AVT:SetAVTransportURI.

> Requirement [7.3.99.8]: The MediaDuration output argument of the AVT:GetMediaInfo 
> action must match the value specified in the AVT.CurrentMediaDuration instance state 
> variable.
> Note that this guideline places no mandatory requirement on the accuracy of the total 
> media duration.  If the MediaRenderer is not able to provide a value, then the 
> "NOT_IMPLEMENTED" value is used.

PR #905, opened 2016-05-29
Adjust AC-3 audio bitrate for wireless networks when transcoding

SubJunk:
This change is [Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/905)

### DIFF
--- a/src/main/java/net/pms/encoders/AviSynthFFmpeg.java
+++ b/src/main/java/net/pms/encoders/AviSynthFFmpeg.java
@@ -81,7 +81,7 @@ public class AviSynthFFmpeg extends FFMpegVideo {
 		if (configuration.isFfmpegAviSynthMultithreading()) {
 			threads = " -threads " + configuration.getNumberOfCpuCores();
 		}
-		return configuration.getMPEG2MainSettingsFFmpeg() + " -ab " + configuration.getAudioBitrateAC3() + "k" + threads;
+		return configuration.getMPEG2MainSettingsFFmpeg() + " -ab " + configuration.getAudioBitrate() + "k" + threads;
 	}
 
 	@Override
--- a/src/main/java/net/pms/encoders/AviSynthFFmpeg.java
+++ b/src/main/java/net/pms/encoders/AviSynthFFmpeg.java
@@ -81,7 +81,7 @@ public class AviSynthFFmpeg extends FFMpegVideo {
 		if (configuration.isFfmpegAviSynthMultithreading()) {
 			threads = " -threads " + configuration.getNumberOfCpuCores();
 		}
-		return configuration.getMPEG2MainSettingsFFmpeg() + " -ab " + configuration.getAudioBitrate() + "k" + threads;
+		return configuration.getMPEG2MainSettingsFFmpeg() + " -ab " + configuration.getAudioBitrateAC3() + "k" + threads;
 	}
 
 	@Override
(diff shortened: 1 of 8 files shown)


SubJunk:
@UniversalMediaServer/developers I think this can be merged pretty quickly but since it's not a minor change I wanted a PR. Basically it makes it so the "AC-3 bitrate" config setting accepts "Automatic" as a value, and if that is set then we use a slightly lower bitrate on wireless networks.

The change won't be audible for users but will save about 256Kbps while transcoding on wireless networks.

Are you guys happy for this to be merged?

src/main/java/net/pms/encoders/AviSynthFFmpeg.java
```diff
@@ -81,7 +81,7 @@ public String initialString() {
 		if (configuration.isFfmpegAviSynthMultithreading()) {
 			threads = " -threads " + configuration.getNumberOfCpuCores();
 		}
-		return configuration.getMPEG2MainSettingsFFmpeg() + " -ab " + configuration.getAudioBitrate() + "k" + threads;
+		return configuration.getMPEG2MainSettingsFFmpeg() + " -ab " + configuration.getAudioBitrateAC3() + "k" + threads;
```
valib: @SubJunk are you sure that this is correct? What about if the `audio_bitrate_ac3` will be set `Automatic` the result will be e.g. ` -ab Automatick -threads 2`

SubJunk: @valib nice catch, it should be CodecUtil.getAC3Bitrate

SubJunk: @valib nice catch, I'm fixing this now

Sami32:
@SubJunk Do you plan to do some more work on that or it could be closed ?

SubJunk:
@Sami32 I think it's in a good state for merging but I'm not sure if I think it's good. Do you guys have any opinions @UniversalMediaServer/developers ?

Sami32:
@SubJunk Do you have the right "bible" ? just kidding :wink: 

> 8.1.1.2
> [GUIDELINE]    Main characteristics of the AC-3 audio stream are defined in ATSC Standard A/52.
> Sampling rate
>  32 kHz
>  44,1 kHz
>  48 kHz
> Content audio channel modes
> A bitstream conformant to this Media Format Profile may contain the following AC-3 formats
>  1/0 (Mono)
>  2/0 (Stereo)
>  3/0 (Multichannel)
>  2/1 (Multichannel)
>  3/1 (Multichannel)
>  2/2 (Multichannel)
>  3/2 (Multichannel)
> Changing audio channels among mono and stereo is allowed.
> Bit rates (variable)
>  64 kbit/s to 640 kbit/s.
> Payload format
> Payload format is raw bitstream.
> 8.1.2       AC-3 audio format: MIME type definition
> 8.1.2.1
> [PROFILES]
> AC3
> 8.1.2.2
> [GUIDELINE]    MIME type "audio/ac3" shall be used for this Media Format Profile.

Sami32:
Interesting AC3 parameters poorly documented: http://www.ffmpeg.org/doxygen/trunk/ac3dec_8h_source.html

SubJunk:
Thanks for the approval @Sami32 :) I decided not to merge this since I like my audio higher quality heh, and it doesn't make that much difference

PR #909, opened 2016-06-06
Update Panasonic-VieraE6.conf

Sami32:
Some format support adjustements

---
This change is [Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/909)

### DIFF
--- a/src/main/external-resources/renderers/Panasonic-VieraE6.conf
+++ b/src/main/external-resources/renderers/Panasonic-VieraE6.conf
@@ -31,8 +31,8 @@ KeepAspectRatioTranscoding = true
 H264Level41Limited = false
 SendDateMetadata = false
 AutoExifRotate = true
-ThumbnailWidth = 320
-ThumbnailHeight = 236
+ThumbnailWidth = 120
+ThumbnailHeight = 120
 MediaInfo = true
 
 # ============================================================================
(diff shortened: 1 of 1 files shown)


SubJunk:
@Sami32 thanks for the PR. Why 120x120?

Sami32:
@subjunk https://github.com/UniversalMediaServer/UniversalMediaServer/files/267064/UPNP_Analyse.txt
Maybe not pertinent, but it is because i saw that in the UPnP analyse of my Renderer :
```
<icon>
				<mimetype>image/jpeg</mimetype>
				<width>48</width>
				<height>48</height>
				<depth>24</depth>
				<url>dlna_icon_48.jpg</url>
			</icon>
			<icon>
				<mimetype>image/jpeg</mimetype>
				<width>120</width>
				<height>120</height>
				<depth>24</depth>
				<url>dlna_icon_120.jpg</url>
			</icon>
```
So, if it's not correct, feel free to change it ;-)

valib:
@Sami32 this analyse only shows what is TV sending to other devices. The only way how to find out the best size is to test it. You can start with the default size and compare it with smaler size if it looks similar on the screen or the picture quality is degraded.

Sami32:
@valib Thanks for the tip :-)
I only saw that it's look like a square and that the quality don't drop, to me.
So, i will make more tests with lower size...

SubJunk:
Typically it's best to leave it at the defaults, or at least leave 320 as the width and customize the height if you want a different aspect ratio.

Sami32:
@SubJunk After testing and mesuring the size of my TV compared to it's definition, my TV will not make any displaying difference with thumbnails superior to 120x90.pixels.
I guess your default values are for UHD TV ?

SubJunk:
@Sami32 the defaults are for PS3 which has huge thumbnails. In UMS we don't have different thumbnails for each renderer, so if you browse on your TV and that creates 120x90 thumbnails, then you use a PS3, the thumbnails will look bad on PS3

Sami32:
@SubJunk Oh i see. Thank you very much for the explanation :+1: 
If one day i will have some PlayStation i need remember that ;-D

SubJunk:
Hopefully in the future we will have different thumbnails per renderer. Thanks for the PR :)

PR #923, opened 2016-06-19
Add thai constants

Sami32:
This change is [Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/923)


SubJunk:
@Sami32 it looks like there's some formatting stuff in here, it would be better to have two commits: one with the change you want to make, then another with the formatting. Then it's easier to review.

Sami32:
@SubJunk Sorry about that, i am really a newbie with good will :-)
Next time i will do it, because now i think it's too late for that...

Sami32:
Done :-) I hope this time it the correct way ?
I was so afraid to make trouble with many commits or PR because i don't know how to use Github correctly yet :-(

PR #936, opened 2016-06-23
DLNA standard

Sami32:
## cf. _upnp-av-contentdirectory-v3-service.pdf_
**For UPnP Device Architecture defined boolean data types**, it is **strongly RECOMMENDED to use the value “0” for false, and the value “1” for true**. However, when used as input arguments, the values “false”, “no”, “true”, “yes” may also be encountered and MUST be accepted. Nevertheless, it is strongly RECOMMENDED that all boolean state variables and output arguments be represented as “0” and “1”. 
**For XML Schema defined Boolean data types**, it is **strongly RECOMMENDED to use the value “0” for false, and the value “1” for true**. However, when used as input properties, the values “false”, “true” may also be encountered and MUST be accepted. Nevertheless, it is strongly RECOMMENDED that all Boolean properties be represented as “0” and “1”.

---
This change is [Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/936)

### DIFF
--- a/src/main/java/net/pms/dlna/DLNAResource.java
+++ b/src/main/java/net/pms/dlna/DLNAResource.java
@@ -2292,7 +2292,7 @@ public abstract class DLNAResource extends HTTPResource implements Cloneable, Ru
 	 *            Media Renderer for which to represent this information. Useful
 	 *            for some hacks.
 	 * @return String representing the item. An example would start like this:
-	 *         {@code <container id="0$1" childCount="1" parentID="0" restricted="true">}
+	 *         {@code <container id="0$1" childCount="1" parentID="0" restricted="1">}
 	 */
 	public final String getDidlString(RendererConfiguration mediaRenderer) {
 		// Use device-specific pms conf, if any
@@ -2349,7 +2349,7 @@ public abstract class DLNAResource extends HTTPResource implements Cloneable, Ru
 		}
 
 		addAttribute(sb, "parentID", id);
-		addAttribute(sb, "restricted", "true");
+		addAttribute(sb, "restricted", "1");
 		endTag(sb);
 		StringBuilder wireshark = new StringBuilder();
 		final DLNAMediaAudio firstAudioTrack = media != null ? media.getFirstAudioTrack() : null;


SubJunk:
@Sami32 it's good to be aware of the standards, especially when developing new things. For existing things that already work, I'm hesitant to make a change. We have seen many times in UMS and PMS that sometimes a tiny change can have a big effect so I think we should only change things when we know something good will come from it like a feature or bugfix.

"If it isn't broken, don't fix it" is a good saying for us because when we don't follow that philosophy we are stuck with angry users when their things break :)

Sami32:
@SubJunk Yes, it's make sens. i need to don't forget that point :-)
Thanks.

Nadahar:
@Sami32 Now you're touching areas where I disagree with the current practice. There are tons of things that's broken, so "don't fix it" doesnt apply. The problem is that one have to prove that a given error causes problems even though it clearly is an error. Following a recommendations like this is very likely to improve compatibility with renderers, and it's very hard to see how it could harm. Any renderer not accepting "0" and "1" would be in serious trouble, and yet this is rejected.

Sami32:
@Nadahar It always better to follow the specifications to avoid later problem, and if it will get worse for the TV for sure, don't about PS3 though...
I agree with you on that point, as @SubJunk, if i understood him.
Maybe he get borred to fix all broken things and don't take a chance ? or wait to fix more of them before a new change ?
For now, the DLNA part, for Panasonic, is not very "healty", and it's a very important part.
And as @SubJunk  and @ExSport, if i remember well, was telling that the users could get bored and move from UMS, i believe that can be reallistic, since the users want something working straitht away.

An experimental version on the website, where the people will chose with wise knowledge, could to be a choice where developpers can have a real laboratory with many customer returns; a special section on the forum will need though.

Just some ideas, trying to give more place to freedom, creativity and wills of the developpers with happy users.
I saw lot of skills and wills between the developpers, but it look like these last month, disagreement and personal susceptibility,  curb the development potential.
Just my opinion anyway, as a newbie without any Github and Java skill ;-)

Sami32:
7.3.2.31     DDC UPnP boolean types
[GUIDELINE]    UPnP endpoints (devices and control points) shall use "0" for false and "1" for true 
when using the UPnP Boolean type.
[COMMENT]   This simplifies control point  implementations and  also  reduces the size  of some 
UPnP traffic.

Nadahar:
To repeat myself I think this should be changed and should only improve compatibility.

SubJunk:
Okay let's do this change. I can't reopen this PR for some reason so feel free to resubmit it

Nadahar:
@SubJunk I'm guessing it's because @Sami32 has deleted the PR branch. The changes are tiny though, so it would be very easy for anyone to redo. What I'm wondering is how complete this is. Is this just one random place in the code, or does this cover the use of "true"/"false" in contentdirectory?

PR #911, opened 2016-06-06
Change Image parsing to Imaging instead of MediaInfo and Sanselan.

valib:
MediaInfo doesn't provide sufficient information and Sanselan is quite old so the follower Imaging is used.

---
This change is [Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/911)

PR #943, opened 2016-07-12
Image transcode

valib:
Divided from https://github.com/UniversalMediaServer/UniversalMediaServer/pull/911#issuecomment-232196310

---

This change is [Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/943)


Sami32:
@valib When compiling i get warning and checked your code
```
        <dependency>
+			<groupId>com.twelvemonkeys.imageio</groupId>
+			<artifactId>imageio-tiff</artifactId>
+			<version>3.2.1</version>
+		</dependency>
```
Get coded 2 times in _pom.xml_

valib:
@Sami32 it is strange. I have the code in the Eclipse OK. I see that here is the code `imageio-tiff` twice but at the page https://github.com/UniversalMediaServer/UniversalMediaServer/blob/image_parsing/pom.xml you can see that everything is OK.

Sami32:
@valib I just verified in the DLNA specification that GIF have only one profile, GIF_LRG, that need to not exceed:1600 x 1200; i was not sure before reading the official documentation.

> 7.3.1.2 [GUIDELINE]   The DLNA GIF_LRG profile shall follow the requirements of GIF 89a ARIB STD  B-1, and the following restrictions and guidelines described in the entries below. 
> File format: GIF
> Copyrighted material licensed to BR Demo by Thomson Reuters (Scientific), Inc., subscriptions.techstreet.com, downloaded on Nov-27-2014 by James Madison. No further reproduction or distribution is permitted. Uncontrolled when printeMaximum resolution the resolution of the image shall not exceed 1 600 pixels in the Width field and shall not exceed  1 200 pixels in the Height field.

So now, i'm wondering if width and height parameters should be added to the supported image lines in the renderer configuration file ? to be used later by your image transcode/resize engine.

What do you think ?

valib:
@Sami32 good idea.

Nadahar:
I think it would be better to try to move away from manual configuration of renderers towards something like @atamariya's "zeroconfig" solution.

Sami32:
@Nadahar I would not agree on that.
My first thought was to doing that automatically depending on the format used but as i read in some others forums, some renderers are poorly DLNA implemented, so the size could depend of the renderer.

SubJunk:
@Nadahar almost none of the renderers I've looked at have properly identified their capabilities so that automatic feature will never be more than a good initial start before we have a better manual one

Sami32:
https://github.com/UniversalMediaServer/UniversalMediaServer/issues/1084

Sami32:
@valib Do this PR still actual ?

PR #950, opened 2016-07-18
WEBVTT

Sami32:
@subjunk @Valib The Samsung renderers have the wrong value, so if you agree to change it...

---
This change is [Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/950)


valib:
@Sami32 WEBVTT is the official name see https://w3c.github.io/webvtt/ and we use `SubtitleType.WEBVTT` in the code so Samsung renderers confs should be updated. :smile:

valib:
@Sami32 thanks

Issue #738, opened 2015-11-23
Stuttering during playback on LG WebOS 2.0 TV


porst17:
I have the issue that almost all my videos are heavily stuttering (every second or so) on my LG Web OS 2.0 TV. First, there is no proper profile for this particular TV, but this doesn't seem to be the issue. For testing, I created a new profile which just streams all media supported by the TV without transcoding. The videos I tested are encoded in formats supported by the TV (e.g. .mkv h.264/ac3).

I also shared the movies via a simple webserver that runs on my box (I can watch all of them in the web browser of my TV without any problems) and via minidlnad (not all videos are listed, but those that are listed are playing well).

Today, I connected my smart phone to UMS using the transcoding-disabled renderer profile I used for testing my TV and magically, the movies are also playing without any stuttering on my smartphone via UMS.

Can anyone please advice me how to debug this issue? It's not a wifi issue and not a codec issue in my opinion. I looks like a buffering issue to me, but I can not find any setting in UMS related to that.

The debug files can be found here: https://gist.github.com/porst17/ad84e0061795bb5b71de
(this is only for the TV connected, not for my smart phone)

porst17:
I just took a look onto the status page of UMS while streaming the files to my TV. The peak bitrate is at about 40Mbps when I start the video, but then goes down to 2Mbps on average. However, when I stream to my smart phone, the bitrate is about a factor of 10 to 20 higher (same router, same network). Since minidlnad streams fine to the TV and the TV plays 1080p on youtube without noticeable lag, I doubt that the bitrate limitation is due to my wifi.

valib:
@porst17 your tested video is not streamed but transcoded because you need to add ac3 audio to the Supported line. Try to change in your LG-UF85.conf line ```Supported = f:mkv v:h264|mp4|vp9 a:aac``` to
```Supported = f:mkv        v:h264|mp4|vp9                  a:aac|aac-he|ac3           m:video/x-matroska```

porst17:
Thanks for your prompt reply. Sorry, I overlooked the missing `ac3` part. Your suggestion does not work either, but when I remove ` m:video/x-matroska` the movies play well.

But does audio transcoding cause such problems usually? Should be pretty lightweight compared to video transcoding. The questions remains: Why is the transcoding to my smart phone fast enough? I used the same (wrong) renderer profile for it.

yesoos:
[...]

PR #953, opened 2016-07-18
LG-UFxxxV.conf

Sami32:
New renderer conf, stricly theorical because i don't have any LG for test it.
Because i saw #738 and other threads in the forum.

**EDIT**: i have done it using LG 65UF850V datas as reference and adjusted it with LG 49UF770V datas.
It's look like, at minima, it will do all ..UF...V serie, but user's computers need to be powerfull to transcode in 4K, so l'm not sure leaving this definition will be advised ? These renderers having integrated upscaler maybe it will be more safe to # them.
The hell is paved with good intention ;-)

**N.B.** The name of this renderer need to be changed _LG-xxUFxxxV.conf_ but i didn't find a way to do it, yet...

---
This change is [Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/953)

### DIFF
--- a/src/main/external-resources/renderers/LG-WebOS4K.conf
+++ b/src/main/external-resources/renderers/LG-WebOS4K.conf
@@ -32,11 +32,11 @@ MaxVideoHeight = 2160
 AutoExifRotate = true
 MediaInfo = true
 
-# Supported video formats:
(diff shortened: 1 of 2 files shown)


Sami32:
Did this line of code is correct `UserAgentSearch = \d{2}UB\d{3}V` ?

Issue #581, opened 2015-06-21
"Prevent Windows From Sleeping"  keeps computer awake even if not streaming


nradisch:
This may be by design but it would be nice to have the option of only
keeping the computer awake while a moving is being streamed.

--Neil

SamyCookie:
It would be a nice feature to have :)

valib:
Now it is implemented to prevent Windows from sleeping when any renderer or device in the network is communicating with the UMS otherwise the sleeping after 40 seconds is re-enabled.

Sami32:
https://github.com/UniversalMediaServer/UniversalMediaServer/pull/959

scriptorron:
I am sure there is a logical bug in the WinUtils.java:
```java
   @Override
   public void disableGoToSleep() {
      // Disable go to sleep (every 40s)
      if (configuration.isPreventsSleep() && System.currentTimeMillis() - lastDontSleepCall > 40000) {
         LOGGER.trace("Calling SetThreadExecutionState ES_SYSTEM_REQUIRED");
         Kernel32.INSTANCE.SetThreadExecutionState(Kernel32.ES_SYSTEM_REQUIRED | Kernel32.ES_CONTINUOUS);
         lastDontSleepCall = System.currentTimeMillis();
      }
   }

   @Override
   public void reenableGoToSleep() {
      // Reenable go to sleep
      if (configuration.isPreventsSleep() && System.currentTimeMillis() - lastGoToSleepCall > 40000) {
         LOGGER.trace("Calling SetThreadExecutionState ES_CONTINUOUS");
         Kernel32.INSTANCE.SetThreadExecutionState(Kernel32.ES_CONTINUOUS);
         lastGoToSleepCall = System.currentTimeMillis();
      }
   }
```

The "if" prevents calling SetThreadExecutionState faster than every 40 seconds. That's a good idea but there are situations where it has to be called more often! For instance:

1.     disableGoToSleep is called (streaming started)
2.     after 1 minute reenableGoToSleep is called (streaming finished)
3.     a few seconds later disableGoToSleep is called (streaming is started again)
4.     a few seconds later reenableGoToSleep is called (streaming finished again)

The second disableGoToSleep will be executed but the second reenableGoToSleep will not! Even though the streaming is finished the PC will not go to sleep anymore.
You can also find examples where a needed disableGoToSleep is suppressed.


I would suggest to change the conditions in the "if" clauses:
```java
   @Override
[...]

PR #959, opened 2016-07-21
Try to fix the problem when http V2 prevents the PC sleeping even if the UMS is not streaming.

valib:
@UniversalMediaServer/developers it needs a comprehensive testing.

---

This change is [Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/959)

### DIFF
--- a/src/main/java/net/pms/io/WinUtils.java
+++ b/src/main/java/net/pms/io/WinUtils.java
@@ -83,7 +83,7 @@ public class WinUtils extends BasicSystemUtils {
 	public void disableGoToSleep() {
 		// Disable go to sleep (every 40s)
 		if (configuration.isPreventsSleep() && System.currentTimeMillis() - lastDontSleepCall > 40000) {
-			LOGGER.trace("Calling SetThreadExecutionState ES_SYSTEM_REQUIRED");
+			LOGGER.trace("Enable Windows sleep mode prevention");
 			Kernel32.INSTANCE.SetThreadExecutionState(Kernel32.ES_SYSTEM_REQUIRED | Kernel32.ES_CONTINUOUS);
 			lastDontSleepCall = System.currentTimeMillis();
 		}
@@ -96,7 +96,7 @@ public class WinUtils extends BasicSystemUtils {
 	public void reenableGoToSleep() {
 		// Reenable go to sleep
 		if (configuration.isPreventsSleep() && System.currentTimeMillis() - lastGoToSleepCall > 40000) {
-			LOGGER.trace("Calling SetThreadExecutionState ES_CONTINUOUS");
+			LOGGER.trace("Disable the Windows sleep mode prevention");
 			Kernel32.INSTANCE.SetThreadExecutionState(Kernel32.ES_CONTINUOUS);
 			lastGoToSleepCall = System.currentTimeMillis();
 		}
--- a/src/main/java/net/pms/io/WinUtils.java
+++ b/src/main/java/net/pms/io/WinUtils.java
@@ -61,9 +61,10 @@ public class WinUtils extends BasicSystemUtils {
 		);
 
 		int SetThreadExecutionState(int EXECUTION_STATE);
+		int ES_CONTINUOUS = 0x80000000;
 		int ES_DISPLAY_REQUIRED = 0x00000002;
 		int ES_SYSTEM_REQUIRED = 0x00000001;
-		int ES_CONTINUOUS = 0x80000000;
+		int ES_AWAYMODE_REQUIRED = 0x00000040;
 
 		int GetACP();
 		int GetOEMCP();
@@ -84,7 +85,7 @@ public class WinUtils extends BasicSystemUtils {
 		// Disable go to sleep (every 40s)
 		if (configuration.isPreventsSleep() && System.currentTimeMillis() - lastDontSleepCall > 40000) {
 			LOGGER.trace("Enable Windows sleep mode prevention");
-			Kernel32.INSTANCE.SetThreadExecutionState(Kernel32.ES_SYSTEM_REQUIRED | Kernel32.ES_CONTINUOUS);
+			Kernel32.INSTANCE.SetThreadExecutionState(Kernel32.ES_CONTINUOUS|Kernel32.ES_AWAYMODE_REQUIRED);
 			lastDontSleepCall = System.currentTimeMillis();
 		}
 	}
--- a/src/main/java/net/pms/network/Request.java
+++ b/src/main/java/net/pms/network/Request.java
@@ -265,10 +265,12 @@ public class Request extends HTTPResource {
 
 		if ((method.equals("GET") || method.equals("HEAD")) && argument.startsWith("console/")) {
 			// Request to output a page to the HTML console.
+			PMS.get().getRegistry().disableGoToSleep();
 			output(output, "Content-Type: text/html");
 			response.append(HTMLConsole.servePage(argument.substring(8)));
 		} else if ((method.equals("GET") || method.equals("HEAD")) && argument.startsWith("get/")) {
 			// Request to retrieve a file
+			PMS.get().getRegistry().disableGoToSleep();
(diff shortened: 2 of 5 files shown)


valib:
@SubJunk you disabled pushing directly to master but there is a lot of PRs whose need to be evaluated and possibly merged like this. Users are asking for it like http://www.universalmediaserver.com/forum/viewtopic.php?f=9&t=4110 and there is no respose of developers for this PR and no decision to merge it or not.

src/main/java/net/pms/io/WinUtils.java
```diff
@@ -83,7 +83,7 @@ boolean GetVolumeInformationW(
 	public void disableGoToSleep() {
 		// Disable go to sleep (every 40s)
 		if (configuration.isPreventsSleep() && System.currentTimeMillis() - lastDontSleepCall > 40000) {
-			LOGGER.trace("Calling SetThreadExecutionState ES_SYSTEM_REQUIRED");
+			LOGGER.trace("Disable the PC sleeping mode");
```
Nadahar: I'd call it something like "Enable Windows sleep mode prevention" instead and corresponding "Disable Windows sleep mode prevention" below. This is more correct, as UMS cannot enable and disable sleep mode which is done by the user in the OS configuration. What UMS can do is to prevent sleep mode while some thread is running.

Nadahar:
I don't quite get GitHub's new review system, it seems my line note can't be submitted - or they might already be. I just know that whatever I do it still says "Finish your review" beneath the line note. That's why I submitted a "forced comment" above, but it didn't change anything.

src/main/java/net/pms/io/WinUtils.java
valib: Done

valib:
This PR was rebased on the current master.

Sami32:
@Nadahar Do you still want block this PR ?

Nadahar:
Has there been proper testing?

Sami32:
@Nadahar I already compiled it, so i'll test it later.

Sami32:
@Nadahar It worked for me.
Did you tried it ? It would be more safe to have an other test with Windows 8/10, IMO

Nadahar:
@Sami32 No, I didn't try it - I don't use sleep mode. Instead I turn of the monitor, harddrives, throttle down CPU etc - which means that the computer is always available. The only place I find sleep mode useful is for laptops. I always has so much stuff open that I usually don't reboot for weeks or months unless the power goes out. Getting everything opened again is a hazzle and takes me several minutes, so I don't want to risk anything by testing sleep mode as it can often mess things up (not all software and drivers handle a wake up from sleep mode very well).

Sami32:
@Nadahar As you know my old one don't use the sleep mode as well, though i shutdown it every day.
But i did for test this PR only.
Can you get a simulation on your virtual machine ?

I guess that @valib did already tested, so @SubJunk ? Do you want test it or it"s fine for you to be merged ?

https://github.com/UniversalMediaServer/UniversalMediaServer/issues/1218

Nadahar:
I could test at some VMs, but it takes time to test something like this properly. It's not enough to test it one time and check that it doesn't go to sleep, you need to start and stop with multiple renderers to check that it doesn't get "confused".

I'm currently trying to work on the memory consumption of #1061, which I think is better use of my time.

Nadahar:
Taking a quick look at the diff, I can't imagine that this works correctly, there doesn't seem to be corresponding ```reenableGoToSleep()``` calls for each ```disableGoToSleep()```. It also disables sleep for a lot of operations that shouldn't - like exchanging information with renderers that happens periodically. The whole approach is to simple IMO, what needs to be done is some kind of (thread safe) reference count implementation that disables sleep mode as soon as the reference > 0 and only reenables sleep when the reference is at 0 again. Now, starting two playbacks and stopping one will re-enable sleep mode. Things like seeking and similar can start and stop multiple playbacks and will completely confuse this implementation.

I also think it would be more appropriate to use ```ES_AWAYMODE_REQUIRED``` instead of ```ES_SYSTEM_REQUIRED```.

Sami32:
@Nadahar That seem a good idea.
Did i implmented `ES_AWAYMODE_REQUIRED` correctly ?

Nadahar:
@Sami32 I don't know - the replacement of the value seems to be correct, but implementing something like that also involves studying how it works and verifying that the expected result is achieved.

Regardless of which "mode" is sent to the Windows API, the important thing is to properly control it's use. That's where I think reference counting would be the way to go. This is what needs to be solved, reference counting or not - to make sure that UMS keeps properly track of when to enable and disable sleep mode.

Sami32:
https://github.com/UniversalMediaServer/UniversalMediaServer/issues/581#issuecomment-286546163

Nadahar:
@Sami32 The problem isn't to see the need, the problem is that nobody has done the work (to make it properly).

Sami32:
@nadahar "No problems, only solutions" ;)
I don't know if i translated the citation correctly though...

Nadahar:
@Sami32 Nomatter how you put it or what you call it, somebody has to actually make a working solution. The current solution can't work properly, it's too simple and doesn't take into account that there will be multiple calls to both enable and disable sleep mode.

Sami32:
@Nadahar It's the every day life on earth, each generation make things better. I hope that the human race do it as well ? :wink:

Nadahar:
@Sami32 I don't know where you're looking if you think the world keeps getting a better place. In some aspects I agree, but in many it keeps getting worse IMO.

I'm not sure I see the connection to this though, it's no problem to fix this - but it must be done by somebody. Talking/writing about it won't contribute though.

Sami32:
@nadahar I guess that today is a day without "humour" ?

valib:
@Sami32 have you test it? The solution I made is realy simple. When any renderer is calling any of the requests which are relevant then it is disabling go to sleep for the 40 seconds (works only for WIN OS). If none of that calls are received then OS is going to sleep after the time which is set in the general WIN setting.

Nadahar:
@valib The functionality I thought we were trying to achieve was to prevent sleep while media was playing, and that is also what the existing implementation attempts. The problem with the current implementation as I see it, is that it doesn't keep track of the "commands" it sends to Windows so the result can be both that Windows never goes to sleep or that it goes to sleep while playing media.

What is the point of just disabling sleep for a given time from a certain even? During e.g a movie playback, no such "events" will occur unless you pause, seek etc. That means that the computer can go to sleep while you're playing media. But, I also saw that you "reset the timer" on some UPnP communication. This communication is repeated with given time intervals, which means that if the OS "go to sleep" time is longer than one of the intervals, it will never go to sleep as long as the renderer is turned on. You can't assume that people will turn off the renderer for the computer to go to sleep, the renderer can have other unrelated tasks, be a computer, mobile phone etc.

Sami32:
@valib Yes, i did and it worked for me. Also with the latest commit.

So as @Nadahar like repeat everytime ;), it's not perfect, but since it's an improvement asked by users i'll not mind if your merge it, as improvements could be added later.
I don't see how things could be worst with that PR.
Better now than never, so i approve it :wink:

Nadahar:
@Sami32 @valib I think we need to estabilish what the **intended behaviour** is. My understanding is/was that UMS was supposed to prevent sleep while media is playing. This doesn't do that.

Do you have an alternative behaviour you see as the intended? If so, please explain to me.

valib:
@Nadahar that is why I set disableGoToSleep() only for the requests which are relevant for browsing or plaiyng media. If the 40 seconds is not enough for plaiyng media then we can simply increase the time to the 5 or more minutes.

Nadahar:
@valib I don't understand - playing a movie can be 2 or 3 hours - how will it help to increase it to 5 minutes?

Sami32:
https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1226

Sami32:
Should have been fixed in #1226

PR #964, opened 2016-07-27
Update INSTALL.txt

Sami32:
Linux installation documentation. Please read it and make sure i didn't write something wrong !

---

This change is [Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/964)

### DIFF
--- a/INSTALL.txt
+++ b/INSTALL.txt
@@ -2,20 +2,94 @@ Installers for Windows and Mac OS X and tarballs for Linux can be found at:
 
     http://sourceforge.net/projects/unimediaserver/files/Official%20Releases/
 
-To install UMS from the tarball on Linux, open a terminal and enter the
-following commands (replace <version> with the version you're installing):
-Note: These instructions have been tested on Ubuntu 12.04, but something similar
-should work on most Unix distributions)
+To install UMS from the tarball on Linux, open a terminal and enter the following commands (replace <version> with the version
+you're installing):
+Note: These instructions have been tested on Ubuntu 12.04, but something similar should work on most Unix distributions)
 
 1) Install the dependencies (this only needs to be done once):
 
-    sudo apt-get install mediainfo openjdk-7-jre
-    # you can also optionally install dcraw and VLC: sudo apt-get install dcraw vlc
+    # UMS requires a Java Virtual Machine to run. Which version depends on which binary package was downloaded. UMS for Java 7
+    would need version 7 or newer, UMS for Java 8 would need version 8 or newer. The most common JVM's to use on Linux is
+    openjdk. Which, if any, of the packages that's available depends on your distro, it's version and what sources are enabled.
+
+    # Install JVM on Debian/Ubuntu and derivates:
+      Debian 8 need to use backports to install openjdk-8 at the time of writing. It's not possible to cover all the sources
+      needed for the different versions in this document. If the package isn't found, make sure to run `apt-get update` to
+      make sure the sources are refreshed.
+      # To install openjdk-7 run:
+      sudo apt-get install openjdk-7-jre
+      # To install openjdk-8 run:
+      sudo apt-get install openjdk-8-jre
+
+    # Install JVM on Fedora, OpenSUSE, Mandriva, Red Hat/Centos and derivates: The epel repository might be needed on some
+    platforms.
+      # To install openjdk-7 run:
+      sudo yum install java-1.7.0-openjdk
+      # To install openjdk-8 run:
+      sudo yum install java-1.8.0-openjdk
+
+    # Install JVM on Arch:
+      # To install openjdk-7 run:
+      sudo pacman -S jre7-openjdk
+      # To install openjdk-8 run:
+      sudo pacman -S jre8-openjdk
+
+    # UMS use MediaInfo to get information about media files. It can be run without it, but media handling will suffer as UMS
+    won't have a detailed information available.
+    # Install MediaInfo on Debian/Ubuntu and derivates:
+    sudo apt-get install mediainfo
+    # Install MediaInfo on Fedora, OpenSUSE, Mandriva, Red Hat/Centos and derivates:
+    sudo yum install mediainfo
+    # Install MediaInfo on Arch (community repository must be enabled):
+    sudo pacman -S mediainfo
+
+    # You can optionally install dcraw (to decode raw images from digital cameras) and VLC (to use as a transcoding engine).
+    Just omit one if you don't want that.
+    # Install dcraw and VLC on Debian/Ubuntu and derivates:
+    sudo apt-get install dcraw vlc
+    # Install dcraw and VLC on Fedora, OpenSUSE, Mandriva, Red Hat/Centos and derivates:
+    sudo yum install dcraw vlc
+    Note: Before installing VLC, you'll need to enable the free RPM Fusion repository in your system:
+    sudo yum install --nogpgcheck https://download1.rpmfusion.org/free/fedora/rpmfusion-free-release-$(rpm -E%fedora).noarch.rpm
+    # Install VLC on Fedora 22 and later:
+    sudo dnf -y install vlc
+    Note: Before that, you'll need to enable the free RPM Fusion repository in your system:
+    sudo dnf install http://download1.rpmfusion.org/free/fedora/rpmfusion-free-release-$(rpm -E %fedora).noarch.rpm
+    # Install dcraw and VLC on Arch (extra repository must be enabled):
+    sudo pacman -S dcraw vlc
+
+    # UMS uses tsMuxeR as a transcoding engine. tsMuxeR is 32 bit only, so installation of some 32 bit libraries are necessary
+    on 64 bit platforms if you want to use tsMuxeR as a transcoding engine.
+    # Install tsMuxeR dependencies on 64 bit Debian/Ubuntu and derivates:
+    sudo dpkg --add-architecture i386
+    sudo apt-get update
+    sudo apt-get install lib32z1 lib32ncurses5 libbz2-1.0:i386 libstdc++6:i386
+    # Install tsMuxeR dependencies on 64 bit Fedora, OpenSUSE, Mandriva, Red Hat/Centos and derivates:
+    sudo yum install zlib.i686 ncurses-libs.i686 bzip2-libs.i686 libstdc++.i686
+    # Install tsMuxeR on Fedora 22 and later;
+    sudo dnf install zlib.i686 ncurses-libs.i686 bzip2-libs.i686 libstdc++.i686
+    # Install tsMuxeR dependencies on 64 bit Arch (multilib repo must be enabled):
+    sudo pacman -S lib32-zlib lib32-ncurses lib32-bzip2 lib32-libstdc++5
+
+    # You can optionally install P7Zip, UMS will use it for archives browsing.
+    # Install P7Zip on Debian/Ubuntu and derivates:
+    sudo apt-get install p7zip
+    # Install P7Zip on Fedora, OpenSUSE, Mandriva, Red Hat and derivates:
+    sudo yum install p7zip
+    # or on Fedora 22 and later:
+    sudo dnf install p7zip
+    # Install P7Zip on Arch:
+    sudo pacman -S p7zip
+    # Install P7Zip on RHEL/CentOS:
+    sudo yum install p7zip
+
+    # Other distros and/or package managers not covered here will need to have the same/corresponding packages installed.
+    If a package isn't available, the binary must be downloaded and installed or built from source.
 
 2) Download the tarball e.g.:
 
     # if wget isn't installed, run: sudo apt-get install wget
-    wget http://sourceforge.net/projects/unimediaserver/files/Official%20Releases/Windows/UMS-<version>.tgz/download
+    wget https://sourceforge.net/projects/unimediaserver/files/Official%20Releases/Linux/UMS-<version>.tgz/download
 
 3) Extract the tarball into a ums-<version> directory:
 
@@ -26,5 +100,15 @@ should work on most Unix distributions)
     cd ums-<version>
     ./UMS.sh
 
+5) The user may have to manually copy these files to his profile folder:
+
+    UMS.conf
+    WEB.conf
+    ffmpeg.webfilters
+    VirtualFolders.conf
+
 UMS accesses some files in the ums-<version> directory (the working directory).
 Other files will be looked for in ~/.config/UMS
+
+Note: On a Linux 32-bit system, use files bigger than 2GB is generally not possible by the current default implementation even
+with a file system supporting bigger files size, you will need to add Large File Support (LFS).
--- a/src/main/assembly/assembly-linux.xml
+++ b/src/main/assembly/assembly-linux.xml
@@ -89,11 +89,15 @@
 				<include>WEB.conf</include>
 				<include>logback.headless.xml</include>
 				<include>logback.xml</include>
+				<include>documentation/**</include>
 				<include>linux/*.txt</include>
 				<include>plugins/**</include>
 				<include>renderers/*</include>
 				<include>VirtualFolders.conf</include>
 				<include>DummyInput.ass</include>
+				<include>DummyInput.jpg</include>
+				<include>web/**</include>
+				<include>ffmpeg.webfilters</include>
 			</includes>
 		</fileSet>
 
--- a/src/main/assembly/assembly-osx.xml
+++ b/src/main/assembly/assembly-osx.xml
@@ -91,6 +91,7 @@
 				<include>DummyInput.ass</include>
 				<include>DummyInput.jpg</include>
 				<include>web/**</include>
+				<include>ffmpeg.webfilters</include>
 			</includes>
 		</fileSet>
 
--- a/src/main/assembly/assembly-osx_java7_8.xml
+++ b/src/main/assembly/assembly-osx_java7_8.xml
@@ -81,6 +81,7 @@
 				<include>DummyInput.ass</include>
 				<include>DummyInput.jpg</include>
 				<include>web/**</include>
+				<include>ffmpeg.webfilters</include>
 			</includes>
 		</fileSet>
                 


Nadahar:
@Sami32 I think it's hard to interpret what to do from that. Here's my suggestion, although it whould be checked for correctness on different distros:

```
    # UMS requires a Java Virtual Machine to run. Which version depends on which binary package was downloaded. UMS for Java 7 would need version 7 or newer, UMS for Java 8 would need version 8 or newer. The most common JVM's to use on Linux is openjdk. Which, if any, of the packages that's available depends on your distro, it's version and what sources are enabled. 

    # Install JVM on Debian/Ubuntu and derivates: Debian 8 need to use backports to install openjdk-8 at the time of writing. It's not possible to cover all the sources needed for the different versions in this document. If the package isn't found, make sure to run ```apt-get update``` to make sure the sources are refreshed. 
      # To install openjdk-7 run:
      sudo apt-get install openjdk-7-jre
      # To install openjdk-8 run:
      sudo apt-get install openjdk-8-jre

    # Install JVM on Fedora, OpenSUSE, Mandriva, Red Hat/Centos and derivates: The epel repo might be needed on some platforms. 
      # To install openjdk-7 run:
      sudo yum install java-1.7.0-openjdk
      # To install openjdk-8 run:
      sudo yum install java-1.8.0-openjdk

    # Install JVM on Arch:
      # To install openjdk-7 run:
      sudo pacman -S jre7-openjdk
      # To install openjdk-8 run:
      sudo pacman -S jre8-openjdk

    # UMS use MediaInfo to get information about media files. It can be run without it, but media handling will suffer as UMS won't have a detailed information available.
    # Install MediaInfo on Debian/Ubuntu and derivates:
    sudo apt-get install mediainfo
    # Install MediaInfo on Fedora, OpenSUSE, Mandriva, Red Hat/Centos and derivates:
    sudo yum install mediainfo 
    # Install MediaInfo on Arch (community repository must be enabled):
    sudo pacman -S mediainfo

    # You can optionally install dcraw (to decode raw images from digital cameras) and VLC (to use VLC as a transcoding engine). Just omit one if you don't want that.
    # Install dcraw and VLC on Debian/Ubuntu and derivates:
    sudo apt-get install dcraw vlc
    # Install dcraw and VLC on Fedora, OpenSUSE, Mandriva, Red Hat/Centos and derivates:
    sudo yum install dcraw vlc
    # Install dcraw and VLC on Arch (extra repository must be enabled):
    sudo pacman -S dcraw vlc

    #UMS uses tsMuxeR as a transcoding engine. tsMuxeR is 32 bit only, so installation of some 32 bit libraries are necessary on 64 bit platforms if you want to use tsMuxeR as a transcoding engine.
    # Install tsMuxeR dependencies on 64 bit Debian/Ubuntu and derivates:
    sudo dpkg --add-architecture i386
    sudo apt-get update
    sudo apt-get install lib32z1 lib32ncurses5 libbz2-1.0:i386 libstdc++6:i386
    # Install tsMuxeR dependencies on 64 bit Fedora, OpenSUSE, Mandriva, Red Hat/Centos and derivates:
    sudo yum install zlib.i686 ncurses-libs.i686 bzip2-libs.i686 libstdc++.i686
    # Install tsMuxeR dependencies on 64 bit Arch (multilib repo must be enabled):
    sudo pacman -S lib32-zlib lib32-ncurses lib32-bzip2 lib32-libstdc++5

    # Other distros and/or package managers not covered here will need to have the same/corresponding packages installed. If a package isn't available, the binary must be downloaded and installed or built from source. 
```
This is just a draft, this should be corrected by someone who knows the individual platforms, but it's my attempt to give somewhat broader installation hints. I've verified that the packages exist on Debian 8 and Centos 7, but there are many distros I haven't tested. Arch commands are completely untested and relies on online references only.

Sami32:
@Nadahar It look great to me :+1: 
As i'm not a Linux user so my ability to write or review it is near to zero.
I doesn't have problem if you want make a PR, on the contrary i will enjoy it, so don't hesitate.
You work is much better than mine ;-)

Nadahar:
@Sami32 I'd much rather that you do the PR, as I'm not really committed to this at the moment and just throw in some sentences when I have the time. You can use what you want of my text "as yours", the only thing that matters is a good end result imo. I know that you will see it through while I might not.

When it comes to all the different Linux distros and versions, it's simply hopeless to give a full description. My thinking is: Let the users know WHAT they need to install. Even if the name of the lib isn't an exact match for their platform, it will give them a pretty good idea of what to look for. When it comes to exactly which repository they have to add/enable to find those packages, or if they have to download the binaries or build them from source manually they have to figure that out themselves. That's just the way it is with Linux.

It makes some sense to try to give the commands to install for the most widespread distros, since some users might "get away with" using one of them without knowing to much about where to find packages, their naming etc. Users of less "mainstream" flavors will have to know how to do this in any case, or they won't be able to get anything running on their platform. To them it's most important to know *what* to install, not in detail *how* to.

When it comes to BSD, it's really just way out of my comfort zone and I consider it "non-mainstream". I do have a couple of FreeBSD servers running, but that is for custom purposes only (pfSense). I never actually do anything "OS level" on those. Technically Free/OpenBSD isn't even Linux, and I'm sure there would be a host of other issues getting all the non-Java supporting libraries and programs to work with UMS. My recommendation is simply to leave BSD out until someone actually does the job of testing UMS on BSD and finding all the quirks needed to get things running.

Sami32:
@Nadahar Thank for your nice work :+1: 
I updated the _INSTALL.txt_ file and the post on the forum http://www.universalmediaserver.com/forum/viewtopic.php?f=10&t=8518
Like usual, you maked good point, as your job since i didn't saw anything to change ;)

Sami32:
@SubJunk OK ?

Nadahar:
@Sami32 It seems more has to be done with this. There's some uncertainty to exactly what dependencies tsMuxeR has under Linux, and I'm not sure how to figure that out. I also think the presentation itself is a bit messy, it might be better to first give general instructions and then give examples for each included distro together instead of having each distro listed for each step as it is now.

Sami32:
@Nadahar No problem, i just need feedback.
So it's why i didn't get answer when i asked you for which one choose ;-)
I read that last one that i told you is only used in Ubuntu 16, but since i don't have any Linux experience...

Concerning the presentation, for me the important is that is will easily understood and simple, i leave it to you, no problem.

INSTALL.txt
```diff
@@ -101,3 +101,6 @@ Note: These instructions have been tested on Ubuntu 12.04, but something similar
 
 UMS accesses some files in the ums-<version> directory (the working directory).
 Other files will be looked for in ~/.config/UMS
+
+Note: On a Linux 32-bit system, use files bigger than 2GB is generally not possible by the current default implementation even
+with a file system supporting bigger files size, you will need to add Large File Support (LFS).
```
Nadahar: This is premature, wouldn't it be a good idea to figure out the exact limitations before adding the documentation? I for one don't know if it only applies to 32 bit Linux or if it also applies to 64 bit Linux with 32 bit JVM. Neither do I know if the 32 bit JVM supports LFS, chances are that adding LFS won't make any difference.

Nadahar:
I'm currently trying to install on Fedora 24. Here are the complications not covered by the above instructions:

* yum is deprecated, dnf is the way to go now. Most command work if you replace ```yum``` with ```dnf```.
* Fedora has it's own definition of "free" software meaning that commercial use must also be allowed. A lot of OSS software is therefore not available like VLC. To get around this, the following is needed:
```dnf install https://download1.rpmfusion.org/free/fedora/rpmfusion-free-release-$(rpm -E %fedora).noarch.rpm https://download1.rpmfusion.org/nonfree/fedora/rpmfusion-nonfree-release-$(rpm -E %fedora).noarch.rpm``` 
For Fedora 22 and earlier the command is:
```yum install --nogpgcheck https://download1.rpmfusion.org/free/fedora/rpmfusion-free-release-$(rpm -E %fedora).noarch.rpm https://download1.rpmfusion.org/nonfree/fedora/rpmfusion-nonfree-release-$(rpm -E %fedora).noarch.rpm```
* The 32 bit libs must be specified by version to be found, e.g:
```dnf install zlib-1.2.8-10.fc24.i686 ncurses-libs-6.0-6.20160709.fc24.i686 bzip2-libs-1.0.6-20.fc24.i686 libstdc++-6.2.1-2.fc24.i686```

**Comment by Sami32** [comment]
@Nadahar Thanks for reporting :+1: 
I updated the wiki part for VLC, but i just include the strict necessary free library.

About the 32-bit library version, i'm suspicious as i never read or listen that before.
Do you have these libraries already installed on your system as 64-bit version before doing that ?
If you use `ncurses-libs*.i686`, for example, did it work ?

**Comment by Nadahar** [comment]
@Sami32 I'd like to add that non-free has nothing to do with what normal people think of as non-free. There aren't that many packages in the library, the majority is nvidia drivers, and there are things like skype, steam and spotify there, but as VLC is found in the "free" library I agree that's all we need.

About the 32 bit libraries, those are all needed for tsMuxer only - since it is 32 bits only. Those packages already exist (by default) in 64 bits versions on a 64 bit system yes. This is what is already covered above by:

```
    #UMS uses tsMuxeR as a transcoding engine. tsMuxeR is 32 bit only, so installation of some 32 bit libraries are necessary on 64 bit platforms if you want to use tsMuxeR as a transcoding engine.
    # Install tsMuxeR dependencies on 64 bit Debian/Ubuntu and derivates:
    sudo dpkg --add-architecture i386
    sudo apt-get update
    sudo apt-get install lib32z1 lib32ncurses5 libbz2-1.0:i386 libstdc++6:i386
    # Install tsMuxeR dependencies on 64 bit Fedora, OpenSUSE, Mandriva, Red Hat/Centos and derivates:
    sudo yum install zlib.i686 ncurses-libs.i686 bzip2-libs.i686 libstdc++.i686
    # Install tsMuxeR dependencies on 64 bit Arch (multilib repo must be enabled):
    sudo pacman -S lib32-zlib lib32-ncurses lib32-bzip2 lib32-libstdc++5
```

**Comment by Sami32** [comment]
@Nadahar I didn't know that implied that 64-bit libraries version are necessary already installed...
So, if they really already are, that could explain why you need to precise the libraries version.

But you didn't answered my question: Do using something like that `ncurses-libs*.i686` will work ?
As the libraries update sometimes, better to not get version specific in the documentation, if possible of course.

**Comment by Nadahar** [comment]
@Sami32 It doesn't mean that the 64 bit libraries are necessarily installed, but I think those libraries are very basic in Linux and generally are available. That is the reason it's only necessary to install these manually when running 64 bit Linux, on a 32 bit Linux they will already be there. But, nothing is really standard in Linux so I can't say that this is always the case. It's not relevant here though, as tsMuxer which need these, can only use the 32 bit versions.

I didn't need to install these specific versions, it was simply due to Fedora's naming convention (putting ```.i686``` after the version number).

The reason I didn't answer your question was that I didn't understand it, but now I do: Yes, I tested and ```dnf install ncurses-libs*.i686``` does work and that's great because it's bad to document specific versions. My general Linux knowledge is limited, and when it comes to Fedora and dnf this was the first time I used it, so I simply didn't think of trying ```*```. Thanks :+1:

**Comment by Sami32** [comment]
@Nadahar Thanks for you feedback, as you know my Linux knowledge is theorical...meaning from internet documentation.

In fact it was relevant in this case, as a rule say that the 32-bit versions need to have the same version as the 64-bit. It's also why they normally ask users to upgrade their 64-bit libraries before installing the 32-bit one.
Without 64-bit versions installed i read that install `dnf ncurses-libs.i686` or `ncurses-devel.i686`, for example, will be fine.

**Comment by Nadahar** [comment]
@Sami32 Ok, I did not know :smiley:

**Comment by Nadahar** [comment]
@SubJunk I just discovered when doing a Linux install that MEncoder isn't distributed. Is that intentional, and if so: why?

If people are supposed to install MEncoder manually it should be in the installation description, and I don't know if there are any configuration needed to make it work..?

What about OS X - is it the same there?

**Comment by SubJunk** [comment]
@Nadahar it's not intentional but I couldn't get a fully static version of MEncoder to build for Linux. I tried really hard to get that done because I don't like that we don't have it. Windows and OS X both have static MEncoder included.

**Comment by Nadahar** [comment]
@SubJunk Ok, I don't know how much of a difference your custom build makes, but the documentation should also instruct them to install MEncoder then - that would at least make MEncoder available.

**Comment by SubJunk** [comment]
@Nadahar my custom build is mostly useful for DVD playback, but also audio/video synchronization. The docs should definitely mention it, you're right. Also we need a Linux static builds wiz to come build for us because maintaining support for MEncoder when people could be using any version would be a nightmare given how unstable the MEncoder project has been at times

**Comment by Nadahar** [comment]
I've never built any C program on any platform (except ready made "makefile" builds on Linux, but that doesn't count), so I'm not in a position to contribute to that. Static in C means that the used parts of dependencies is compiled "into" the executable so that you don't need them to be available and isn't relying on the versions existing on the computer, right?

If that's the case, I definitly agree. Imo that's one of the biggest hassle with Linux, the endless dependency issues and conflicts - I don't get why more isn't compiled static. I know there are some license issues in some cases, but it seems to me to be a general idea of using dynamic linking to save harddrive space - which simply isn't worth it in most cases in my view. The size of compiled executables is negligible most of the time today with the current cost per megabyte, and I'd prefer something that works a thousand times to something that only complains of missing dependencies or doesn't work as intended that saves me 10 MB.

**Comment by Nadahar** [comment]
@Sami32 As discussed above, mencoder also needs to be installed manually on Linux.

I have only tested that on Fedora and Debian, but it's straight forward on those so we can assume the same for the other flavours. By straight forward I mean ```apt-get install mencoder``` for Debian/Ubuntu, ```yum install mencoder``` or ```dnf install mencoder``` for Fedora > 22, ```yum install mencoder``` for Fedora <= 22 and Centos etc.

**Comment by Sami32** [comment]
@Nadahar I read, but didn't want to, as i thought perhaps a static build could be done.
So i like more wait and if i cannot make it before Christmas i'll add it in comments.

But if i do that, i'll perhaps no write your way, but instead use the very recent version up to date available in RPMFusion repository ;-)

**Comment by Nadahar** [comment]
@Sami32 I don't think you can make a static Linux build from Windows, so to do that you'd first have to install some Linux version.

PS - I just checked on Centos 7, and MEncoder isn't available in the standard repo's there :disappointed:

**Comment by Sami32** [comment]
@Nadahar Sorry, i meaned only for Fedora users, so you are lucky ;-)

It's why i need to wait my new HDD :wink:

**Comment by Sami32** [comment]
@Nadahar In fact, it's look that you can do that for Cent OS.
But it's like for RHEL, meaning that you need to enable EPEL before usin RPM Fusion repository.

**Comment by Nadahar** [comment]
@Sami32 Yes, RPMFusion is available for Centos, but I have no reason to enable it for this installation as it's a server doing something completely different. I wasn't going to install mencoder here, I just tried to see if I found the package.

**Comment by Sami32** [comment]
@Nadahar If i understood correctly, when the packages are in RPM Fusion it's because they are unsupprrted in standard repository...so it'll be same for RHEL and Fedora.

**Comment by Nadahar** [comment]
@Sami32 You are correct - I forgot that RPM Fusion is already enabled on my Fedora installation, so that is why MEncoder was available there.

**Comment by Sami32** [comment]
I'm trying to document for MAC users, but i never used one myself, so any help is welcome.

How it's UMS work on MAC ? I mean they need to run _ums.app_ ?
When doing a Control-click on the UMS icon link, do the "Properties" and "Shortcut" tab and "Target" appear as well ?

**Comment by Nadahar** [comment]
@Sami32 Here's a quick overview over the process:

open_dmg
After downloading the file, click on the downloaded ```dmg``` (here ```UMS-6.5.1-Java7.dmg```)

view_dmg
The ```dmg```'s content is shown. Once there, drag ```Universal Media Server``` into ```Applications```.

copy_to_applications
The files are copied from the ```dmg``` into the file system. Once the copying is finished it is "installed". 

If you try to open UMS by doubleclicking on the icon, the following blackmail message appears:
osx_blackmail
This is designed to scare users away from using software from developers that don't pay Apple money to "trust" them.

To be allowed to still run the application either Apple-click or right-click (for advanced users that can handle more than one mouse button) and select ```Open``` from the appearing menu. The following confirmation will appear:
confirm
Once you accept the huge risk of running something not "trusted" by Apple, UMS will open normally:
running

OS X will remember the confirmation, so the confirm dialog shouldn't appear when launching UMS later.

**Comment by Sami32** [comment]
@Nadahar Sorry, i guess i wasn't clear.
How to get this on MAC ? Are the steps same as Windows ?
mac

**Comment by Nadahar** [comment]
@Sami32 No, I don't even know if there is something equivalent to "Windows shortcut" in OS X. What you will find in "Applications" is the application files itself, not a shortcut.

I just tested starting it from ```Terminal``` with the ```headless``` argument, but UMS is wrapped in something called ```JavaAppLauncher``` that doesn't seemt to take or forward arguments.

That said, I doubt there are many OS X users out there that's even advanced enough to understand the concept of running headless. When people choose Apple products for their "user friendliness" that means that they don't want to be exposed to anything remotely technical and prefer that a big corporation make all the decisions for them over being forced to think. I just don't see why anyone with this philosophy would consider running UMS headless.

**Comment by Sami32** [comment]
@Nadahar OK. Thanks anyway to tested and report it to me :-)
Could you try `--headless`  or `-vm-display-null` ?

**Comment by Nadahar** [comment]
> Could you try with --headless or -vm-display-null parameter ?

Neither had any effect.

**Comment by Sami32** [comment]
@Nadahar So, i will leave the MAC users found the answer by themself ;-)
Thanks for your help.

**Comment by Sami32** [comment]
@Nadahar I just saw that alias can be created:
Example: alias vmrun=/Applications/VMware\ Fusion.app/Contents/Library/vmrun

And that from the alias it will be possible to run headless:
vmrun -T fusion start "/path/to/vm.vmx" nogui

Do that really work ?

**Comment by Nadahar** [comment]
@Sami32 VMware Fusion isn't a Java application, so it's a whole different ball game. The problem with Java is that they have to be launced inside a JVM, so to pass an argument to UMS itself the JVM has to "forward" the argument.

The ```JavaAppLauncher``` used handles starting the JVM and then tells the JVM to run UMS. The problem is getting ```JavaAppLauncher``` to forward the argument to the JVM that again has to forward it to UMS. It doesn't seem like ```JavaAppLauncher``` supports this.

**Comment by Sami32** [comment]
@Nadahar I see.
So, you think that for MAC users, only a modified UMS version with `-Djava.awt.headless=true` will make it work headless ?

**Comment by Sami32** [comment]
@Nadahar I think i foung 2 way for doing that:
Rename /usr/bin/java to /usr/bin/java.ori

Create the following `/usr/bin/java` replacement with 755 permissions:
```
    #!/bin/bash
    java.ori $JAVAOPT "$@"
```
Then you can use the environment variable JAVAOPT (or perhaps `JAVA_OPTIONS` or `JAVA_OPTS` will be better ?) to set global java options. These will also be propagated to any java subprocess that may be spawned:
```
    export JAVAOPT=-Djava.awt.headless=true
    adt ...
```

Any feedback is welcome :wink:

**Comment by Nadahar** [comment]
@Sami32 
> So, you think that for MAC users, only a modified UMS version with -Djava.awt.headless=true will make it work headless ?

I don't think that would work either as UMS doesn't check this because it is said to be unreliable. Instead, UMS simply tries to create a window and if that fails concludes that a graphical environment isn't available. In addition UMS can be forced to run headless via the command line argument ```headless``` or ```console```.

> I think i found 2 way for doing that, here the first one (the "Java global one"):

The first way is a no-go, you don't rename the Java binary it will make all kind of things fail. If this suggestion would work, you could also just as well run it this way directly. The problem is that UMS isn't launched this way, it's launched using ```JavaAppLauncher``` because that's "the Apple way". It's obviously possible to get UMS to run headless on OS X - the main problem is the way UMS is packaged. If we'd just distribute the JAR in an archive like on Linux, I assume that the approach would be very much like Linux. But, that wouldn't make it runnable as an "Application" as OS X defines it.

The second way will fail for the same reason as above.

I'm sure we could find a way around all the "OS X stupidification" and actually pass an argument to UMS. My point is that I don't think it's worth the effort. I doubt anyone with a Mac will use it that way, most of the user base doesn't even know what a terminal does and think it's a scary place never to be opened. The whole reason to choose OS X is to never "see" anything remotely technical. If they are the type of users that would appreciate a headless installation (running it in the background/as a daemon) they wouldn't be running OS X or paying a premium for hardware that has nothing special about it except that it is "approved" by Apple.

**Comment by Nadahar** [comment]
@Sami32 I hope this makes you sleep better, but I found a way to launch it headless under OS X:
```
cd /Applications/Universal\ Media\ Server.app/Contents/Resources
java -jar /Applications/Universal\ Media\ Server.app/Contents/Java/ums-6.5.1-jar-with-dependencies.jar headless
```

The name of the JAR must be adapted to the version used obviously.

**Comment by Sami32** [comment]
@Nadahar Nice that you found one other :+1: 

My second way will work perfectly :wink: 
> Holding the Control key, click on the application's icon.
> Select "Show Package Contents".
> Open the "Contents" folder.
> Open the Info.plist file in PropertyListEditor, located under /Developer/Applications, or with a text editor.

From there, you will be able to add `-Djava.awt.headless=true`.
https://www.dbvis.com/doc/4.1/doc/ug/admin/admin.html

**Comment by Nadahar** [comment]
> "My" second way will work perfectly :stuck_out_tongue_winking_eye:

Have you tested this? I know that UMS doesn't "respect" the JVM's headless setting, but it might be that his setting makes the graphical environment unavailable for UMS and as such *forces* it into headless mode. I haven't tested that (that can be tested on Windows as well). I just know that the JVM's headless state is ignored by UMS because it is unreliable under some circumstances (reporting that a graphical environment is available when it's really not AFAICR).

I did try to edit ```Info.plist``` but adding anything to either JVMOptions or JVMArguments made it fail with a "class not found" (or something similar) error. I really didn't care enough to figure out why. 

Launching it without ```JavaAppLauncher``` as I suggested is cleaner if running headless anyway imo, but it's important that the current folder is set correctly before starting UMS or relative paths won't work and UMS won't find any bundled resources.

**Comment by Sami32** [comment]
@Nadahar No Apple to test...i was teasing you.
The values can be changed, but i guess that adding some need to be done with "Project Builder" ?
I tried already on Windows and it worked headless this way as well, but as you can understand i didn't told the Windows or Linux users this way.

As you say, now that i have a Macintosh to add to the wiki i will sleep better :wink: 
I was near to tell MAC users to throw it by the windows to make it fully headless :-D

**Comment by Nadahar** [comment]
@Sami32 I'd like to tell Mac users to do a lot of things, but I'm trying not to :stuck_out_tongue_closed_eyes:

**Comment by Sami32** [comment]
@Nadahar Did you add `-Djava.awt.headless=true` it under _VMOptions_ in the _Info.plist_ file ?
_Xcode_ editor help: http://help.apple.com/xcode/mac/8.0/#/dev3f399a2a6
```
<key>VMOptions</key>
  <array>
    <string>-Djava.awt.headless=true</string>
  </array>
```
right before the final occurrence of `</dict>`
I don't like give up ;-) That will be be a more easy step for MAC users.

In your solution using the JAR do _**ums-*-jar-with-dependencies.jar headless**_ work ?

**Comment by Nadahar** [comment]
@Sami32 I did use proper XML with ```<string>```, ```<array>``` and the lot, yes. I don't remember exactly which combinations I tested, but it didn't seem to matter. Whenever I had *anything* under either ```VMOptions``` or ```VMArguments``` it wouldn't start. I obviously didn't understand ```-D``` since that normally forwards the argument to the application and shouldn't keep the JVM from starting.

In any case it's a bad solution, because if this were to work there would be no terminal window in which UMS is running. When running headless that terminal window is the only "connection" with UMS, it's where any output appear and it's where you have to click Apple+C to end UMS. Without it, there would be no way to quit UMS.

> In your solution using the JAR, do ums-*-jar-with-dependencies.jar headless work as well ?

I very much doubt that would work, the argument is a path to a specific file.

I'm not going to fire up the VM to test every strange little thing, anyone can install a hypervisor and some VM's (with some effort) if they want to. If you feel the itch to test things, go with Virtualbox and read [this article](http://www.redmondpie.com/how-to-install-mac-os-x-snow-leopard-in-virtualbox-on-windows-7/).

**Comment by Sami32** [comment]
@Nadahar I miss some RAM, CPU and space for VM use ;-)

**Comment by Nadahar** [comment]
@Sami32 I would never buy an Apple product, haven't you understood that by now? :stuck_out_tongue_winking_eye: 

That said, the virtualization is perfectly good and the OS has no idea that it's not actually running real hardware. I'm even running a specialized BMW diagnostics software with a custom Linux version made specificly to only run on one particular IBM laptop from the 1990's. 

Rest assured, the reason things don't work/is hard to do an OS X has nothing to do with the virtualization, but the fact that Apple doesn't want users to be able to do things. Their vision of "user friendliness" is that everybody should only do what they tell them, and they should open their wallet while doing so. They simply don't care about things like this, it's not generating revenue for them - and if anything, they make it harder so that the users will buy something from their "store" instead of using free software.

**Comment by Nadahar** [comment]
@Sami32 You have no idea how bad it is to use that shit. They've even [rearranged the keys](http://lowendmac.com/2015/the-mac-users-guide-to-using-a-pc-keyboard/) so that their users should be forced to buy their "special keyboards" at a very expensive price or face utter frustration every time they want to find a symbol key. After a lot of searching I finally found that for my Norwegian keyboard, the ```|``` can be achieved by pressing ```Left Alt + 7```. It's small things like that that really get my blood boiling, they don't care of they make life difficult for people as long as they think they can blackmail some more money out of their pray/users/customers.

I fired it up one more time and
```xml
<key>VMOptions</key>
  <array>
    <string>-Djava.awt.headless=true</string>
  </array>
```
actually worked. However, it was just as I anticipated: You have no indication that UMS is running or if there are any errors, and you have no way to stop it.

I had to go into terminal and run:
```
ps -ax | grep Universal
kill <pid from previous command>
```
To stop UMS again. A bad solution imo, but yeah - it is doable if it's very important to do it in a confusing way.

Sami32:
@Nadahar Thank you for confirming :+1:
Actually i taken some of these informations from archived Apple developper documentation, so i was confident that it should work but i didn't about their keyboard trick...

Not surprising me though, as they are a cash machine...i love their design but that's all. AFAIK, they become less and less financial healthy since 1 or 2 year.

I will leave "your" JAR way solution in the wiki documentation, but i didn't like leave users on their own just because they make a "fruit" choice.

That said, i need to kill the process as well if i want close UMS while using `-Djava.awt.headless=true` way on Windows. Same while using `console` or `headless` command line arguments.
Did i missed something ?

Nadahar:
@Sami32 No, it's the same problem with Windows (or actually it's worse in Windows as even if you run it from a terminal Window you can't stop it with Ctrl+C). Headless only makes sense for Linux, as it can be made into a Deamon with some scripting. 

For Windows it should be run as a service instead, but that also has it's problems - although they are due to flaws in UMS' design.

OS X also has something in that direction (I'm not sure what it's called), but I'd never waste my time on figuring out how to configure that.

Btw, I do think people should be "punished" for going with "fruit". The problem is that you vote with your wallet, and when somebody has inacceptable business practices people should stop "voting" for them. But, that wasn't my intent here, I just can't imagine that there's any OS X users out there that even understand the "headless" concept, and if they do they should be more than able to figure out how to launch it "manually" ("my way") using the ```headless``` argument.

Nadahar:
I just installed dependencies for ```tsMuxeR``` on 64-bit Fedora 24, and all I needed was:
```
sudo dnf install freetype.i686
```
That automatically installed dependencies:
* bzip2-libs.i686
* libpng.i686
* zlib.i686

It seems that's all it took, as ```tsMuxeR``` no longer complain about missing dependencies. It seems making a general description for installing tsMuxeR dependencies can be quite hard because it varies a lot between distributions and versions. 

What I did to find out what was missing was very simple though, I simply ran ```tsMuxeR``` from the command line. It's not in the ```$PATH``` when "installed" by UMS though, so one would need to know the path to where the executable is located. That should be quite easy though, as it will generally be in the same folder as ```UMS.sh``` needed to start UMS. 

What I did was simply (being in the correct folder) to run ```./tsMuxeR```. It would then complain that ```libfreetype.so.6``` was missing, and I had to use my non-existing Linux skills to figure out that what I needed was ```freetype.i686``` (as we already know that it's the 32-bit versions tsMuxeR needs). That might be the simplest general approach, I'm guessing the same would apply to any other missing dependencies.

Sami32:
@Nadahar That's a nice finding :+1: that could make the Linux install/documentation easy, simpler.

But before changing the doc, i would appreciate if @SubJunk and @atamariya could confirm that's work on their Linux OS as well.

Sami32:
I don't know which way is the best to install tsMuxeR on Linux ?

Some user install it only with `lib32-gcc-libs` and `lib32-freetype2` libraries (Ubuntu, OpenSuse...)
Some 64-bit OS users only with `ia32-libs-dev` library (Debian).
As i can't test it myself, i'm perplex...

Nadahar:
@Sami32 I have VMs running Debian, Ubuntu, Centos, Fedora and Arch. The problem is the way dependencies work though: One package will install whatever other packages it needs to run. Some packages are preinstalled when you install the OS, and who that is depends on the intallation choices you make. Other software you install installs the packages it needs.

That means that testing it on a system where you have already installed unrelated software isn't really viable as some of the dependencies can already be in place. You'd ideally need a fresh "minimum" install to make sure you find every dependency needed for that distribution, which is why my existing VMs arent of that much help. Given the [chaos that is Linux distributions](https://en.wikipedia.org/wiki/List_of_Linux_distributions) it's really an impossible task to get this 100%.

SubJunk:
I deleted my Linux install so I can't test on it sorry

INSTALL.txt
```diff
@@ -2,20 +2,94 @@ Installers for Windows and Mac OS X and tarballs for Linux can be found at:
...
+    Note: Before installing VLC, you'll need to enable the free RPM Fusion repository in your system:
+    sudo yum install --nogpgcheck https://download1.rpmfusion.org/free/fedora/rpmfusion-free-release-$(rpm -E%fedora).noarch.rpm
+    # Install VLC on Fedora 22 and later:
+    sudo dnf -y install vlc
+    Note: Before that, you'll need to enable the free RPM Fusion repository in your system:
+    sudo dnf install http://download1.rpmfusion.org/free/fedora/rpmfusion-free-release-$(rpm -E %fedora).noarch.rpm
+    # Install dcraw and VLC on Arch (extra repository must be enabled):
+    sudo pacman -S dcraw vlc
+
+    # UMS uses tsMuxeR as a transcoding engine. tsMuxeR is 32 bit only, so installation of some 32 bit libraries are necessary
```
SubJunk: Maybe add a note about tsMuxeR being disabled by default, since this is potentially unnecessary work for a user installing UMS

INSTALL.txt
```diff
@@ -2,20 +2,94 @@ Installers for Windows and Mac OS X and tarballs for Linux can be found at:
...
+      # To install openjdk-8 run:
+      sudo yum install java-1.8.0-openjdk
+
+    # Install JVM on Arch:
+      # To install openjdk-7 run:
+      sudo pacman -S jre7-openjdk
+      # To install openjdk-8 run:
+      sudo pacman -S jre8-openjdk
+
+    # UMS use MediaInfo to get information about media files. It can be run without it, but media handling will suffer as UMS
```
SubJunk: use -> uses

INSTALL.txt
```diff
@@ -2,20 +2,94 @@ Installers for Windows and Mac OS X and tarballs for Linux can be found at:
...
+    won't have a detailed information available.
+    # Install MediaInfo on Debian/Ubuntu and derivates:
+    sudo apt-get install mediainfo
+    # Install MediaInfo on Fedora, OpenSUSE, Mandriva, Red Hat/Centos and derivates:
+    sudo yum install mediainfo
+    # Install MediaInfo on Arch (community repository must be enabled):
+    sudo pacman -S mediainfo
+
+    # You can optionally install dcraw (to decode raw images from digital cameras) and VLC (to use as a transcoding engine).
```
SubJunk: It could be good to split the optional dependencies into an "optional" section so it's a bit less daunting

Nadahar:
@SubJunk I think it need to be organized a different way. I've since discovered that this is already outdated for some distros, and they keep changing package managers, packages, dependencies and repositories so that it's virtually impossible to give a list of commands people can run without understanding what's happening.

I think it must be abstracted somewhat, and instead explain what software and dependencies are needed - the users will have to figure out the exact commands themselves unless we want to keep updating this once a month.

SubJunk:
That sounds smart to me. In that case I suggest this PR gets closed since the changes here don't match with that suggestion.

PR #970, opened 2016-08-01
Minor update of IEC61937AudioOutputStream.java

Sami32:
@SubJunk While looking at your issue enhencement i discovered that.
https://msdn.microsoft.com/en-us/library/windows/desktop/dd316761%28v=vs.85%29.aspx
http://www.ffmpeg.org/doxygen/3.1/spdif_8h_source.html

**P.S.** I know it's a minor modification than can be commit without PR, BUT i don't want make more problems than help, so for now i will appreciate a supervision if you don't mind 8-/

---
This change is [Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/970)


SubJunk:
@Sami32 thanks for making the pull request. There are other comments in the file that lead me to believe that DTS-HD isn't working so maybe we should leave it

ExSport:
@SubJunk @Sami32 As Shagrath, author of the code and PMS mentioned, it didn't worked due to bandwidth limitation.
In links you provided there is info about sending more streams together to achieve this so I am almost sure that current code doesn't support HD part.
In past I had a discussion with him and copied it somewhere in forum where he states it is undoable due to bandwidth limits of used LPCM standards. But at that date maybe there was no possibility to decode multistream?

Sami32:
Thanks you guys for your feedback :+1:

Issue #954, opened 2016-07-18
VLC on iOS doesn't always browse first directory


SubJunk:
It shows up as recognized and connected, then when I click the UMS icon in the VLC app, it displays an empty directory. Other times it works like normal.

Sami32:
http://www.universalmediaserver.com/forum/viewtopic.php?f=9&t=4060 ;-)

SubJunk:
It seems like this has been fixed in #975 though I'm not sure I know which part of that fixed it. I think it was something from the initial commit 6b38ab1

SubJunk:
@UniversalMediaServer/developers that was a false alarm, it is still happening. It's just inconsistent. It may just be that it expects us to respond within a certain time and we aren't.

I would appreciate some help on this because I've tried using Wireshark to spot differences between us and Serviio (which works 100% of the time) and I can't see what the difference is.

Ideas? Comments? @atamariya ?

By the way I have traced this bug all the way back to PMS so it has always existed in UMS, and I think always in PMS too though although I haven't tried the first release of that.

Sami32:
@SubJunk Some ideas, but before i would like ask you if you tried UMS with only one shared folder containing only 1 or 2 files ? did you get this issue as well ?

Nadahar:
As per DLNA spec you have quite a while to answer a browsing request (30 seconds or something like that from what I remember), so I doubt that's the problem. I'd guess that there's some small different in the response when it works and doesn't, I'd try to pinpoint that. With trace logging the whole response is dumped right? Copy a working and a not working response into two separate text files and compare them in a diff tool. If they are identical, add some debug timestamps and start looking at timings.

SubJunk:
@Sami32 tried it with just one folder with one small file in it and it still happened

@Nadahar I've been trying to do that for weeks and maybe I'm just not good at it.
I think it always happens when VLC is being run a second time since UMS started, I guess that is a good clue too.

Nadahar:
@SubJunk I understand that you've compared them, I just thought that using a diff tool could be helpful for picking up tiny details that could be enough for VLC to dismiss it. I don't remember the details, but I looked at some "fishy" calculations in the response message a long time ago - some calculation of remaining items, bytes or something of that kind that looked flawed to me - but I lacked the overview to do something about it. My guess is that while many renderers accepts such small errors and just "makes the best out of it", others might not be so forgiving.

Sami32:
[...]

PR #975, opened 2016-08-04
Fix VLC browsing (#954)

SubJunk:
This is to see if we can get VLC for iOS to consistently work with UMS, as described in #954 

I have been using Wireshark to see the differences between what we do vs what our competitors who don't have this bug do, and dumping the changes into here until they work. It may already have been fixed with the current changes but I need to test it more, and then to discard the unnecessary changes.

---

This change is [Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/975)


Sami32:
@SubJunk `CACHE-CONTROL: max-age=1800` :+1:

SubJunk:
Using getAliveIntervalMillis instead of doing regular search() doesn't make Panasonic TVs respond with their UPnP identifying information unfortunately, so in its current state this branch identifies Panasonic TVs as just the generic profile instead of their specific one

Sami32:
@SubJunk What a busy you are ;-)
I'm really impatient to see the result, really !

Sami32:
Concerning Panasonic (Matsushita Electric), i'm not surpised because AFAIK they have done many parts of DLNA specification so they are strict on that. Perhaps LG is concerned as well ?

But like one people told, DLNA is one of the worst standard, in the mean like everyone made it's own implementation variant...

SubJunk:
I've merged this branch into the 7.0.0 branch since I want to test both branches without having to switch between them all the time. We can continue to discuss it here or in that PR, either way I'll probably commit the changes to the 7.0.0 branch to make it easier. That will mean there are temporary commits in there but it can be cleaned up before merging.

Sami32:
@SubJunk I suspect that many renderers could profit from it, as Roku. When i checked their miniDLNA logs where they told it was working: First i saw 12 ssdp:byebye and after 6 ssdp:alive ...not too standard compliant but better than UMS.
So let's see what could happen with the Vizio renderers as well.

Sami32:
I didn't know about that:
> Generally event notifications are sent to the registered callback URL as a unicast. The receiver acknowledges the notification with the appropriate HTTP response. However, **there is also an alternate mechanism to publish events as a multicast to 239.255.255.246:7900**.

Sami32:
Hoping not bothering anyone, just letting your know:

> Requirement [7.2.3.4]: UPnP control points must send M-SEARCH messages using a source port greater than 1024 and not 1900.

IF issues arise with phones, perhaps detecting if it's a mobile and lowering the value if it's the case ?
My phone is as old as my computer ;-) but i read in Nokia study that they test it Nokia  6600 with value of 600 as the worst case scenario to 1800.
> Comment: Most devices that remain on the network for long periods should have CACHE-CONTROL value of 1800.  **However, some devices (mobile, wireless, etc.) that may want a smaller CACHE-CONTROL value**.

> Requirement [7.2.4.5]: The interval of sending these advertisement groups on a single 
> network interface must be less than ½ the CACHE-CONTROL value.
> The first advertisement set and the duplicate sets (transmitted on a single network 
> interface) make up an advertisement group.
> Comment: For consistency and interoperability, devices need to advertise more often 
> than their notification cycle.  However, implementers should remember that 
> advertising too frequently runs the risk of flooding the SSDP channel.

Nadahar:
> My phone is as old as my computer ;-) but i read in Nokia study that they test it Nokia 6600 with value of 600 as the worst case scenario to 1800.

@Sami32 I use an [old Nokia](http://www.gsmarena.com/nokia_n95_8gb-2088.php) as well, I don't like the thouch screens (too slow and fragile). In any case, I wasn't aware there were DLNA clients for Symbian...? And if so, why? I guess Audio is OK, but I'd never consider wathing pictures or video on that tiny screen. I have many other screens more suitable so I don't see the need either, I was just wondering it the test study really is about the same thing?

Sami32:
@SubJunk It's correctly detected on my Viera:
[debug.log_PR975.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/409265/debug.log_PR975.txt)
**N.B.** I can see many `Ignoring self-originating request` from my router appear in the log that was not here before these changes.

Sami32:
@Nadahar It was at the beginning study on DLNA, some years ago now, so i guess that they used what they have :-)
[1-N02-02.pdf.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/409531/1-N02-02.pdf.txt)

@SubJunk Concerning the M-SEARCH request you could be inrested by the LG way:
http://developer.lgappstv.com/TV_HELP/index.jsp?topic=%2Flge.tvsdk.references.book%2Fhtml%2FUDAP%2FUDAP%2FM+SEARCH+Request.htm
The UDAP/2.0 used in the USER-AGENT optional header field is a LG flavor.

SubJunk:
@Sami32 when it detected your TV correctly, was the TV on when you started UMS?

Sami32:
@SubJunk No, my TV was off.
But this one was with TV on when UMS start.
[ums_dbg_PR975.zip.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/409985/ums_dbg_PR975.zip.txt)

SubJunk:
@Sami32 hmm maybe my TV has a bug yours doesn't... Which is weird since I'm using the firmware update released last month

Sami32:
@SubJunk That could be the opposite as well ;-)

Nadahar:
@SubJunk In my experience the amount of time since release is about proportional with the probability for bugs, so I'm not sure that's so weird. I always prefer older versions that's proven their reliability unless I have a particular need for some feature of the latest.

SubJunk:
@Sami32 @Nadahar fair points but in this case the issue has existed for this TV in all firmware since its release AFAIK

Sami32:
@SubJunk Sad to learn that, but maybe this guy could help you: Hirokazu Ishikawa
http://www.dlna.org/news/news/2014/04/15/dlna-member-achievements-commended

But as a developper and maintener of this project you will be happy of that: ;-)
http://www.businesswire.com/news/home/20150818005244/en/DLNA%C2%AE-Announces-3.0-Certification-Program-Updated-Guidelines

**P.S.** You could find this link of some use for helping making or debugging renderers configuration:
http://coherence.beebits.net/wiki/DeviceSpecificities

Sami32:
http://www.dlna.org/news/2016/4/19/dlna-solves-html5-conformance-testing-challenge-in-tvs-and-mobile-devices

SubJunk:
The changes here didn't work unfortunately

PR #988, opened 2016-08-18
Include the CharsetDetector code from ICU4J to not import whole ICU4J library

valib:
Use only a part of the ICU4J library code to decrease the UMS size. The only problem is that in this case we must follow possible ICU4J changes and update it manually.

---

This change is [Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/988)


Nadahar:
How will updates to ICU4J be handled when parts of the lib is copied? Is the license compatible with UMS'?

valib:
@Nadahar I forgot to mention in the comment that the disadvatage of that soulution is that somebody would take care about changes made by ICU4J team and manually maintain it. 
Regarding the licence I read it and if I understand it well it is compatible with UMS. But please can you and other developers confirm that? The ICU4J licence is here http://www.unicode.org/copyright.html#License

Sami32:
> provided that either (a) this copyright and permission notice appear with all copies of the Data Files or Software, or (b) this copyright and permission notice appear in associated Documentation.

Sami32:
@SubJunk We win around 40% weight compared to the full version by default !

Nadahar:
To be honest I think #717 is a much better and more general solution to the problem, even if it requires some configuration to work properly.

Sami32:
@UniversalMediaServer/developers What will the best way to keep UMS size's control fro now, `maven-shade-plugin` ?

Any suggestions are greatly welcome ;-)

valib:
Tested the #717 PR and I must to say that it saves about 11 MB in the UMS ums-xxx-SNAPSHOT-shaded.jar but the size of the UMS-setup.exe is still the same about 82 MB. This PR saves about 10 MB not only in the ums-xxx-SNAPSHOT-jar-with-dependencies.jar but also in the UMS-setup.exe which is reduced to 72 MB. Maybe the #717 PR needs some improvement. Anyway it seem promissing.

Nadahar:
@valib I never finished #717 as I didn't get any response. As stated before more changes are needed in the POM to make it work. The resulting installer you get is therefor probably not the correct one. The good thing about using the shade plugin is that it will automagically leave out unused stuff, but I no longer remember the details.

SubJunk:
I would rather have larger files but not have to track changes to their library

Nadahar:
@valib A better approach would be to exclude parts of their library need needed in the POM. It achieves much of the same, and doesn't require maintenance,

Issue #981, opened 2016-08-10
Renderer max thumbnail size


squadjot:
Some devices, like my Sony W5500, supports file/folder thumbnails.
Small problem, if the folder.jpg is too large (larger than 512x512px) the thumbnail won't show.

Would it be possible to let UMS make sure that the thumbnail are in supported size?

atamariya:
I've added the DLNA compliant thumbnail support in #987 
`if (size.equals("TN"))
				size = "160x160";
			else if (size.equals("SM"))
				size = "640x480";
			else if (size.equals("MED"))
				size = "1024x768";
			else if (size.equals("LRG"))
				size = "4096x4096";`

SubJunk:
@atamariya very cool, if you can make that part its own branch we can merge it :)

squadjot:
Cool, looking forward to test it ^^

atamariya:
Created PR #989

SubJunk:
@squadjot please test the fix :)

squadjot:
Sorry, for being a newbie.. how do i test? Can i just use UMS builder?

Sami32:
@squadjot Just follow these step: http://www.universalmediaserver.com/forum/viewtopic.php?f=25&t=1782 
And while asked to choose a pull request[3], after having selected your OS, select 989.

Hoping that helped...

squadjot:
I cant get it to work =/, i installed java SDK and Maven, added the environment paths etc but got stuck at:

```
[INFO] Downloading artifact jwbroek.cuelib:cuelib:1.2.1-2008-06-13:jar from URL
http://cuelib.googlecode.com/files/cuelib-1.2.1-2008-06-13.jar
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 01:30 min
[INFO] Finished at: 2016-09-12T01:42:56+02:00
[INFO] Final Memory: 13M/31M
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal com.universalmediaserver:external-maven-plugin:0.
1:install (default-cli) on project ums: Failed to download artifact jwbroek.cuel
ib:cuelib:1.2.1-2008-06-13:jar: File: http://cuelib.googlecode.com/files/cuelib-
1.2.1-2008-06-13.jar , ReasonPhrase:Not Found. -> [Help 1]
[ERROR]
[...]

PR #989, opened 2016-08-20
Thumbnail image is DLNA compliant.

atamariya:
This change is [Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/989)


Sami32:
@SubJunk https://github.com/UniversalMediaServer/UniversalMediaServer/issues/981#issuecomment-246376857

SubJunk:
Cool, I'm happy to merge this if the conflicts are resolved

Sami32:
@atamariya :-)

valib:
I also like this PR but it uses currently deprecated methods `getThumbSize()` and `getThumbBG()` so we should decide if we should revert those methods and deprecate (delete) the parameters `ThumbnailWidth` and `ThumbnailHeight` used in the renderers conf.

atamariya:
@SubJunk conflict is resolved.

SubJunk:
@valib good point, we should make a decision about that... The main relevance of ThumbnailWidth and ThumbnailHeight in the renderer was when we added black bars to the thumbnails. Now we just output them at their actual aspect ratio so that doesn't matter. Do you agree?

@atamariya thanks

Nadahar:
I am completely confused by this PR. Why is there 32 commits in this where only 3 (and one merge) is done by @atamariya? There must be some git error that should be figured out and corrected here. The easiest might be to create a new branch and cherry-pick the relevant commits to the new one.

SubJunk:
Yeah I'm not sure how some branches end up with those extra commits either. It shouldn't happen.

Nadahar:
IMO it should be cleaned up before merge. I don't have write access to this branch so I can't do it without copying it to a new branch. Either @atamariya do it in his branch or I can do a copy and clean operation - just let me know. If I do this I'll have to create a new PR though.

SubJunk:
Maybe if @atamariya pulls from master into this branch it will fix that weird diff. I agree it should be cleaned up since there are a lot of unrelated changes showing up here

atamariya:
Created PR #1024

PR #1008, opened 2016-09-19
Restart the server if the network connection is lost

SubJunk:
I ended up developing this for myself because my wireless network extender has an issue where it loses connection every now and again. When that happens, UMS gets into a state where it can't do anything until it is restarted, so this just restarts it quietly instead of getting stuck.

Does anyone object to it being merged?

---
This change is [Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/1008)


Nadahar:
It shouldn't come as a surprise that I don't fancy such solutions (restarting instead of figuring out the underlying issue), but it's up to you what you want to do. It's easy to see situations where this might result in unwanted behaviour. I'd look into why UMS gets into this state, I can't see why loosing the network connection should be an "unrecoverable" situation. Does the NIC release it's IP when this happens, or is the interface in itself working as if it were simply "alone in the world"=

SubJunk:
@Nadahar I agree it's not a perfect solution and ideally we should find a better way to do it, but also it's not really our bug - it's a bug with my wireless extender. I don't think we should spend too much time coming up with a workaround for a bug caused by a problematic device. Do you have motivation to work on a more specific fix ?

Nadahar:
@SubJunk I lack both motivation and time at the moment. I just gave my input. I agree that there's no reason to go to great lengths to solve a specific bug in a specific device. I'm not sure if this applies here though, it depends a little bit on how the issue with your extender manifests itself. If it simply results in temporarily loss of connectivity, that should not result in UMS being able to recover. If that is the case, this will be a much broader issue than can be behind various problems experienced by users.

If on the other hand the issue with your extender creates a very rare/unique situation were your OS generally "gets confused" it probably isn't worth persuing. I'm afraid though that this workaround can create other problems in other situations. You should at the very least make this a configuration option, and I'd recommend it to default to false. I'd also think it could be solved by rebinding the UPnP socket instead of restarting the whole application if a more "logical" solution isn't viable.

```IOException``` is a quite general exception, what is the exact exception this generates? It would probably be wise to narrow it down as much as possible..?

Sami32:
@SubJunk As a string parameter in the configuration file, it will be fine for me.

Maybe we could do better with a log and Wireshark capture, but honestly it's not a search/check that i want to do now, so let's go for your workaround.

Sami32:
That's said i don't mind for the default value that will be affected to it, as it will be user configurable.

Nadahar:
@Sami32 The reality is that the majority of users will never change a default setting, especially if you only don't implement it in the GUI. That means that a default will effectively decide how a feature is used. Getting the defaults right is critical imo.

I don't see how restarting UMS can help debugging or router issues, a restart will normally just erase states and make everything harder to figure out.

Sami32:
@Nadahar As i told i don't care for the default value :)
Very easy, if users with connection problem use this option and don't have anymore issue that problably mean that UMS can be withdrawed from the equation...

Nadahar:
@Sami32 ? I don't know how you can draw that conclusion, that doesn't prove anything as I see it. Also, users will have "issues" in any case if UMS keeps restarting during use. This automatic restart is only ever "useful" is the event is rare.

ExSport:
Today tested UMS with non static IP address and UMS is not aware about IP change. When DHCP changes IP to the client, UMS still sits on interface with wrong IP and doesn't reflect the change. Maybe it is your problem @SubJunk ?
You or me, when I will have time, we can test if problem arise with IP change only or it is enough to release and renew also same IP on interface.
You can test it with IPCONFIG /release & IPCONFIG /renew if UMS will loose connection...
I spotted this behavior when I rebooted router with DHCP and no reserved IP for the UMS client.

Nadahar:
There probably are some issues with the socket so it would need reinitialization. When it comes to changing IP address "on the fly" there's not many "server" programs that support this. You'd have to redo a lot of things when the IP changes like, rebind sockets, inform the other DLNA control points of the changed address and do a lot of internal reinitialization for UPnP listeners, DLNA HTTP server and the web interface HTTP server. I'd say that wouldn't be worth the trouble, the basic idea is normally that any computer with listening ports/that accepts incoming traffic stays on one IP address.

Normally the DHCP server would just reassign the same IP, but it depends on the lease time configured in the DHCP server. If a DHCP subnet has few addresses compared to the number of clients using short leases can make sense, but normally a home network will have a C class subnet (254 IP pool) available for relatively few devices and thus lease time can be very long without ever running out of addresses. 

I'm betting that a release/renew is enough to cause problems, but even that shouldn't be something that would normaly happen. I'm not sure how long a client will keep it's current IP when contact with the DHCP server is lost, but I'd think the client would just stick to the assigned address until it's told different from the DHCP server. There shouldn't be any need for the DHCP server and clients to have any further communication once the initial assignment was done, but I don't know the details of how the DHCP protocol works.

ExSport:
@Nadahar As a server admin I know exactly what happens, I only wanted to describe possible problem @SubJunk has. I am using static or reserved IP so no problem for me or want to request changing anything in the UMS logic.
It was question if his buggy router loose DHCP lease info so when UMS disconnects is is due to IP change or it doesn't change so renewing same IP happen which can be enough to loose connection.
Btw. every client receives IP for some specific lease time. Every half period it tries renewal, it is like I am still here. If there is no response from DHCP it stays on same IP but will wait for next renewing until about 90% of lease time when it tries to renew again continuously until lease expires or any DHCP server will respond. When lease time expires client must release IP and restart renewing process from start=broadcasting to the network for DHCP server.
I will try renewing with same address if it is enough for UMS to loose socket binding and let to know here... Hopefully during next week.

Nadahar:
@ExSport My explanation wasn't meant specificly for you but for anyone trying to follow the logic here :wink: I'm just saying that a "server" normally isn't meant to handle changing IP, even though I saw in the DLNA specs that there is support for a DMS informing their clients of IP change. It's a bit odd though, because it requires the server to keep answering also on the old IP for a while. I don't know what practical scenario that's applicable in.

I don't know the nature of the bug in @SubJunk's extender, but an extender normally shouldn't manipulate the packets in any way so I'd guess that all it could do is loose connectivity. If however the extender is running a "shadow" of the WiFi router so that the NIC could pick up traffic from the router when the extender is down, I can imagine there being some confusion on the OS level. It's hard to see how simply loosing connectivity could make much difference to the OS interface, but who knows what the drivers do and if it "invalidates" the current interface and starts scanning for something new after some timeout. If so, it could be a new or reinitialized interface available after this, and I'd guess that UMS would have to rebind the sockets. It should still be much cleaner to rebind the sockets than to restart the whole application.

PR #1013, opened 2016-09-22
Remove tags from srt subs

valib:
@Sami32 good idea because also my Panasonic TV doesn't support SRT tags. Can anybody test it if it works?

---
This change is [Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/1013)


Sami32:
Lightening fast and efficient, as always ;-)
Work perfectly :lgtm_strong: 

---

Reviewed 6 of 6 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/1013#-:-KSJl8MJLnmRXcaDUqRG:boj5ka5)*

src/main/java/net/pms/network/Request.java
```diff
@@ -327,10 +329,16 @@ public void answer(
...
 							if (sub.isExternal()) {
-								inputStream = new FileInputStream(sub.getExternalFile());
+								if (sub.getType() == SubtitleType.SUBRIP && mediaRenderer.isRemoveTagsFromSRTsubs()) { // remove tags from .srt subs when renderer doesn't support them
+									inputStream = new FileInputStream(SubtitleUtils.removeTagsFromSubs(sub.getExternalFile()));
+								} else {
+									inputStream = new FileInputStream(sub.getExternalFile());
+								}
+							} else {
+								LOGGER.trace("Not loading external subtitles because they are not external: " + sub);
 							}
```
Nadahar: That text is hard to understand...

src/main/java/net/pms/network/Request.java
```diff
@@ -327,10 +329,16 @@ public void answer(
...
-						} catch (NullPointerException npe) {
-							LOGGER.trace("Could not find external subtitles: " + sub);
+						} catch (IOException npe) {
```
Nadahar: NullPointerException isn't a subclass of IOException, which means the NPE will no longer be caught.
```java
catch (NullPointerException | IOException e)
```
Would be a better solution given that the previous logic is there for a reason.

src/main/java/net/pms/network/RequestV2.java
```diff
@@ -327,13 +329,18 @@ public ChannelFuture answer(
...
+									inputStream = new FileInputStream(sub.getExternalFile());
+								}
+								
 								LOGGER.trace("Loading external subtitles: " + sub);
 							} else {
 								LOGGER.trace("Not loading external subtitles because they are not external: " + sub);
 							}
-						} catch (NullPointerException npe) {
-							LOGGER.trace("Not loading external subtitles because we could not find them at " + sub);
+						} catch (IOException ioe) {
```
Nadahar: Same as in V1

src/main/java/net/pms/util/SubtitleUtils.java
```diff
@@ -622,4 +622,28 @@ public static String convertColourToASSColourString(int colour) {
...
+	 * @param file the source subtitles
+	 * @return Temporary file with the converted subtitles. This file will be deleted when the UMS will be stopped.
+	 */
+	public static File removeTagsFromSubs(File file) throws IOException {
+		BufferedReader input = FileUtil.bufferedReaderWithCorrectCharset(file);
+		File tempSubs = new File(configuration.getTempFolder(), file.getName());
+		BufferedWriter output = new BufferedWriter(new OutputStreamWriter(new FileOutputStream(tempSubs), Charset.forName(CHARSET_UTF_8)));
+		String line;
+		while ((line = input.readLine()) != null) {
+			line = line.replaceAll("\\<.*?>","") + "\n";
```
Nadahar: This will affect text content containing ```< >```. I can't find a proper format description so I don't know if there is an escape mechanism, but since the allowed tags are very few I'd think it would be better to only match the specific tags.

From [Wikipedia](https://en.wikipedia.org/wiki/SubRip#SubRip_text_file_format):
```
Formatting

Unofficially the format has very basic text formatting, which can be either interpreted or passed through for rendering depending on the processing application. Formatting is derived from HTML tags for bold, italic, underline and color:[10][dubious – discuss]

    Bold – <b> ... </b> or {b} ... {/b}
    Italic – <i> ... </i> or {i} ... {/i}
    Underline – <u> ... </u> or {u} ... {/u}
    Font color – <font color="color name or #code"> ... </font> (as in HTML)

Nested tags are allowed; some implementations prefer whole-line formatting only.
```

Nadahar:
@SubJunk Reviewable and GitHub's new review functionality is very confusing together, maybe it would be best to disable Reviewable for now?

valib:
@Nadahar thanks for comments. I will check it.

src/main/java/net/pms/network/Request.java
valib: @Nadahar thats true. Coppied from the original code in RequestV2 :smile:

src/main/java/net/pms/network/Request.java
valib: @Nadahar I will add it but I can't imagine which part of the code can throws NPE

Nadahar: @valib Catching NPE's is an absolute no-no as I see it, it should be null checked instead. But it can happen here, as the comments above explains. PMS forums are down, so I can't check the link, but it's easy enough to accept that ```sub.getExternalFile()```. This was added in 27a277b4f300961e266eb3a30d664c2d5af1ecef.

src/main/java/net/pms/util/SubtitleUtils.java
```diff
@@ -622,4 +622,28 @@ public static String convertColourToASSColourString(int colour) {
...
+	/**
+	 * Remove tags like <"b"> <"/b"> <"i"> <"/i"> <"u"> <"/u"> <"/font> <"font color="....">
+	 * from subtitles file for renderer not capable to show that tags correctly.
+	 *
+	 * @param file the source subtitles
+	 * @return Temporary file with the converted subtitles. This file will be deleted when the UMS will be stopped.
+	 */
+	public static File removeTagsFromSubs(File file) throws IOException {
+		BufferedReader input = FileUtil.bufferedReaderWithCorrectCharset(file);
+		File tempSubs = new File(configuration.getTempFolder(), file.getName());
```
Nadahar: This is a potential bug - the same filename could already exist and be open for writing by another thread or by incorrect termination of UMS. This will result in an IOException and abort.

SubJunk:
@Nadahar yes I agree, I will disable reviewable. Also I've enabled the setting now that requires at least one review with no requested changes before merging a PR.

PR #1019, opened 2016-09-24
Remove tags from srt subs reloaded again

valib:


Nadahar:
@valib :+1: I'll merge my commits into this branch as well, and then we can review/test and hopefully make the final merge.

Nadahar:
@valib I had to do a lot of "git stunts" to get this together since my former commits were overwritten. I've had to do a couple of rebases and manual conflict resolution, so I had to force push this branch. All your commits are still here though, but I've removed the merge with master to avoid complexity (as my changes would need another merge with master). I'll do one final merge with master when all is "ready".

My changes are really minor except that I've changed the regex/what is being stripped. I studied the format, and from what I could find out there is no tag support in SubRip at all. It seems that this all started with MPC implementing support for some HTML tags and then [some other followed](http://ale5000.altervista.org/subtitles.htm). The same happened later with some of the ASS tags. There is no definition as to what is "supported" and not, it all comes down to what subbers use and that probably comes down to what their "defacto standard" player/engine supports. If that's MPC, DirectVobSub or VLC I have no idea. Since there is no standard for tags in SubRip, I doubt it's implemented in many renderers if any at all. VLC and Kodi might support it (from what I read on the Kodi forum they didn't seem very eager to implement such an undefined mess), but I doubt any of the hardware devices would implement something that isn't properly defined. Because of this, I think the stripping should be standard.

Trying to figure out what ASS tags are actually supported in SubRip is quite hard, as there is no such thing as supported. I used [this ASS tags definition](http://docs.aegisub.org/3.2/ASS_Tags/) and came to the conclusion that the best thing would be to treat anything between ```{\``` and ```}``` as ASS tags. The curly brackets is also said to be used for inline comments, but at the same time [SubRip_capability_tester.srt.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/490686/SubRip_capability_tester.srt.txt) dictates that ```{normal text}``` should be displayed. This is contradictory but I chose to not strip them since this isn't actually ASS files but SubRip files and I believe the [SubRip_capability_tester.srt.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/490686/SubRip_capability_tester.srt.txt) is the closes thing to a standard that exists.

There is no "open" and "close" concept with ASS tags, tags either apply to the line they're on or affect the text from that tag onwards. 

I also did a little optimalization on the regex compiling it only once instead of for each line.

We need proper review and testing this time around before any merging, let's let this be the last one. Because there might be more changes I want to wait until all is done before making the final merge with master.

Nadahar:
I pushed another commit where I changed the default for RemoveTagsFromSRTSubtitles to true. This is because I'm fairly confident that the support for these tags are close to non-existant as there is no defined standard. Having them stripped by default is therefore most likely to give the best user experience.

If you disagree about this, let me know and I'll remove the commit.

valib:
@Nadahar agree with you that removing tags could be general.

valib:
@SubJunk there is something wrong with the setting because I can't approve reviuwed changes.

Nadahar:
@valib You can't approve reviews because you created the PR. It's so one person can't both create and approve the PR. I can approve reviews on this even though I have some commits, so it's only depending on who created it. I don't want to approve it yet though, I'll wait until it's tested and ready.

Nadahar:
@valib What do you think about the changes to the regex? Does it make sense to you?

valib:
@Nadahar they are good. I tested them and the result is reasonable. I think that this PR could be merged. Acording the review of this code somebody like you should approve it and merge.

src/main/java/net/pms/util/SubtitleUtils.java
```diff
@@ -639,7 +639,7 @@ public static InputStream removeSubRipTags(File file) throws IOException {
 		BufferedReader input = FileUtil.bufferedReaderWithCorrectCharset(file);
 		ByteArrayOutputStream os = new ByteArrayOutputStream();
 		Writer writer = new OutputStreamWriter(os, Charset.forName(CHARSET_UTF_8));
-		Pattern pattern = Pattern.compile("\\</?(?:b|i|s|u|font[^\\>]*)\\>|\\{\\\\[^\\}]+\\}|\\\\h|\\\\N");
+		Pattern pattern = Pattern.compile("\\</?(?:b|i|s|u|font[^\\>]*)\\>|\\{\\\\.*?}|\\\\h|\\\\N");
```
Nadahar: I don't know if I agree that this make it simpler, but it makes is a few characters shorter :stuck_out_tongue:  I don't like to use the non-greedy unless I have to, I never really trust that it does what i want. Escaping the closing ```}``` is something I do because I think it makes reading the regex easier (it's obvious that it's a literal). But, the result should be the same :wink:

Nadahar:
@valib I can approve and merge this but shouldn't we wait for at least one more opinion/review?

src/main/java/net/pms/util/SubtitleUtils.java
valib: @Nadahar of course I know that the resullt will be the same but I think that to use `.*?` is a little bit more general to say that any characters between `{\...}` should be replaced. It is about the preferences so feel free to revert this commit.

Nadahar: @valib As you say, it comes down to preference. There's no need to revert, there's no "correct" way to write a regex :smile:

Nadahar:
@valib I'm trying to test if Kodi supports this so that stripping could be turned off for Kodi if it does. But, I can't get Kodi to display .srt subtitles at all. Is there something I must enable in the renderer configuration for this to work?

src/main/java/net/pms/util/SubtitleUtils.java
valib: @Nadahar maybe we can test which of the approach is faster to distinguish between them. 😄

Nadahar:
@valib Maybe you have already thought of this, but what about the transcoding engines? I just tested FFmpeg and it does support tags, so they should not be stripped when transcoding. Is the same true for the other transcoding engines?

Nadahar:
I finally got subtitles to work with Kodi, I'm not sure what did it but I think it was that I sat ```MediaInfo = true``` in the renderer configuration (and made a supported line to match the test video). It looks to me like subtitles streaming doesn't work when MediaInfo is off - which is unrelated to this PR but still a bug if that's true. Also, a great renderer like Kodi should have updated its renderer configuration so it works properly.

It turns out that Kodi supports a few ot the tags, but most from the test file doesn't work. It does however strip the tags itself, so we don't have to. That way the few that is implemented works, and if they implement more of them in future versions they will be supported too.

Nadahar:
I'm pretty confident that VLC supports the tags as I know it does so when playing the file itself. There should be no reason for them to use a different SubRip rendering when playing via UPnP. Since we have no renderer configuration for VLC, and it doesn't work with UMS due to another bug I can't add the option to VLC's renderer configuration. The day one is created, it should very likely contain ```RemoveTagsFromSRTSubtitles  = false```.

src/main/external-resources/renderers/DefaultRenderer.conf
```diff
@@ -749,11 +749,10 @@ OfferSubtitlesAsSource =
 StreamSubsForTranscodedVideo =
 
 # Remove the (HTML) tags: 
-#   <b> </b> <i> </i> <u> </u> <s> </s> <font *> </font>
-# and any ASS tags <code>
-# {\*}
+# <b> </b> <i> </i> <u> </u> <s> </s> <font *> </font> and any ASS tags {\*}
 # from SubRip subtitles files for renderers not capable of showing SubRip tags
```
Nadahar: I was too quick for myself there, thanks :+1:

Nadahar:
@valib Have you tested the VLC profile? VLC doesn't work with UMS... so it does nothing for me.

Nadahar:
I got VLC to work by just sharing one folder with a few files in and turning off all "virual folders" like "New media" etc. It queries the whole media tree to do internal sorting/organization, so it simply times out in most cases. 

I can't get external subtitles to work with it though, does it really not support that?

valib:
@Nadahar for me all shared folders works.  As you said VLC is checking all folders tree so it takes long time to start working. I didn't tested yet if external subs work. I used the Apple-iDevice-VLC.conf and only changed the UserAgentSearch setting. What is strange for me that VLC doesn't respond at the UPnP level and only DLNA protocol is working for me.

Nadahar:
@valib I'm not sure how you define "UPnP" and "DLNA" levels. DLNA is just a restriction of UPnP, everything DLNA is also UPnP but UPnP allows much more - it can share any kind of files without being restricted by DLNA's profiles among others. 

DLNA defines device classes:
* DMS - A Digital Media Server with the role of exposing and distributing content.
* DMP - A Digital Media Player with the role of finding content exposed by a DMS and playing the content locally on the DMP.
* DMR - A Digital Media Renderer with the role of playing content it receives after being setup by another network entity.
* DMC - A Digital Media Controller with the role of finding content exposed by a DMS and matching it to the rendering capacities of a DMR and setting up the connections between the DMS and DMR.
* DMPr - A Digital Media Printer  with the role of printing images.
* M-DMS - A Mobile Digital Media Server with the role of exposing and distributing content.
* M-DMP - A Mobile Digital Media Player with the role of finding content exposed by an M-DMS and playing the content locally on the M-DMP.
* M-DMU - A Mobile Digital Media Uploader with the role of sending content to an M-DMS with upload functionality.
* M-DMD - A Mobile Digital Media Downloader with the role of finding and downloading content exposed by an M-DMS and playing the content locally on the M-DMD after downloading.
* M-DMC - A Mobile Digital Media Controller with the role of finding content exposed by an M-DMS and matching it to the rendering capabilities of a DMR and setting up the connections between the server and renderer.

The first 4 are what are important to us. UMS is a DMS, what we call a renderer can either be a DMP or a DMR (and some even have DMC capabilites). A DMP can only PLAY content, while a DMR can be remote controlled by a DMC. With Cling and the "renderer panel"/bump UMS implements DMC. Not all DMPs are DMR although most are. That's why you don't get the "control panel" on every device. VLC doesn't seem to have any DMR capabilities implemented.

Is this what you mean?

valib:
@Nadahar exactly. You explain it very well.

Nadahar:
@valib Ok, but its nothing strange involved (or a bug in VLC), they simply haven't implemented it. It shouldn't have any implications for us other than that VLC can't be remote controlled via DLNA.

The only unresolved issue I see with this PR is how to handle transcoding engines. It seems to not strip tags when using FFmpeg regardless of the renderer configuration as it is now, which works well. Is the same true for other transcoding issues?

atamariya:
@valib @Nadahar Shouldn't config values follow inheritance too? e.g. if the value is defined in default conf and not in renderer specific conf, default conf value should apply.

Nadahar:
@atamariya That would probably have been a good solution, but it's not the way it's currently implemented. All defaults are hard coded, and only non-default values need to be in the configuration file. ```DefaultRenderer.conf``` isn't a default/parent configuration, it's more a reference configuration. AFAICR it's also used for "unknown"/unrecognized renderers. The plus side with the hardcoded default is that it will start and do basic tasks without any configuration files.

Nadahar:
@valib Regarding VLC I see what you mean now, it has a very weak implementation. It seems it doesn't even implement the DMP role but only does the absolutely necessary to find a media server with no concern that others should see it. The good thing is that the UPnP/DLNA part of VLC is replaced with a completely new one in VLC 3.x. This version won't browse the whole directory tree but only list one folder at a time, and hopefully be more compliant in other ways as well. I love VLC generally, i handles almost anything I throw at it, so having the DLNA part work properly as well would be great. When that is working I'd say that would be a better way to access a UMS from a computer than using the web interface. Kodi does a great job too, but it's interface isn't really desktop friendly.

Nadahar:
@valib **IMPORTANT** I just did a rebase and force push on this branch. If you are to do another commit in this branch you must reset your local branch or delete it and check it out again. If you already have some work in progress, push it to a temporary/new branch and let me know and I will rebase it into this branch.

Sami32:
@valib I'm not sure you will smile, but i just discovered that FFmpeg and MEncoder (little tricky) are able to do that by themself; it's not documented...yet.
https://lists.ffmpeg.org/pipermail/ffmpeg-user/2015-December/029591.html
https://ffmpeg.org/doxygen/trunk/srtenc_8c-source.html

So that should easily work with internal and external subtitles ;-)

Nadahar:
@Sami32 The stripping is used when streaming, not transcoding. FFmpeg or MEncoder isn't used when streaming, so they aren't relevant. It would be much more complicated and slow to use one of them to just do the stripping.

Sami32:
@Nadahar If i believe your own words:
> stripping is only applied to external SubRip subtitles when requested separately via HTTP

But we can make it work with internal subtitles as well, without "complicated" things.
Anyway, i just give the information...

src/main/external-resources/renderers/VLC-for-desktop.conf
```diff
@@ -0,0 +1,51 @@
...
+Supported = f:dts    m:audio/vnd.dts
+Supported = f:flac   m:audio/flac
+Supported = f:ogg    m:audio/ogg
+
+# Supported image formats:
+Supported = f:jpg   m:image/jpeg
+Supported = f:png   m:image/png
+Supported = f:gif   m:image/gif
+
+SupportedInternalSubtitlesFormats = ASS,SUBRIP
```
Sami32: @valib Why not this instead ?
`SupportedInternalSubtitlesFormats = SUBRIP,ASS,TEXT,MICRODVD,WEBVTT,SAMI`

https://wiki.videolan.org/subtitles
http://git.videolan.org/?p=vlc.git;a=blob;f=modules/demux/subtitle.c

Nadahar:
@Sami32 Please explain how to do this with internal subtitles without demuxing/remuxing (transcoding). I don't understand what you mean. When a file is "streamed" that means it's delivered "as is" without any changes to the file by UMS. That means, per definition, that it's impossible to do anything to internal subtitles as I understand it.

I don't know if you're familiar with the term used here, and actually it's wrong, but "streaming" is used about files that's not run through a conversion (transcoding) in "UMS terminology". Technically both untouched and transcoded files are streamed, so I see how some misunderstanding could arise.

src/main/external-resources/renderers/VLC-for-desktop.conf
Nadahar: @Sami32 AFAIK @valib simply copied and modified the VLC for IOS profile so the subtitles support is an inheritance from that. It might very well be that more types are supported, but I didn't bother to test. I tested external subtitles and found nothing to work with VLC, but I didn't want to take the trouble of muxing different internal subtitle types into files for testing.

It's likely that the support for internal subtitles are the same as when VLC plays a file directly, but there's no way to be sure without testing. The current DLNA implementation in VLC also means it's not that relevant imo, but if you want to test and confirm that's the best solution.

Sami32: @Nadahar I was lazzy as you, so i didn't tested it myselt, i just believe the words of the main VLC developper that internal subtitles over network are supported and maintened since long ago.

Nadahar: @Sami32 We will never know unless someone tests it..

Sami32:
@valib `<br />` tag is not taken in consideration, right ?
I just re-discover that one in a video test file right now.

Nadahar:
@Sami32 This is what defines what's removed:
```java
Pattern pattern = Pattern.compile("\\</?(?:b|i|s|u|font[^\\>]*)\\>|\\{\\\\.*?}|\\\\h|\\\\N");
```

Sami32:
@Nadahar Is your comment useful ?
I saw that and didn't understood that `<br />` was handled. Do you mean that it is ?

Nadahar:
@Sami32 No, it isn't. Since you asked if ```<br />``` was handles, I replied with what was, implying that nothing else is.

Sami32:
@Nadahar I was just reporting that is wasn't, and could be encounter in some recent subtitles.

Nadahar:
@Sami32 I interpret a sentence ending with ", right?" like a question. The question is if it should be removed, since there is no standard for this we can't remove everything anyone happens to use in a subtitle file. It's a fine line between removing what shouldn't be there and removing something that should be there.

Sami32:
@nadahar For me, it is like `correct ?`.
So next time i'll write french, it should be better...

Nadahar:
@Sami32 That's the same interpretation I have, and it's still a question. You suggest something, and then you ask for confirmation or refutal of that suggestion. My intention was to give that confirmation.

When it comes to the tag, according to [this](http://ale5000.altervista.org/subtitles.htm) (which is the closest thing to a "standard" for these tags I could find) the correct tag for line break is ```\N```.

Sami32:
@Nadahar So your answer was not clear to me as "yes" or "not" ;-)

`<s>` is only found in VLC or , but not in once instance of subtitles if i believe the statistics that have be done recently in subtitles website.

Nadahar:
@Sami32 I'm sorry if it weren't clear, maybe [this](https://regex101.com/r/4DoJ16/1) is better.

According to [the same page as my previous reply](http://ale5000.altervista.org/subtitles.htm) ```<s>``` is supported by MPC, MPC-HC and VLC while BSPlayer and MPlayer is untested. I'm not sure what you mean, but if you have better information than what this PR was based on, please present it. I'm just saying that our current implementation is based on the links given in [this post](https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1019#issuecomment-249378292).

Nadahar:
Supporting it is easy, this regex will remove it:
```java
Pattern pattern = Pattern.compile("\\</?(?:b|i|s|u|br\\s*/|font[^\\>]*)\\>|\\{\\\\.*?}|\\\\h|\\\\N");
```
The question is what is the correct thing to do, if we're just going to add anything any subber happen to use at one stage it will be a never ending endeavor.

Sami32:
@Nadahar Thanks, i read that already. Telling that some players support it doesn't mean that it is used...
I'll remember to not use this form anymore in future to avoid misunderstanding, as for me it need answer only if it was wrong.

Sami32:
@Nadahar Yes, i agree, but if i attract attention on this one it's because i thought that line ending is "important" part of displaying and asked feedback.
Do you mean that it doesn't have to be added ? (i know it is not standard, like many others)

Nadahar:
@Sami32 I mean that it doesn't automatically need to be added just because someone has used it. Whether or not it should be added I don't have enough information to evaluate, can you link to the statistics you mentioned for example? It it's just one guy that doesn't know what he's doing, I don't think it should be added. If it's a new "trend" using ```<br />``` instead of ```\N``` it's probably smart to add it. The thing is that I don't quite understand why using ```<br />``` is easier/better ```\N``` in any way, so it looks to me like there might simply be someone that assumed that this was supported since some other HTML tags are without actually checking. Do you know if any players support this tag?

Sami32:
@Nadahar Maybe you don't know this Addic7ed website, it a subtitles afficionado site where most of translator are the same for a serie since many years. I understand that they know about subtitles, it's why i report it.

But since i always check on my computer, i never saw that before.

If i can found again this subtitles statistics website i'll leave you the link

Nadahar:
I'm making this comment in this PR because this is where the VLC renderer profile was introduced even though it really shouldn't be related to this PR.

I just tested the new [VLC 3.0 preview version called 360](http://people.videolan.org/~jb/Builds/360/) and indeed as I hoped their DLNA implementation is very different. I haven't tested it thoroughly, but browsing works properly with UMS now. I have problems with playback of some files, so we probably need to update the renderer configuration, but I don't know if it's premature to do that based on a preview version or if we should wait for the 3.0 release before making any changes.

PR #1031, opened 2016-10-08
Re-submit PR #936

Sami32:
I did deleted the anterior branch found in https://github.com/UniversalMediaServer/UniversalMediaServer/pull/936

But perhaps you should test it with PS3 before merge it ? never know for sure with these crappy things...

**EDIT**: Already implemented in the "concurrent" media server programs...


Sami32:
Thanks :+1:

Issue #1032, opened 2016-10-08
Log encoding issue ?


Sami32:
MB (StatusTab.12) is displayed in the log as `ÐœÐ‘`.
> INFO  2016-09-11 19:00:19.964 [main] OS: Mac OS X 64-bit 10.11.6
> INFO  2016-09-11 19:00:19.964 [main] Encoding: UTF-8
> INFO  2016-09-11 19:00:19.964 [main] Memory: 683 **ÐœÐ‘**
> INFO  2016-09-11 19:00:19.971 [main] Language: Russian

https://github.com/UniversalMediaServer/UniversalMediaServer/blob/86502e6a72005cf18055f8eb118f85d2d833a004/src/main/resources/i18n/messages_ru.properties#L454-L455

**EDIT:** It's remind me my first post on the forum about log encoding issue with french accents. Maybe the codepage is not correctly handled ?
> TRACE 2016-09-11 19:00:23.925 [mplayer-0] Successfully initialized buffer to **52Â 428Â 800** bytes.
...
TRACE 2016-09-11 19:00:24.128 [main] Adding new child "Ð’ÑÐµ Ñ‚Ñ€ÐµÐºÐ¸ Ð°ÑƒÐ´Ð¸Ð¾" with class "net.pms.dlna.virtual.MediaLibraryFolder"

Nadahar:
@Sami32 That's very strange given: https://crowdin.com/translate/universalmediaserver/1/enus-ru#q=statustab.12

Nadahar:
@Sami32 It's more likely that whatever software you use to read the log file doesn't support Unicode. The log file is supposed to be UTF-8 AFAICR, but if you open it in a non-unicode capable software like ```Notepad.exe``` or similar it will be interpreted like a ANSI document.

Unless you are sure that your current editor supports UTF-8, try [Notepad++](https://notepad-plus-plus.org/). It is my favorite editor in general, handles syntax highlighting for a lot of programming languages, different encodings and a lot of other goodies, AND it is lightweight and can run on your computer. Try reading the same logfile and report the result back.

Sami32:
@Nadahar Thanks, you was right, Word Pad did not displayed correctly.

And for one time we share something, Notepad++ is very usefull and light !
Since most often the users interface have european language, i opened directly from inside the ZIP without issue. So i just changed his default viewer.

PR #1035, opened 2016-10-08
Unlocalize log messages

Nadahar:
Inspired by #1032 I threw together this little PR. The log isn't supposed to be localized but are a few places, so I removed the ones I found with the exception of the two error messages for shared folders in ```PMS.java``` that I just rewrote some. The reason I kept the localization there is that they refer to the ```Navigation/Share Settings``` tab, and it's easier for the user if the reference match the name in the GUI.

I don't know if I found them all, I simply searched the whole code with:
```
\.(INFO|WARN|ERROR|TRACE|DEBUG)\([^;]*Messages.getString\([^;]*;
```

I think I've found most of them though.

@SubJunk I know I used "folder" instead of "directory" and that you dislike that. To me the most important thing is to be consistant. I used to dislike ```folder``` to, simply because I was more used to ```directory``` from the olden days, but I've grown accustomed to it and now I prefer it. The reason I prefer it is that I think it's more "correct", as I see "directory" to be the name of the whole file/folder structure while "folder" is merely one of those folders containing other files and folders. If you disagree, just change it.


Sami32:
@Nadahar It's look good to me, since it's make this part of the code more easily readable as well and avoid 2 lines of translation.

What about the numbers encoding ? Is there is a way to avoid that ?
> TRACE 2016-09-11 19:00:23.925 [mplayer-0] Successfully initialized buffer to 52Â 428Â 800 bytes.

When you use "use automatic maximum bandwitdth " with norvegian language, do you get encoding displaying issue with that extra informations logging ?

Nadahar:
@Sami32 The numbers encoding would have to be changed on each line of code that logs the numbers in question. Java use the OS locale settings by default, so the number formatting will be as defined in the OS setting unless specificly overridden.

When it comes to that particular log message it is generated [here](https://github.com/UniversalMediaServer/UniversalMediaServer/blob/95b529762e7fc92d2522fd3d751226975774cfb3/src/main/java/net/pms/io/BufferedOutputFileImpl.java#L159) and the format is defined [here](https://github.com/UniversalMediaServer/UniversalMediaServer/blob/95b529762e7fc92d2522fd3d751226975774cfb3/src/main/java/net/pms/io/BufferedOutputFileImpl.java#L83). It seems to me that the format is overridden so that it's supposed to generate the same output without regard to the local number formatting, but apparantly it doesn't work as intended. The instructions for how to use ```DecimalFormatter``` and generate patterns can be found [here](https://docs.oracle.com/javase/7/docs/api/java/text/DecimalFormat.html) and anyone willing to read that can come up with a better definition.

I don't know how you got the example above from, but it seems that you again didn't use a Unicode capable editor. In any case, it's strange that a simple ```,``` end up being a double-byte character...  it must be ```¸``` (U+00B8) which is ```c2 b8``` in UTF 8. ```c2``` is ```Â``` in Windows 1252 but that would mean that the second character ```b8``` should display as ```¸``` so I really don't know how you got that result.

Regarding the your last sentence I'm not sure what you mean. I don't use "automatic maximum bandwidth" though as I used wired a network.

Sami32:
@Nadahar I repeat myself, but that not a viewer issue but why it's encoded like that in first place.
[debug.log.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/518275/debug.log.txt)

Thanks for pointing me to the Java classes.
Why not use `public void applyPattern(String pattern)` it will avoid number localisation in the log ?

Concerning "Use automatic maximum bandwidth" you teach me that it's only working for Wireless connection ?! Are you sure about that ?
Whatever, when used with a non AINSI language, you can see strange characters in your interface log concerning this bandwidth extra log while UMS running, so that's mean an encoding issue for me.

Nadahar:
@Sami32 When I open the file you attached with WordPad (non-unicode) I get the same result as you pasted. When I open it with Notepad++ it is shown with a space as a number formatter.

Why the defined format results in Unicode separators at all is an enigma to me, but I see from the code that one "rule" is already broken - you're not supposed to call the ```DecimalSeparator``` constructor directly but use the factory - but we call the constructor directly. I don't know what consequences breaking that rule will have - it might or might not be the cause. I'll have to look further at it to try to understand what happens here.

As far as I can see using ```applyPattern``` wouldn't change anything, you will still need to have an instance of ```DecimalFormatter``` to call that method and it probably does the same thing passing it to the constructor does.

I'm not saying that "Use automatic maximum bandwidth" only works with wireless - I'm just saying that it's not needed when you have a proper (1 Gbps) wired connection. I doubt I have any media that can't be streamed on my available network bandwidth. What happens if the bandwidth is deemed to low, is that it is transcoded to a lower quality format, and there's no situation where that's what I want to happen. Also, running the "automatic maximum bandwidth" has a cost both in CPU cycles and potential bug situations, so I generally avoid using things I don't need. I will try to turn it on and see if I get any of those log entries.

I'm not sure what you mean by "ANSI language", my use of the term might not be entirely accurate, what I mean by [ANSI](https://en.wikipedia.org/wiki/ANSI_character_set) is 8 bit character encoding with different [code pages](https://en.wikipedia.org/wiki/Windows_code_pages) depending on the local language/character set. In short, it's "pre Unicode" character encoding. 

It's common to refer to this as ANSI in developers terms even if it's not techincally accurate. I guess the main reason for this is Microsoft's practice of naming their methods with a appending A (ANSI) for non-unicode versions a W (WIDE) for unicode versions. Example: ```GetClassInfoExW (Unicode) and GetClassInfoExA (ANSI)```. You can find more on the discussion of the (mis)use of the word [here](http://stackoverflow.com/questions/701882/what-is-ansi-format).

I'm guessing you mean e.g. Windows 1252 with "non-ANSI language", but I mean exactly things like Windows 1252. This isn't the issue here though, UMS doesn't apply the local code page when writing logs (if it did, any character not in that code page wouldn't be valid, like file names from users with non-latin characters), instead it uses UTF-8 which is one of several possible unicode encodings. The reason UTF-8 has become so popular is that while it can encode any unicode character, the first 127 characters is identical to ASCII - which again is the basis for many of the most popular 8 bit code pages. That basically means that the normal latin letters and the most common signs will work even if you read it as if it was ANSI/8 bit. It also means that the latin letters (which is what the the dominant charaters used on the internet) only use 8 bits while still being able to encode other characters with a higher bith width. The BAD thing about UTF-8 is data the byte-length of a string isn't proportional to the number of characters in a string. UTF-16 (encoding all characters with 16 bits) was supposed to solve this, where the byte length would always be twice that of the number of characters - but from what I understand UTF-16 ran out of address space as well as unicode expanded and today UTF-16 has no real benefit over UTF-8 and will simply waste bits/memory/bandwidth under normal circumstances. Windows (and many other systems AFAIK) still use UTF-16 (Wide character) internally though, so it's something we'll have to live with many years to come. UTF-32 still has the capacity to produce known length strings, but at the cost of wasting a lot of buts/space when encoding most characters.

All this means that when you view an UTF-8 encoded text with a 8-bit editor, most characters will be the same (the ASCII range) but anything above that will be displayed as two or more characters as the 8-bit editor doesn't understand multi-byte characters. That's why one character displayed as multiple characters is a pretty safe bet that a non-unicode capable viewer is trying to display unicode text. If applies to all characters your source is probably UTF-16, if it only applies to some your source is probably UTF-8. Get it?

What I still don't know is why the number separator (not the decimal sign) end up being outside the ASCII range in the first place. I think the format is supposed to be ```nn,nnn,nnn```. Those pesky yankees think that ```.``` is the decimal separator and the ```,``` is the thousands separator ([reference](https://docs.oracle.com/cd/E19455-01/806-0169/overview-9/index.html)) so I think that's what has been the intent of this formatting.

Nadahar:
@Sami32 I've finally figured out what happens. In the definition ```#,###``` the ```,``` isn't a literal, it's a symbol for the locale thousands separator. 

Because of another "flaw" in UMS, the OS locale is overridden globally when UMS starts up and set to "UTF-8" - which isn't a real locale. The reason for doing this is because it's a lazy way of making sure that reads and writes are UTF-8 by default. The bad thing about such laziness is that you actually loose the OS locale information and thus have no way to use the correct locale when that's needed. I started to "fix" this a long time ago, but it's a lot of work (there's maany places in the code that need to be changed) and there was no general interest for this among the others so it stalled at some stage. 

The "UTF-8" locale in Java is clever, or it might be that Java applies this cleverness to all locales that use space as thousands separator (French does, Norwegian doesn't but because of the locale override I do get the same result as you do), but the cleverness is that it's not a normal space that's used, but a magical [no-break space U+00A0](http://www.fileformat.info/info/unicode/char/00a0/index.htm). This is clever because it means that number wont be broken up but moved to the next line as a whole when word wrapped. The less clever part though is that this character ends up being a 16 bit characer in UTF-8 encoded as ```0xC2 0xA0``` (194 160). Those bytes in [Windows 1252](https://en.wikipedia.org/wiki/Windows-1252) results in the characters [Â](https://en.wikipedia.org/wiki/%C3%82) and [NBSP](https://en.wikipedia.org/wiki/Non-breaking_space) - which is exactly what you see when opening it in Wordpad.

Since the log is supposed to be unlocalized US English I'll change it to use ```,``` as thousands separator (even though I hate that as for us Europeans that is the decimal separator).

Sami32:
@Nadahar Thanks to have looked at it.

**Ok, so i will try to make it clear one time for all: Do short short short comments for me ! 
As i don't understand what you are telling i just give up and read the last 5 line...**

French accents give problem so i guess norvegian special characters will do as well in the "use automatic maximum bandwidth" extra log.

**P.S.** The life is too short for "hate" ;-)

Nadahar:
@Sami32 2aa158c84de2741401c16718430027b8a93fcd29 fixes this for me, please confirm.

Nadahar:
@Sami32 I can't explain things in short comments, I can only make assertions then. My experience is that making assertions often isn't very helpful, but if you don't read my explanations anyway I'll try to do so. It will probably just increase our misunderstandings though.

Sami32:
@Nadahar Build is on his way, will need to wait 45 minutes ;-)

Sorry, i didn't want offense you, but it's really difficult for me to understand you, so i think that we can do like for translation, shorter better. Yourself have difficulty to understand me when i get long so...

Nadahar:
@Sami32 No offence, there's just a lot of things I can't say in a short comment. I generally explain (to you and anyone else) to avoid misunderstandings - which in my view is a common outcome what people just make short assertions/claims.

About my interpretation of what you write, it doesn't come down to the length - but the fact that some sentences are constructed in such a way that I can't figure out what they intended meaning is - or that I can think of several possible intended meanings (ambiguity). I'm think in a very "strict" way, so when something is ambiguous to me my brain refuses to process any further as I might pick the wrong meaning rendering the thoughts useless. I know I'm a bit strange :wink:

Sami32:
@Nadahar I don't think "strange" when i think about you, but "perfectionnist", that could make 500 lines comments to explain or wait for ever that the PR look perfect for you :wink: 

**P.S.** Perfection is a human "dream" and have a price...

Sami32:
@Nadahar Ok, that's work for me as well. So i suppose that it will for russian computer locale as well...

I joint you 2 log related to the "Use maximum bandwidth" extra log encoding issue. It use the computer locale not the UMS locale.
[debug_rus.log.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/518553/debug_rus.log.txt)
[debug.log.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/518554/debug.log.txt)

Nadahar:
@Sami32 I'm definitly suffering from perfectionism and some times it ends up costing me a lot of time and frustration, but it also saves me a lot of problems "down the road" if things are a bit complicated. It's hard to say if the total (in time and effort) is positive or negative when it comes to computer programming. There are other areas in life where it doesn't have much benefit but that's another story :wink:

Believe it or not but I'm usually trying pretty hard to "relax" and leave imperfections alone - so I usually don't see what I do as perfectionistic :nerd_face: 

I'm not sure what to make of your attached logs. The "rus" one doesn't seem to be generated with this branch - is it? The other one seems fine to me. The chosen UMS language is not the same as the [locale](https://en.wikipedia.org/wiki/Locale_(computer_software)), the locale is normally defined by the the OS configuration but as stated in UMS' case is overridden and set to a fictional locale called "UTF-8".

Sami32:
@Nadahar Nobody is perfect, so just try to don't ask the others to be :wink: 

My log was just an example of what you will see by yourself in your log when you'll select "Use maximum bandwidth" option. The extra verbose use ignore UMS locale and use computer locale. And strangely the characters encoding is not correctly readable in UTF-8 editor not either in the interface log.
Better than you try yourself than my poor explanation :-(

**P.S.** Yes, one was done with 6.5.0 that i normally use, as i already said that 6.5.1 have issue recognizing Panasonic Viera when running before the TV. And it's look like the problematic change was merged in 7.0.0 now.

Nadahar:
@Sami32 I don't understand. I have tried it myself, and I got the same result as you when opening the files in WordPad (non-unicode capable). I made a fix in this PR, and it now works fine for me. Are you saying that there's still an encoding issue for you with this PR?

Sami32:
@Nadahar Your PR is "perfect" ;-)

Just trying to say that "Use maximum bandwidth" generate extra verbose in the log that is localized when it shouldn't.

Nadahar:
@Sami32 Ah, thanks for the examples - now I understand. 

Trying to be short, there is nothing we can do about the language as that is output generated by the OS forwarded to UMS' log. You will find other examples of that in the log like som error messages.

Regarding the encoding of the OS messages, this is due to the aforementioned "flaw" in UMS where the OS locale is overridden. That means we don't know the OS locale (code page) and can't interpret it correctly. It is read as UTF-8 when it is (for French OS) Windows 1252.

Nadahar:
I think all issues that can be resolved are resolved then, and I'm going to merge this PR.

@Sami32 I don't know what you mean that this PR is "perfect". I'm sure one could find possible improvements if one looked hard. I don't want to merge something before it has a certain standard and I believe that I have done a reasonable effort to make sure it will solve what it's designed for and doesn't have any unintended consequences. I'm not in a rush to merge either, unless a release is around the corner a few days give or take means nothing.

Do I have to throw in a couple of bugs to meet your standard? :stuck_out_tongue:

Sami32:
@Nadahar It's "perfect" (just kidding you) for a merge :wink:

Sami32:
https://github.com/UniversalMediaServer/UniversalMediaServer/blob/d259bfc070165a2d49e51e4e0e47ee8338519885/src/main/java/net/pms/io/BufferedOutputFileImpl.java#L287

Nadahar:
@Sami32 It's very hard to guess what you mean with those links. Both those places are taken care of by this PR - are you saying that there's something wrong with the output?

Sami32:
@Nadahar Don't worry, it's just a reminder for myself, as i discovered these lines in an other log with totally different issue.
Just making me remember which part use "formatter" classe.

PR #1036, opened 2016-10-09
Respect the renderer setting H264Level41Limited when deciding whether to stream or transcode

SubJunk:


SubJunk:
@UniversalMediaServer/developers I think this one can be merged pretty quickly, I've tested it and it works well

Sami32:
@SubJunk Cool, that's a good idea.

Nadahar:
@SubJunk I've pushed a minor tweak. The reasons for the changes are:
* The reason to use ```{}``` instead of concatenating strings when making logging calls isn't just aesthetic - the main point is to avoid having do run a StringBuilder when the results are not going to be used (when debug level != trace). Whenever you concatenate strings with ```+``` in Java the compiler makes it into a StringBuilder, which is said to be very expensive in Java. When passed unconcatenated LogBack makes sure to only build the finished string *if* the message is actually used.

Nadahar:
@Sami32 I've already changed the null handling in 9569f7de044264495152f999a84ffcd31275cabf :wink:

SubJunk:
@Nadahar and @Sami32 thanks for the improvements :)

PR #1037, opened 2016-10-09
FrameAppender synchronization

Nadahar:
I forgot to shut down UMS last night, and when I woke up today there were a lot of stalled threads that had thrown exceptions like this:
```logtalk
INFO  15:42:40.369 [Thread-460] Exception in thread "Thread-226" Exception in thread "Thread-332" Exception in thread "Thread-372" Exception in thread "Thread-172" Exception in thread "Thread-460" Exception in thread "Thread-299" Exception in thread "Thread-224" Exception in thread "Thread-241" Exception in thread "Thread-173" Exception in thread "Thread-606" Exception in thread "Thread-442" Exception in thread "Thread-443" Exception in thread "Thread-552" Exception in thread "Thread-192" Exception in thread "Thread-659" Exception in thread "Thread-227" Exception in thread "Thread-298" Exception in thread "Thread-281" Exception in thread "Thread-583" Exception in thread "Thread-461" Exception in thread "Thread-533" Exception in thread "Thread-390" Exception in thread "Thread-205" java.lang.Error: Interrupted attempt to aquire write lock
INFO  15:42:40.371 [Thread-460] 	at javax.swing.text.AbstractDocument.writeLock(AbstractDocument.java:1349)
INFO  15:42:40.375 [Thread-460] 	at javax.swing.text.AbstractDocument.insertString(AbstractDocument.java:702)
INFO  15:42:40.391 [Thread-460] 	at javax.swing.text.PlainDocument.insertString(PlainDocument.java:130)
INFO  15:42:40.398 [Thread-460] 	at javax.swing.JTextArea.append(JTextArea.java:477)
INFO  15:42:40.404 [Thread-460] 	at net.pms.newgui.TracesTab.append(TracesTab.java:186)
INFO  15:42:40.412 [Thread-460] 	at net.pms.newgui.LooksFrame.append(LooksFrame.java:523)
INFO  15:42:40.419 [Thread-460] 	at net.pms.logging.FrameAppender.append(FrameAppender.java:109)
INFO  15:42:40.425 [Thread-460] 	at ch.qos.logback.core.UnsynchronizedAppenderBase.doAppend(UnsynchronizedAppenderBase.java:84)
INFO  15:42:40.430 [Thread-460] 	at ch.qos.logback.core.spi.AppenderAttachableImpl.appendLoopOnAppenders(AppenderAttachableImpl.java:48)
INFO  15:42:40.432 [Thread-460] 	at ch.qos.logback.classic.Logger.appendLoopOnAppenders(Logger.java:270)
INFO  15:42:40.742 [Thread-460] 	at ch.qos.logback.classic.Logger.callAppenders(Logger.java:257)
INFO  15:42:40.746 [Thread-460] 	at ch.qos.logback.classic.Logger.buildLoggingEventAndAppend(Logger.java:421)
INFO  15:42:40.759 [Thread-460] 	at ch.qos.logback.classic.Logger.filterAndLog_0_Or3Plus(Logger.java:383)
INFO  15:42:40.762 [Thread-460] 	at ch.qos.logback.classic.Logger.debug(Logger.java:482)
INFO  15:42:40.765 [Thread-460] 	at net.pms.io.PipeIPCProcess.run(PipeIPCProcess.java:86)
```

After a quick peek I discovered that FrameAppender's call to update the GUI is run from whatever thread makes the logging call. Since Swing isn't multithreaded that isn't allowed and every update of the GUI must be done via the event dispatcher thread. I see this as the likely cause of the issue, and corrected that.

Ideally I'd like for the "separation" between other threads and Swing to be outside the "GUI classes" (net.pms.newgui.*) since I think that's the logical place to do the separation and because any GUI component can safely call such a method without queuing the event dispatcher as they already run in the event dispatcher thread. That isn't feasable here because of the IFrame construct though (because when running headless there *is* no event dispatcher thread), so I had to put the change in LooksFrame even though I dislike it.

I've also commented out the rolling file appender configuration from the logback configuration files since some upgrade of logback has started throwing errors because it was configured with the same file name as the non-rolling file appender. Commenting it out should have no practical effect since the rolling file appender isn't used by default and only serves as a configuration example for anyone that want to configure it differently. Coming up with a separate name for the rolling file name would require changes in several classes and further complicating the configuration file with another setting, so I found it easiest to simply comment out the example configuration. This takse care of the errors thrown by logback during LogBack configuration when UMS starts.

I've attached the logfile containing both errors (It's compressed as it is too big to attach uncompressed):
[debug.zip](https://github.com/UniversalMediaServer/UniversalMediaServer/files/518409/debug.zip)

---
This change is [Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/1037)

### DIFF
--- a/src/main/external-resources/logback.headless.xml
+++ b/src/main/external-resources/logback.headless.xml
@@ -50,9 +50,9 @@
 		Appender for the default logfile that will roll the logfile after 10 MBytes
 		and will keep a maximum of 5 old and compressed logs.
 	-->
+	<!--
 	<appender name="default.log.10MB" class="ch.qos.logback.core.rolling.RollingFileAppender">
-		<!-- No threshold filtering, log everything the root logger allows -->
-		<file>${logFilePath}${file.separator}${logFileName}</file>
+		<file>${logFilePath}${file.separator}${logFileName}.rolling</file>
 		<rollingPolicy class="ch.qos.logback.core.rolling.FixedWindowRollingPolicy">
 			<fileNamePattern>${logFilePath}${logFileName}.%i.zip</fileNamePattern>
 			<minIndex>1</minIndex>
@@ -66,6 +66,7 @@
 			<pattern>%-5level %d{yyyy-MM-dd HH:mm:ss.SSS} [%thread] %msg%n</pattern>
 		</encoder>
 	</appender>
+	-->
 
 	<!--
 		For debugging purposes it can be handy to define custom levels for
--- a/src/main/external-resources/logback.xml
+++ b/src/main/external-resources/logback.xml
@@ -59,8 +59,8 @@
 		Appender for the default logfile that will roll the logfile after 10 MBytes
 		and will keep a maximum of 5 old and compressed logs.
 	-->
+	<!--
 	<appender name="default.log.10MB" class="ch.qos.logback.core.rolling.RollingFileAppender">
-		<!-- No threshold filtering, log everything the root logger allows -->
 		<file>${logFilePath}${file.separator}${logFileName}</file>
 		<rollingPolicy class="ch.qos.logback.core.rolling.FixedWindowRollingPolicy">
 			<fileNamePattern>${logFilePath}${logFileName}.%i.zip</fileNamePattern>
@@ -75,6 +75,7 @@
 			<pattern>%-5level %d{yyyy-MM-dd HH:mm:ss.SSS} [%thread] %msg%n</pattern>
 		</encoder>
 	</appender>
+	-->
 
 	<!--
 		For debugging purposes it can be handy to define custom levels for
--- a/src/main/java/net/pms/newgui/LooksFrame.java
+++ b/src/main/java/net/pms/newgui/LooksFrame.java
@@ -520,7 +520,13 @@ public class LooksFrame extends JFrame implements IFrame, Observer {
 
 	@Override
 	public void append(final String msg) {
-		tt.append(msg);
+		SwingUtilities.invokeLater(new Runnable() {
+
+			@Override
+			public void run() {
+				tt.append(msg);
+			}
+		});
 	}
 
 	@Override


Nadahar:
@Sami32 Please don't push merge commits into PRs - they make such a mess!

Sami32:
@Nadahar Ok, sorry about that :-/

Sami32:
@Nadahar Could you explain me what is the best way to merge from Github interface, for UMS project case, as i don't really understand their technical word ?
- Create a merge commit
- Squash and merge
- Rebase and merge

valib:
Reviewed 3 of 3 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/1037)*

Nadahar:
@valib Thank you.

@Sami32 I've rebased away the merge commit so it's "fixed" now :wink: It's not that these merge commits do any harm to the code itself, but they make the history very difficult to read when you need to see how something evolved - and since I spend a lot of time doing exactly that when I try to figure something out, I've learned to avoid that as much as possible.

When it comes to the 3 different "merge" options GitHub offers it's mostly a matter of taste - the resulting code is the same for all, what differs it the git history:

* Create a merge commit: Creates a new commit in master that include all the commits from the PR as separate commits. This creates a "split" in the history because the merge commit will have 2 parents, one being the history of master as it was without the PR while the other is the history of the PR. When backtracking history you have to choose which path to follow, and you might have to follow both to figure out what was really going on. One such split isn't a big problem, but when you have to traverse many the number of different routes quickly become overwhelming and impossible to follow.
* Squash and merge: Creates a new commit in master that include all changes from the PR in one generated commit that consists of the sum of all the PR commits. It has the same issue with 2 parents as the above merge commit, but the history is significantly simpler since all changes from the branch is combined (squashed) into one commit. You loose the individual commits though making it harder to track how the branch developed and separate different changes done by the same commit. I don't fancy this method in general as that one commit can have a lot of changes making it very difficult to understand. If a PR contains several small commits that "works on the same change" and would make more sense as a single commit, it could be a good option though.
* Rebase and merge: This is the only one that doesn't make a merge commit and thus doesn't create the "two parents" problem. What rebase does instead is that it takes all the commits from the PRs and place them after the commits in master as if the branch and the merge never happened. This generates the cleanest history, but it has other problems in that a rebase rewrites history. This is a bit complicated, so instead of me trying to explain it you can read [this](https://git-scm.com/book/en/v2/Git-Branching-Rebasing) or [this](https://www.atlassian.com/git/tutorials/merging-vs-rebasing/conceptual-overview).

Projects usually does one or the other - merges or rebasing. UMS uses merging, so we should stick with that as a general rule, but in my opinion rebasing could be a better option at least for small single-commit PRs - although that's not my decision to make.

Sami32:
@Nadahar Thanks for lightening my lantern, i get a better understanding now, but i need to read your interesting links :+1:  Their schemas and my bad english will perhaps give a meaning to that ;-)

SubJunk:
@Nadahar yeah I often squash simple commits. The rebase and merge button removes merge commits by the way so I don't mind them being in the branch - they will disappear upon merge

Nadahar:
@SubJunk I haven't really used "rebase and merge" so I didn't know. If that is actually the case that is good news, but what if the merge had manually resolved conflicts? I'm guessing that the rebase won't remove it in that case?

PR #1038, opened 2016-10-14
Automatically rescan shared folders if there is a change in a monitored folder

SubJunk:
In my opinion this feature is an important part of keeping the program up to date with the functionality of our competitors, and it especially makes sense given the changes in 7.0.0.

The only real change to this I can think of is to make a config variable to be able to disable it, but then I think it makes sense to include it in the functionality of "monitor" which we already have.

### DIFF
--- a/src/main/java/net/pms/dlna/RootFolder.java
+++ b/src/main/java/net/pms/dlna/RootFolder.java
@@ -150,6 +150,13 @@ public class RootFolder extends DLNAResource {
 			addChild(r);
 		}
 
+		/**
+		 * Changes to monitored folders trigger a rescan
+		 */
+		for (DLNAResource r : getConfiguredFolders(tags)) {
+			FileWatcher.add(new FileWatcher.Watch(r.getSystemName(), LIBRARY_RESCANNER));
+		}
+
 		for (DLNAResource r : getVirtualFolders(tags)) {
 			addChild(r);
 		}
@@ -279,8 +286,12 @@ public class RootFolder extends DLNAResource {
 	}
 
 	private List<RealFile> getConfiguredFolders(ArrayList<String> tags) {
+		return getConfiguredFolders(tags, false);
+	}
+
+	private List<RealFile> getConfiguredFolders(ArrayList<String> tags, boolean monitored) {
 		List<RealFile> res = new ArrayList<>();
-		File[] files = PMS.get().getSharedFoldersArray(false, tags, configuration);
+		File[] files = PMS.get().getSharedFoldersArray(monitored, tags, configuration);
 		String s = configuration.getFoldersIgnored(tags);
 		String[] skips = null;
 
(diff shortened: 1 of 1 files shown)


valib:
@SubJunk I don't think that this is a good idea to rescan all shared folders. The only folder where file was changed should be updated.

Nadahar:
I haven't really paid attention to the 7.0 branch, but triggering a full library rescan is a bad idea. For many users this can take hours.

I don't get what rescanning has to do with monitored folders either, isn't the natural thing to just scan whatever is changed? I'm not entirely familiar with the filewatcher and how it works, but given that you get a "change" event for every changed, created or deleted file or folder, scanning should only be of the changed element and any children if the element is a folder. It would probably be wise to make some logic that queued events for a few seconds and then consolidated changes in the same element into just one scan as I've read somewhere that such change elements can be reported in "chains" by the OS for some operations with multiple events for one actual event.

Another thing is to make sure that this is only done if the cache/medialibrary is enabled. Scanning is just a waste of time otherwise.

SubJunk:
@valib yeah that would be a good refinement to make

@Nadahar for me it only takes about 1 second or less to re-scan all folders with thousands of videos since files that are already in the database are skipped. This change does only happen if the cache is enabled

Nadahar:
@SubJunk Audio and picture collections are typically much larger in size than video collections, but if you have literally thousands of video files it should still take some time. Even if only the ones not in the cache is actually scanned, all folders need to be enumerated - which in itself can take time. It depends on a number of things, your hardware and the HDD's you use (SSD's don't count, they will enumerate the content more or less instantly) and if what you're asking is already cached or not (cached in the HDD buffer). There's also things like network drives which can be significantly slower - I moved my local git repos to to a network drive that has RAID and backup a while back - but I recently moved the most used ones back to my local HDD since I just couldn't stand the time it took to do almost any git operation. Even getting Eclipse to refresh the project (check for changed files if I've checked out another branch) took close to a minute. In short, I don't believe that it will be that quick under all circumstances.

Another thing is that the scanning system is made so that you don't have to scan from the top. A library scan is simply a scan initiated at the root folder, which scans every child recursively. Why not just do the exact same thing but start the scan from the changed element instead of from the rootfolder?

SubJunk:
@Nadahar I agree that a finer level of targeting would be best, as I said to @valib . That would be a good improvement to this feature.

I think having this feature is better than not having it though even in its current state. Do you think this branch as it is now is worse than master?

The hard drive I use for video storage is a 5400k RPM drive and doesn't have a large cache so it's a pretty decent indicator of average or less than average speed I think.

Nadahar:
@SubJunk I haven't studied how the file watcher works, but given that it could report back what file/folder was changed I'd think it shouldn't be too hard to scan just that.

When it comes to better or worse than master, I don't know what to say. It will probably work better for those with relatively few files, fast storage and where files change relatively seldom - and vice versa. I'm just arguing: Why not make it work better for everybody instead of better for some and worse for others?

SubJunk:
@Nadahar the `FileWatcher.listener` returns a filename variable which is the path to the folder that contains the file that was modified. I'm having difficulty figuring out how to make the scanner only scan that folder. Any ideas?

SubJunk:
Or any @UniversalMediaServer/developers @atamariya

Nadahar:
@SubJunk See #1065 for my ideas.

atamariya:
@SubJunk Calling RealFile.run() for a folder will scan that particular folder. But the larger problem is - it will scan the complete subtree!! As @Nadahar pointed out - the scanning piece needs a rewrite if it were to benefit from FileWatcher or you'll end up putting band-aids all over the place.

If we can efficiently analyze a single file with [this code](https://github.com/atamariya/UniversalMediaServer/blob/dev/src/test/java/net/pms/io/IOTest.java#L54-L63), that would be an ideal solution. It has following benefits:
- Enable two step scan (First, add all the files to DB. Second, analyse metadata.) If one has lot of video files, thumbnail generation blocks browsing currently. Plus first use experience will also improve. 
- TaskRunner limits the number of files (in a queue) that are analyzed instead of hogging the CPU
- Only anyalse the files that are changed
- Enable Media Library / cache folder view (browse by album, artist etc.) even when one has NOT navigated the associated folder
- Additionally, this will enable one to trigger indexing as a separate job without running PMS.

That's my two cents.

SubJunk:
I'm closing this since the work can be continued in 7.0.0 or elsewhere

PR #1039, opened 2016-10-15
Fix audio LPCM Standard respect.

Sami32:


Nadahar:
Did you test this on PS3? It could be a bug in the PS3 making L16 not working...

Sami32:
@Nadahar Of course, i just buy one for you ;-)
I just read the standard and the PS3 renderer from others media servers.
That's said, i thought as you when i accidentally discovered that.

You will need to test it or believe the one that will tell you that it's work...

Nadahar:
@Sami32 Did anybody confirm that it works? In that case there should be no problem.

Nadahar:
As a side note, WAV support for the PS3 is currently terrible. None of my test files would play correctly (wrong pitch) and many wouldn't play at all. It seems like it ONLY supports 48kHz 16bits stereo when the PS3 is configured for 48kHz. Therefore, much more should be transcoded then it currently is. But, that has nothing to do with this PR.

Issue #565, opened 2015-06-04
disable_transcode_for_extensions can't be enforced if subtitles are not explicitly disabled


Iridias:
**Problem**: even setting disable_transcode_for_extensions to "*" will sometimes result in UMS transcoding a file!

**Investigation**: after having a look at the source-code it seems, that this config-entry will only be used to determine compatibility, but not to enforce not transcoding the matched files.
In `net.pms.dlna.DLNAResource#resolvePlayer` its used to determine compatibility, but when deciding if transcoding is preferred, it might get overruled by the existence of subtitles:
`boolean preferTranscode = isIncompatible || hasSubsToTranscode;`
(Unless of course they're not disabled as well)

In the final decision there is a check for `isSkipTranscode()`, but this member is never set!

**possible solution**: allow setting the skipTranscode-member by configuration.

**temporary workaround for users**: disable subtitles in the config (key is: disable_subtitles )

Sami32:
@Iridias Please, could you post your full trace log when that's append, following these steps:
http://www.universalmediaserver.com/forum/viewtopic.php?f=9&t=556

Nadahar:
@Sami32 There's no need for a log, the problem is fully explained. This isn't a bug, it's a "design flaw".

Sami32:
Using `disable_transcode_for_extensions` in UMS.conf, alone, will not have effect on files with subtitle(s).` StreamExtensions = ` in the renderer configuration will not either.
But using the two together will do the trick, AFAIU.
https://github.com/UniversalMediaServer/UniversalMediaServer/blob/master/src/main/java/net/pms/dlna/DLNAResource.java#L607-L621

It need to use `VideoFormatsSupportingStreamedExternalSubtitles = `, `SupportedInternalSubtitlesFormats = ` and `SupportedExternalSubtitlesFormats = ` to get the job done without disabling the subtitles (`disable_subtitles`).

**Note**: `disable_transcode_for_extensions` is given as a test, debugging function, and the renderer configuration parameters should be preferred.
https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1040

Nadahar:
@Iridias This should finally be fixed now with #1053. This will be a part of the next release.

PR #1040, opened 2016-10-15
Information complémentaire

Sami32:


Nadahar:
@Sami32 Why won't it work with subtitles? Renderer supported subtitles should play fine..? Anything not supported by the renderer obviously won't play...

Sami32:
@Nadahar Did you tried it ?

Nadahar:
@Sami32 No, I haven't. I just don't understand why subtitles would make a difference as long as the renderer understand the subtitles. If it doesn't, it will need to be transcoded and since that is turned off it will obviously fail, but that's logical and the only way it can be.

Sami32:
https://github.com/UniversalMediaServer/UniversalMediaServer/issues/565

Nadahar:
@Sami32 If that's what you mean, it's not accurate to say simply ```Will not work with files having subtitle(s).```. Subtitles that can be streamed will be streamed, the problem is when there's a subtitle that can't be streamed. While one could argue that UMS should then instead stream the file without the external subtitle or stream it with an internal subtitle the renderer doesn't understand, that's not the choice that's been made. I think this logic should be looked at, but I don't think adding this line of text will solve any problems - especially since it's not entirely truthful.

Sami32:
> Problem: even setting disable_transcode_for_extensions to "*" will sometimes result in UMS transcoding a file!

Waiting that it's solve itself, or commenting or modify the logic...that's the question.

PR #1041, opened 2016-10-16
Search for subtitles in alternative folder even when a subtitle was already found

tdcosta100:
The original code does not search for subtitles in alternative subtitles folder if a subtitle is already found, because of `[short-circuiting evaluation](https://en.wikipedia.org/wiki/Short-circuit_evaluation)`.

### DIFF
--- a/src/main/java/net/pms/util/FileUtil.java
+++ b/src/main/java/net/pms/util/FileUtil.java
@@ -808,7 +808,7 @@ public class FileUtil {
 			}
 
 			if (subFolder.exists()) {
-				found = found || browseFolderForSubtitles(subFolder, file, media, usecache);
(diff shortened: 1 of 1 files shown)


Nadahar:
@Sami32 Short-circuiting evaluation is used "all the time" but usually the other way. Whoever wrote it obviously forgot about it at that time.

We usually do null checking with that:
```java
if (foo != null && foo.bar() == 3) {
}
```
```foo.bar()``` would throw a NPE if ```foo``` was null, but because the first evaluation turns out false in that case, and the operator ```&&``` requires both to be true to result in true, the second evaluation is never done and the NPE is avoided.

tdcosta100:
It's a very common mistake we do when programming, and usually it's not so easy to identify and locate. But it's very relevant depending on the feature it affects. In this case, it affected a relevant feature to me =P.

PR #1047, opened 2016-10-16
Update Sony-Bravia5500.conf

Sami32:
Correct an incorrect parameter since UMS 4.0.0 version.

### DIFF
--- a/src/main/external-resources/renderers/Sony-Bravia5500.conf
+++ b/src/main/external-resources/renderers/Sony-Bravia5500.conf
@@ -17,7 +17,7 @@ TranscodeExtensions = dvr-ms,dvr,mkv,dv,ty,mov,ogm,hdmov,hdm,rmv,rmvb,rm,asf,evo
 AccurateDLNAOrgPN = true
 ForceJPGThumbnails = true
 ThumbnailAsResource = true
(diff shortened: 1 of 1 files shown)


Sami32:
@Nadahar Should i revert it ?

Nadahar:
@Sami32 I have no idea, I'm completely confused as to what has been done here and why. What do these settings do first of all? The most important thing imo is to don't merge/revert etc before you think it's the "final" solution. There's no rush, take the time to figure it out and maybe test it if needed, so that whatever you end up doing is the best solution..

Sami32:
@as you saw it wasn't used since 4.0.0 and when restablished it given the issue the was referred by the phantom commit.
It's why i ask you that, but it's OK for me, i leave Git things to you.

Nadahar:
@Sami32 I still don't understand the phantom commit, so I'm as lost as you. I don't know if I appreciate you using @as as my nickname though.... or did you mean @ass?

Sami32:
@nadahar Sorry :flushed: i didn't saw it !
Sometimes my browser become so slow that it doesn't display the name or smiley automatically...
Be assured that i will never write it here, again less that i'm thinking it !

Issue #597, opened 2015-07-09
Disable transcoding


garygreen:
Just getting to grips with the media server, so my understanding maybe lacking. Does it always have to transcode or can you disable it? Basically I don't want files transcoded that can be natively played on TV. It should just instead 'push' to tv and that is all

Nadahar:
What formats are transcoded or streamed is decided by your renderer configuration. If you select the status tab in UMS you hopefully see your renderer (your TV). If you click it, you get some additional information about it. On the very top there's a ```<filename>.conf```. If you hold your mouse over this, you will see the full path to the configuration file, unless your TV is "unknown renderer" as simply running a default configuration. This file can be a bit cryptic to understand, so look at the ```DefaultRenderer.conf``` in the same folder for an explanation of the different parameters.

Iridias:
Keep in mind, that UMS may transcode files even if they're natively supported by your renderer, if those files contain subtitles. And even if you have configured UMS to not transcode files with those extensions. See: https://github.com/UniversalMediaServer/UniversalMediaServer/issues/565

Issue #1043, opened 2016-10-16
"High Profile Level 4.2" missing parameter (PS3/PS4)


Sami32:
While i looked at PlayStation renderers i saw that they was incompletes, but the most important part was a missing implementation of a parameter for _High Profile Level 4.2_ that make unnecessary transcoding for 4.2 profile (or giving error with a 4.2 superior profile if >4.1 profile was set to `true`).

SubJunk:
Yeah it would probably be good to have a renderer setting like MaximumH264Level instead of the existing 4.1 one

SubJunk:
This has been implemented for a long time with `H264LevelLimit` and `H265LevelLimit`

PR #1053, opened 2016-10-18
Disable transcoding #565

Nadahar:
This both fixes the existing option to disable transcoding for specific extensions and creates a new configuration option that disables transcoding completely. #565 #597 #1040 

For as long as I've been involved I've seen requests from people from time to time for the option to disable transcoding completely. Some people simply doesn't have a computer that can handle transcoding, other prefers that the file doesn't play if the renderer doesn't support it - or that the subtitle is ignored instead of transcoding the video.

When trying to fix "disable transcoding for specific extensions" it didn't cost much to add it. Regarding that feature, I don't think this can have been working properly for a very long time. As far as I can see this was broken way back in 43c0c12a7b417c69d5e5e7034f83c216c963c155 where basicly `skipTranscode` was set for the parent of the resource if was checked for (the parent would be a folder, and thus the information had no impact). There were other problems as well that has probably trickled in as time went by since the feature didn't work anyway.

I've done some testing and it seems to work for me, but since `DLNAResource` is, in my opinion, the worst mess in the whole codebase, it is extremely dangerous to change anything there as the side effects are very hard to predict. From what I can see the class variable `skipTranscode` isn't used for anything and could be removed, but since it's there I'm guessing some obscure plugin has used it once upon a time so I left it there. As a result I think we need to test this pretty thoroughly before a potential merge.

---
This change is [Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/1053)

### DIFF
--- a/src/main/external-resources/UMS.conf
+++ b/src/main/external-resources/UMS.conf
@@ -596,6 +596,15 @@ disable_transcode_for_extensions =
 # troubleshooting when you are not able to edit the renderer configuration.
 # Default: ""
 force_transcode_for_extensions =
+
+# Disable all transcoding
+# -----------------------
+# Do not transcode under any circumstances.
+# Warning: This will probably mean that your renderer will fail to play many
+# files or fail to show subtitles, but there are cases where this is still 
+# preferable.
+# Default: false
+disable_transcoding = 
  
 # ---< Audio settings tab >---------------------------------------------------
 
--- a/src/main/java/net/pms/newgui/LooksFrame.java
+++ b/src/main/java/net/pms/newgui/LooksFrame.java
@@ -448,7 +448,11 @@ public class LooksFrame extends JFrame implements IFrame, Observer {
 		tabbedPane.addTab(Messages.getString("LooksFrame.20"), gt.build());
 		tabbedPane.addTab(Messages.getString("LooksFrame.27"), pt.build());
 		tabbedPane.addTab(Messages.getString("LooksFrame.22"), nt.build());
-		tabbedPane.addTab(Messages.getString("LooksFrame.21"), tr.build());
+		if (!configuration.getDisableTranscoding()) {
+			tabbedPane.addTab(Messages.getString("LooksFrame.21"), tr.build());
+		} else {
+			tr.build();
+		}
 		tabbedPane.addTab(Messages.getString("LooksFrame.24"), new HelpTab().build());
 		tabbedPane.addTab(Messages.getString("LooksFrame.25"), new AboutTab().build());
 
--- a/src/main/java/net/pms/configuration/PmsConfiguration.java
+++ b/src/main/java/net/pms/configuration/PmsConfiguration.java
@@ -130,6 +130,7 @@ public class PmsConfiguration extends RendererConfiguration {
 	protected static final String KEY_DISABLE_FAKESIZE = "disable_fakesize";
 	public    static final String KEY_DISABLE_SUBTITLES = "disable_subtitles";
 	protected static final String KEY_DISABLE_TRANSCODE_FOR_EXTENSIONS = "disable_transcode_for_extensions";
+	protected static final String KEY_DISABLE_TRANSCODING = "disable_transcoding";
 	protected static final String KEY_DVDISO_THUMBNAILS = "dvd_isos_thumbnails";
 	protected static final String KEY_DYNAMIC_PLS = "dynamic_playlist";
 	protected static final String KEY_DYNAMIC_PLS_AUTO_SAVE = "dynamic_playlist_auto_save";
@@ -366,6 +367,7 @@ public class PmsConfiguration extends RendererConfiguration {
 			KEY_AUDIO_THUMBNAILS_METHOD,
 			KEY_CHAPTER_SUPPORT,
 			KEY_DISABLE_TRANSCODE_FOR_EXTENSIONS,
+			KEY_DISABLE_TRANSCODING,
 			KEY_ENGINES,
 			KEY_FOLDERS,
 			KEY_FORCE_TRANSCODE_FOR_EXTENSIONS,
@@ -2376,6 +2378,10 @@ public class PmsConfiguration extends RendererConfiguration {
 
 	public void setDisableTranscodeForExtensions(String value) {
 		configuration.setProperty(KEY_DISABLE_TRANSCODE_FOR_EXTENSIONS, value);
+	}
+
+	public boolean getDisableTranscoding() {
+		return getBoolean(KEY_DISABLE_TRANSCODING, false);
 	}
 
 	public String getForceTranscodeForExtensions() {
--- a/src/main/java/net/pms/configuration/PmsConfiguration.java
+++ b/src/main/java/net/pms/configuration/PmsConfiguration.java
@@ -538,9 +538,8 @@ public class PmsConfiguration extends RendererConfiguration {
 	 * @param loadFile Set to true to attempt to load the PMS configuration
 	 *                 file from the profile path. Set to false to skip
 	 *                 loading.
-	 * @throws org.apache.commons.configuration.ConfigurationException
 	 */
-	public PmsConfiguration(boolean loadFile) throws ConfigurationException {
+	public PmsConfiguration(boolean loadFile) {
 		super(0);
 
 		if (loadFile) {
--- a/src/main/java/net/pms/configuration/RendererConfiguration.java
+++ b/src/main/java/net/pms/configuration/RendererConfiguration.java
@@ -282,7 +282,7 @@ public class RendererConfiguration extends UPNPHelper.Renderer {
 							if (selectedRenderers.contains(rendererName) || selectedRenderers.contains(renderersGroup) || selectedRenderers.contains(pmsConf.ALL_RENDERERS)) {
 								enabledRendererConfs.add(r);
 							} else {
-								LOGGER.debug("Ignored " + rendererName + " configuration");
+								LOGGER.debug("Ignored \"{}\" configuration", rendererName);
 							}
 						} catch (ConfigurationException ce) {
 							LOGGER.info("Error in loading configuration of: " + f.getAbsolutePath());
@@ -2192,12 +2192,13 @@ public class RendererConfiguration extends UPNPHelper.Renderer {
 	public boolean isCompatible(DLNAMediaInfo mediainfo, Format format) {
 
 		PmsConfiguration configuration = PMS.getConfiguration(this);
-		if (configuration == null) {
-			LOGGER.error("Can't find configuration in isCompatible()");
-			return false;
-		}
 
-		if (format != null && format.skip(configuration.getDisableTranscodeForExtensions())) {
+		if (
+			configuration != null &&
+			(configuration.getDisableTranscoding() ||
+			(format != null &&
+			format.skip(configuration.getDisableTranscodeForExtensions())))
+		) {
 			return true;
 		}
 
--- a/src/main/java/net/pms/configuration/RendererConfiguration.java
+++ b/src/main/java/net/pms/configuration/RendererConfiguration.java
@@ -2190,25 +2190,24 @@ public class RendererConfiguration extends UPNPHelper.Renderer {
 	 * 				otherwise.
 	 */
 	public boolean isCompatible(DLNAMediaInfo mediainfo, Format format) {
+
+		PmsConfiguration configuration = PMS.getConfiguration(this);
+		if (configuration == null) {
+			LOGGER.error("Can't find configuration in isCompatible()");
+			return false;
+		}
+
+		if (format != null && format.skip(configuration.getDisableTranscodeForExtensions())) {
+			return true;
+		}
+
 		// Use the configured "Supported" lines in the renderer.conf
 		// to see if any of them match the MediaInfo library
 		if (isUseMediaInfo() && mediainfo != null && getFormatConfiguration().match(mediainfo) != null) {
 			return true;
 		}
 
-		if (format != null) {
-			String noTranscode = "";
-
-			if (PMS.getConfiguration(this) != null) {
-				noTranscode = PMS.getConfiguration(this).getDisableTranscodeForExtensions();
-			}
-
-			// Is the format among the ones to be streamed?
-			return format.skip(noTranscode, getStreamedExtensions());
-		} else {
-			// Not natively supported.
-			return false;
-		}
+		return format != null ? format.skip(getStreamedExtensions()) : false;
 	}
 
 	public int getAutoPlayTmo() {
--- a/src/main/java/net/pms/configuration/RendererConfiguration.java
+++ b/src/main/java/net/pms/configuration/RendererConfiguration.java
@@ -2186,12 +2186,15 @@ public class RendererConfiguration extends UPNPHelper.Renderer {
 	 * @param mediainfo The {@link DLNAMediaInfo} information parsed from the
 	 * 				media file.
 	 * @param format The {@link Format} to test compatibility for.
+	 * @param configuration The {@link PmsConfiguration} to use while evaluating compatibility
 	 * @return True if the renderer natively supports the format, false
 	 * 				otherwise.
 	 */
-	public boolean isCompatible(DLNAMediaInfo mediainfo, Format format) {
+	public boolean isCompatible(DLNAMediaInfo mediainfo, Format format, PmsConfiguration configuration) {
 
-		PmsConfiguration configuration = PMS.getConfiguration(this);
+		if (configuration == null) {
+			configuration = PMS.getConfiguration(this);
+		}
 
 		if (
 			configuration != null &&
@@ -2209,6 +2212,22 @@ public class RendererConfiguration extends UPNPHelper.Renderer {
 		}
 
 		return format != null ? format.skip(getStreamedExtensions()) : false;
+	}
+
+	/**
+	 * Returns whether or not the renderer can handle the given format
+	 * natively, based on its configuration in the renderer.conf. If it can
+	 * handle a format natively, content can be streamed to the renderer. If
+	 * not, content should be transcoded before sending it to the renderer.
+	 *
+	 * @param mediainfo The {@link DLNAMediaInfo} information parsed from the
+	 * 				media file.
+	 * @param format The {@link Format} to test compatibility for.
+	 * @return True if the renderer natively supports the format, false
+	 * 				otherwise.
+	 */
+	public boolean isCompatible(DLNAMediaInfo mediainfo, Format format) {
+		return isCompatible(mediainfo, format, null);
 	}
 
 	public int getAutoPlayTmo() {
--- a/src/main/java/net/pms/dlna/DLNAResource.java
+++ b/src/main/java/net/pms/dlna/DLNAResource.java
@@ -794,6 +794,11 @@ public abstract class DLNAResource extends HTTPResource implements Cloneable, Ru
 			return resolvedPlayer;
 		}
 
+		if (configurationSpecificToRenderer.getDisableTranscoding()) {
+			LOGGER.trace("File \"{}\" will be streamed since transcoding is disabled", getName());
+			return null;
+		}
+
 		String configurationSkipExtensions = configurationSpecificToRenderer.getDisableTranscodeForExtensions();
 		String rendererSkipExtensions = renderer == null ? null : renderer.getStreamedExtensions();
 
@@ -802,7 +807,7 @@ public abstract class DLNAResource extends HTTPResource implements Cloneable, Ru
 		setSkipTranscode(skipTranscode); // Hard to see the point in doing this, but assuming some plugins rely on this information.
 
 		if (skipTranscode) {
-			LOGGER.trace("File \"{}\" will be forced to skip transcoding by configuration", getName());
+			LOGGER.trace("File \"{}\" will be forced to be streamed by configuration", getName());
 			return null;
 		}
 
--- a/src/main/java/net/pms/dlna/DLNAResource.java
+++ b/src/main/java/net/pms/dlna/DLNAResource.java
@@ -1598,6 +1598,7 @@ public abstract class DLNAResource extends HTTPResource implements Cloneable, Ru
 		}
 
 		if (
+			!configurationSpecificToRenderer.getDisableTranscoding() &&
 			hasExternalSubtitles() &&
 			!isNamedNoEncoding &&
 			media_audio == null &&
--- a/src/main/java/net/pms/dlna/DLNAResource.java
+++ b/src/main/java/net/pms/dlna/DLNAResource.java
@@ -605,21 +605,6 @@ public abstract class DLNAResource extends HTTPResource implements Cloneable, Ru
 				}
 
 				if (child.format != null) {
-					String configurationSkipExtensions = configuration.getDisableTranscodeForExtensions();
-					String rendererSkipExtensions = null;
-
-					if (defaultRenderer != null) {
-						rendererSkipExtensions = defaultRenderer.getStreamedExtensions();
-					}
-
-					// Should transcoding be skipped for this format?
-					boolean skip = child.format.skip(configurationSkipExtensions, rendererSkipExtensions);
-					skipTranscode = skip;
-
-					if (skip) {
-						LOGGER.trace("File \"{}\" will be forced to skip transcoding by configuration", child.getName());
-					}
-
 					// Determine transcoding possibilities if either
 					//    - the format is known to be transcodable
 					//    - we have media info (via parserV2, playback info, or a plugin)
@@ -809,6 +794,19 @@ public abstract class DLNAResource extends HTTPResource implements Cloneable, Ru
 			return resolvedPlayer;
 		}
 
+		String configurationSkipExtensions = configurationSpecificToRenderer.getDisableTranscodeForExtensions();
+		String rendererSkipExtensions = renderer == null ? null : renderer.getStreamedExtensions();
+
+		// Should transcoding be skipped for this format?
+		boolean skipTranscode = format.skip(configurationSkipExtensions, rendererSkipExtensions);
+		setSkipTranscode(skipTranscode); // Hard to see the point in doing this, but assuming some plugins rely on this information.
+
+		if (skipTranscode) {
+			LOGGER.trace("File \"{}\" will be forced to skip transcoding by configuration", getName());
+			return null;
+		}
+
+
 		// Try to match a player based on media information and format.
 		resolvedPlayer = PlayerFactory.getPlayer(this);
 
@@ -852,7 +850,6 @@ public abstract class DLNAResource extends HTTPResource implements Cloneable, Ru
 								media_subtitle.setSubsStreamable(true);
 								LOGGER.trace("This video has external subtitles that should be streamed");
 							} else {
-								forceTranscode = true;
 								hasSubsToTranscode = true;
 								LOGGER.trace("This video has external subtitles that should be transcoded");
 							}
@@ -860,7 +857,6 @@ public abstract class DLNAResource extends HTTPResource implements Cloneable, Ru
 							if (renderer != null && renderer.isEmbeddedSubtitlesFormatSupported(params.sid)) {
 								LOGGER.trace("This video has embedded subtitles that should be streamed");
 							} else {
-								forceTranscode = true;
 								hasSubsToTranscode = true;
 								LOGGER.trace("This video has embedded subtitles that should be transcoded");
 							}
--- a/src/main/java/net/pms/dlna/DLNAResource.java
+++ b/src/main/java/net/pms/dlna/DLNAResource.java
@@ -794,8 +794,50 @@ public abstract class DLNAResource extends HTTPResource implements Cloneable, Ru
 			return resolvedPlayer;
 		}
 
+		boolean hasSubsToTranscode = false;
+
+		boolean hasEmbeddedSubs = false;
+		for (DLNAMediaSubtitle s : media.getSubtitleTracksList()) {
+			hasEmbeddedSubs = (hasEmbeddedSubs || s.isEmbedded());
+		}
+
+		/**
+		 * At this stage, we know the media is compatible with the renderer based on its
+		 * "Supported" lines, and can therefore be streamed to the renderer without a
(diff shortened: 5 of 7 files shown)


src/main/external-resources/UMS.conf
```diff
@@ -596,6 +596,15 @@ disable_transcode_for_extensions =
 # troubleshooting when you are not able to edit the renderer configuration.
 # Default: ""
 force_transcode_for_extensions =
+
+# Disable all transcoding
+# -----------------------
+# Do not transcode under any circumstances.
+# Warning: This will probably mean that your renderer will fail to play many
+# files or fail to show subtitles, but there are cases where this is still
```
Sami32: Telling that it's for testing and debugging purpose should be more appriopriate, pertinent, imo.
Better to strongly advertise the users, just in case...

That's said, i don't see the point to keep 2 similar test/debug functions: "disable_subtitles" and "disable_transcode_for_extensions"

src/main/java/net/pms/newgui/LooksFrame.java
```diff
@@ -448,7 +448,9 @@ public JComponent buildMain() {
 		tabbedPane.addTab(Messages.getString("LooksFrame.20"), gt.build());
 		tabbedPane.addTab(Messages.getString("LooksFrame.27"), pt.build());
 		tabbedPane.addTab(Messages.getString("LooksFrame.22"), nt.build());
-		tabbedPane.addTab(Messages.getString("LooksFrame.21"), tr.build());
+		if (!configuration.getDisableTranscoding()) {
+			tabbedPane.addTab(Messages.getString("LooksFrame.21"), tr.build());
```
Sami32: :+1:

Sami32:
@Nadahar As far i as understood, `skipTranscode` is used in the case where `mediaInfo=false` and `StreamExtensions =` are used.

src/main/external-resources/UMS.conf
Nadahar: @Sami32 I don't think it's a testing and debugging function, it is for those that never want transcoding.

Sami32: So, why not set this as an easy access option in the GUI ?

Nadahar:
@Sami32 That has probably been the original use yes, but as stated above I can't see how it could have been working because:
```java
                    boolean skip = child.format.skip(configurationSkipExtensions, rendererSkipExtensions);
                    skipTranscode = skip;
```
skipTranscode above is an instance variable belonging to the current instance, while skip is evaluated for the ```child```. This means that whatever instance is marked with the skip is a folder - the parent of the instance to be skipped. In addition this status will be overwritten for every other child that is evaluated. When ```isSkipTranscode()``` is called in ```resolvePlayer()``` it is evaluated for the current instance - which has never been set.

Sami32:
@Nadahar So, i was the one that misunderstood the code ;-)

That's said, i'm not for keeping this part of the code as "disable_transcode_for_extensions" code as it look totally useless now.
And keeping it for an eventual plugin use since 2011...as i guess you already checked, the main plugins used are not using it.

src/main/external-resources/UMS.conf
Sami32: So finally, we share the idea that it's nice to have but need to be used with care and be widely advertised. Different way same obective.

So that's fine for me, as i share that it need to be carefully used, it's why i though about commenting this function as "testing/debugging" to make users carefull.

Nadahar:
@Sami32 The "per extension" setting has a different use imo, it's for when there's a particular reason that you never want e.g flac to be transcoded regardless of the renderer configuration used. The "correct" place to do this would obviously be in the renderer configuration, but that can be too complicated for some (to figure out which file to edit and where). It isn't ideal, but I merely fixed a feature that was already there and I don't see it's use the same as for disabling transcoding alltogether.

When it comes to checking for plugin use I've been down that road several times before, so I haven't checked this time around. Previous experience has shown that it doesn't matter if it's actually used, it's the potential that it's being used that's considered interesting.

Sami32:
@Nadahar Yes, "disable_transcode_for_extensions" have nothing to do there, but that not come from PR though ;-)

And let's hope for `skipTranscode` use.

Sami32:
@Nadahar It'll be nice if you could move "disable_transcode_for_extensions" to the renderer configuration file IMO.

Nadahar:
@Sami32 Actually it's already there - but the name in the renderer configuration is ```StreamExtensions```. The change with this PR though is that it actually works, also when MediaInfo is true.

Both ```disable_transcode_for_extensions``` and ```disable_transcoding``` should also work in the renderer configuration, but I'll have to test it to be sure.

Nadahar:
@UniversalMediaServer/developers I'm missing some feedback from the rest of you, what do you think of this - do you even want this?

I'd also appreciate some testing by you all, I've done some testing and found nothing, but as stated before, the probability of side-effects is way too high for comfort as soon as anything is changed in ```DLNAResource```.

Sami32:
@Nadahar That's great :+1: 
So, it just need to move the `disable_transcode_for_extensions` comment to the renderer configuration file.
On other hand, `disable_transcoding` will be safer, and more logic imo, in the UMS.conf only.

Nadahar:
@Sami32 There's no need to move anything, they both work all places. In theory ALL configuration options in UMS.conf also work in the renderer configuration file, but it's not always true in the real world - that's why I had to test. 

To properly document that, you'd have to copy everything from the default UMS.conf into DefaultRenderer.conf - and that would be a bad idea.

Sami32:
So just leave `disable_transcode_for_extensions` as it is and hope that the users know that they can use it in their renderer configuration file ? Everything possible...just not my test though.

Nadahar:
@Sami32 You can argue for moving it, but it will lead to changes in the GUI and will be a breaking change that some old configurations or plugins might rely on. The bad thing about it is that there will be no way to reach it from the GUI - and a lot of users don't know how to edit text files - or are scared of that, and don't know where files are located, how to gain write privileges to the files etc.

In many cases there is only one renderer in use, and for those using the existing setting will work just fine. I agree that it's far from ideal, but I think we have to realize that editing the renderer configuration files isn't something every user will do. They will probably just find the hole thing to difficult and move on to another media server.

My main point isn't that this can't be changed, it's not exactly elegant the way it is, it's just that I think it's outside the scope of this PR. This PR fixes what is there and adds an extra option that's been requested for a long time. Changing the current logic should be discussed separately from this PR.

Sami32:
@Nadahar I got an NPE:

> DEBUG 2016-10-18 20:13:24.984 [main] Reading engines: 
> "ffmpegvideo,tsmuxer,mencoder,vlctranscoder,avsffmpeg,avsmencoder,ffmpegaudio,tsmuxeraudio,ffmpegwebvideo,mencoderwebvideo,vlcwebvideo,vlcvideo,vlcaudio" (default:"ffmpegvideo,mencoder,tsmuxer,ffmpegaudio,tsmuxeraudio,ffmpegwebvideo,vlcwebvideo,vlcvideo,mencoderwebvideo,vlcaudio,ffmpeg dvrmsremux,rawthumbs")
> ERROR 2016-10-18 20:13:24.984 [main] A serious error occurred during Universal Media Server initialization: null
> TRACE 2016-10-18 20:13:25.000 [main] 
> java.lang.NullPointerException: null
> 	at **net.pms.newgui.TranscodingTab.addEngines(TranscodingTab.java:329)**
> 	at net.pms.newgui.LooksFrame.addEngines(LooksFrame.java:572)
> 	at net.pms.PMS.init(PMS.java:722)
> 	at net.pms.PMS.createInstance(PMS.java:1101)
> 	at net.pms.PMS.main(PMS.java:1261)
> DEBUG 2016-10-18 20:13:25.031 [mplayer.exe-0-2] File not found: 'dummy'
> DEBUG 2016-10-18 20:13:25.031 [mplayer.exe-0-2] Failed to open dummy.
> TRACE 2016-10-18 20:13:25.093 [mplayer.exe-0] EOF
> TRACE 2016-10-18 20:13:25.093 [mplayer.exe-0] Ready to Stop: true

Nadahar:
Thanks @Sami32 and @valib, I didn't notice that when I tested. I don't quite understand how I could have missed that, that error should mean that the GUI never appeared right?

valib:
@Nadahar it is funny because when I tested your PR for the first time it looks that it was working correctly but when I retested it after the @Sami32  comment I found the same error.

Sami32:
After this commit i get a compilation error:
capture_10192016_020725

Nadahar:
@UniversalMediaServer/developers I've rebased and pushed a new version of this PR. Both the GUI initialization issue and the tests are taken care of.

The tests didn't really fail because of this PR, but because they were dependent on the local UMS.conf file and it's settings. When transcoding was turned off the backwardscompatibility tests for PS3 obviously failed. Tests should never rely on local configuration even though it can be hard to avoid the way configuration is used in UMS. I did modify it so the local configuration shouldn't impact this test at least (default configuration is used).

Please test again and/or give some feedback.

valib:
Reviewed 3 of 6 files at r1, 1 of 2 files at r2, 3 of 3 files at r3.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/1053)*

Nadahar:
@UniversalMediaServer/developers I'd appreciate some more testing of this, I'm a bit paranoid when making changes in ```DLNAResource```.

Sami32:
@Nadahar I'm not sure if it was your intention but `disable_transcoding` and `disable_transcode_for_extensions` doesn't work with files having internal subtitle.
[ums_Nadahar.zip.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/544749/ums_Nadahar.zip.txt)
BUT it give error with MP4/3G2/ files having AAC audio track : icon and message "File not supported"
That's said i never used it before this testing.

These 2 string parameters used in the same time give a warning: (It's look to me that it affect only the "Format_profile" having a space character)

> WARN  2016-10-21 21:10:40.687 [New I/O worker #1] Could not parse AvcLevel value Main@High 1440.
> TRACE 2016-10-21 21:10:40.781 [New I/O worker #1] Parsing results for file "TS-MPEG2-720-Main@High_1440-MP2-2ch.ts": container: mpegts, bitrate: 5297148, size: 3844976, video tracks: 1, audio tracks: 1, video codec: mpeg2, duration: 00:00:05.00, width: 1280, height: 720, frame rate: 23.976, mime type: video/mpeg Audio track id: 257, lang: und, audio codec: mpa, sample frequency:44100, number of channels: 2, bits per sample: 16
> WARN  2016-10-21 21:11:23.578 [New I/O worker #2] Could not parse AvcLevel value Main@High 1440.
> TRACE 2016-10-21 21:11:23.578 [New I/O worker #2] Parsing results for file "M2TS-MPEG2-1080-Main@High_1440-MP2.m2ts": container: mpegts, bitrate: 4297672, size: 10249196, video tracks: 1, audio tracks: 1, video codec: mpeg2, duration: 00:00:18.00, width: 1440, height: 1080, frame rate: 29.970, mime type: video/mpeg Audio track id: 257, lang: und, audio codec: mpa, sample frequency:44100, number of channels: 2, bits per sample: 16

@SubJunk The awaiting flags are great :+1:

Nadahar:
@Sami32 Thank you for testing :+1:  - and btw. I created the "awaiting testing" flag :wink: 

I'm not sure what you mean by doesn't work for files with internal subtitles. The log doesn't tell me that much as I don't know what to look for. What is the result? Are they still transcoded, or are there some other error?

Regarding "File not supported" I'm guessing that was on your renderer? That is as designed, when transcoding is disabled any files the renderer can't play natively will fail.

The last thing you're describing I don't understand at all: What to strings give this warning, and how is this warning related to disabled transcoding? Does this warning only appear if transcoding is disabled? Disabling transcoding per extension or completely should make no difference, they are triggered in the same code locations and the handling is the same.

SubJunk:
Nice addition with the `awaiting testing` label. I also just removed some unused labels, and added `not ready` to let people know which PRs are not finished.

@UniversalMediaServer/developers I'm open to suggestions for label changes too. I don't think we should get too complicated but a few of them can be helpful

Sami32:
@Nadahar I mean that internal subtitles are streamed.

Of course, when i told that give this icon and message "Files not supported" it was with well supported files. For me it's just the opposite way it should work: making unsupported files that are normally supported by the renderer.

Sorry if i'm not clear, in that case you'll need to try it yourself with internal subtitle and with the formats as told.

When you parameter are used at the same time with the other similar parameter, i discover that this warning appear in the log. I tried many combination and only that one throwed this warning.

Nadahar:
@Sami32 When you disable transcoding everything is streamed - that's the whole point. If your renderer can't handle the subtitles I guess it will simply ignore them?

Are you saying that files that are normally streamed to your renderer doesn't work when disabling transcoding, or are you simply saying that they usually work because they are transcoded? If it's the latter, that's how it's supposed to work.

The two parameters you talk about, is that ```disable_transcoding``` and ```disable_transcode_for_extensions```? Is there an extra warning when both are enabled at the same time? If so, what do you have to do to get that warning?

I could try it myself if I had any such files, but I'm not sure I do. Do you have some links for me to any of the files that fails?

Sami32:
@Nadahar Sorry for my bad english.
- Yes, for me it disabled external sutitiles streaming that are normally supported by my renderer.
{External subtitle} mesage is displayed on the renderer instead of {Sub:SubRip}
- Files having m:video/mp4 (3GP/3G2/MP4) and AAC audio become unsupported.
- This "Format_profile" warning appear only when i use these 2 parameters together, i never saw "Could not parse AvcLevel value Main@High 1440" before.

Nadahar:
@Sami32 I've finally had the time to look at this. I can confirm that the "special Samsung" hack for external subtitles doesn't work when transcoding is disabled. I'll see if I can find out how it ties together, but I'm pretty sure why: Some code checking for this is mixed in with a lot of other code that isn't run when transcoding is turned off (I'm skipping all the checking for which transcoding engine can be used etc.). Some vital piece of code for this is probably among that code.

When it comes to your other issues, I can't reproduce them. I can play MP4 with AAC just fine, and I can't get that AvcLevel warning for the TS files I've tested. You probably have to give me a link to some files with this problem for me to figure out what's happening.

Sami32:
@Nadahar  **AVClevel error** while `MediaInfo`, `disable_transcoding` and `disable_transcode_for_extensions` are set to `true`.
[TS-MPEG2-720-Main@High_1440-MP2-2ch.ts.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/587262/TS-MPEG2-720-Main.High_1440-MP2-2ch.ts.txt)
[debug.log.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/587258/debug.log.txt)

Nadahar:
@Sami32 Yes I agree that the PR just reveals existing bugs, but I could try to solve it anyway. As I said, I've reproduced the subtitle bug with Kodi, so I have that one "under control". I'll check out your file.

Nadahar:
@Sami32 @SubJunk I've found the reason for the AvcLevel warning, and it has nothing to do with this PR. I get the error regardless of the transcoding settings, and that's only natural as it happens during MediaInfo parsing. The bug was introduced in 86a261f725a5ee55066f0c2e95fa77e437f5d47e.

What happens in basicly that AVC level is attemted parsed without checking if the file is actually AVC. In the test file you send me here, the format is ```MPEG Video``` and thus the level has a different format ```Main@High 1440```. The AVC level parser is expecting strings in the form of ```High@L3.0```, ```High@L4.1``` etc and fails. It's actually just looking for anything following ```@L``` so it's a pretty unprecise parser that could "successfully" parse other things that would make no sense later on. In any case, since there is no such combination in this string the parsing "fails". The real fail though is that it's trying to interpret a non-AVC video stream as AVC.

I don't think this should be corrected in this PR, but it absolutely should be corrected.

Sami32:
@Nadahar I didn't mean that you need to, just letting you know as i only saw that while i test your PR. As i let you know that it affect external subtitles on my Panasonic, i don't have any Samsung.
That could be done in an other PR, like having a more general AVCLevel parameter (4.0, 4.2, 5.0...) https://github.com/UniversalMediaServer/UniversalMediaServer/issues/1043

About the MP4 and AAC issue i'm not able to reproduce it myself :-) I guess i forgotten to close and reopen your PR build after having turn on the TV, meaning that UMS used the default renderer configuration.

Nadahar:
@Sami32 The subtitle bug wasn't really because of this PR, it would happen whenever a compatible transcoding engine wasn't found (if none of those enabled was compatible with the file). Since transcoding is disabled in this case, I didn't do the search for a compatible transcoding engine and thus the same bug would occur. 

I think I've fixed the original bug with the last commit, so it should work both when transcoding is disabled and if no capable engine is available now. The underlying cause was that whoever put the subtitle evaluation code there simply put it in the wrong place. 

I've also "fined tuned" the logging somewhat, adding "Final verdict" to all situations so that it is easy to search the log for this.

@UniversalMediaServer/developers The downside to this is that any change in ```DLNAResource``` requires rigorous testing, I hope as many as possible will help test this. You don't have to particularly test disabling of transcoding that much, I'm more worried about any general side effects of touching anything in that class.

Nadahar:
I just did a rebase to sync with master, so any local copy of this branch must be either fetch/reset or deleted and checked out again.

Sami32:
@Nadahar Before your PR, i saw an avertissement saying that MediaUtil library have a cookie issue, but i was able to build fully anyway.

But now, the build fail.
mediautil

I don't know if it could be helpful, but a guy did make a "Mavenised" version:
https://github.com/fhoehnel/webfilesys/blob/master/maven-repository/mediachest/mediautil/1.0.0/mediautil-1.0.0.pom

```
<groupId>mediachest</groupId>
  <artifactId>mediautil</artifactId>
  <version>1.0.0</version>
```
Any idea ? suggestion ?

SubJunk:
@Nadahar yeah as you have found, that issue with AVCLevel has been fixed in master :)

Nadahar:
@Sami32 I just deleted my local mediautil and run ```mvn external:install``` again - I don't get any issue at all. I don't understand what's causing the error you get... if you run ```mvn -X external:install``` you'll get detailed debug logging for everything that happens, maybe that can shed some light on this (if you post the output as text, not as an image it would be helpful).

I suddenly realized a part of why your builds take such an extreme long time. If you had installed Maven instead of using the one "bundled" with UMSBuilder, any downloaded artifacts would be stored in your user profile (```C:\Documents and Settings\<username>\.m2\repository```). That means Maven builds would go much quicker, as only those artifacts that were new or had a new version would need to be downloaded.

Nadahar:
@Sami32 A couple of things that should be eliminated before we do anything else since this is a certificate error: Is your computer clock correct? If not, I'd advise [configuring NTP](http://superuser.com/questions/288404/how-to-synchronize-to-a-ntp-server-on-windows-xp). Second, I don't know if this impacts "normal Java" or just Java in the we browser, but check the Java applet in "Control Panel" under the security tab and make sure "High" is selected, not "Very High". 

This whole certificate scheme that the "big guys" want to enforce is nothing but a scam anyway, what they forget to tell you is that it isn't some rigorous testing scheme of the developer or the code that decides if he is "trusted", but whether the developer pays money to the "certificate authority". A revoked certificate simply means that they stopped their subscription.

Using certificates can have benefits even if the are expired or not issues by a "trusted authority" because the certificates are encryption keys and thus enable content to be encrypted, but with all the "scary" warnings they make about running expired of self-signed certificates a lot of users will deny it. It's sad, but it's also not that relevant to this.

I don't understand exactly how/where a certificate is involved here. The link itself is for an unencrypted artifact, and I can't really understand why Java would evaluate any certificate for the artifact itself, which is why I'm thinking that seeing the full debug output *might* offer a further clue to exactly where this certificate comes into play.

Sami32:
@Nadahar Don't ask me why, but after building an other PR before retry your, it worked fine !
That's said, i'm surprise you tell me that i'm the only one getting this invalid cookie header advertissement on _MediaUtil_ artifact...

If i want faster and keep the artifacts, i just need to do a Ctrl+C when the _UMS-builder.exe_ build come or modify the _.bat_ file for not clean; but i need to don't forget to delete all while one of you rebase a PR though !

I will test it tomorrow.

Nadahar:
@Sami32 It's good its working now. I don't know if nobody else has a cookie error, I only know I don't. 

About maven artifacts, we're talking about different things. When Maven is properly installed it has it's own local artifact cache that's independent from the project you work on. Whenever Maven needs an artifact it will check the cache first and only download if it isn't already there.

When I do a rebase it doesn't mean you have to delete everything and start over. That's just a git issue and isn't related to Maven or it's local repository. All you need to do is (with a command line window being in the UMS "base" folder):
```
git fetch <remotename>
git reset -- hard <remotename>/<branchname>
```

I don't know what name UMSbuilder gives the remote, you can list your remotes with ```git remote```. The "git default" remote name is ```origin```. This branch is called ```DisableTranscoding```. That would make it:
```
git fetch origin
git reset --hard origin/DisableTranscoding
```
Then your branch is "up to date" again, but the ```reset``` will mean that you loose any local changes you have made. Once that is done, simply run:
```
mvn package
```
and UMS will build.

Sami32:
@Nadahar Here the text capture you asked for:
[mvn-X2.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/587749/mvn-X2.txt)
The invalid cookie header avertissement appear only when you build from "scratch", not the way that any of you use so often, except perhaps when a rebase was done.
mediautil_cookie

Sami32:
@Nadahar That's great, so for me you can merge it when you want.

But, i take the occasion to point you that an other external subtitles code flaw is around since long time:
img_0602
[data0001.srt.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/587919/data0001.srt.txt)
[debug.log.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/587935/debug.log.txt)
[debug.log2.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/587937/debug.log2.txt)
https://we.tl/PLrLyLICon (video sample - 18MB)

> TRACE 2016-11-13 18:43:11.320 [Library Scanner] Parsing results for file "data0001.ts": container: mpegts, bitrate: 4693857, size: 18825216, video tracks: 1, audio tracks: 3, subtitle tracks: 6, video codec: h264, duration: 00:00:31.00, width: 1920, height: 1080, frame rate: 25.000, mime type: video/mp4, matrix coefficients: BT.709
> 	Audio track id: 530, lang: fr, audio codec: eac3, sample frequency:48000, number of channels: 2, bits per sample: 16
> 	Audio track id: 531, lang: qaa, audio codec: eac3, sample frequency:48000, number of channels: 2, bits per sample: 16
> 	Audio track id: 532, lang: qad, audio codec: eac3, sample frequency:48000, number of channels: 2, bits per sample: 16
> 	Subtitle track id: 550, type: UNKNOWN, lang: fr
> 	Subtitle track id: 551, type: UNKNOWN, lang: fr
> 	Subtitle track id: 102, type: MICRODVD, lang: fr, externalFile: C:\Documents and Settings\Philippe\Mes documents\Téléchargements\UMS Video Testing Suite (20150121)\M2TS\data0001.sub, external file character set: ISO-8859-1
> 	Subtitle track id: 103, type: SUBRIP, subtitles track title from metadata: _same_fixed, lang: fr, externalFile: C:\Documents and Settings\Philippe\Mes documents\Téléchargements\UMS Video Testing Suite (20150121)\M2TS\data0001_same_fixed.srt, external file character set: ISO-8859-1
> 	Subtitle track id: 104, type: TEXT, lang: fr, externalFile: C:\Documents and Settings\Philippe\Mes documents\Téléchargements\UMS Video Testing Suite (20150121)\M2TS\data0001.txt, external file character set: ISO-8859-1 
>         Subtitle track id: 105, type: SUBRIP, lang: fr, externalFile: C:\Documents and Settings\Philippe\Mes documents\Téléchargements\UMS Video Testing Suite (20150121)\M2TS\data0001.srt, external file character set: ISO-8859-1

..
> TRACE 2016-11-13 17:52:31.961 [New I/O worker #1] Adding new child "data0001.ts **{External Subtitles}**" with class "net.pms.dlna.FileTranscodeVirtualFolder"

Nadahar:
@SubJunk @valib I'd like for this version to be tested a bit more before merge. Could you try to just use it normally a little bit, no special testing needed, just to see if everything behaves as usual?

@Sami32 I'll have to read what you write some more times to see if I manage to understand it.

valib:
Reviewed 2 of 2 files at r4, 1 of 1 files at r5.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/universalmediaserver/universalmediaserver/1053)*

Nadahar:
@Sami32 About the cookie warning I'm sure you're correct that it's there for everybody under certain circumstances. I don't see it as very important as long as it works though. The file is hosted at SourceForge - and they concern is making money, which has led to them being under heavy criticism at times and a lot of projects have slowly moved away from them. They do almost anything to present a lot of ads, and my guess is that they try to serve a cookie to any client downloading anything from them for some ad-related use, and that Maven doesn't "approve" or understand this cookie. As long as the file is downloaded it really doesn't matter to me, it becomes interesting first when the file doesn't download. Since this is no longer the case and I atleast don't really understand what caused it, I think we'll juse have to let that go until it surfaces again. 

That said, I very rarely do a Maven build as I don't have to do that when I use Eclipse. If I do a Maven build it's mostly to run the tests or build the installer/distributable. I almost never use ```clean``` though, as I see it ```clean``` is only needed if the file structure has been "corrupted" in some way. If the build fails for some non-obvious reason, I use ```clean``` to see if that helps, but that has been a very rare event. Also, unlike on your computer, a Maven build generally means that a lof of text scrolls over the screen much too fast for me to be able to read it, so I mostly look at the end result and only in the event of a failure scrolls up to see what has happened.

Nadahar:
@valib We don't use Reviewable anymore because it overlaps too much with Githubs own reviewing functionality, if you want to make a "review" just select the "code" tab on the top of this PR and create a review on the button on the right side. That said, a normal comment will do fine for me in this case since Sami has already given the review needed for merge.

valib:
@Nadahar it means that @SubJunk should remove the Reviewable from this project.

Nadahar:
@valib He has tried, but there seems to be something still lingering somewhere. When you go into Reviewable you will see that it says that the repository is disconnected though, so I don't know why these links keep being inserted into the first post in every PR. I think that's the only thing still happening.

As to this PR, have you tested it in the current state?

Nadahar:
@SubJunk When I look at "Authorized Applications" Reviewable is still authorized for this repository. Are you sure you have [revoked it's authorization](https://help.github.com/articles/denying-access-to-a-previously-approved-application-for-your-organization/)?

Also, have you tested using this PR at all?

Nadahar:
@Sami32 It's not clear to me what the other problem with subtitles that you mention is. What would be the correct behaviour in contrast to what is actually happening?

Sami32:
@Nadahar Do the "{External Subtitles}" displaying is normal ?

Nadahar:
@Sami32 I haven't really studied when what is displayed in ´´´{}´´´, it might be a bug but I wouldn't know. When I test the attached file on this branch that's not what I see though:
data0001 ts
Have you tested it with this branch? Could moving the subtitle code here also have fixed this by chance? Or is there simply some other circumstances that differ?

I think @valib knows much more about this than me.

Nadahar:
@UniversalMediaServer/developers Although I think it's scary to touch anything in ```DLNAResource``` because I consider it to be a "minefield", I haven't seen any ill effects of these changes. My ability to test is limited to explicit testing, since I hardly ever use UMS to watch video content (I generally don't watch fiction very much). What I wanted was for you that actually use UMS for streaming video content to just use this branch a little bit to see if you discovered anything.

I do realize that not a lof of time has went by and you might be doing this as I write, but the lack of response to my request leaves me unsure if this is going to happen. Am I simply being over-cautios, should we consider this safe enough to merge?

Sami32:
@Nadahar That's weird ! Same file, this same latest branch build and different result.
I'm perplex...maybe the difference come because you're using the Java 8 version ?

Nadahar:
@Sami32 I'm using Java 7, but there are probably a lot of factors that can influence this outcome. Is the renderer configured to stream or transcode this file? In my case it's streaming..

Sami32:
@Nadahar I get the same result with any configuration, streaming or transcoding, with the default renderer as well.

Nadahar:
Try Kodi, that's what I used. You only need to add ```eac3``` to the mpegts line in the renderer configuration, and you have an identical configuration to what I had.

Sami32:
@Nadahar I don' know if Kodi will be able to run on my computer, but right now i don't have time, i'll try it this evening.

Sami32:
@Nadahar How is your ?
[UMS.conf.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/592881/UMS.conf.txt)

Nadahar:
@Sami32 It's a mess, it's used for testing all kind of things and is a complete chaos. What parameters are you looking for?

valib:
@Nadahar I made the PR  #1097 which is a part of the subs fix and should be merged to this PR  and combine with the commit I made.  This PR works for me but I would recommend to rename the `getDisableTranscoding()` to `isDisableTranscoding()` to keep the rules of this project.

valib:
@Nadahar good work.

Nadahar:
@valib Why did you push those commits and then merge? What was the point of creating #1097 if you're going to "highjack" this PR to get it merged?

First of all, there were no reason to make 7f1e99c143a28ff09bcb6d91418d0662f6ed918e at all. This branch were without any conflicts, the only difference with master was the Kodi profile and since that file wasn't touched in this branch it would have merged nicely.

Second, 6fdd1538642f1f45d842102ae7282242cbbab06a seems like a bug to me. I don't exactly follow the logic, but why can't there be external subtitles if transcoding is off? I don't get it... and also, assuming the logic made sense, you have missed one options - if transcoding is disabled for that format/extension only.

As I see it though, the "rules" in general should be identical if a file is "streamed normally" or streamed because transcoding is disabled. Please explain to me how 6fdd1538642f1f45d842102ae7282242cbbab06a makes any sense...

PR #1055, opened 2016-10-18
Update WebRender.java

Sami32:
Their png icon was not added yet, as i'm not sure which Vivaldi one should be the best choice...
https://vivaldi.com/
https://www.chromium.org/
Why these 2 only ? **ChromeCast able**, and Chromium becoming the open source base for many browsers perhaps it deserved to have it's own icon ;-)
vivaldi_browser_logo_2015
vivaldi200-175
vivaldi
chromium


Sami32:
@nadahar Only for Vivaldi from the log of the ChromeCast user issue that you already commented.
But it can be seen there as well:
http://www.webapps-online.com/online-tools/user-agent-strings/dv/browser725166/vivaldi

PR #1065, opened 2016-10-26
Loosely implemented subtree scanning concept

Nadahar:
@SubJunk As usual you're too much in a hurry for me :wink: 

I hadn't forgotten to answer you, I simply haven't had the time to look into the details. As a result, here is how I'm thinking it should be done. This is untested and incomplete, I've left several todos that need to be implemented before this can have a chance to work.

I'm realizing that just creating a new `RealFile` like I'm doing here probably won't work, instead the resource must probably be looked up in the hierarchy starting from `Rootfolder` but this is all a bit unclear to me. It seems like UMS is actually trying to hold the entire tree in memory once it's been used/scanned, which I'd think could pose some issues for large libraries with thumbs and all. Peeking at how browsing via the media library locates its resources when something selected from browsing the database only is probably a good way to figure this out.

From the quick look I've had now, I can already see that your current implementation won't work. The existing logic doesn't take into account that anything can change, so as soon as something "is discovered" it won't be rediscovered. That obviously have to change.

The concept itself should be easy enough to understand, but the details can be a bit tricky to figure out given the mess this logic is. The whole idea of only scanning based on folders should be changed, it should be possible to rescan a simple file - given there will be events for each changed file, finding its parent folder and scanning with the current logic will be a bad solution.

I see other problems with the current system as well, there should be a proper queing of objects up for scanning and a delay for consolidation, without this with recursive scanning I see the possibility for massive rescanning of the same objects.

---

This change is [Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/1065)

### DIFF
--- a/src/main/java/net/pms/dlna/DLNAResource.java
+++ b/src/main/java/net/pms/dlna/DLNAResource.java
@@ -4363,4 +4363,40 @@ public abstract class DLNAResource extends HTTPResource implements Cloneable, Ru
 		scaleHeight = Player.convertToModX(scaleHeight, 4);
 		return scaleWidth + "x" + scaleHeight;
 	}
+
+	public void scan() {
+		//TODO This method must be rewritten to scan the current resource if it's a file and its children recursively if it's a folder
+		for (DLNAResource child : getChildren()) {
+			if (child.allowScan()) {
+				child.setDefaultRenderer(getDefaultRenderer());
+
+				// Display and log which folder is being scanned
+				String childName = child.getName();
+				if (child instanceof RealFile) {
+					LOGGER.debug("Scanning folder: \"{}\"", childName);
+					PMS.get().getFrame().setStatusLine(Messages.getString("DLNAMediaDatabase.4") + " " + childName); //TODO - needs resetting
+				}
+
+				if (child.isDiscovered()) {
+					child.refreshChildren();
+				} else {
+					if (child instanceof DVDISOFile || child instanceof DVDISOTitle) { // ugly hack
+						child.syncResolve();
+					}
+					child.discoverChildren();
+					child.analyzeChildren(-1);
+					child.setDiscovered(true);
+				}
+
+				int count = child.getChildren().size();
+
+				if (count == 0) {
+					continue;
+				}
+
+				child.scan();
+				child.getChildren().clear();
+			}
+		}
+	}
 }
--- a/src/main/java/net/pms/dlna/RootFolder.java
+++ b/src/main/java/net/pms/dlna/RootFolder.java
@@ -29,6 +29,7 @@ import java.nio.charset.StandardCharsets;
 import java.text.Collator;
 import java.text.Normalizer;
 import java.util.*;
+import static java.nio.file.StandardWatchEventKinds.*;
 import net.pms.Messages;
 import net.pms.PMS;
 import net.pms.configuration.DownloadPlugins;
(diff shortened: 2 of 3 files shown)


SubJunk:
@Nadahar great, thanks for your work on this. I agree about your notes too, I had encountered the same things when I attempted an implementation. The code in this area is quite patchy and it would be ideal to make it more cohesive and predictable.

SubJunk:
Is this going to continue to be worked on? Because my first implementation is still working well for me months later, albeit slowly and inefficiently, but it works and this is holding up 7.0.0

Nadahar:
@SubJunk It wasn't my intention to do more than answer [your question](https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1038#issuecomment-255507951). The problem here isn't really "your implementation" but the way the existing logic works. It seems that the existing scanning/discovery logic never took into consideration that anything could change while UMS/PMS was running, and changing it from a "static" to a "dynamic" resource tree isn't trivial. As I see it, it's either to rediscover/reparse way too much information or to do some heavy refactoring to the basic logic/structure. I don't think I need to say what I think is the right solution, but if you (or anybody else) want to do it is up to you.

I you choose the path of simply rescanning a lot of already discovered resources, I think you still need to address this:
> The existing logic doesn't take into account that anything can change, so as soon as something "is discovered" it won't be rediscovered.

Wouldn't you need to invalidate the "resolved" (or what ever the name is) and "thumbsready" status of the resources you're going to rescan, or is that already implemented in #1038?

SubJunk:
@Nadahar do you mean in the case of a file being replaced by another file with the same filename?

Nadahar:
@SubJunk Yes. Any situation where the content of a file is changed/replaced, AFAICR it won't help if you rescan; UMS simply won't parse it again until you've restarted.

SubJunk:
@Nadahar it's a good point but I think that will be a rare case. Different versions of the same downloaded video will have different filenames (like `The.Movie.1080p-UMS` vs `The.Movie.1080p-Nad`).
Music files may be more likely to be similar since it's often just the folder name that is unique instead of the filenames, but I don't think it is common for someone to replace any files.
It may be an incorrect assumption based on my use of the program though so if you have different ideas about that let me know.

Nadahar:
I'm closing it some nobody gets tempted to press the merge button, this was never intended to be merged as it is.

@SubJunk I guess it won't be that common for any types of files, but my idea of a "server" is somewhat different. In my mind a "server" is something providing a service, ideally running on it's own 24/7 and for months at the time without being touched. Seen from that perspective, you'd want it to detect any changes to the shared files, it could be your family photos that got a photoshop touch-up, it could be that you replaced some MP3 files with better encoded versions etc. In any case, you'd want the "server" sharing these media to your devices to properly handle the updated versions.

SubJunk:
@Nadahar I agree with you in terms of ideal use. I would love UMS to be able to sit there for months too. Just trying to prioritize so we can get things moving and it would be good to do change-detection later :)

I think in the example you gave, of MP3 files, a replaced file would probably still work even though we didn't notice it. The filename, container, codec and duration would be the same so nothing important would be discovered by updating

atamariya:
@subjunk MP3 metadata eg. embedded cover-art might change - just saying.

Nadahar:
@SubJunk I think it all comes down to that we're actually talking about different things. When you wrote "Rescan shared folders if there is a change" in the title of #1038 I interpreted it to be about actual change. I think what you're after is only detecting if video files are added, and that's a very narrow part of what I think of as "change".

SubJunk:
@atamariya yes the cover art change wouldn't show up. It would be ideal if it did.

@Nadahar to be even more clear, a rescan is already triggered in my implementation when a file is changed, I have tested and confirmed it, but as we have been discussing, existing files are skipped by the actual scan.
I agree with you that ideally we would detect changes too and I hope we do.

PR #1067, opened 2016-10-28
Revert "ChromeCast update and fix"

Sami32:
Reverts UniversalMediaServer/UniversalMediaServer#1066

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -380,7 +380,7 @@
 		<dependency>
 			<groupId>su.litvak.chromecast</groupId>
 			<artifactId>api-v2</artifactId>
-			<version>0.9.3</version>
+			<version>0.9.2</version>
 		</dependency>
 		<dependency>
 			<groupId>org.javassist</groupId>


SubJunk:
Why?

Sami32:
@SubJunk For now it doesn't improve issue while using the latest Chromecast firmware but only give a NPE add...so i'm afraid that it could affect all the others Chromecast  users as well.

They are working on though...so maybe better to wait next version.
Users having this kind of issue can just revert they firmware update until it get solved.

Nadahar:
You guys are too trigger happy - test first, THEN merge :wink:

vitalidze:
Wait, I don't consider that NPE an issue in library. Please read this: https://github.com/vitalidze/chromecast-java-api-v2/issues/51#issuecomment-256838206 for an explanation about what happened. The thing is that there are no running application at the moment of sending that request. So most probably the fix in next version will be to throw a more descriptive error than NPE. This will not fix the root cause in UMS. 

I don't know how the code is being used right now in UMS, haven't got a time to read it. But I think it should check for a running application before sending request to play some media on ChromeCast.

Nadahar:
@vitalidze That sounds reasonable, thank you for informing. I just understood @Sami32's description of the issue like it was a library issue.

I don't think we will be able to get this up and running as there seems to be no developer that owns a Chromecast.

SubJunk:
I own a Chromecast and am happy to do testing. I've used it before with UMS and it worked well but I haven't done heaps

Sami32:
That's a great new :-)

Sami32:
The NPE has been already fixed in 0.9.4 SNAPSHOT:
https://github.com/vitalidze/chromecast-java-api-v2/commit/99250880bb0ba0246abb18248fc11df52cc51ba9

Nadahar:
@Sami32 UMS code will still have to be changed, an exception will still be thrown - we will just get a ```ChromeCastException``` instead of a ```NullPointerException```. The result will be the same, UMS will have to keep track of whether an "app" (whatever that might be) is running or not before making the call.

Sami32:
@Nadahar It's why in my first post, i asked to wait this [PR](https://github.com/vitalidze/chromecast-java-api-v2/pull/40) to be realized.

PR #942, opened 2016-07-03
7.0.0

SubJunk:
This branch is for working towards the 7.0.0 release. My vision for 7.0.0 is to focus on the media library, which right now is called the Cache folder.
It seems that it's becoming more popular for servers to organize content and users enjoy that convenience, so I think we should focus on offering that experience alongside our current experience which is to mostly show the folder structure with smart features.

TODO:
- [x] Rename Cache back to Media Library as it was called in the past
- [x] Make Media Library display by default
- [x] Add new options in there like "TV Shows" and "Movies". We already detect those types via our filename prettifying and opensubtitles implementation so a lot of this is already written and just needs to be plugged in to the database
- [x] Fix bug where removing a directory in UMS does not remove its contents from the Media Library.
- [x] Change our "Hide" options to "Show", e.g. "Hide Media Library folder" becomes "Show Media Library folder"
- [x] Fix sorting in dynamic folders
- [x] Prepend "Season " (translatable) to the season virtual folders so they say e.g. "Season 2" instead of just "2". This will require a new feature for those folders since they currently don't have the ability to display different text to what is in the database.
- [x] Fix TV show names being different in different database entries for the same show. This happens when OpenSubtitles returns a result for some episodes but not all of them, so we can end up with for example "American Dad" from the filename and "American Dad!" from OpenSubtitles.
- [x] Add fully played tracking to the SQL database
- [x] Import the old style fully played tracking into SQL
- [x] Right-click to mark folder as fully played
- [ ] Properly parse external subtitles in realtime when they are added to monitored folders, so that they can be instantly used in the media library
- [ ] Change file names when viewed in the Media Library to reduce duplicate data (e.g. in the TV Shows folder, the file name doesn't need the show name in it, because they are in that show's folder already)
- [ ] Display a friendly log message when a file is added on the fly

PR #1068, opened 2016-10-28
Enable database debug logging

Nadahar:
This is **not** a PR to be merged, it's for investigating database behaviour. 

This enabled debug logging for the database both to a separate "trace.db" file and in the regular UMS log file. Logging in the regular UMS log file is very useful at it shows when the database events take place in relation to what happens in UMS.
- [ ] Do not merge!

### DIFF
--- a/src/main/java/net/pms/dlna/DLNAMediaDatabase.java
+++ b/src/main/java/net/pms/dlna/DLNAMediaDatabase.java
@@ -88,7 +88,7 @@ public class DLNAMediaDatabase implements Runnable {
 		dbName = name;
 		File profileDirectory = new File(configuration.getProfileDirectory());
 		dbDir = new File(profileDirectory.isDirectory() ? configuration.getProfileDirectory() : null, "database").getAbsolutePath();
-		url = Constants.START_URL + dbDir + File.separator + dbName;
+		url = Constants.START_URL + dbDir + File.separator + dbName + ";TRACE_LEVEL_FILE=3;TRACE_LEVEL_SYSTEM_OUT=3";
 		LOGGER.debug("Using database URL: " + url);
 		LOGGER.info("Using database located at: " + dbDir);
 


Sami32:
@Nadahar A simple Yes was fine !
But i appreciate your move ;-) So let's throw some datas that i just get: (Without folders monitoring)
[medias.trace.db.7z.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/559276/medias.trace.db.7z.txt)
Around all the 2 mn after the scan the database file size changed, also sometimes i didn't saw the trace file size changed when that happen.

**P.S.** I didn't saw any special database datas in UMS log...
[Database_dbg.zip.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/559280/Database_dbg.zip.txt)

Nadahar:
@Sami32 Looking at your trace.db file I can see that there has been no database disconnects during use, so the theory that the varying size is due to automatic cleanup at during disconnects seems to be a dead end. The file has two disconnects, it seems like the file cover two UMS sessions (this logfile isn't wiped when UMS is restarted).

When it comes to logging in the normal log file, I guess I was a bit trigger happy. I'm using Eclipse (you should too), and in Eclipse it was merged in. I didn't check the actual debug.log. I think some extra code is needed to get them to appear in the log file.

Sami32:
I also saw that the trace file is not reinitialiset, but i yet didn't understand why the file size increase/decrease/increase...
I'm not familiar with database so i'm wondering what are for all this rollback around ?

Nadahar:
@Sami32 There shouldn't normally be any rollbacks, so that is probably caused by bugs.

Short explanation: Proper databases are so called "transactional databases" (but a lot isn't, like the standard database engine in MySQL). A transactional database has a concept called a transaction that are supposed to be commited completely or not at all.

Let's say you have a table with users and another with user status (this example would be a bad table design, it's just to make it easy to understand):

username | full_name
---------- | ----------
sami32 | Samuel Jackson
user2 | Username 2

username | role
---------- | ----
sami32 | administrator
sami32 | moderator
user2 | spectator

As you can see, "username" is what ties these together. If you wanted to change your username from "sami32" to "supersami" you'd have to make changes in both tables.

This is where transactions come in: You want to make sure that the changes are either made to both tables or not at all, because otherwise you would "loose" the connection between the two (you would loose your roles). Before you change anything, you tell the database that you want to start a transaction. You then make changes to both tables with separate commands, and finally you tell the database to "commit" the transaction. If the connection to the database is lost or some error happens in the middle of this transaction, the transaction will be rolled back. You can also tell the database to roll it back at any time until the commit command is sent. That way you can make checks and make sure everything is ok before you commit.

There is no point in using transactions if you just perform one change to one table that's not related to anything else, but transactions can also be used to protect multiple modifications to the same table that need to be done together or not at all.

As you can see, rollbacks shouldn't normally appear, that's a "safety net" if something goes wrong or doesn't add up.

Nadahar:
It seems that whoever made a lot of that code didn't understand transactions, transactions are started but never commited and then the connection is closed. That means that the changes are rolled back and whatever is supposed to be stored isn't. That can also explain the yo-yo file size as transactions aren't applied until commit but are probably temporarily stored last in the file and then cleaned up when there is time.

Sami32:
Thank you very much for the explaination :+1: 
I found it suspect as well, and possibly related to the size fluctuation.
The fact that the trace file size didn't change when this weird variation was seen on the database file size give me a stronger suspicion that it could be the culprit. 

I remember seen an open issue telling that the files already scanned was again scanned each time while browsing; that could possibly be related to these rollbacks as well. 
I will try to do more tests tomorrow.

And you, did you get these rollbacks in your database trace file as well ?

Nadahar:
@Sami32 I didn't scan anything and I didn't keep it - I just started up to check that the logging worked. But, I'm pretty sure anyone will get these, this has nothing to do with Windows version etc.

Sami32:
Just to be SURE that it's not a windows XP issue ;-)

Sami32:
> 2016-10-29 12:44:46 database: closing C:/Documents and Settings/All Users/UMS/database/medias from shutdown hook
> 2016-10-29 12:44:46 database: disconnecting session #3
> 2016-10-29 12:44:46 database: disconnected session #3
> 2016-10-29 12:44:46 database: closing C:/Documents and Settings/All Users/UMS/database/medias
> 2016-10-29 12:44:46 database: closed
> 2016-10-30 16:40:41 jdbc[3]: exception
> org.h2.jdbc.JdbcSQLException: Database is already closed (to disable automatic closing at VM shutdown, add ";DB_CLOSE_ON_EXIT=FALSE" to the db URL) [90121-192]
> 	at org.h2.message.DbException.getJdbcSQLException(DbException.java:345)
> 	at org.h2.message.DbException.get(DbException.java:179)
> 	at org.h2.message.DbException.get(DbException.java:155)
> 	at org.h2.message.DbException.get(DbException.java:144)
> 	at org.h2.jdbc.JdbcConnection.checkClosed(JdbcConnection.java:1482)
> 	at org.h2.jdbc.JdbcConnection.checkClosedForWrite(JdbcConnection.java:1467)
> 	at org.h2.jdbc.JdbcConnection.rollback(JdbcConnection.java:500)
> 	at org.h2.jdbcx.JdbcXAConnection.getConnection(JdbcXAConnection.java:107)
> 	at org.h2.jdbcx.JdbcConnectionPool.getConnectionNow(JdbcConnectionPool.java:229)
> 	at org.h2.jdbcx.JdbcConnectionPool.getConnection(JdbcConnectionPool.java:199)
> 	at net.pms.dlna.DLNAMediaDatabase.getConnection(DLNAMediaDatabase.java:131)
> 	at net.pms.dlna.DLNAMediaDatabase.getData(DLNAMediaDatabase.java:356)
> 	at net.pms.dlna.RealFile.resolve(RealFile.java:175)
> 	at net.pms.dlna.DLNAResource.syncResolve(DLNAResource.java:1507)
> 	at net.pms.dlna.DLNAResource.run(DLNAResource.java:1306)
> 	at net.pms.dlna.RealFile.isValid(RealFile.java:66)
> 	at net.pms.dlna.DLNAResource.addChild(DLNAResource.java:573)
> 	at net.pms.dlna.DLNAResource.addChild(DLNAResource.java:533)
> 	at net.pms.dlna.MapFile.manageFile(MapFile.java:114)
> 	at net.pms.dlna.MapFile.analyzeChildren(MapFile.java:171)
> 	at net.pms.dlna.RootFolder.scan(RootFolder.java:260)
> 	at net.pms.dlna.RootFolder.scan(RootFolder.java:218)
> 	at net.pms.dlna.DLNAMediaDatabase.run(DLNAMediaDatabase.java:814)
> 	at java.lang.Thread.run(Unknown Source)
>

Nadahar:
@Sami32 Please explain what you point to instead of just referring to something (whether it's a log or an URL). If you're referring to the error closing the database, my guess is that H2 has its own shutdown hook running in parallel. The order shutdown hooks are run in is arbitrary, which is why shutdown hooks can be very tricky. Some times UMS' shutdown hook will run first and the error will (actually they probably check first) happen in H2's shutdown hook, or if the H2 shutdown hook has already closed it UMS' shutdown hook will throw an exception.

If this is the case (it should be verified), we should just remove any access to the database from the shutdown hook, or if we need database access in the shutdown hook ```DB_CLOSE_ON_EXIT=FALSE``` should be used.

Sami32:
Sorry, i posted it with for keeping a trace of the error in mind, as i was afraid to forget it.
[medias.trace.db.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/560653/medias.trace.db.txt)
It happened while closing UMS and a scan was running.

Nadahar:
@Sami32 Ok, I doubt closing UMS while scanning is handled very gracefully as threads are generally not closed down gracefully in UMS and scanning is running in a separate thread.

Nadahar:
I think I've found the source for all the rollbacks in ```JdbcXAConnection```:
```java
        public synchronized void close() throws SQLException {
            if (!isClosed) {
                try {
                    rollback();
                    setAutoCommit(true);
                } catch (SQLException e) {
                    // ignore
                }
                closedHandle();
                isClosed = true;
            }
        }
```

This is called when a connection is closed. I'm not sure what to think of that, it seems that they just call a rollback "to be sure" even if no uncommitted transaction exists. That means that even non-transactional simple queries will generate this if the connection is just used for that, so this is simply "logspam"...?

I'm not sure what to think at the moment, but this looks like it could be nothing.

Sami32:
After you [jdbcXAConnection](https://github.com/h2database/h2database/blob/7071b60505786cbf2e73d06fe0e088f972a27c37/h2/src/main/org/h2/jdbcx/JdbcXAConnection.java) comment i do some search and found that:
http://stackoverflow.com/questions/218350/does-java-connection-close-rollback

Nadahar:
@Sami32 I just did a scan myself and the database file doesn't fluctuate when I do it. I didn't test #942 though - do you get this issue only in that branch?

Also, you haven't by any chance turned on compression on your harddrive to save space? That will slow down the computer a lot and make sizes generally unreliable.

Sami32:
I used the master branch and other as well.
My 20GB main partition is not compressed (FAT32), only my secondary 5GB NTFS partition is compressed. 
Maybe an XP issue ?
Did you waited for few minutes while looking at the database file size ?

Nadahar:
@Sami32 I sat here waiting and refreshing (F5) for several minutes until the database file had grown to ~165 MB. I never saw any reduction in size. Even after quitting UMS there was no reduction.

I haven't used FAT32 in many years, it's easy to loose data with FAT32 (no journaling) and there's no security/permission support, so I tested on uncompressed NTFS. I don't think much has happened to neither FAT32 nor NTFS since WinXP, so I doubt it's an XP issue on a filesystem level at least.

The big question is obviously if the database resides on your uncompressed FAT32 or the compressed NTFS partition. I assume it is on the FAT32 partition. There's so many years since I used FAT32 for anything but thumbdrives/flash cards that I honestly don't remember if there's any size calculation issues there. I don't think I've used that on a harddrive since around year 2000.

Sami32:
I told you, some dinosaurs live again around you ;-)

Yes, the NTFS partition is for saveguard copy.

Nadahar:
To enable database logging in the normal logfile (```debug.log```), add the following to the database connection URL: ```";TRACE_LEVEL_FILE=4"```.

Sami32:
Thank you :-)
I saw that but understood it incorrectly, as i tryied with "ALL" in the logging mode.

PR #1083, opened 2016-11-04
Update Google-Android-Chromecast.conf

Sami32:
Need to be tested on latest _Chromecast_ firmware or latest Chromecast device.

The _Chromecast Audio_ can now stream up to 96KHz/24bit lossless audio, but since some features could be limited to them, maybe create a different renderer file will be more appropriate.
A test will answer that, meaning that we depend on _Chromecast_'s users ;-)
http://162.243.219.67

### DIFF
--- a/src/main/external-resources/renderers/Google-Android-Chromecast.conf
+++ b/src/main/external-resources/renderers/Google-Android-Chromecast.conf
@@ -24,7 +24,7 @@ RendererIcon = chromecast.png
 # User-Agent: Android/4.2.1 UPnP/1.0 Cling/2.0
 # User-Agent: Apache-HttpClient/UNAVAILABLE (java 1.4)
 # ============================================================================
-#
+# https://developers.google.com/cast/docs/media
 
 UserAgentSearch = Android|CrKey
 
@@ -40,13 +40,23 @@ CustomFFmpegOptions = -async 1 -fflags +genpts -c:a libmp3lame -ac 2 -b:v 35000k
 MediaInfo = true
 
 # Supported video formats:
-Supported = f:mp4     v:mp4|h264   a:aac-lc|he-aac|mp3   n:2   m:video/mp4
+Supported = f:mp4    v:mp4|h264    a:aac-lc|he-aac|mp3                 n:2       m:video/mp4
+Supported = f:mkv    v:mp4|h264    a:aac-lc|he-aac|mp3|opus|vorbis     n:2       m:video/webm
 
 # Supported audio formats:
-Supported = f:aac         n:2      m:audio/aac       a:aac-lc|he-aac
-Supported = f:mp3         n:2      m:audio/mp3
-Supported = f:m4a|mp4     n:2      m:audio/mp4       a:aac-lc|he-aac
+Supported = f:flac        n:2      m:audio/x-flac                                s:96000
+Supported = f:mp3|mpa     n:2      m:audio/mp3
+Supported = f:m4a         n:2      m:audio/mp4       a:aac-lc|he-aac|mp3
+Supported = f:ogg         n:2      m:audio/ogg       a:flac|opus|vorbis
 Supported = f:wav         n:2      m:audio/x-wav     a:lpcm
+Supported = f:weba        n:2      m:audio/webm      a:opus|vorbis
+
+# Supported images formats:
+Supported = f:bmp         m:image/bmp        w:1280     h:720
+Supported = f:gif         m:image/gif        w:1280     h:720
+Supported = f:jpg         m:image/jpeg       w:1280     h:720
+Supported = f:png         m:image/png        w:1280     h:720
+#Supported = f:webp        m:image/webp       w:1280     h:720
 
 # Supported subtitles formats:
 SupportedExternalSubtitlesFormats = WEBVTT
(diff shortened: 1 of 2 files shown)


Nadahar:
How are chromecast users expected to test this? They'd have to use UMSBuilder or build it themselves - isn't that too much to ask?

Sami32:
@Nadahar I was thinking about @SubJunk as he told that he have a Chromecast.

Sami32:
As per an other user request:
http://www.universalmediaserver.com/forum/viewtopic.php?f=3&t=9223#p29791

Sami32:
One more request:
http://www.universalmediaserver.com/forum/viewtopic.php?f=3&t=9223#p29791

Near to all of them, ask for FLAC support.

Sami32:
OGG audio files confirmed as working as well.
https://github.com/UniversalMediaServer/UniversalMediaServer/issues/1030#issuecomment-286929722

Nadahar:
@Sami32 Shouldn't 
```powershell
Supported = f:aac         n:2      m:audio/aac       a:aac-lc|he-aac
```
be ```f:adts``` instead? Is there such a thing as an ```aac``` format?

Sami32:
@Nadahar As you already know, before your audio formats separation from video ones PR the deleted that part, it was OK, but now all the renderers have that issue. I guess it where using an IDE is very usefull for more of one hundred configuration files :wink: 
`aac` was used for M4A(AAC) not for ADTS.

I'll delete this line as it shouldn't have been write with that MIME type anyway:
https://developers.google.com/cast/docs/media

Thanks :-)

Nadahar:
@Sami32 What do you mean "all the renderers have that issue"? I only changed the name of the codecs, not the formats.

I actually used regex to update the configuration files, I use regex a lot for "advanced search and replace".

Nadahar:
@Sami32 Are you saying that I should replace all ```f:aac``` with ```f:m4a```? That wouldn't take much time.

Sami32:
@Nadahar Since we already have words on that in my AACfix PR, and we disagree, or you didn't understand, i prefered shade up and move on.
The origin:
https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1285/files#diff-ea2b6ebc35654523b71294f0907eef82L327
The result:
https://github.com/UniversalMediaServer/UniversalMediaServer/search?utf8=%E2%9C%93&q=%22Supported+%3D+f%3Aaac%22&type=

Nadahar:
@Sami32 I probably didn't understand and I still don't quite understand. I didn't introduce ```aac``` as a "format", if you use blame in the renderer configurations you will see that it goes back to PMS. 

I wasn't aware that ```aac``` meant M4A, so if #1285 broke something there I wasn't aware of it. As far as I'm concerned M4A didn't work as it were, as it were recognized as video.

If there's a problem in all the renderer configurations now, we should fix it. Just make sure there are no more misunderstandings. What is it that we disagree on, what do you think should have been done differently?

Sami32:
@Nadahar You still don't understand, `aac` was fine, now we loose ability to have a different MIME type for AAC and M4A.
Anyway that can be fixed later, in the AACfix PR.
I don't or don't say that you should do something, as i don't like argue to much.

Since the `Supported f:aac` is not used anymore, it'll not hurt to delete that line from renderers, IMO.
M4A files (audio/x-m4a, not AAC) based, are more rarely used, IMO, so that will doesn't give much issues report i think, and can be fixed later.

Nadahar:
@Sami32 You're right that I still don't understand. I didn't know that there exists as separate "AAC" file format. I thought AAC was either in a MP4 container or as ADTS.

Sami32:
@nadahar Fundamently you're right, there are M4A or MP4, but only the AAC family codecs have an audio/mp4 MIME type (though some renderers wait for audio/x-aac or audio/aac or some more rare ones), and are specified in the DLNA standard.
The other MP4/M4A audio files should have audio/x-m4a or audio/m4a, i guess that's it's why it has been done. And perhaps also because we call them AAC and that it was more easy to users to recognize such appelation in their renderer displaying or renderer configuration file.

As long as these lines are replaced, for now, by `f:m4a`, the negative effect should be limited, IMO.

Sami32:
@Nadahar For me the fix should be to rename all the `f:aac` to `f:m4a`, and to modify the existing m4a line by:
`Supported = f:m4a        n:2      m:audio/x-m4a       a:(?!aac-|he-aac)`

I think that should be fine, but the problem is that there are around one hundred files to modify...

Nadahar:
@Sami32 I'm not sure about modifying the audio codec, some may support only aac-lc and he-aac, some only aac-lc and others more variants. I think that would need to be done on a per-renderer basis. Some of the ```f:aac``` and ```f:m4a``` lines already exclude ALAC for example, so I think the best thing would be to not touch the audio codec. 

Why do you want to exclude ```aac-``` and ```he-aac``` form the ```f:m4a``` lines? What are they supposed to match then, ALAC?

Sami32:
Yes, ALAC need audio/x-m4a MIME type, but some MP4 can have MP3 codec or other that will not be accepted/played with audio/mp4.

At this stade i don't want anything, i just prefer forget about that and leave it.

PR #1090, opened 2016-11-08
Added the ability to transcode to MP4

SubJunk:


src/main/java/net/pms/dlna/DLNAResource.java
```diff
@@ -2257,6 +2257,8 @@ private String getDlnaOrgPnFlags(RendererConfiguration mediaRenderer, int locali
 				} else if (mime.equals("video/vnd.dlna.mpeg-tts")) {
 					// patters - on Sony BDP m2ts clips aren't listed without this
 					dlnaOrgPnFlags = "DLNA.ORG_PN=" + getMPEG_TS_EULocalizedValue(localizationValue, media.isHDVideo());
+				} else if (mime.equals(MP4_TYPEMIME)) {
+					dlnaOrgPnFlags = "*";
```
atamariya: @SubJunk In most cases for DLNA flags, the defaults work. If it is mentioned, it has to be correct. You can verify the same by removing these lines. 
The other thing to try is - copy the ffmpeg command line and generate an output from terminal in a file. Check if that works.

SubJunk: The converted file plays, just not while streaming. You said you use it all the time, so can you let me know how the following is different to what you use?

```
"C:\Program Files (x86)\Universal Media Server\win32\ffmpeg64.exe" -y -loglevel info -i D:\Klaus\Unsorted\Unwatched\Adventure.Time.With.Finn.And.Jake.S02E18.1080p.BluRay.x264-DEiMOS.mkv -bufsize 30000k -maxrate 29000k -crf 23 -ab 320k -ar 48000 -c:a aac -c:v libx264 -tune zerolatency -preset ultrafast -level 31 -pix_fmt yuv420p -f mp4 -movflags frag_keyframe+empty_moov file.mp4
```

atamariya: @SubJunk If the converted file plays fine in the renderer, it would be difficult to guess what else might be failing. Do you see anything in the logs?

SubJunk: @atamariya UMS acts like it is sending it so it doesn't realise anything has gone wrong

PR #979, opened 2016-08-07
Implement OSHI to obtain CPU and memory info

valib:
This CPU info could be used for https://github.com/yakka34/Unofficial-cpubenchmark.net-API offered by @Sami32 to obtain the CPU score discussed in https://github.com/UniversalMediaServer/UniversalMediaServer/commit/70914a8c03a455d02dc6b9fbc883638cf6c08f10

---

This change is [Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/979)

PR #1091, opened 2016-11-12
Implement OSHI to obtain CPU and memory info (#979)

Nadahar:
I just learned that if you force push while a PR is closed, the PR is invalidated and can't be reopened for some strange reason.

This is therefore the continuation of #979. That doesn't mean that I think it should be merged - I actually dislike several aspects of this. In addition to my reservation regarding catering just to those with spare CPU capacity and bad networks:
* I think storing the CPU information in the configuration file defeats all purpose of this. That means that whenever someone moves UMS from one computer to another (or upgrades their hardware) UMS will still use the information from the previous hardware. That won't make sense to the users, that I'm sure of.
* This doesn't actually measure the speed of the computer in any way, but tries to get a "CPU score" from a benchmarking website. The average reported benchmark score of a CPU can not be translated to "how much CPU resources we have available in the current enironment".
* The lookup is based on a text string of the CPU name. This will only work as long as everything is kept updated, and is bound to fail for more rare cases or when some of the involved libraries are abandoned.
* Retrieving the results is done by scraping [cpubenchmark.net](http://www.cpubenchmark.net). Scraping is bad in so many ways I don't know where to begin. It's slow, it WILL break and it's often against the will of the host. In addition the current scraping code is very "lazily" written and doesn't handle when things doesn't go as planned. One basic example is that if UMS doesn't have internet connection, it fails to start. This also breaks with UMS' "use external network" permission, it will contact the internet regardless.

@valib I had to do a lot of rebasing work to get this branch to compile at all. I don't know what had happened, but there were some serious git corruptions in this branch. I think I managed to figure it all out. In addition I've squashed simple spell correction or white space commits, removed the commit that was reversed and it's reversal and removed the merge commits. The result is that this shrunk from 12 to 3 commits!

For me, this is a major reason why I takes a long time for me to review a PR - when there is so much "mess" with the commits, I need a long time to figure it out before I can actually review the changes. It took me almost 3 hours to get this to a state where I could actually get an overview. You should really start using rebasing to clean up all your "work in progress" commits, and at the very least use ```amend```.

One further issue is that the ```Oshi``` is Java 8 only. I've told you a long time ago that you should configure Eclipse to use JDK 7 as long as we're working with Java 7, you would have detected it right away. So, I had to revert the version back to the last Java 7 release ```2.6-m-java7```. 

Then I could finally figure out the ```low``` error. The error is simple enough, it is on line 122 in ```CbmAPI```: 
```java
        int cpuScore = Integer.parseInt(table.select("span").last().text());
```

The reason it fails is because it is a scraper and a lousy one as such. It expects that element to be the cpu score, while it's actually the word "low". The reason is bound to be because they have done some slight adjustment to the web page. In addition the code is so terribly lazy that it doesn't check for a ```NumberFormatException``` when parsing scraped data! I can't fix it because I don't know what to do when the score can't be found. That could happen even if I updated the scraper to find the (currently) correct element. It's also guaranteed to send CPU strings that [cpubenchmark.net](http://www.cpubenchmark.net) can't parse and the scraper would probably fail even worse.

A better approach (although I still consider it wrong to assume that UMS can "spend" all available CPU resources to save network bandwidth) would be to run a short CPU intensive code or external program and then time it. That would give an indication of speed that is a much better approximation. @Sami32 would probably have to look at the splash image for quite some time though :stuck_out_tongue:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -407,6 +407,16 @@
     		<artifactId>oshi-core</artifactId>
     		<version>2.6-m-java7</version>
 		</dependency>
+		<dependency>
+			<groupId>org.jsoup</groupId>
+			<artifactId>jsoup</artifactId>
+			<version>1.8.3</version>
+		</dependency>
+		<dependency>
+			<groupId>com.googlecode.json-simple</groupId>
+			<artifactId>json-simple</artifactId>
+			<version>1.1.1</version>
+		</dependency>
 	</dependencies>
 	<build>
 		<defaultGoal>assembly:assembly</defaultGoal>
--- a/pom.xml
+++ b/pom.xml
@@ -402,6 +402,11 @@
 			<artifactId>coverartarchive-api</artifactId>
 			<version>2.1.0</version>
 		</dependency>
+		<dependency>
+    		<groupId>com.github.dblock</groupId>
+    		<artifactId>oshi-core</artifactId>
+    		<version>2.6-m-java7</version>
+		</dependency>
 	</dependencies>
 	<build>
 		<defaultGoal>assembly:assembly</defaultGoal>
--- a/src/main/java/net/pms/PMS.java
+++ b/src/main/java/net/pms/PMS.java
@@ -67,6 +67,10 @@ import net.pms.newgui.*;
 import net.pms.remote.RemoteWeb;
 import net.pms.update.AutoUpdater;
 import net.pms.util.*;
+import oshi.SystemInfo;
+import oshi.hardware.CentralProcessor;
+import oshi.hardware.GlobalMemory;
+import oshi.hardware.HardwareAbstractionLayer;
 import org.apache.commons.configuration.ConfigurationException;
 import org.apache.commons.configuration.event.ConfigurationEvent;
 import org.apache.commons.configuration.event.ConfigurationListener;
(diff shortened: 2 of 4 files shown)


valib:
@Nadahar thanks for your effort. I agree that using [cpubenchmark.net](http://www.cpubenchmark.net) to get CPU score is not the best idea due to reasons you perfectly described. I fixed the code of `CbmAPI` which is not mine but  recommended by @Sami32. Agree that could be possible to create our own bechchmark and get some indication of the CPU speed but I don't have any idea what would be trigger (the benchmark result value) to start thinking about changing the  automatic UMS setting to get the best output result for the given CPU.

Nadahar:
@Valib I have to keep saying that I still think it's a bad idea to use the CPU speed for this, but I guess if we had a setting "Automatically adjust transcoding quality" it could be ok.

When it comes to doing the "benchmark" we wouldn't have to be very precise. We would just have to do a short test that took a second or less on a "normal" computer and then scale it from there. Faster computers would spend less time and slower more. We could do a simple percentage (100) from something we picked as a "reference computer". The reference computer, which should be one of hours, should be tested so that we found the limit if what it can manage to transcode, and then maybe go with 80 % of that. Then we could do the same with a slower computer, let's say @Sami32's (if we can make it transcode anything at all) and give that the value 10. With these two "data points" we would just make it a linear function for everything else and try to guess at what value different transcoding settings would work. x264 can actually do lossless transcoding, so it could be fun to test that on mine to see what it can achieve (I have quite a powerful CPU), maybe we could come up with a third point.

I think this would be a much better approach if this is to be done, but I'd like to repeat that it should be optional as I'm sure I'm not the only one that don't want my CPU to work harder than strictly necessary.

Sami32:
For me the goal was to help users that report issue while trying to stream or transcode H265 or 4K movies and complaining that it stutter or whatever, as they just forget that transcoding it make their computer on it kneel.

Sami32:
@Nadahar That's a nice catch :+1: I totally missed that OSHI was only Java 8 since 2.6.1 version, it's why i can't get it working.

PR #1093, opened 2016-11-12
Implement CPU benchmark

valib:
As @Nadahar requested in #1091 this is the first atempt to evaluate the CPU  to be used to set the transcoding  setting to meet the CPU capabilities
EDIT: the automatic transcoding configuration based on this PR should be implemented

### DIFF
--- a/src/main/java/net/pms/PMS.java
+++ b/src/main/java/net/pms/PMS.java
@@ -1386,12 +1386,17 @@ public class PMS {
 	 * warnings where appropriate.
 	 */
 	private void logSystemInfo() {
-		long memoryInMB = Runtime.getRuntime().maxMemory() / 1048576;
-
 		LOGGER.info("Java: " + System.getProperty("java.vm.name") + " " + System.getProperty("java.version") + " " + System.getProperty("sun.arch.data.model") + "-bit" + " by " + System.getProperty("java.vendor"));
 		LOGGER.info("OS: " + System.getProperty("os.name") + " " + getOSBitness() + "-bit " + System.getProperty("os.version"));
+		try {
+			LOGGER.info("CPU benchmark time: {} milliseconds.", Long.toString(CPUbechchmark.run(5)));
+		} catch (Exception e1) {
+			// TODO Auto-generated catch block
+			e1.printStackTrace();
+		}
+
 		LOGGER.info("Encoding: " + System.getProperty("file.encoding"));
-		LOGGER.info("Memory: {} MB", memoryInMB);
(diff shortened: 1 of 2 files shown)


SubJunk:
@valib isn't it better to use a library like we used for previous PRs? What was wrong with that? It seems more accurate to do a lookup since this will be affected by what else the computer is doing.

When I tested it last time it seemed good and just needed a couple of tweaks like making it not add to the initial load time

Nadahar:
@SubJunk If you read my post in #1091 you'll see some of my argument against using the other solution. I think that's what behind @valib making this. I personally think this is a much better approach, exactly the point that everything else in the environment (like if it's otherwise engaged, used for virtualization, load of background tasks/OS, speed and amount of RAM etc) is taken into account. Basing it just an a CPU score is oversimplifying things imo, pluss the fact that the other way is very bug prone being based on CPU name strings, external web sites and scraping.

Let's not pretend this whole "automatic tuning" feature will have much accuracy in any case. There's no good way to translate CPU score into transcoding settings. It will have to be a pretty rough guess with a decent safety margin in any case, this will at least have a shot at measuring the computer's actual performance level in it's real world environment.

SubJunk:
That's a fair point, you've persuaded me that this approach is better, and yes FFmpeg transcodes multithreaded

Nadahar:
This won't work, the benchmark takes a whopping 92 seconds on my computer and my CPU utilization is only about 10% with none of the cores running high (I'd expect one core to work near full capacity). 

I think the best approach woud to bundle some short sample file, spawn a transcoding session early during startup transcoding that file to ```/dev/null``` on Linux/OS X or ```> NUL``` on Windows and then timing that. FFmpeg is bundled for all platforms, so it should be available. Still, the code needs to handle if either FFmpeg or the sample file isn't available without crashing, although it would leave us without test results. I don't think the test needs to run much more than a second on an "up to date" system to get a reasonable clue about the speed.

Sami32:
@Nadahar Thanks for the link :-) It's a nice tool.

Sami32:
Should this PR still stay open ?

valib:
@Sami32 I closed this PR because the idea to use FFmpeg to benchmark CPU mentioned by @Nadahar seems better. Maybe we can add the benchmark feature to the code which is checking the fontconfig cache in the background and do it in one step.

Nadahar:
@valib Either that or just create a separate thread for it. If you look at the code that gets the hardware information, it's done it its own thread. Code that is truly independent, like this would be, is ideal for running in its own thread - the only place which needs thread-safety is when you're "storing" the result back in UMS.

PR #1094, opened 2016-11-13
Updated/Created new renderer profile for Kodi

Nadahar:
This isn't perfect, trying to configure ```Supported``` lines correctly looking up file format and codec combinations and how they are interpreted by UMS makes me loose the will to live, so I'm simply uploading it as it is now. It's certainly a whole lot better than the current Kodi profile, which to be honest, doesn't work very well at all.

Kodi is brilliant for testing in my opinion, but the severely lacking renderer profiles always drive me crazy and introduces issues whenever I am to do that. This will remedy that situation a lot I'd think.

I've tested this on quite a few audio and video formats, most seem to work well. I haven't tested all the subtitle formats since I don't have test files for them, and there are some problems with a few of the audio formats (some M4A files aren't correctly recognized, but that is a UMS bug that's been there a long time). Generally it will play most audio well. 

When it comes to images I haven't really managed to get it to show JPG at all - it just displays a black screen. I don't quite understand how that could be UMS' fault so I'm assuming that is a bug with Kodi and haven't pursued it any further.

Feel free to make corrections to this PR if you see anything that could be improved, and I hope this can be merged relatively quickly so that all the issues I have when testing other stuff can be a little less frustrating.

---
This change is [Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/1094)

### DIFF
--- a/src/main/external-resources/renderers/Kodi.conf
+++ b/src/main/external-resources/renderers/Kodi.conf
@@ -16,9 +16,65 @@ RendererIcon =
 
 UserAgentSearch = Kodi
 UpnpDetailsSearch = Kodi
+
+DefaultVBVBufSize = true
+H264Level41Limited = false
+MediaInfo = true
+MuxDTSToMpeg = true
 RemoveTagsFromSRTSubtitles = false
+StreamSubsForTranscodedVideo = true
+SubtitleHttpHeader = CaptionInfo.sec
+TranscodeAudio = WAV
+TranscodeFastStart = true
 
-# Include all settings from this parent configuration:
-# Note: this must come last for the overrides above to maintain priority
-include = XBMC.conf
+Supported = f:3gp             v:cvid|divx|h263|h264|h265|mjpeg|mp4|mpeg1|mpeg2|vc1|wmv   a:aac|aac-he|ac3|dts|dtshd|lpcm|mp2|mp3|mpa|truehd|wma        m:video/3gpp
+Supported = f:avi             v:cvid|divx|h263|h264|h265|mjpeg|mp4|mpeg1|mpeg2|vc1|wmv   a:aac|aac-he|ac3|dts|dtshd|lpcm|mp2|mp3|mpa|truehd|wma        m:video/avi
+Supported = f:flv                                                                                                                                  m:video/flv
+Supported = f:mjpeg																									                               m:video/mjpeg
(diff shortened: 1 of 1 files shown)


src/main/external-resources/renderers/Kodi.conf
```diff
@@ -18,7 +18,66 @@ UserAgentSearch = Kodi
...
 
-# Include all settings from this parent configuration:
-# Note: this must come last for the overrides above to maintain priority
-include = XBMC.conf
+MediaInfo = true
+TranscodeAudio = WAV
+DefaultVBVBufSize = true
+MuxDTSToMpeg = true
+H264Level41Limited = false
+StreamExtensions = mkv,hdmov,hdm,flac,fla,dts,ogg,asf,asx,m2v,avi
```
SubJunk: Why use `streamextensions` as well as `supported`?

SubJunk: Why use `streamextensions` as well as `supported`?

src/main/external-resources/renderers/Kodi.conf
```diff
@@ -18,7 +18,66 @@ UserAgentSearch = Kodi
...
+SubtitleHttpHeader = CaptionInfo.sec
+TranscodeFastStart = true
+StreamSubsForTranscodedVideo = true
 
+Supported = f:3gp             v:divx|h263|h264|h265|vc1|mjpeg|mp4|mpeg1|mpeg2|wmv|cvid   a:aac|aac-he|ac3|dts|dtshd|lpcm|mpa|mp2|mp3|wma|truehd        m:video/3gpp
```
SubJunk: This looks good but I usually try to keep the parameters in alphabetical order. Not that important just a note.

Nadahar: I haven't paid any attention to the order, I was slowly loosing the will to live as I was trying to test all combinations...

When you say parameters, do you mean the order of the supported lines, to order of the other lines or the "match groups" in the supported lines?

SubJunk: Both, but like I said it's not important, it can just make it a bit easier with certain profiles since officially renderer documentation is usually listed in alphabetical order. It probably doesn't make much difference here

src/main/external-resources/renderers/Kodi.conf
SubJunk: I'm happy for this to be merged after that's removed

Nadahar:
@SubJunk I think I've addressed your comments, please see if there's anything more.

PR #1061, opened 2016-10-22
Show generic icons for media that don't provide thumbnails

valib:
Show generic icons (thumbnails) for media whose don't provide  thumbnails or  thumbnails are not generated for them like  audio files wihout CoverArt. It can be simply updated by addding images with convention "media format".png and "media format".jpg to the resources/images folder. Images I provided in this commit are only for testing because I think  that they could be changed to something better.

PR #1097, opened 2016-11-15
Renamed PmsConfiguration.getDisableTranscoding() to PmsConfiguration.isDisableTranscoding()

valib:


valib:
Finally I did it properly.

Nadahar:
@valib I've rebased and corrected this commit as this didn't even compile (you can't just rename the method in one place). 

I've also reversed 6fdd1538642f1f45d842102ae7282242cbbab06a for now as I can't understand how it's not a bug.

@valib @Sami32 When it comes to the "bug" this was supposed to rectify I still don't know what the bug was. Looking at the code, it seems to me like the reason ```{External Subtitles}``` would display instead of the actual subtitle information is because ```media_audio``` and ```media_subtitle``` is ```null```.
```java
		if (
			hasExternalSubtitles() &&
			!isNamedNoEncoding &&
			media_audio == null &&
			media_subtitle == null &&
			!configurationSpecificToRenderer.hideSubsInfo() &&
			(
				player == null ||
				player.isExternalSubtitlesSupported()
			)
		) {
			nameSuffix += " " + Messages.getString("DLNAResource.1");
		}
```
The above also seems to be a bug to me, why is ```media_audio``` taken into account? The "detailed" subtitle info can only be generated if ```media_subtitle``` isn't ```null``` so that makes sense, but ```media_audio``` isn't used here and if we have a case where ```media_audio``` is ```null``` but ```media_subtitle``` isn't, we will end up with no ```{}``` section even though valid external subtitles exists. Is that the intention, and if so why?

What happens for @Sami32 since he gets ```{External subtitles}``` while I don't must be that the subtitles and audio isn't parsed properly, so as far as I can understand that's where the "bug" must be. Was MediaInfo used? Given that we don't have the information needed to generate the detailed subtitle information, it's hard for me to see how showing ```{External subtitles}``` is a bug...?

Sami32:
It's a good point, that make me remember the people in Asia view Charlie Chaplin movies with subtitle.
I guess that movies or videos with subtitle(s) but without audio was not taken in consideration maybe because they are very rare ?

@Nadahar As you saw in my log !, MediaInfo 0.7.81 was used. My parsing results seem to be correct, but you didn't told me if you get the same parsing ?

Sami32:
I missed the fact that external subtitles was parsed with some issues; if they begin with the same name their are parsed and added to the movie:
Ex. data0001_same_fixed.srt and data0001.srt are for different video but the first one will be added to the data0001.ts video.

valib:
@Nadahar, @Sami32 I made the PR #1098 which can solve the problems you mentioned but it need testing because it seems to me that it makes problems for the web interface. For my TV it works well.

Nadahar:
@valib I like to finish things before I get involved in new ones - I'm still waiting for your response to #1097 and #1061 before I look at #1098.

Nadahar:
@Sami32 I tested with your SubRip file and it displayed as well. I haven't kept my logs, but obviously it parsed fine for me.

About Charlie Chaplin you really miss the point. I'm not talking about *movies* with subtitles and no audio, I'm talking about any video content with subtitles where the audio can't be parsed/recognized. It could either be because there are no audio, or because our parsing doesn't recognize it. Even though it's probably not a very common occurrence, I'm wondering why there is a check for audio there - and if it serves any purpose.

Sami32:
@Nadahar I only missed the part where the audio was not recognised, meaning that i thought delete this audio check will be a good thing.

The exact SRT file name is in the log i leave to [you](https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1053#issuecomment-260208813) or [the one](https://github.com/UniversalMediaServer/UniversalMediaServer/files/596448/debug.log.txt) i leaved to @valib . But any external subtitle beginning with the same name, as related [here](https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1097#issuecomment-261101193), will give the same inadequat parsing.

Nadahar:
@Sami32 Sorry, I don't see where you left me the names. I already used ```data0001.srt``` that you gave me when I tested, but unless I misunderstood you there were more subtitle files starting with ```data0001``` in your folder and you think that has to do with the problem? Or did I misunderstand you..?

Sami32:
@Nadahar Sorry, maybe i should have been clear saying that it's a different subtitles parsing issue.
External subtiles with same name at the beginning are incorrectly handled.
Ex. _Nadahar.srt_ and _Nadahar_two.srt_ coming from different movie (or not) will be parsed with _Nadahar_ movie.

Nadahar:
@Sami32 I'm still confused - I've understood the issue you point to now (any file starting with the same) and I can see how that one can be a challenge to "fix" because in most cases that's the correct thing to do.

Are you saying that this is not related to the issue discussed here where you for some strange reason get ```{External subtitles}``` for the example files you posted where I get the actually parsed subtitle information?

If those are not related, the "common starting name" issue should be discussed in a separate issue.

PR #1100, opened 2016-11-17
link update

Sami32:
That's said since Java 6 issues related in the reference link, the things could have changed:
http://docs.oracle.com/javase/8/docs/api/java/lang/Runtime.html#exec
http://docs.oracle.com/javase/8/docs/api/java/lang/ProcessBuilder.html

### DIFF
--- a/src/main/java/net/pms/util/ProcessUtil.java
+++ b/src/main/java/net/pms/util/ProcessUtil.java
@@ -49,7 +49,7 @@ public class ProcessUtil {
 	private static final int ALRM_TIMEOUT = 2000;
 
 	// work around a Java bug
-	// see: http://kylecartmell.com/?p=9
+	// see: http://www.cnblogs.com/abnercai/archive/2012/12/27/2836008.html
 	public static int waitFor(Process p) {
 		int exit = -1;
 


Nadahar:
@Sami32 Do you know that the new link has the same content as the old, dead link? Any, UMS isn't using Java 8 but Java 7.

Sami32:
Of course, it's just an update for the dead link.
You can look at Java 7 if you want, but AFAIK UMS Java 8 version exist, right ?

Nadahar:
Isn't it better to use [this link](https://web.archive.org/web/20121201070147/http://kylecartmell.com/?p=9) so it won't die again? 

Regarding Java 7/8 the point is that Java 7 code can be run by Java 8 (in all but a very few situation) but Java 8 code can't be run by Java 7. Even if we compile a version for Java 8, it doesn't mean that we use what's available in Java 8. If so, we would have to develop to parallel versions, since there's a lot of things you can do in Java 8 that doesn't exist in Java 7.

As long as UMS is to support Java 7 we're effectively limited to Java 7, and it's the Java 7 "rules" that apply.

Sami32:
@nadahar Yes, it was my first thought. But i was afraid they loose the data as well:

> In August 2012, the Archive announced[20] that it has added BitTorrent to its file download options for over 1.3 million existing files, and all newly uploaded files.[21][22] This method is the fastest means of downloading media from the Archive, as files are served from two Archive data centers, in addition to other torrent clients which have downloaded and continue to serve the files.[21][23] On November 6, 2013, the Internet Archive's headquarters in San Francisco's Richmond District caught fire, destroying equipment and damaging some nearby apartments.[24] According to the Archive, it lost:[25] a side-building housing one of 30 of its scanning centers; cameras, lights, and scanning equipment worth hundreds of thousands of dollars; and "maybe 20 boxes of books and film, some irreplaceable, most already digitized, and some replaceable". The nonprofit Archive sought donations to cover the estimated $600,000 in damage.[26]

That's said perhaps, it'll a better choice ? I was not sure about.

Do you mean that we cannot implement the latest `ProcessBuild` or `runtime` classes improvments in UMS Java 8 version ?

Nadahar:
@Sami32 We could, but that would generate a lot of extra work and would require separate testing of each version etc. Whenever kode was changed, you'd have to come up with different changes for each version. So, in reality: no.

Sami32:
@Nadahar You make me confuse. What for the UMS Java 8 was made ? As you told me that the Java 7 version can be run on Java 8 system.

Nadahar:
@Sami32 You'd have to ask @SubJunk, I'm kinda wondering the same thing. I know there are very rare cases where Java 7 code won't work, it might simply work better to compile it on Java 8 for running on Java 8 - even though our code is Java 7. 

We did have some differences between the Java 6 and Java 7 versions, and the it quickly turned into a lot of work to maintain. The differences were very few though. Effectively 99% of UMS' code is still Java 6, as relatively few Java 7 features have been implemented in relatively few places.

Sami32:
@Nadahar Thank you for the explanation :-)

Java 9 will arrive, maybe at the end of the next year, as it's delayed nothing is sure yet...
The nice thing i did read, is the load modularity of the dependancies, it could lightening Java applications.

That's said some Linux OS only support Java 7, as Windows XP is more confortable with it, but 2017 could be the last year of Java 7 support for UMS main build IMO,  because perhaps less than 5% will use these OS.

Newbie question: Did some Oracle tool converting efficiently Java 7 to Java 8 code exist ?

Nadahar:
@Sami32 I seem to remember touching some of those "Java 7 only" Linuxes eariler, but I don't really remember now. I find it a bit strange that they can't installed e.g [Oracle Java 8](http://www.oracle.com/technetwork/java/javase/downloads/jre8-downloads-2133155.html) even if the only version of OpenJDK.

When it comes to "coverting" code I doubt that is possible in any reasonable way. As said, in almost all cases Java 7 code will work just fine on Java 8 - that's basicly what we do when we compile our Java 8. Used this way, Java 8 has no advantages over Java 7. To take advantage of Java 8 you have to use the new features it offers - and that means doing what you do in a different way or doing things you couldn't do with Java 7. I can't even imagine how a tool should be able to do that, that would basicly mean you had an "automatic refactoring tool" that could refactor existing code into "better" code - that would be fantastic, but would also exclude developers from the process :wink: 

That said, I'm sure there are some "helper tools" available to tell you when simple things can be written in a "Java 8 way", but I honestly don't see the point in doing that unless there is something specific to gain by doing it. What you loose is obvious, the ability to compile the code with Java 7.

Sami32:
@Nadahar I was suspecting that.
Seing the _processbuilder_ and _runtime_ weaked from Java 1.6 used, make thing that perhaps a lot of "weaked" code could be upgraded when Java 1.8 trunk update will be done.
You just wiped my "automatic" upgrade/convert hope :wink:

SubJunk:
The Java 8 branch is there so that it's easier when we do drop support for Java 7, and because it's maybe slightly faster to run with certain operations. We could do a poll to see how many people are stuck on Java 7. We did polls before we dropped Java 6 support and got good feedback.

SubJunk:
These are the differences between our Java 7 and 8 releases. Mostly it's GUI-related but it's not nothing https://github.com/UniversalMediaServer/UniversalMediaServer/compare/Java8

Nadahar:
About missing Java 8 on Linux, I'm wondering if that was my finding and it that was Debian. Debian is extremely slow to adopt changes because of their philosopy where providing something rock solid trumps new features. That means that anything released in their "stable" version will need thousands of hours of use before it is deemed "stable".

That said, one can install Java 8 on Debian, either [OpenJDK 8 via backports or "testing"](https://packages.debian.org/search?keywords=openjdk-8-jre) or simply by installing Oracle Java 8. I'm not saying that Debian is the only Linux distribution that doesn't offer Java 8 "by default" because there are hundreds of distributions, but I think it's the only major distribution that doesn't. But, Debian isn't a desktop distribution (you can use it as a desktop, but that's not the targeted usage case) - it's meant for server use where stability and predictability is "all that matters". What I'm trying to say is that I don't think the extra "work" involved in installing Java 8 on Debian is a valid reason stick to Java 7. Debian 9 "Stretch" , which is currently considered "testing" but will become the next "stable", offers OpenJDK 8 and is up for release [in a not too distant future](https://lists.debian.org/debian-devel-announce/2016/11/msg00002.html)*.

PR #1095, opened 2016-11-13
Make web server interface binding configurable

zero0cool0:
Changes to RemoteWeb to make the web server interface (port 9001) binding configurable, similar to the server (port 5001)

PR #1102, opened 2016-11-24
Make all network services use network configuration

valib:
Formatted PR #1095 and this PR replace it.

### DIFF
--- a/src/main/java/net/pms/remote/RemoteWeb.java
+++ b/src/main/java/net/pms/remote/RemoteWeb.java
@@ -2,8 +2,7 @@ package net.pms.remote;
 
 import com.sun.net.httpserver.*;
 import java.io.*;
-import java.net.InetAddress;
-import java.net.InetSocketAddress;
+import java.net.*;
 import java.security.AccessController;
 import java.security.GeneralSecurityException;
 import java.security.KeyStore;
@@ -22,6 +21,7 @@ import net.pms.configuration.WebRender;
 import net.pms.dlna.DLNAResource;
 import net.pms.dlna.RootFolder;
 import net.pms.newgui.DbgPacker;
+import net.pms.network.NetworkConfiguration;
 import org.apache.commons.configuration.ConfigurationException;
 import org.apache.commons.lang.StringUtils;
 import org.slf4j.Logger;
@@ -39,6 +39,10 @@ public class RemoteWeb {
 	private RemoteUtil.ResourceManager resources;
 	private static final PmsConfiguration configuration = PMS.getConfiguration();
 	private static final int defaultPort = configuration.getWebPort();
+
+	private InetAddress iafinal;
+	private String hostname;
+	private NetworkInterface networkInterface;
 
 	public RemoteWeb() throws IOException {
 		this(defaultPort);
@@ -65,7 +69,29 @@ public class RemoteWeb {
 		//readCred();
 
 		// Setup the socket address
-		InetSocketAddress address = new InetSocketAddress(InetAddress.getByName("0.0.0.0"), port);
+
+		hostname = configuration.getServerHostname();
+		InetSocketAddress address;
+
+		if (StringUtils.isNotBlank(hostname)) {
+			LOGGER.info("WEB: Using forced address " + hostname);
+			InetAddress tempIA = InetAddress.getByName(hostname);
+
+			if (tempIA != null && networkInterface != null && networkInterface.equals(NetworkInterface.getByInetAddress(tempIA))) {
+				address = new InetSocketAddress(tempIA, port);
+			} else {
+				address = new InetSocketAddress(hostname, port);
+			}
+		} else if (isAddressFromInterfaceFound(configuration.getNetworkInterface())) { // sets iafinal and networkInterface
+			LOGGER.info("WEB: Using address {} found on network interface: {}", iafinal, networkInterface.toString().trim().replace('\n', ' '));
+			address = new InetSocketAddress(iafinal, port);
+		} else {
+			LOGGER.info("WEB: Using localhost address");
+			address = new InetSocketAddress(port);
+		}
+
+		LOGGER.info("WEB: Created socket: " + address);
+
 
 		// Initialize the HTTP(S) server
 		if (configuration.getWebHttps()) {
@@ -147,6 +173,21 @@ public class RemoteWeb {
 			}
 		});
 		return server;
+	}
+
+
+	// this sets iafinal and networkInterface
+	private boolean isAddressFromInterfaceFound(String networkInterfaceName) {
+		NetworkConfiguration.InterfaceAssociation ia = StringUtils.isNotEmpty(networkInterfaceName) ?
+				NetworkConfiguration.getInstance().getAddressForNetworkInterfaceName(networkInterfaceName) :
+					NetworkConfiguration.getInstance().getDefaultNetworkInterfaceAddress();
+
+		if (ia != null) {
+			iafinal = ia.getAddr();
+			networkInterface = ia.getIface();
+		}
+
+		return ia != null;
 	}
 
 	public String getTag(String user) {


Nadahar:
@valib I've squashed your commit c9bea11e47bdf3d8abd76e5eaa34ed25cf5ad6c3 into @zero0cool0's ff3421d7c0c26fa997e9094da0374ed9d9f4449e. The new commit is 0922b12e7543869a0f9a739127a3f9d2d727ba96,

I have started to try to address the problems with this implementation, but I don't have time to finish it now.

Nadahar:
I've committed some changes for this. I'm still short on time and have only very briefly tested it. The commit message explains the changes.

Many of the bugs/issues fixed here is inherited from ```HTTPServer.start()``` so corresponding fixes should be implemented there as well. It's a bit more complicated there however.

Nadahar:
@UniversalMediaServer/developers This is a bit of a nightmare - I guess there's a reason why nobody has wanted to touch the network configuration for so long. The web interface itself is easy enough, but for it to have any point I figured that the UPnP configuration should adhere to the configuration as well, and this is where it all starts to get complicated.

Cling is a chapter in itself, I THINK I have that covered now but I'm not quite sure if the changes I made apply both to Cling's UPnP ports and Cling's HTTP server or not. Actually I'm not sure if Cling spawns a HTTP server or not. Any help with this is welcome.

The real nightmare is the "internal" UPnP classes, they are a mess of another dimension. Not only do I have to control what socket they bind to (that in itself is achievable), but I have to make them restartable. This doesn't seem to ever have been a thought, and the shutdown mechanisms are appalling not to say non-existing. The general disregard for interrupts thoughout the code is depressing, and I can't really see any clear strategy for closing down threads. It doesn't help to call ```thread.interrupt()``` when the thread code itself ignores interrupts... and we have blocking sockets that doesn't even respond to interrupts. In all, it's a mess.

This isn't completed by any means, I pushed it simply to let you see a bit of what I'm trying to do. I don't know if this will be successful, and we'll need to test it well. To having any hope of ever completing this, I have to try to look another way regarding a lot of issues found along the way like thread safety, generics etc. Imo the internal UPnP stack should be removed and Cling be used to handle all of this. The current "hybrid solution" doesn't deserve to be called a solution...

Nadahar:
@UniversalMediaServer/developers This turned out to be a big one. This was far from easy, and I can't be sure that I haven't created some new bugs. There are a couple of areas I'm particularly "worried" about:
* I have removed the exclusion of virtual network adapters both in our "internal" UPnP and Cling. From what I could find the reason for this exclusion was simply that some OS X users running some VM's on their desktop OS (VMWare Workstation and similar) had some issues if these virtual adapters were used for multicasting. As I see it, that's not a good enough reason to exclude them. Excluding them means that UPnP won't work properly from any VM either running on a virtualisation server or alongside a desktop OS. It also means that even if the issue with the virtual adapters on OS X is fixed (which it might be already) UPnP still won't work for the VM adapters, neither will it on other platforms without the bug. I think the only correct thing to do is to remove the exclusion and maybe make a note in the documentation that if OS X users have issues with this, they should configure their network to use a specified network interface.
* I have implemented a slightly different multicast socket binding logic, also for Cling. Special arrangements have been made for Windows since the "proper" binding is reported to not work correctly for Windows. One multicast listening socket (and thread) is created PER active interface. This could solve current issues with hosts having multiple network interfaces, but it could also introduce unknown side effects.

There has been a lot of changes here and I expect this to be a bit hard to review. I've tried to split the commits into "logical units of change", even though in reality most of this was done without commiting as it is all very entangled. I've also tried to make descriptive commit messages, but I'm sure there are more that can be said about them. Please don't hesitate to ask if anything is unclear. I might add further details later, but I know from experience that it won't be read anyway so it might be better to just answer any questions you might have.

This will need a lot of testing with different network configurations, it might simply be that this should be released as a beta to gather feedback.

atamariya:
@Nadahar Quick question - did you account for accessing the web socket via multiple interfaces? The problem I'd noticed was that the multicast discovery works. But since web socket used to run on a particular socket, future communication for media used to fail.

Nadahar:
@atamariya Yes, both the "main" HTTP interface and the web interface should reply or multiple interfaces now.

atamariya:
@Nadahar The UPnP discovery returns a device XML (PMS.xml). This has a web URL which is used by the client for future communication. So if the client is in 192.168.0.x, the URL should be in the same subnet. Else the communication fails as no route is found.

Nadahar:
@atamariya I'm aware of that aspect, and I haven't addressed that. This doesn't only apply to the "device" but to every other link/address sent. All the URLs for content would need the same "address aware" logic.

My goal has been to limit this PR to the network configuration itself. I've bumped into several more or less related issues that I've left alone, simply because it would otherwise become huge and untestable/unmergable.

If the client has a route to the other network it should work as it is. If not, if will fail. I'm not sure what address UMS "picks" as it's reference address. From what I've seen in the Cling code, Cling handles this. It figures out which of its own IP addresses is on the same subnet as the client, and then substitutes it's outgoing "reference" address accordingly. From what I can see, that can also fail if there is routing involved, but I haven't studied that in detail. In any case, routing multicast is too complicated for all but the very few, so I'm guessing that is a very rare scenario.

As I see it, UMS' internal UPnP handling should be removed and device and services configured via Cling instead. That would take care of this (maybe except for routed networks) "for free".

Nadahar:
I just tested this on a Ubuntu 12.04 LTS VM and it seems to work fine except that if detects an "unknown renderer" on 127.0.0.1. I'm not sure why, as loopback interfaces are supposed to be excluded.

Nadahar:
I just tested this on OS X 10.8.5, all seems to work fine - no loopback device discovered here. Both of these tests are with a single network card and IP address, but that's my primary concern since that's all the current implementation can handle anyway.

Nadahar:
Same test on Windows 10, no apparant issues.

Nadahar:
Tested on Windows XP 32 bit, no issues.

src/main/java/net/pms/network/HTTPServer.java
```diff
@@ -40,35 +43,97 @@
...
+				socketAddress = new InetSocketAddress(port);
+			}
+		} else {
+			socketAddress = tmpSocketAddress;
+		}
+
+		if (StringUtils.isBlank(tmpHostname) && socketAddress != null) {
+			if (socketAddress.getAddress().isAnyLocalAddress()) {
+				try {
+					tmpHostname = InetAddress.getLocalHost().getHostName();
```
valib: @Nadahar host name doesn't work for my Panasonic TV. This TV understand only the IP address.

src/main/java/net/pms/network/HTTPServer.java
```diff
@@ -40,35 +43,97 @@
...
+					tmpHostname = InetAddress.getLoopbackAddress().getHostName();
+				}
+			} else {
+				tmpHostname = socketAddress.getHostString();
+			}
+		}
+		hostname = tmpHostname;
 	}
 
 	public String getURL() {
```
valib: @Nadahar maybe this should be changed to 

    public String getURL() {
		return "http://" + socketAddress.getAddress().toString() + ":" + port;
	}

Nadahar: @valib I just pushed a commit that takes care of the hostname/IP address issue. 

Are you describing another problem in the comment above a different problem?

valib:
@Nadahar great. Now it works 👍

src/main/java/net/pms/network/HTTPServer.java
valib: @Nadahar that problem was related to the hostname using and it was only the description how it  has an influence to the UMS renderers recognition. Now you fixed it and the problem is gone.

SubJunk:
Wow lots of work went into this. Thanks :) Will test on Windows 7 when I get a chance

SubJunk:
But if you're confident that the testing has already been comprehensive enough, and if a code review is done, don't wait for me to test :)

Nadahar:
@SubJunk I'm developing on Windows 7, so it fundamentally works on Windows 7. However, what mostly needs to be tested here is different network configurations, so anyones test can reveal something (like @valib without local name resolution).

I've found the reason for the "localhost" unknown renderer on Ubuntu. It was Cling that was "discovered", the ```isSelf``` definitions in the request handlers just got a bit more complicated now that we listen to multiple interfaces with the HTTP server. The reason it only happened on Ubuntu is that Debian (Ubuntu upstream) has decided to have a "default IP" of ```127.0.1.1```, and the existing ```isSelf``` definitions didn't see that as a local address. By definition anything ```127.*``` is local.

It's such "side effects", things that aren't really related to the changed code but behave unexpected under new circumstances, we need to find.

zero0cool0:
For what it's worth, I've tested this on a dual-interfaced Ubuntu 16.04 LTS using `network_interface=<internal interface>` and it behaves as expected. As far as I can tell, interface bindings are IP address specific for ports 5001, 9001 and 33439 (which I suspect is somehow linked UPNP?). More importantly, none of the external interfaces is bound (also not the loopback interface 127.0.0.1 if that is of interest). This is in-line with what I originally wanted to achieve with this PR.

Nadahar:
@zero0cool0 Thank you for testing 👍 As long as you bind to an interface or IP I think this works well. My testing has shown some issues when not binding (that is, trying to serve all interfaces) on Linux, I'm trying to figure it out.

It might be that I have to bite the bullet and implement dynamic "reference" address as @atamariya described, even though I wasn't ready for that extra work in this PR.

Nadahar:
Just tested on Arch, same results as Ubuntu (detecting itself as a renderer on loopback before last commit).

Nadahar:
@UniversalMediaServer/developers I just pushed my current "work in progress" on this. To fix the problems detected, I see no other solution than to make a "proper" network configuration where individual network interfaces and/or IP addresses are selectable.

This means that I have to do GUI redesign, and as much as I hate that I've started to do it. It took me a long time to get my custom ```JTree``` up and working properly. 

This is in no way complete, in fact the network configuration itself is quite bugged in this state, I'm only focusing on the GUI for now. I'd like to get feedback on what I'm doing on the "General" tab, as GUI work is so time consuming that I don't want to continue if I have to redo it later.

Currently I have some test code in there that generates some random "sub ip addresses" to show in the tree. This is for testing the tree only, and won't be there in the final code. Changes aren't saved to the configuration either at this stage.

What is new/I want feedback on are:
* Grouping the network configuration the way I've done it. 
* The interface/address selection tree
* The new ```DefaultTextField``` I've created and is currently used for hostname and port. The way it works is that it's like a normal text field except that when it's blank, the default is shown in a "muted color". 

Also, I'm not very happy with the icons used in the interface/address selection tree, but I decided to use them for now and move on, because I have a tendency of being stuck with such things for a loong time. Any suggestions there is very welcome.

Nadahar:
What an overwhelming response - doesn't anybody have ANY views on the current GUI "solution"?

valib:
@Nadahar I am not looking for the changes all over the time so sometimes the respond would take a time. This looks good but I am not sure if there will be any other IP addresses for the particular interface.

Nadahar:
@valib Normally there won't be multiple IP addresses per interface, but an interface can have multiple IP addresses (and sub interfaces) so I've tried to model the configuration to cater for all possibilities.

SubJunk:
Testing this is on my todo list :)

Nadahar:
@SubJunk There's not much to test at this stage, a quick look at the GUI changes is all that's needed atm. I need to know it the direction is endorsed or not before I put more work into it, GUI simply takes soo much time to get right.

SubJunk:
@Nadahar that looks awesome, the GUI seems like a big improvement to me

Nadahar:
I've reset the branch to it's original state, and this work will continue in #1195 instead. I vote for not merging this as it is, as it won't work if network configuration is changed. #1195 will solve what this PR intended.

SubJunk:
Closing this as it was superseded by #1195

PR #1105, opened 2016-11-26
Fix bug for the FFmpeg setting of the CPU cores

valib:
@SubJunk this is really annoying to make a PR for the evident bugs fix.


valib:
Sorry guys I made a mistake

Nadahar:
@valib Maybe the PR wasn't such a bad idea then, we just saved ourselves a reversal in master.

valib:
There must be something wrong with my WIN 10 because during testing the UMS offered only 1 CPU core instead of 4 available. When I retested this PR againn the 4 CPU cores are available and everything is OK.

SubJunk:
That was good timing for the PR to prove its usefulness ;)

Issue #554, opened 2015-05-05
dcrawMS.exe breaks scanning the library and makes DB huge in size + JPEG ISO exception


ExSport:
I have some .CR2 files. UMS tries to extract thumb when scanning library or manually browse them and store them in DB.
In my case scanning is stucked because no output is generated, no timeout, so UMS waits endlessly for the output. Killing process manually unpause this stuck but next file again pause whole scan process.
Running dcrawMS.exe -e -c photo.cr2 generates in cmd "Will not write an image to the terminal!" and without -c param thumb jpg file is generated with size larger than 4MB. If it is stored in library, DB enormeously grows (killing processes manually). I tried to update exe file with latest v9.25 64bit with no change.
I can provide sample somewhere if needed. Also it is the question if thumbs should not be cached outside DB as separate files as it generates huge DB (limited by java heapsize set) and also making small thumb instead of full thumb generated by dcrawMS (15MB cr2 includes 4MB jpeg as thumb=jpeg in full size, not real small thumb)

ExSport:
Btw. when I disabled image thumb generation, dcrawMS is still loaded so it freezes scan process on cr2 files again

ExSport:
It freezes on (log from library scan with unticked image thumbs generation):
```
TRACE 22:21:38.427 [Library Scanner] Matched format RAW to "E:\_MIX\_RAID\_FOTO (preber)_\_Fotky z Icka(externi)\fotooooooooooooooooo\DCIM2\100CANON\IMG_1103.CR2"
DEBUG 22:21:38.427 [Library Scanner] Starting C:\Program Files (x86)\Universal Media Server\win32\dcrawMS.exe -i -v E:\_MIX\_RAID\_FOTO (preber)_\_Fotky z Icka(externi)\fotooooooooooooooooo\DCIM2\100CANON\IMG_1103.CR2
DEBUG 22:21:39.987 [dcrawMS.exe-176-1] 
DEBUG 22:21:39.987 [dcrawMS.exe-176-1] Filename: E:\_MIX\_RAID\_FOTO (preber)_\_Fotky z Icka(externi)\fotooooooooooooooooo\DCIM2\100CANON\IMG_1103.CR2
DEBUG 22:21:39.987 [dcrawMS.exe-176-1] Timestamp: Sat Aug 02 02:24:43 2008
DEBUG 22:21:39.987 [dcrawMS.exe-176-1] Camera: Canon EOS 400D DIGITAL
[...]

PR #1109, opened 2016-12-04
Update the RAW images support.

valib:


Nadahar:
@valib Why is 06298aef3b49dbd1e2dc7f67f9026d62f2a4f6d2 here? Isn't that what #959 is for?

valib:
@Nadahar it was a mistake. I was working on both problems in the one branch and I forgot to make two different PR's.

Nadahar:
@valib Ok, if you do a rebase you can easily remove both commits (the wrong one and the reversal). If you pick the commit id/sha before the first one (78727617308731681bafbe20ab11f03b4a58d1f0 in this case) and do:
```
git rebase -i <commit id>
```
...a text document will open in whatever text editor your git is configured to use containing all commits made after <commit id>. By editing and saving that document and closing the text editor, git will do a rebase based on your edits. If you change nothing in the text document, no changes will be made, but it's very easy to change earlier commits, squash some or even reorder them by editing the text document. If you delete the lines for the two unwanted commits and leave the rest alone, the rebase will simply remove those two commits.

If you haven't configured a text editor for git, you'll have to [do that first](http://stackoverflow.com/a/773973/363791). Pick anything you like - I use Notepad++. I would stay away from ```notepad.exe``` because it doesn't handle neither UTF-8 nor "Linux" line feeds (LF without CR).

Sami32:
@valib :+1: 

About [PPM ](http://netpbm.sourceforge.net/doc/ppm.html), _TwelveMonkeys_ is able to handle it, as [Advanced Imaging Image I/O](http://www.oracle.com/technetwork/articles/java/advancedimage-140153.html) if i understood correctly. Since they're already implemented in UMS...
That's said i never tried PPM format.

EDIT: I discovered that one: https://github.com/tosmun/Java-PPMTool , just in case...

Nadahar:
@valib Why is it better to use streams than a bytearray?

valib:
@Nadahar it could be faster to use streams than use bytearray do you think that it is wrong?

Nadahar:
@valib My guess is that most stream implementations is a wrapper for a byte array, so I doubt it's faster. I'm not so fond of streams because of the "cursor" concept, that is that once something is read you can't read it again unless the stream supports reset, and then you have to start over.

A byte array enables you to read and write to any byte you want. As long as the size doesn't change I think arrays are great for storage, as they are simply a "wrapper" for directly accessed memory. Adding to an array is bad though, because you have to move the whole content to do it. Then a more intelligent wrapper like a list is better, because it allocates more memory than needed and keeps track of what is used and not.

I think streams are mostly useful when you need to have the possibility to handle an actual stream (e.g. network or file access).

valib:
@Nadahar OK. I will try to test both variant and will see.

Sami32:
@valib It's better than before, scanning without ending as "freezed" on SRW file, but have again issue with SRW files.
[debug.log_SRW.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/659731/debug.log_SRW.txt)

Sami32:
@valib A NPE is throwed:
> DEBUG 2016-12-27 22:29:10.250 [Library Scanner] Caught exception
> java.lang.NullPointerException: null
> 	at net.pms.formats.RAW.parse(RAW.java:149)

https://github.com/UniversalMediaServer/UniversalMediaServer/blob/master/src/main/java/net/pms/formats/RAW.java#L149
```
catch (Exception e) {
LOGGER.debug("Caught exception", e);
```

[debug.RAW.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/674619/debug.RAW.txt)

Sami32:
https://github.com/UniversalMediaServer/UniversalMediaServer/issues/554

Nadahar:
Given all the changes in #1061 I don't know how much of these issues will still be there.

Issue #1111, opened 2016-12-17
100% CPU usage while doing nothing. LG TV cpu usage bug resurface?


orangpelupa:
my issue sounds really similar with this old issue https://github.com/UniversalMediaServer/UniversalMediaServer/issues/316

basically after around 4 hours idle, or sometimes after around 1 hour of playing video without transcoding, UMS will eat 100% CPU. Even after i turned off my LG TV, it stays at 100%.

the last half of the log is attached

[ums.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/658901/ums.txt)

thankyoiu

Specs;
windows 10 64 bit
UMS 6.5.2 
LG UH610T

Nadahar:
@orangpelupa There's a lot of different errors in your partial log, unfortunately it's not enough to figure out what's going on. You need to post full debug files [as described here](http://www.universalmediaserver.com/forum/viewtopic.php?f=9&t=556). 

What I can see from your log is that you have a lot of exceptions that "crash threads" and my guess is that their resources arent properly deallocated because ot that. As a result, you run out of memory at the end of the log. Once out of memory things WILL go wrong. You can increase Java heap space to make it take longer to run out, but the real resolution would be to track down the causes of all those exceptions - which is why proper debug files are needed.

orangpelupa:
UMS wont eat 100% CPU when i started trace mode :(

have been trying to get it to 100% CPU, cant. weird.

valib:
@orangpelupa it seems like wrong UMS or Java instalation. What is your OS?

orangpelupa:
@valib im able to reproduce it while running trace mode :D  
[ums_dbg.zip](https://github.com/UniversalMediaServer/UniversalMediaServer/files/662912/ums_dbg.zip)

Nadahar:
@orangpelupa I can see none of those "interesting" errors in your trace log that your previous log had. There is one error there caused by a bug in UMS, but it shouldn't cause a lot of CPU usage as far as I can tell.

I think you need to let it "record" in trace mode until you get a similar situation as the previous log.

orangpelupa:
@Nadahar its a similar situation. it was idle, and it hogs the CPU 100%. Hmm, i'll try running it without trace again and see whether the normal logs will have the previous errors. and then with trace again.

orangpelupa:
after a few days it finally 100% cpu again but it was not in trace mode. now running in trace mode.

[...]

PR #1112, opened 2016-12-20
Rudimentary synchronization of InfoDb

Nadahar:
I just quickly did some redimentary synchronization of ```InfoDb``` based on the log in #1111. I don't know if the CPU issue is related to this, but it's extremely thread-unsafe the way that class is today. 

This implementation won't be very efficient (but I don't know if the difference is noticable). A more refined implementation should be made, but some heavy refactoring would be needed. As it is now this code almost qualifies as obfuscated to me, so it would take some effort to sort it out.

### DIFF
--- a/src/main/java/net/pms/util/InfoDb.java
+++ b/src/main/java/net/pms/util/InfoDb.java
@@ -36,25 +36,30 @@ public class InfoDb implements DbHandler {
 
 
 	private void askAndInsert(File f, String formattedName) {
-		try {
-			String[] tmp = OpenSubtitle.getInfo(f, formattedName);
-			Object obj = db.nullObj();
-			if (tmp != null) {
-				obj = create(tmp, 0);
+		synchronized (db) {
+			try {
+				String[] tmp = OpenSubtitle.getInfo(f, formattedName);
+				Object obj = db.nullObj();
+				if (tmp != null) {
+					obj = create(tmp, 0);
+				}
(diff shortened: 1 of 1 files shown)


Nadahar:
I'd like to add that I haven't tested it, as I'm not sure how ```FileDB``` is used. Although this should require massive testing, it should be verified as it (still) works as intended.

SubJunk:
@Nadahar some if not all of this is unused or due to be unused in 7.0.0 anyway, in favor of SQL

Sami32:
@Nadahar Do you think it can be merged now for see if it solved #1111, until 7.0.0 become a stable release ? or closed while waiting for 7.0.0 ?

Nadahar:
@Sami32 I'd ideally want for this to be tested, but I guess that's not really going to happen. The changes aren't what I'd consider likely to cause bugs, neither FindBugs nor Threadsafe have any complaints so I guess I might as well merge it.

Sami32:
@Nadahar I understood that, but i didn't know how to test it correctly/efficiently, as database use is not my strong use...
:-/

Nadahar:
@Sami32 It's database only in the name. It is really only a text file where you can add and remove entries, much like a "file-backed" ```List<String>```. That part is enough, I'm just not sure how it's actually being used and what implications a bug here would have. I could find out by studying the code, but isn't *that* interested, especially when 7.0.0 will replace its use with a real database.

PR #1113, opened 2016-12-21
Converts images to jpeg

valib:
Try to send transcoded images not supported by renderer in the JPEG format. This PR extends the PR #1061 and it needs testing before it should be merged


Nadahar:
@valib I'm sorry to make trouble for you, but I'm still working on #1061 so this will probably need to be updated/changed accordingly.

valib:
@Nadahar I am aware of it and I will do it if necessary. This commit works but there is something strange when some images are shown only partially and the full image is shown when it is open for the second time. That is why I made the PR to be tested by others.

Sami32:
I don't know if be able to choose to convert to PNG, instead of JPEG, to keep alpha channel need an other PR or can be done in this one ?

The convertion doesn't take **progressive** or **IPTC** statut in consideration, AFAIK.

Nadahar:
#1061 will handle most of (if not all) of this guys, sorry that I'm taking so long but I keep bumping into trouble that needs to be solved. I don't think any of this is necessary, ```DLNAImage.toDLNAImage()``` should do any conversion as needed.

@Sami32 When it comes to keeping alpha channel or not, "the way of #1061" is to let the renderer choose the format. Everything is offered in both JPEG and PNG. If the original is a (compliant) PNG the CI flag will be 0. An intelligent renderer should then choose this version. If that is what happens in every case is another thing, I doubt that all renderers are "intelligent", but that's how #1061 works.

Sami32:
@Nadahar Don't worry, i guess that serialization give you trouble; i just post here because i thought that perhaps it was more appropriate than post in in #1061 

I saw that you send in the 2 formats but i was just thinking about a configuration way to set the prefered transcoding format, not having too much hope in the "intelligent" TV.
That's said, perhaps the renderers will be smarter than me :wink:

Nadahar:
@Sami32 At the moment I'm not struggeling with serialization. That work isn't complete, but I think I know what I need to do to fix that now (I have a couple of alternatives in mind). 

Right now I'm struggeling with something completely different, and it goes right to the core of buggy UMS/PMS code - I've discovered that ```ProcessWrapperImpl``` is seriously bugged if run in the same thread. Basicly, it creates a fixed size buffer and then starts the external process (like DCRaw). It then expects another thread to process its buffer so that it doesn't overflow. When run in the same thread, there's no other thread to process its buffer, so if the output of the process is larger than the fixed sized buffer (500 kB) the thread will hang. Since DCRaw will produce images larger than that, it will hang. This is a general problem though, not only concerning DCRaw. Basicly ```ProcessWrapperImpl``` is only written with transcoding in mind where you have a consumer that always consumes the buffer. So, I have to figure out a proper solution to that other than just raising the buffer size.

When it comes to the intelligence of the renderers (or rather, the developers to its software) I don't have very high hopes either. I have a solution in mind, but it will need implementation and testing. My assumption is that the less intelligent renderers will simply pick the first resource they get that they can play. That means that if they can play both PNG and JPEG, they will probably pick whatever is presented first. An easy solution would be to always offer PNG first, but: Since UMS holds so much in memory (all thumbnails of resolved resources for example) and since I've seen that PNGs generally many times as big as the corresponding JPEG when held in memory, I don't think it's a good idea to always put PNG's first. My idea is that it should be dependent on the source. If the source is alpha capable (or even better, actually has alpha), PNGs should be offered first. If not, JPEGs should be offered first.

Its all guessing how this will work really, and it won't be the same from renderer to renderer, but I think that's something worth testing out. The less renderer configuration that is needed, the better as I see it. Even though it's nice with options and all, a huge problem is that most of the renderer profiles aren't correct, and we mostly only updates the profiles of the renderers we have (and can test). That leaves a huge amount of renderers with bad or missing profiles, and for every extra option this problem becomes a little bit bigger.

Sami32:
@Nadahar It could explain why the SRW give me "freezing"...
If i remember the last time i checked it, you told me that it was using the old Java 6 structure ? unsecure and "weak".
That's said i was not aware about this limitation.

I totally share that PNG should be used only if alpha channel is detected, saving memory and keeping transparency :+1: 

Sadly true, it's why a protocolInfo implementation, as @atamariya did, could be a good thing, IMO, as it should relieve users and devoloppers from many lost energy...

Sami32:
As a side note: the transcoding for progressive, alpha channel should be avoid for Web interfaces, and i guess for software renderers as well ? (Kodi...)

Sami32:
http://www.universalmediaserver.com/forum/viewtopic.php?f=4&t=9109&p=29230#p29230

valib:
Done much better in #1061

PR #1115, opened 2016-12-25
Update Maven dependencies

valib:


pom.xml
```diff
@@ -912,7 +912,7 @@
 					<!-- Plugin to assemble a jar with dependencies -->
 					<plugin>
 						<artifactId>maven-assembly-plugin</artifactId>
-						<version>2.6</version>
+						<version>3.0.0</version>
 							<configuration>
```
Nadahar: @valib A new major version might bring big changes. Have you checked what the impact is?

valib: @Nadahar tested without any problem.

SubJunk:
I can't remember if there was a reason for not updating the assembly plugin, but I guess we can find out

Nadahar:
Has anyone tried building for Linux and OS X?

valib:
@Nadahar not me I have only WIN OS.

Nadahar:
@valib There are multiple free virtualization solutions, so anyone that want to can test for other platforms just like I do. My development computer is also just Windows. The easiest to get up and running is probably [VMWare Workstation Player](http://www.vmware.com/products/player/playerpro-evaluation.html). The reason I say it's the easiest isn't because I think the others are difficult to install/configure, but because the VMWare products are so widely used that you can find almost any information you need very easily using a search engine. Even though VMWare tries as hard as they can to route you to the pay-version, if you click carefully you can get a fully working free version.

> The free version is available for non-commercial, personal and home use. We also encourage students and non-profit organizations to benefit from this offering.

In any case, I think it should be verified that this builds for other platforms as well.

Nadahar:
@SubJunk @valib I've tested this on Ubuntu 12.04 LTS and OS X 10.8.5 and it worked fine on both as far as I could tell. I build it with Maven, extracted the tar.gz on Linux and started the generated "packages" under target on both. 👍

SubJunk:
Awesome, thanks. Looks like you're good to merge @valib

PR #1116, opened 2016-12-26
Fixed some bugs already solved in #1061

valib:
- Fixed a bug in getFFmpegThumbnail() where thumbnail position could be after EOF
- Fixed some missing "thumbready = true"
- Fixed a bug where FFmpeg thumbnail generation would be attempted for non-video content
- Fixed missing media name for the TRACE log when resolving player


Nadahar:
@valib There's not much left until I'm done with #1061, so I'm not sure it's worth the extra merge trouble to merge this PR in the meantime..?

valib:
@Nadahar those bugs are so general that should be merged also to master doesn't matter if the #1061 PR will be merged or not.

PR #1120, opened 2016-12-30
Some fixes to TableFilesStatus

Nadahar:
I have some suggested changes - I'll make comments in the code about why.

### DIFF
--- a/src/main/java/net/pms/database/TableFilesStatus.java
+++ b/src/main/java/net/pms/database/TableFilesStatus.java
@@ -95,10 +95,10 @@ public final class TableFilesStatus extends Tables{
 			if (fileId == 0) {
 				query = "SELECT ID FROM FILES WHERE FILENAME = " + sqlQuote(fullPathToFile);
 				if (trace) {
-					LOGGER.trace("Searching for file with \"{}\" before update", query);
+					LOGGER.trace("Looking up file id for file \"{}\" with \"{}\" before update", fullPathToFile, query);
 				}
 
-				TABLE_LOCK.writeLock().lock();
+				TABLE_LOCK.readLock().lock();
 				try (Statement statement = connection.createStatement()) {
 					try (ResultSet result = statement.executeQuery(query)) {
 						if (result.next()) {
@@ -106,7 +106,7 @@ public final class TableFilesStatus extends Tables{
 						}
 					}
 				} finally {
-					TABLE_LOCK.writeLock().unlock();
+					TABLE_LOCK.readLock().unlock();
 				}
 			}
 
@@ -128,7 +128,7 @@ public final class TableFilesStatus extends Tables{
 							} else {
 								query = "UPDATE " + TABLE_NAME + " SET MODIFIED=CURRENT_TIMESTAMP, ISFULLYPLAYED=" + isFullyPlayed + " WHERE FILEID=" + fileId;
 								if (trace) {
-									LOGGER.trace("Found the file entry so we will update it with {}", query);
+									LOGGER.trace("Found the file entry so we will update it with \"{}\"", query);
 								}
 
 								Statement statement2 = connection.createStatement();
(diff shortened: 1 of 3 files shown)


src/main/java/net/pms/database/TableFilesStatus.java
```diff
@@ -95,18 +95,18 @@ public static void setFullyPlayed(final String fullPathToFile, final boolean isF
 			if (fileId == 0) {
 				query = "SELECT ID FROM FILES WHERE FILENAME = " + sqlQuote(fullPathToFile);
 				if (trace) {
-					LOGGER.trace("Searching for file with \"{}\" before update", query);
+					LOGGER.trace("Looking up file id for file \"{}\" with \"{}\" before update", fullPathToFile, query);
 				}
 
-				TABLE_LOCK.writeLock().lock();
+				TABLE_LOCK.readLock().lock();
 				try (Statement statement = connection.createStatement()) {
```
Nadahar: You should only use a write lock when you're actually going to write. The difference between a read and a write lock is that a write lock is exclusive (only one can be held at any time) while read locks can be held in parallel with each other and will always be granted unless a write lock is held. Acquiring a write lock blocks until all other read- and write locks are released, and will block any new read- or write locks from being aquired while held. That means that as long as you hold a write lock, you have exclusive access to that table, so it should only be used when needed.

src/main/java/net/pms/database/TableFilesStatus.java
```diff
@@ -181,11 +181,12 @@ public static void setDirectoryFullyPlayed(final String fullPathToFile, final bo
 			TABLE_LOCK.writeLock().lock();
 			try (Statement statement = connection.createStatement()) {
 				try (ResultSet result = statement.executeQuery(query)) {
-					TABLE_LOCK.writeLock().unlock();
 					while (result.next()) {
```
Nadahar: This is wrong, you should never release the lock while you're iterating a result. That means that other threads could alter the table in the meanwhile rendering the ResultSet you're iterating invalid. You want an operation like this to be "atomic" in the sense that the operation is done without concurrent changes by other parties, and thus the lock should be held from you start the operation until it's finished or fails. The lock is a ```ReentrantReadWriteLock``` meaning that it's ok for the same thread to aquire the lock recursively. As long as it already holds the lock, it will always be granted the recursive lock. I'm guessing this is the reason you thought you had to release it before aquiring it again in ```setFullyPlayed```. 

My "best practices" tells me that it's ok to aquire a read lock while holding a write lock, but not the other way. I never ask for a write lock while holding a read lock. 

When I think about it now, I'm not sure if that's actually necessary here - I'd think that aquiring a write lock while holding a read lock would just wait until all other read- and write locks were released and then proceed, but I'd have to make really sure that's the way it works before changing my "best practice". All I know is that I've been burned by doing that before, but that was in a different language with a different locking mechanism so it might not apply here. I guess I'll have to investigate that when the need to do that present itself, in the meantime I'll just avoid doing it.

src/main/java/net/pms/database/TableFilesStatus.java
```diff
@@ -203,7 +204,7 @@ public static FilesStatusResult isFullyPlayed(final String fullPathToFile) {
 		FilesStatusResult result;
 
 		try (Connection connection = database.getConnection()) {
-			String query = "SELECT ISFULLYPLAYED FROM FILES JOIN " + TABLE_NAME + " ON FILES.ID = " + TABLE_NAME + ".FILEID WHERE FILENAME = " + sqlQuote(sqlLikeEscape(fullPathToFile));
+			String query = "SELECT ISFULLYPLAYED FROM FILES JOIN " + TABLE_NAME + " ON FILES.ID = " + TABLE_NAME + ".FILEID WHERE FILENAME = " + sqlQuote(fullPathToFile);
```
Nadahar: "LikeEscape" should only be used with ```LIKE```, not when using ```=```. The escaped syntax isn't necessarily valid when using ```=```.

src/main/java/net/pms/database/TableFilesStatus.java
```diff
@@ -300,21 +301,20 @@ private static void upgradeTable(final Connection connection, final int currentV
...
+				"CREATE TABLE " + TABLE_NAME + "(" +
 					"ID            IDENTITY PRIMARY KEY, " +
 					"FILEID        INT      NOT NULL UNIQUE, " +
 					"MODIFIED      DATETIME, " +
 					"ISFULLYPLAYED BOOLEAN  DEFAULT false" +
 				")"
 			);
 
-			statement.execute("CREATE INDEX ID_IDX ON " + TABLE_NAME + "(ID)");
 			statement.execute("CREATE INDEX FILEID_IDX ON " + TABLE_NAME + "(FILEID)");
```
Nadahar: The primary key always has an "implied index" so there's no need to create one explicitly.

Nadahar:
@SubJunk That's entirely up to you - merge it if you want it. I haven't tested it, but I can't see how any of these changes could pose a problem.

PR #1125, opened 2017-01-04
Fixed regression in a4f837958f623c178d0132c3d24509e54f87c371

Nadahar:
This is a bit urgent, we should maybe consider to do a release soon because of this. I can't believe we haven't gotten more reports about this, but maybe we have?

Basicly, setting ```codecA``` or ```codecV``` to ```null``` causes a NPE after a4f837958f623c178d0132c3d24509e54f87c371.

This is the [source](http://www.universalmediaserver.com/forum/viewtopic.php?f=10&t=9086).

### DIFF
--- a/src/main/java/net/pms/dlna/DLNAMediaAudio.java
+++ b/src/main/java/net/pms/dlna/DLNAMediaAudio.java
@@ -18,6 +18,7 @@
  */
 package net.pms.dlna;
 
+import java.util.Locale;
 import net.pms.configuration.FormatConfiguration;
 import net.pms.formats.v2.AudioProperties;
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
@@ -412,7 +413,7 @@ public class DLNAMediaAudio extends DLNAMediaLang implements Cloneable {
 	 * @since 1.50
 	 */
 	public void setCodecA(String codecA) {
-		this.codecA = codecA.toLowerCase();
+		this.codecA = codecA != null ? codecA.toLowerCase(Locale.ROOT) : null;
 	}
 
 	/**
--- a/src/main/java/net/pms/dlna/DLNAMediaInfo.java
+++ b/src/main/java/net/pms/dlna/DLNAMediaInfo.java
@@ -865,7 +865,7 @@ public class DLNAMediaInfo implements Cloneable {
 						ffmpeg_parsing = true;
 					}
 				}
-				
+
 				if (configuration.getImageThumbnailsEnabled() && thumbOnly) {
 					LOGGER.trace("Creating (temporary) thumbnail: {}", file.getName());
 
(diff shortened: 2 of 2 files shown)


Nadahar:
@SubJunk You'll have to judge how critical this is, I really can't understand that this hasn't manifested itself sooner (but it probably has but haven't reached us).

SubJunk:
Cool I can do a release soon

SubJunk:
This kind of bug would be quite easy to prevent in the future with automated tests if anyone is interested :)

Nadahar:
@SubJunk I'm not so sure I agree. While it might be true in an ideal world, I think writing useful tests is hard enough when I have full understanding of the code. Writing useful tests for unstructured code where I don't have a chance to obtain the full overview is close to impossible.

I think anyone that sees something useful to test should add a test for that just like we add anything else.

SubJunk:
At my job we have about 80% coverage of the code by automated tests and code is never merged unless it has tests attached. It prevents so many regressions. I wouldn't make it mandatory here because no one would ever contribute code again but I do really encourage it.

In this case all the test would need to do is check that it's possible to pass `null` to `setCodecV` without an error occurring, which is one of the most simple types of tests there is.

Nadahar:
@SubJunk There might be something major I haven't understood about how to create tests, I don't have a lot of experience with the concept, but: I don't know if you're talking about unit tests or integration tests. I haven't really looked into integration tests as it seems like a massive undertaking for UMS. If it were there from the start and were expanded as things got added, it would possible be doable, but to start to write that now...? I have no idea how to.

Unit tests are hard to do for complex classes, but easy for simple ones. You have to create a lot of mock objects to be able to test many of the things in UMS. I suspect that many things would have to be refactored just to be testable. There's a lot of "unwise" use of state storage in UMS, a lot of static state for example. It can be very difficult or impossible to fake those. As I see it, time is spent more effectively by having good coding practices, getting an overview before doing changes and try to test things as you go along then trying to make some actually useful tests for UMS. One important thing you'd need to make any "real" tests were a mock/fake renderer that you could trust to be reasonably reliable. I doubt you'll find that, and I very much doubt it would be worth the time *writing* one.

While writing ```null``` tests is one of the things I actually find easy, there are hundreds if not thousands of ```null``` tests that would have to be written.

While encouraging is fine, I can't remember seeing you making a lot of tests for the code you submit either..? Either you actually agree with me that it's not worth the time most of the time when it comes down to it, or you should do so to inspire and show the rest of us how to do it.

Sami32:
Do anyone like classical music here :notes:

SubJunk:
I agree that I need to put more time into writing automated tests for this project :) and yes all we have right now is unit tests which aren't as good as integration tests, but still far better than nothing in my opinion.

There is pretty good testing around the filename prettifying feature which I wrote and it has saved me from pushing regressions to that feature a couple of times, so in that case it has already been very useful.

PR #1127, opened 2017-01-06
Add MacOS "Inhibit system sleep" hook while running

michaelt-ums:
PMS.java: add MacOS "inhibit system sleep" hook when running


michaelt-ums:
Oops - The head title of this pull request is misleading - it took the first file comment I entered a long time ago (probably the wrong way or time!)
It should be:  Add MacOS "Inhibit system sleep" hook while running

Nadahar:
@michaelt-ums Thank you for contributing 👍 

I have some questions:
* How is changing the required Java version related to running ```caffeinate```?
* Is ```caffeinate``` a standard command available on all OS X versions?
* Don't we need to stop "caffeinating" once playback is stopped to allow the system to go to sleep again?

michaelt-ums:
*** my email server is fighting me. not sure if you got it so here is a copy:

On 2017 Jan 06, at 3:09 PM, Nadar <notifications@github.com> wrote:

@michaelt-ums Thank you for contributing 👍

I have some questions:

**How is changing the required Java version related to running caffeinated?**
Nothing - as I tried to annotate - it was required to allow me to build. I run a system updated with all the latest software revs. I saw this as one of a few checks which has an upper bound. Didn’t make sense to me. Who knows how high the revs will go during the lifetime of a configuration. I don’t.

**Is caffeinate a standard command available on all OS X versions?**
No - not all, just 10.8+. Introduced in Mountain Lion (2012). Obscure but there (it’s what kept me looking for a viable solution. I‘ve been running 10.7 Lion until last Sep)

**Don't we need to stop "caffeinating" once playback is stopped to allow the system to go to sleep again?**
(***Note: if the system sleeps when waiting for another selection, it will break the connection. Not good. Any server, headless or not, would require a user to wake it if idle too long to reconnect. Disabling only when streaming is only a partial fix. )
Nope - that’s the good part: normally you use it like “exec”, just passing another command to ‘watch’. When it terminates, system settings are restored. It does not work for UMS since the java launcher throws PMS and it is PMS which we must wait for to die.
Another option allows handing it a PID to watch. I let PMS spawn it with success. 
I don’t much care for the name but it does exactly (almost) what we want. I know because it use to disrupt our hours of watching movie series. Now it allows the display to sleep while not having the system “alive” 24-7.

michaelt-ums:
++ suggest I add a check for 10.8+

Nadahar:
@michaelt-ums Adding a check for 10.8 doesn't really solve much, it's still only a solution for 10.8+. But, I guess it's better than nothing. 

About preventing sleep mode, people have different wishes. The way it works for Windows is that it only prevents sleep while something is being played back, and that's the way it has been designed to work. Many people actually want their computer to go to sleep if it's not used, but don't want it to go to sleep while they are watching or listening to something. As soon as they're finished playing the media, the computer can go to sleep like normal.

If somebody want the computer to always stay on (like mine does), it's just to disable sleep in the OS. In don't know about OS X, but in Windows you can configure this exactly as you want - you can make it still turn of the screen, spin down harddrives, slow down the CPU to a minimum, turn of power to add-on cards etc, but still "stay awake" and listen for connections. That way, harddrives will spin up and things will turn on when needed. You actually don't need UMS to do anything to achieve this behaviour.

Is there a way to get ```caffeinate``` to behave "the UMS way", that it is possible to "turn off" again? Binding it to the JVM's process or PID obviously won't work, and a Java program doesn't have "its own".

michaelt-ums:
> On 2017 Jan 07, at 4:16 PM, Nadar <notifications@github.com> wrote:
> @michaelt-ums <https://github.com/michaelt-ums> Adding a check for 10.8 doesn't really solve much, it's still only a solution for 10.8+. But, I guess it's better than nothing.
> 
agreed (Note: it is not a fix for 10.8, rather 10.8+. All releases since have it)
> About preventing sleep mode, people have different wishes. The way it works for Windows is that it only prevents sleep while something is being played back, and that's the way it has been designed to work.
> 
The Windows implementation is within the restrictions of the Windows systems. The Mac implementation is the same (as will be linux). 
The Windows functionality is not available. MacOS works by monitoring a process. 
> Many people actually want their computer to go to sleep if it's not used, but don't want it to go to sleep while they are watching or listening to something. As soon as they're finished playing the media, the computer can go to sleep like normal.
> 
The major problem here is not watching on a computer but on an external device. When the computer sleeps, it breaks the network connection. In my case: fatal for the TV (have to power off/on and/or relaunch UMS. Yup - it’s brain dead but it is what I got). 
So this leaves the Mac user the choice of messing with System Preferences before and after, or this fix. 
Been running for over a week. Love it. Eight parts of Poldark without an interruption or hiccup! And I can let the system limit power in a few minutes, not many hours.
> If somebody want the computer to always stay on (like mine does), it's just to disable sleep in the OS. In don't know about OS X, but in Windows you can configure this exactly as you want - you can make it still turn of the screen, spin down harddrives, slow down the CPU to a minimum, turn of power to add-on cards etc, but still "stay awake" and listen for connections. That way, harddrives will spin up and things will turn on when needed. You actually don't need UMS to do anything to achieve this behaviour.
> 
(My systems are off only when the power goes off)
This does not keep everything powered up, just the core + active disk(s). Displays, unused discs, et.al. go to sleep.
Macs are a little more coarse in their power handling. All my drives and displays still sleep but the cpu is alive to keep the DLNA connection serviced. 
Server problem ? Client problem? Don’t know. 
Like the renderers, we have to deal it as best we can with each of their idiosyncrasies. No different than my vehicles.
> Is there a way to get caffeinate to behave "the UMS way", that it is possible to "turn off" again? Binding it to the JVM's process or PID obviously won't work,
> 
No - it reverts to the system settings when the process terminates. 
What is the “UMS way”? From what I researched it looks like it is simply a Windows implementation. 
> and a Java program doesn't have "its own”.
> 
It kinda does - just got to look for it. I prototyped this in a shell script before I found the system function (much simpler).


Your third question prompted me to relook at the approach I took to Implement this. Very limited, simple. (I’m new to java)
I want to withdraw this request and submit another, expanded one shortly. 
The new one makes the sleep optional for all platforms via the “General Configuration” tab, all in the same way. It is also OS revision aware for Macs. And enhancements for Linux and others would be easy to do in the future.

Nadahar:
You can either close this and create a new, or you can simply push new commits to your fork and this PR will be updated.

When it comes to the way it works today, it's not a "Windows implementation" - it's simply only implemented on Windows. You can make it work both ways in Windows, it used to prevent sleep as long as UMS was running but this was changed because people requested if (if I remember correctly). When it comes to OS X the missing implementation is simply because non of the developers use OS X and that Apple are control freaks and generally don't want the users to decide anything making it even less tempting to try an implementation. There is nothing "Windows specific" about the current implementation. Ideally the user could choose which way it should work.

I doubt we will ever see an implementation for Linux, atleast not until Linux get their act together and standardize such basic functions. As it is now, you'd probably have to make one solution per graphics manager per distro. It's simply a mess to try to get anything like this working on Linux.

michaelt-ums:
Finally(?) ready. 
Made changes to check Mac rev valid, retained previous Windows look & feel, simplified.

The upstream changes forced me into a Merge/Rebase resulting in a complete disaster. I had to purge my local project, re-clone, re-edit, test and all is well. This is the push I successfully(?) did.

If I messed up the branch here - tell me how to back it out & start over. I am not comfortable yet with this process.
I have backups of all the changes.

Nadahar:
@michaelt-ums I'm not convinced this functionality is what the others want, as I've tried to tell you. We'll just have to see what their response is.

You very rarely actually loose anything with git, and you never have to re-clone unless your git repo is somehow corrupted (I've never experienced that). If you're getting yourself into a merge or rebase that you can't cope with, you can always abort (e.g ```git rebase --abort```). I'm assuming that what you had was a merge-conflict, and those are not that uncommon and is perfectly solvable. It's too much to explain here, but any search engine will offer you loads of help on that.

The POM change needs to go as I've said before, and I'm sure there's some other small things as well. But, you don't have to worry about that, I can do a little rebase on this if the general functionality is accepted.

@UniversalMediaServer/developers What do you think? This uses a different "prevent-sleep" functionality on OS X than on Windows. On Windows sleep is prevented only when serving media (unless I'm completely confused) but this solution prevents sleep as long as UMS is running. I'm afraid it will be completely confusing to users, and I'm not sure that's the best solution for most users. The ideal solution would be to offer both variants, e.g. a combobox instead of a checkbox where you could choose between:
* Don't prevent sleep
* Prevent sleep during playback
* Prevent sleep while UMS is running

It's hard for me to judge because I would never use such functionality, it's so easy to just turn off sleep in the OS settings. A server should never sleep anyway imo. Do you want this or not? We need to do some "touch ups" to this in any case, the change in the POM must be removed and I think I saw some formatting issue up there as well. I haven't looked at how it works as I don't really know how the Windows variant work, so it might need some touches there as well.

Sami32:
I use OS setting as well, but i guess that at changing the way how it is now, it'll be not be worst adding more options choice as you suggested.

That's said, as i saw many Mac OS users on the forum complaining about not being able to see the series/movies without sleep state interruption, this issue should be addressed, at least for Mac users, IMO.

That's was my 2 cents...

valib:
If I read the code correctly it does nothing to the current Windows prevent sleep implementation which is by the way not working correctly. There is my PR which tries to fix it. 
This code only affects the OS X.

Nadahar:
@valib Yes, I'm not saying it changes the Windows implementation, I'm saying that it works in another way and wondering if that's ok.

Sami32:
For Mac OS noobs like me:

http://www.universalmediaserver.com/forum/viewtopic.php?f=11&t=1714&start=20

https://github.com/newmarcel/KeepingYouAwake
https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man8/caffeinate.8.html
http://apple.stackexchange.com/questions/148206/how-do-i-stop-my-mac-os-x-from-going-to-sleep
https://support.apple.com/kb/PH13808?locale=fr_FR
http://osxdaily.com/2014/06/16/caffeinate-prevent-sleep-while-command-active-mac-os-x/

http://alternativeto.net/software/amphetamine/?platform=mac

michaelt-ums:
@Sami32 - interesting links. Conclusion: same = sleep inhibit on a per process only. This is what I found and is the only solution. Not doing anything is not an option when there is an acceptable implementation, which is optional and stays within the system API. Bad things happen when you break into back doors.

On another note: yes I can enable/disable the inhibit around the streaming but it does not stop the breaking of the DLNA connection. Ugly. Another possibility is to wrap the "session", allowing the server to sleep upon client exit. But the server initiates the connection not the client so you have no way to wake up the server. Catch-22.

As for the ' not like Windows' -  of course, it's a Mac. Some of us don't care. If we did, we would get a PC. 
Kinda like trying to make a Mustang (car) look like 4X4 F150 (truck, 4 wheel drive). They are different platforms - you can't 'make them look or act the same.

Nadahar:
@michaelt-ums I think your argumentation is invalid. This has nothing to do with Windows/Mac, it has to do with what behaviour the users want. Some people want the computer to go to sleep when it's not being used, but not during playback. I can see how this can be a relevant wish. Others, you in particular, want the computer to stay awake as long as UMS is running, so that it is available to deliver new content at any time. I'd say that this scenario would be easy to achieve by simply disabling sleep in the OS, but that doesn't seem to be a viable solution for you. I think the best solution would be to offer both alternatives, although I have a bit of a hard time to understanding why anyone would enable sleep in the OS on a computer they want to stay available.

Sami32:
Some users request and comments:
http://www.universalmediaserver.com/forum/viewtopic.php?f=11&t=1714&p=11596&hilit=mac+system+sleep#p11596
> OS X System Preferences/Energy Saver > Computer Sleep: Never

Nadahar:
The information from the linked thread is new to me. From what I can read there, UMS becomes inactive as soon as the screen shuts of..? This sounds very strange to me, it's hard to see how that could impact UMS' operation - but you never know with Apple I guess. Is that actually the problem (not preventing the computer from sleeping but making UMS work despite the screen being shut off)?

michaelt-ums:
@Nadahar  - 
You are right:  
> don't know about OS X
> I would never use such functionality
> I don't really know how the Windows variant work
> 

False:
> UMS becomes inactive as soon as the screen shuts off

and: 

> "A server should never sleep anyway imo"

 - is exactly what this does for Mac ONLY.

> Others, you in particular

This is not about what I want - it is about what I could do. 
What I want is still in my sandbox.

I am not going to get into the Mac-PC war of words and ideas.
If you don't know how un_x works,  please stop trying to force feed Windows ideas and operations on it. It is obvious you have a low opinion of Apple, don't know how Mac's work and don't care to.

For a PC user it is easy to use the system OS Sleep/No Sleep. Most (all?) users run with admin privileges.
On uni_x platforms we separate admin from user, and most (all?) non-PC users operate this way. So, to use the OS On/Off model, what you want means we must be, or authenticate as, an admin everytime we want to switch - unacceptable when there is a 'better', benign change which does not affect non-Mac users.

This submission had a dual purpose: 
1. process: the process was new, very different from what I am used to, but workable. 
2. management: I find restrictive, closed-minded and offensive. Those who "don't know" should not telling others how it should work and drop insults.

On a personal note: I spent many years running around the bowels of MS-DOS, Windows, and many un_x systems. I know them very well. I am a kernel hack. I run and use un*x based systems for a good reason. I walked away from Windows for a good reason. Don't try to force-feed me. 
Due to this attitude and a few more, I left the corporate predatory environment. 
I expected this to be an "open" development, with "open" minds. I guess not.
I did not want to get personal but you left me no choice.
Lastly: this is my opinion: not right, not wrong - just mine.

This is a useless time-sink. I choose not to continue. I'll spend my time in my sandbox.
Based on what I have experienced, please take down this request for change. 
I will not be submitting any more of the changes I have. (This includes Linux)
I will not lay them out for others to reject now, then later submit and take credit.

I am very disappointed.

Nadahar:
@michaelt-ums I don't know where to start, but you're barking up the wrong tree. I have from the start iterated that this isn't about the OS, it's about what users want/need. You have repeated your argument that what goes on Windows doesn't apply on OS X (now you've actually expanded it to include *nix, but it's used to be about OS X). I disagree with this, and that seems to annoy you a great deal.

I'm not a native English speaker, so I might come of rude without knowing it, but I don't get why you react as you do. When I said "your argumentation is invalid" I didn't mean anything other than that you keep focusing on Windows vs others despite my stance that it's not what it's about. I assume that's what got you so upset. I don't know how to put it in any other way, but I really don't think the OS has anything to do with it. I think it's we're talking about two different implementations of "prevent sleep" where one will be useful for some users and the other for other users. I'm not saying which one is "the best", I'm just stating that one is very easy to work around as you can just turn off sleep in the OS - while the other isn't easily achievable without UMS interacting with the sleep mechanism.

My main point the whole time is that I think UMS should be consistent across platforms. I think it's a very bad idea, that will cause much confusion, to make one option do two completely different things depending on the OS. As I see it, UMS should either prevent sleep while it's running on both Windows and OS X, or it should prevent sleep while playing on both platforms. As I've said before, I think the ideal solution would be to have both options. My opinion on this stops there.

There's a number of facts you seem to be unaware of or are wrong on. Most important, I'm not the one that decides what will be merged and not or how something should work. What you've seen is an attempt at a discussion, not me dictating how things should be. Decisions are made collaboratively where @SubJunk has the final word.

In addition, you can't make something and get insulted if a project doesn't want your changes (I'm not saying that's the case here). I currently have two PRs I've made for another project waiting for approval, and it seems like they will both be rejected. In fact, almost anything I've done has been outside UMS, and my initial PR for UMS was also rejected. While I think it would be good if some projects were a little more open to outside ideas, you must understand that it has to be some form of process where somebody decides what will be accepted and not. If not, nothing would work as people with all kind of knowledge and ideas would all contribute to a complete chaos. I've put down a lot of hours for things that's never come to anything, and I'm trying to tell myself that I should ask the project if they're interested in a given functionality before I start making it.

When it comes to your assumptions about me not knowing anything about *nix and is a Windows fanboy you're simply dead wrong. I think it's safe to say that I'm the one in this project that cares most about how UMS works on OS X and Linux. At home I have (both physical and virtual computers):
* 5 x Windows 7
* 2 x Windows XP
* 1 x OS X 10.8.5
* 2 x FreeBSD
* 2 x ESXi
* 2 x Fedora 25
* 5 x Centos 7
* 2 x Debian 8
* 2 x Ubuntu 16
* 1 x Ubuntu 12
* 1 x Antergos
* 1 x Windows 10
* 2 x Windows 2008
* 1 x Kubuntu 16

It's simply wrong that I nothing about OS X and Linux. I'm not what I would consider an expert at any of these, but I do find my way around and run a lot of different software on Linux. I have no idea where you got the idea that I don't know anything about these OSes from. It's true that I dislike Apple strongly, and I really don't understand why anyone choose to spend their money on Apple products. I dislike Apple because I think they are one of the most unethical companies I know. I dislike their products because they are hyped, overrated, dumbed down and crippled by Apple's "we know best" attitude. But, I don't like Windows very much more. I think Windows 2000 was a great OS, and I think it's mostly gone down hill since then. Microsoft has increasingly copied a lot of Apple's crappy "business practices" in later years and I don't see me upgrading from Windows 7 as my "main platform" probably ever. The day Windows 7 can't fill my needs, I'll probably have to pick one of the Linux distros - but I hope they get their act together when it comes to standardization before that day comes. 

It seems to me like your experience with Windows is quite old. "Everyone" being administrator stopped being the case after Windows XP when they introduced UAC (User Access Control) with Windows Vista. UAC is pretty much exactly the same as how it works in OS X or modern Linux'es - you have to confirm elevation to make any system changes. But crucially, you can **turn it off**. I don't know if you can do the same in OS X (you can in Linux), but it would be very unlike Apple to give the user the option to choose. After all, the user has only bought the hardware and the OS - it's only reasonable that Apple decides how they should use it.

I am tired of trying to circumvent Apple's attempts at blocking any kind of freedom of choice, yes. I've been down that road some times, and I always end up thinking: Why do people choose this voluntarily? I'm not overly eager at doing one of these journeys into the world of unnecessary limitations, but that doesn't mean that I will try to deny anyone else the pleasure.

This has never been about me being against a solution for OS X. *I* didn't write the current Windows only implementation. If I wrote the current implementation I can assure you that I would have gone to great lengths to try to make the solution as universal as possible. I think a solution for OS X would be a good thing. The only argument I've made is that of consistency - I think the functionality should be as consistant as possible across platforms. But then again, it's not up to me, I'm simply presenting my opinion.

I think it's sad if you won't make any more contributions (to UMS or any other software) because I have a different view than you on one particular issue. But in the end, that's entirely up to you.

Sami32:
https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1226

Nadahar:
@michaelt-ums I've finally gotten around to implement "prevent sleep" for OS X. Here's a download link for [OS X (Java 8)](http://superbum.org/downloads/UMS_Sleep_Test.dmg) if you (or anyone else) is interested in testing it. Feedback can be given here.

Nadahar:
I recommend merging #1226 instead as it addresses both Windows and OS X and give the users the option to choose behaviour.

Sami32:
@michaelt-ums It should have been already solved by @Nadahar PR.

Feel free to re-open it if you think it need to.

PR #1129, opened 2017-01-07
Fix UPnP controls for Viera E6 series

Sami32:
I don't know if it's the way to go, in the mean that other Viera model could be unfairly concerned as well, so any comments are welcome...
Perhaps only "banned" the Viera that we are sure about ?


SubJunk:
I suggest we do https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1132 instead and we can then add a renderer configuration setting for broken ones if we need to

PR #1132, opened 2017-01-07
Reverted the manual disabling of UPnP control support on Panasonic TVs

SubJunk:


Nadahar:
Argh, how did this end up having two merge commits in it? It makes such a mess...

Sami32:
@Nadahar Sorry :flushed:  i rebased and merged from the Github interface and didn't saw that before you comment...
Do you want i revert it ?

Nadahar:
@Sami32 No... my comment simply meant that I don't think "hacks" like hardcoding part of a renderer string is a good thing to have in the code - I welcome that it goes away. Another thing is that I doubt that it actually doesn't work with the Vieras in question, we're not exactly following the standard very closely so there is many things that could break stuff for renderers that for some reason are extra picky about some detail we don't fulfill. This was probably just an "easy way" to make a problem "go away". I much prefer to find the actual reason why something doesn't work than to do things like what was done here.

Nadahar:
@Sami32 😆 

Now I understand - you were referring to the message I deleted. I deleted it because I saw that you had used rebase, not merge 👍 That means that the merge commits were removed and my comment was irrelevant 😉

Sami32:
@Nadahar Yes. I was afraid of doing something wrong.
Don't worry, i listen you, so now i rebase everything :wink: 

Totally agree with you about the code, but i didn't want offence anyone, so i just modified the existing one to make my renderer able to use it, and asked about that choice. As a newbie it seem fair to me

Nadahar:
@Sami32 I always seem to offend people whatever I do, so I just say what I think. It doesn't always go down well with people, but at least it avoids a lot of complications 😉

PR #1136, opened 2017-01-08
Fix bitrate logging unit info omission

Sami32:
Bitrate unit was omited in that commit https://github.com/UniversalMediaServer/UniversalMediaServer/commit/078c3055869a1ddb215386c2f31accb29978822d

### DIFF
--- a/src/main/java/net/pms/encoders/FFMpegVideo.java
+++ b/src/main/java/net/pms/encoders/FFMpegVideo.java
@@ -509,7 +509,7 @@ public class FFMpegVideo extends Player {
 					defaultMaxBitrates[0] = 3000;
 				}
 
-				LOGGER.trace("Adjusting the video bitrate limit to " + defaultMaxBitrates[0] + " to make room for audio");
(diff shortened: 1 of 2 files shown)


Nadahar:
@Sami32 I haven't checked that this actually specifies kilo/kibi, but given that's the case, shouldn't it rather be ```kb/s``` or ```Kibit/s``` depending on what the number actually shows.

Sami32:
@Nadahar Right, i was too lazzy, but kbps or kbit/s should be possible as well ;)
Done.

Nadahar:
@Sami32 I usually prefer ```kbps``` but I'm not sure if that's technically "correct". Anyway, when we first supply units they should be correct 😉

Nadahar:
@Sami32 I still don't understand why you do one commit per change. I squashed them for you.

Sami32:
@Nadahar Because it's the only way i found in the Github interface.
But as i already told you, i rebase before merge so i ti's really an issue, right ?

Nadahar:
@Sami32 Rebase doesn't remove those, only the merge commits. You can use "squash and merge" to squash them, but then everything becomes one commit. In this case it would have been ok, but often that's not what you want.

PR #1137, opened 2017-01-09
Sony BRAVIA XD 70/75/80/83/85/93/94 series renderer

Sami32:
Following this issue http://www.universalmediaserver.com/forum/viewtopic.php?f=3&t=9105#p29245 reported by AlGrenadine.


src/main/external-resources/renderers/Sony-BraviaXD.conf
```diff
@@ -0,0 +1,63 @@
...
+" X-AV-Client-Info: av=5.0; cn="Sony Corporation"; mn="BRAVIA KD-65XD7505"; mv="3.0";
+# friendlyName=KD-65XD7505
+# udn=uuid:48454d48-01af-4343-ad82-70e21eb9dee1
+# manufacturer=Sony Corporation
+# modelName=KD-65XD7505
+# modelDescription=BRAVIA
+# manufacturerURL=http://www.sony.net/
+# ===================================================================================
+# https://www.sony.fr/electronics/televiseurs/xd7505-xd7005-series/specifications
+
```
valib: @Sami32 I recommend to add
```
UserAgentAdditionalHeader = X-AV-Client-Info
UserAgentAdditionalHeaderSearch = KD-\\d{2}XD7[0,5]
```

Sami32: @valib I choose to delete them because i feel like it is unnecessary recurrent, and i suspect the people to make copy from the renderer from the old time were http header parser was weak.

Do you have some reason for adding them ?

valib: @Sami32 there could be some network setting restrictions which prevent the DLNA working like I have. This should be much safer way to recognise the renderer when only the DLNA is working.

src/main/external-resources/renderers/Sony-BraviaXD.conf
```diff
@@ -0,0 +1,63 @@
...
+UpnpDetailsSearch = BRAVIA KD-\\d{2}XD7[0,5]
```
valib: @Sami32 this could be also changed to `KD-[0-9]{2}XD[0-9]{4}` to include also the new XD9 series

valib:
@Nadahar I am not an expert of creating the image so feel free to make it better.

Nadahar:
@valib I'm no expert either, but if I'm going to try to remove the white I'll need a slightly higer quality image to start with. What model is used in the pictures? To me the two doesn't even seem to be of the same TV...?

Sami32:
Do we really need to add one more icon instead of the one of the X model already here ?

Nadahar:
I've rebased your commit bonanza and added a more compliant renderer icon.

SubJunk:
Looks good to me

Nadahar:
@Sami32 @valib There is no "DLNA" and "UPnP" level. DLNA is a subset of UPnP.

Sami32:
@Nadahar I guess i'll need to deal with an other more renderer icon :wink:

Sami32:
As related [here](http://www.universalmediaserver.com/forum/viewtopic.php?p=29293#p29293), the user have a better result without the following lines:
```
UserAgentAdditionalHeader = X-AV-Client-Info
UserAgentAdditionalHeaderSearch = BRAVIA KD-\\d{2}XD[7,8,9]
```
So, i become again more suspicious about the necessity of these lines...
Perhaps i should remark them, for the ones with network issues.

valib:
@Sami32 those two lines have nothing to do with the better result the user reported because there are used only for renderer recognition. I am not sure what he was actually reported when the UMS never recognises his TV properly. Because the TV is Android based the UMS incorrectly uses the Google-Android-Chromecast.conf not the confs we provided. Reported  [here](http://www.universalmediaserver.com/forum/viewtopic.php?f=5&t=9035&start=10).

Sami32:
@valib I guess you checked his olg log.
I asked him already to shutdown Chromecast, like write in the renderer configuration file, and it solved his connection issue as expected.

AFTER that, he tried the 2 different renderers and the one without these 2 lines was better for him, if i believe his word.
And not, i didn't paid him :wink:

valib:
@Sami32 to shutdown Chromecast engine doesn't solve the problem. I checked the log he provided after your reccommendation and as you can see the Google-Android-Chromecast.conf is still used.

Sami32:
@valib OUPS, my bad...his last log with the other configuration without the "extra" lines was not posted yet, but i think he will do tomorrow, and it will be correctly recognized ?

When i saw his log, i suspected these lines to be the origin of the bad renderer recognition, but i could be wrong though...let's see tomorrow.

PR #1141, opened 2017-01-11
YAMAHA RX-A2050 conf

Sami32:
As reported by markoma here:
http://www.universalmediaserver.com/forum/viewtopic.php?p=29287#p29287
Following his log:
http://www.universalmediaserver.com/forum/download/file.php?id=3564

Do we really need to add one more icon ?
Personnaly i don't saw the difference in the 2 Yamaha renderers photographies...


Nadahar:
Shouldn't all new renderer configurations use MediaInfo?

Sami32:
@Nadahar Did i missed that DSD format was supported other way than that one ?

Nadahar:
@Sami32 Not that I know of, but there's a lot of "recent" code that assumes that MediaInfo is used so I'm betting a lot of things won't work very well without MediaInfo. Are you sure DSD support is that important..?

Sami32:
@Nadahar For this kind of audio users, yes, i think so.

I saw that @atamariya fork support it.

Nadahar:
@Sami32 Does @atamariya's fork support it using MediaInfo? 

I may be completely out of sync with the world, but isn't SACD a giant flop? I've never seen a SACD myself, and suspect that only a very few people own or use them.

Sami32:
@Nadahar I don't think so, but i'm not sure.

I never saw or tried myself, but my purpose is only trying to give to the users the possibility of such streaming on their renderers able to support such format.

I saw already on the forum and Github some requests for DSD support.

Nadahar:
@Sami32 I've seen those too, but it's not very difficult to convert DSD files into FLAC (manually) - there are free software out there that does it. Either that or the few that needs it could use a renderer configuration without MediaInfo. I'm just saying that I'm not sure all the other disadvantages of not using MediaInfo is worth it.

Sami32:
@Nadahar Which disavantages are you thinking about ?

Nadahar:
@Sami32 Parsing without MediaInfo is the "old" original PMS parsing. Everything done in recent years has been aimed at "the MediaInfo way". Parsing is done completely different, it's mostly done by completely different methods and logic, so the results will be very different. I don't remember which exact things go wrong, but I remember that I had a lot of issues until I understood that I had to rewrite my renderer profiles to use MediaInfo.

It's not that parsing without MediaInfo itself should be such a big disadvantage, at least for music. But, because it has been "silently deprecated" for years, a lot of changes have been made without taking non-MediaInfo parsing into consideration, often breaking things. 

There's also the obvious: You don't have the ability to specify the "Supported" lines so you will never get what should be streamed and what should be transcoded right. Parsing without MediaInfo only looks at the file extension, and that's it.

Sami32:
@Nadahar I believe to think this it's the right for it.
I didn't listen any audio users complain about that...that's only my opinion though...

Nadahar:
@Sami32 It certainly wasn't true for my Onkyo. I had all kind of issues when using the renderer profile currently bundled (it's not using MediaInfo). I have my own "custom" renderer configuration with MediaInfo and it works a lot better. It's not perfect though, so I've never added it to UMS. I still have some files that won't play properly, but I haven't looked at that for ages so I don't remember the details now.

valib:
@Sami32 have you seen @atamariya fork working? Me not. It seems interesting but until the problem with `ehcache` implementation will be solved it is not usable. I made some recommendation but there is not respond from him since 1 Nov 2016.

Sami32:
@valib Sadly not for me, too much CPU or memory intense, i suspect some memory leaks as well, but i saw some users on the forum that tried it with succes and was happy with it. But look that's work only on some specific PC configuration :-(

Yes, i know...maybe he don't have time or he don't want, i really don't know. I only hope that he didn't misunderstood us, as we all agree that he did a very good work, but it's only cannot be merged as it is, IMO.
I totally agree `ehcache` will a great improvement, all these DLNA enhancement as well !!!

Sami32:
@Nadahar Let's forget about Onkyo users, as for them nothing work :stuck_out_tongue_winking_eye:

Nadahar:
@Sami32 It wasn't an assertion, it was a question: 
> Shouldn't all new renderer configurations use MediaInfo?

I thought that was a "rule" so to say, that it's been a goal to get everything over to MediaInfo since 2014 or there about.

atamariya:
> Parsing without MediaInfo only looks at the file extension, and that's it.
@Nadahar That's incorrect assumption - the fallback is ffmpeg parsing which is crude but works well for a far wider variety of formats. It also means if the parsing is successful, media is playable. However, it does not generate rich meta info as well as mediainfo does.

Nadahar:
@atamariya I know that FFmpeg is used for parsing, I was talking about how UMS decides what to stream and what to transcode. Without MediaInfo that's only based on the extension.

atamariya:
@Nadahar Doesn't look like mediainfo plays a role in that decision. That happens when DIDL is generated based on config.
`if (media.isVideo()) {
						if (mediaRenderer.isTranscodeToMPEGTS()) {
							transcodedExtension = "_transcoded_to.ts";
						} else if (mediaRenderer.isTranscodeToWMV() && !xbox360) {
							transcodedExtension = "_transcoded_to.wmv";
						} else {
							transcodedExtension = "_transcoded_to.mpg";
						}
					} `

Nadahar:
@atamariya I don't mean which format to transcode to but which formats to transcode. When MediaInfo is used, the whole ```supported =``` system is used which looks at codecs, bitrate, channels etc. Without MediaInfo, only the file extension (called ```Format``` by UMS) is evaluated.

valib:
@Sami32 agree with @Nadahar that using `MediaInfo=true` and `supported` lines will be better and in the accordace with the rules the @SubJunk set.

Sami32:
@valib Is that not already done ?

valib:
@Sami32 sorry the changes are so fast that I don't follow them 😄

src/main/external-resources/renderers/Yamaha-RXA2050.conf
```diff
@@ -0,0 +1,39 @@
...
+# =============================================
+# http://usa.yamaha.com/products/audio-visual/aventage/rx-a2050_black_u/
+
+UpnpDetailsSearch = RX-A2050
+
+Video = false
+Image = false
+SeekByTime = true
+TranscodeAudio = WAV
+StreamExtensions = dff,dsf,l16,lpcm
```
valib: When you use the MediaInfo=true than this setting is overided and should be deleted.

Sami32: @valib Are you sure about that ?
Because when i looked at the code, as i reported in a commit, i didn't understood that.
https://github.com/UniversalMediaServer/UniversalMediaServer/blob/978bd1584cbce21989bb8c89732442ae81355d91/src/main/java/net/pms/configuration/RendererConfiguration.java#L2210-L2215

valib: @Sami32 if the "Supported" lines are correct than this code is never used. I think that is better to set the supported media correctly than to make an backup which can hide the not proper setting.

Sami32: @valib I would like, if you teach me how to make MediaInfo parse these formats.l16, .lpcm ...formats ?

Sami32:
Now, if someone would be kind enought to delete the not used anymore `yamaharxa1010.png` file it'll be nice, as i don't know yet how to do that...

Nadahar:
@Sami32 ```yamaharxa1010.png``` is used by ```Yamaha-RXA1010.conf```.

Nadahar:
"Someone" have done their magic...

PR #1142, opened 2017-01-12
Fix my own duplicated comments...

Sami32:


src/main/external-resources/UMS.conf
```diff
@@ -803,26 +803,6 @@ use_embedded_subtitles_style =
...
-# The maximum number of live subtitles results to show.
-# Default: 20
-live_subtitles_limit =
-
-# Keep live subtitles
-# -------------------
-# Whether to keep downloaded live subtitles permanently.
-# See live_subtitles_timeout below.
-# Default: false
-live_subtitles_keep =
```
SubJunk: So we don't want `live_subtitles_keep` and `live_subtitles_limit` anymore at all?

Sami32: @SubJunk Of course we do ;-)

These settings comments are already write there:
https://github.com/UniversalMediaServer/UniversalMediaServer/blob/03448e7321761fcc67eaf350fa5ab230162af9eb/src/main/external-resources/UMS.conf#L1414-L1436

SubJunk:
:+1:

PR #1144, opened 2017-01-13
Update Sony-BraviaNX800.conf

Sami32:
One of the mistake i have done that nobody seen as well. I did get confused with the regex and specific Java handling way.


Sami32:
@Nadahar I saw that you got confused, as everyone, as nobody told me that it was wrong in my Sony Bravia renderer configuration that i need to modify now.
Yes, it's tested, and @valib can confirm that, as i used a user for testing it.

As a side note, you will observe that it's the only one like that, sadly the one i did my copy past ;-)

Issue #1150, opened 2017-01-18
defaultMaxBitrates = rendererMaxBitrates ???


Sami32:
When i saw that in a log:
> Using the video bitrate limit from the renderer config (25) which is lower than the one from the program settings (25)

I found it not correct. Why `defaultMaxBitrates = rendererMaxBitrates` was set before logger ?

```
if (rendererMaxBitrates[0] > 0 && rendererMaxBitrates[0] < defaultMaxBitrates[0]) {
			defaultMaxBitrates = rendererMaxBitrates;
			LOGGER.trace("Using the video bitrate limit from the renderer config (" + rendererMaxBitrates[0] + "Mb/s) which is lower than the one from the program settings (" + defaultMaxBitrates[0] + ")");
		} else {
			LOGGER.trace("Using the video bitrate limit from the program settings (" + defaultMaxBitrates[0] + "Mb/s)");
}
```

valib:
@Sami32 fixed.

PR #1151, opened 2017-01-18
Fix bitrate in log bis

Sami32:
I guess that i never saw the end of the lines...

https://github.com/UniversalMediaServer/UniversalMediaServer/issues/1150

### DIFF
--- a/src/main/java/net/pms/encoders/FFMpegVideo.java
+++ b/src/main/java/net/pms/encoders/FFMpegVideo.java
@@ -193,7 +193,7 @@ public class FFMpegVideo extends Player {
 					}
 				} else if (params.sid.isExternal()) {
 					if (params.sid.isStreamable() && renderer.streamSubsForTranscodedVideo()) { // when subs are streamable do not transcode them
-						originalSubsFilename = null; 
+						originalSubsFilename = null;
 					} else {
 						originalSubsFilename = params.sid.getExternalFile().getAbsolutePath();
 					}
@@ -333,7 +333,7 @@ public class FFMpegVideo extends Player {
 				params.aid.isDTS() &&
 				!avisynth() &&
 				renderer.isDTSPlayable();
-			
+
 			boolean isSubtitlesAndTimeseek = !isDisableSubtitles(params) && params.timeseek > 0;
 
 			if (configuration.isAudioRemuxAC3() && params.aid != null && params.aid.isAC3() && !avisynth() && renderer.isTranscodeToAC3() && !isSubtitlesAndTimeseek) {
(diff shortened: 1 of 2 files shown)


src/main/java/net/pms/encoders/FFMpegVideo.java
```diff
@@ -431,10 +431,9 @@ public FFMpegVideo(PmsConfiguration configuration) {
 
 		// Give priority to the renderer's maximum bitrate setting over the user's setting
 		if (rendererMaxBitrates[0] > 0 && rendererMaxBitrates[0] < defaultMaxBitrates[0]) {
-			defaultMaxBitrates = rendererMaxBitrates;
-			LOGGER.trace("Using the video bitrate limit from the renderer config (" + rendererMaxBitrates[0] + "Mb/s) which is lower than the one from the program settings (" + defaultMaxBitrates[0] + ")");
+			LOGGER.trace("Using the video bitrate limit from the renderer config (" + rendererMaxBitrates[0] + " Mb/s) which is lower than the one from the program settings (" + defaultMaxBitrates[0] + " Mb/s)");
```
Nadahar: @Sami32 Where did ```defaultMaxBitrates = rendererMaxBitrates;``` go? Moving it after the logger is ok, but removing it is not. That will probably mean something won't work as expected, ie. a bug.

src/main/java/net/pms/encoders/MEncoderVideo.java
```diff
@@ -693,10 +693,9 @@ private String addMaximumBitrateConstraints(String encodeSettings, DLNAMediaInfo
 
 		// Give priority to the renderer's maximum bitrate setting over the user's setting
 		if (rendererMaxBitrates[0] > 0 && rendererMaxBitrates[0] < defaultMaxBitrates[0]) {
-			defaultMaxBitrates = rendererMaxBitrates;
-			LOGGER.trace("Using the video bitrate limit from the renderer config (" + rendererMaxBitrates[0] + "Mb/s) which is lower than the one from the program settings (" + defaultMaxBitrates[0] + ")");
+			LOGGER.trace("Using the video bitrate limit from the renderer config (" + rendererMaxBitrates[0] + " Mb/s) which is lower than the one from the program settings (" + defaultMaxBitrates[0] + " Mb/s)");
```
Nadahar: The same applies here.

src/main/java/net/pms/encoders/FFMpegVideo.java
Sami32: @Nadahar It's why it's labelled as not ready ;-)
The complement was done by @valib in this PR: https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1152/files

Thanks for reviewing.

src/main/java/net/pms/encoders/MEncoderVideo.java
Sami32: The same applies here ;-)

src/main/java/net/pms/encoders/FFMpegVideo.java
Sami32: @Nadahar Done.

Sami32:
@Nadahar Done.

Nadahar:
@Sami32 I rebased it and changed the syntax to how the "modern" logger works. There's lots of that old form around, but I always update log lines to the newer format when I touch them.

The reason to write it this way is:
* It's much easier to see how the text is going to be formatted (spaces and all)
* It's cheaper when the log mode is such that this line won't be logged. With the "old" (using ```+```) way of writing it, Java will have to concenate the strings (which is very expensive) before the evaluation of the log level is done. Using ```{}``` the concenation will take place in the logger instead, which means that it will only be done when that line as actually being logged. That way all those ```TRACE``` lines won't slow down the code that much when running without ```TRACE```.

Issue #1149, opened 2017-01-17
subtitles_color log informations ?


Sami32:
Reading a log i saw that:
**subtitles_color: -52276**   (=0xFFFFFFFFFFFF33CC)

So i'm wondering why a such value is displayed...and why not 16724940 (255,51,204 ==> 0xFF33CC)

Do the value displayed in the log will not be more natural in hexadecimal ?
Maybe `Integer.toHexString(value)` or `String.format("%x", Integer.parseInt(s))` could be used ?

valib:
@Sami32 fixed

PR #1152, opened 2017-01-18
Store the subs color in UMS.conf as the hexa string

valib:
+ fix the logging when the renderer's maximum bitrate setting has the
priority over the user's setting

This fix issues #1149 and #1150

### DIFF
--- a/src/main/java/net/pms/encoders/MEncoderVideo.java
+++ b/src/main/java/net/pms/encoders/MEncoderVideo.java
@@ -51,6 +51,7 @@ import static net.pms.util.AudioUtils.getLPCMChannelMappingForMencoder;
 import static net.pms.util.StringUtil.quoteArg;
 import org.apache.commons.configuration.event.ConfigurationEvent;
 import org.apache.commons.configuration.event.ConfigurationListener;
+import org.apache.commons.lang3.StringUtils;
 import static org.apache.commons.lang.BooleanUtils.isTrue;
 import static org.apache.commons.lang3.StringUtils.*;
 import org.slf4j.Logger;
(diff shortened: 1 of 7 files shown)


src/main/java/net/pms/encoders/MEncoderVideo.java
```diff
@@ -1408,8 +1409,8 @@ public ProcessWrapper launchTranscode(
 
 				if (override_ass_style) {
 					String assSubColor = "ffffff00";
-					if (configuration.getSubsColor() != 0) {
-						assSubColor = Integer.toHexString(configuration.getSubsColor());
+					if (!configuration.getSubsColor().isEmpty()) {
```
Sami32: @valib Why this way is better ?
Could not an NPE be throwed ?

Sami32:
What a busy bee :+1: 
Nice work :-)

Sami32:
@valib I thought that since it's RGB 0xffffff was fine.
> Reading subtitles_color: "ffff33cc" (default: "ffffffff")
Did the extra byte used for alpha channel ?

Sami32:
@valib No idea ?
0xFFFFFF       (RGB)
0xFFFFFFFF     ?

valib:
@Sami32 this is the Java presentation of the sRGB color see https://docs.oracle.com/javase/7/docs/api/java/awt/Color.html which includes also alpha channel and this color format is used also with FFmpeg.

Sami32:
@valib Thank you very much :smile: 
I was only suspecting it, as my stupid brain was stuck to HTML code representation...

Honestly, i don't think that users will ever change it from their configuration file.
But if they want keep a safe copy of their general configuration for later fast ready use just in case, why not...

PR #1154, opened 2017-01-19
Fix WMP for Windows 10

Sami32:
Following this user report: http://www.universalmediaserver.com/forum/viewtopic.php?f=5&t=9142
And this WMP user-agent strings posted here:
https://forums.plex.tv/discussion/211935/dlna-server-not-accessible
http://www.universalmediaserver.com/forum/viewtopic.php?t=8108
https://emby.media/community/index.php?/topic/42118-dlna-profiles-perhaps/
...

### DIFF
--- a/src/main/external-resources/renderers/Microsoft-WindowsMediaPlayer.conf
+++ b/src/main/external-resources/renderers/Microsoft-WindowsMediaPlayer.conf
@@ -12,13 +12,14 @@ RendererIcon = wmp.png
 #
 # User-Agent: Microsoft-Windows/6.1 UPnP/1.0 Windows-Media-Player-DMS/12.0.7601.17514 DLNADOC/1.50
 # User-Agent: Windows-Media-Player-DMS/12.0.7601.17514
+# User-Agent: Microsoft-Windows/10.0 UPnP/1.0 Windows-Media-Player/12.0.10011.16384
 # (Win7.x) User-Agent: Microsoft-Windows/6.1 UPnP/1.0, Cache-Control: no-cache, Pragma: no-cache
 # (Win8.1) User-Agent: Microsoft-Windows/6.3 UPnP/1.0, Cache-Control: no-cache, Pragma: no-cache
 # (Win8.1) User-Agent: User-Agent: Microsoft-DLNA DLNADOC/1.50
 # ============================================================================
 #
 
-UserAgentSearch = Windows-Media-Player|Microsoft-Windows/6.[13] UPnP/1.0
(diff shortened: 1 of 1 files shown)


Sami32:
I don't understand why these string are set ? Anyone have an idea ?
```
 Video = false
 Audio = false
 Image = false
```

Nadahar:
@Sami32 They are probably set to "disable" WMP to prevent UMS going on a scan frenzy if there's a Windows with DLNA enabled on the network. There are several problems between UMS and WMP. According to @atamariya one of the problems is that UMS doesn't support search (on the UPnP/DLNA level), and WMP requires this. In addition I think that WMP queries the whole directory tree up front (just like VLC pre 3.x) which really doesn't work well with UMS. That means that every folder will be queried, including #TRANSCODE# folders, settings folders etc and the media library. UMS will as a result go crazy trying to scan all this (often parsing each item multiple times) and the whole thing simply times out. 

I once wanted to fix this myself, but when I realized how massive changes would be needed to be able to handle this I abandoned the idea.

Sami32:
@Nadahar Thank you for the explaination :smile:

Sami32:
Infos take from [this user log](http://www.universalmediaserver.com/forum/download/file.php?id=3624):
> INFO  2017-01-22 15:14:19.232 [New I/O worker #8] net.pms.network.RequestHandlerV2 Media renderer was not recognized. Possible identifying HTTP headers:User-Agent: User-Agent: Microsoft-DLNA DLNADOC/1.50

> TRACE 2017-01-22 15:14:19.779 [New I/O worker #6] net.pms.network.RequestHandlerV2 Received on socket: User-Agent: Microsoft-Windows/10.0 UPnP/1.0

> TRACE 2017-01-22 15:14:19.232 [New I/O worker #8] net.pms.network.RequestHandlerV2 Received on socket: User-Agent: User-Agent: Microsoft-DLNA DLNADOC/1.50

> INFO  2017-01-22 15:14:19.195 [New I/O worker #5] net.pms.network.RequestHandlerV2 Media renderer was not recognized. Possible identifying HTTP headers:User-Agent: Microsoft-Windows/10.0 UPnP/1.0, Cache-Control: no-cache, Pragma: no-cache

> User-Agent: DAFUPnP

Should i add this last one ? is it really WMP specific or Windows in general ?

I think that i should add `Microsoft-DLNA` as well. Any objection to that ?

Sami32:
https://github.com/UniversalMediaServer/UniversalMediaServer/pull/697

PR #1155, opened 2017-01-20
Panasonic DMR-BWT740

Sami32:
As saw in the log of this user: http://www.universalmediaserver.com/forum/download/file.php?id=3616

DEBUG 2017-01-19 10:19:25.891 [cling-7] New renderer found: "DMR-BWT740" with dlna details: {friendlyName=DMR-BWT740, address=192.168.1.83, udn=uuid:4D454930-0300-1000-8000-A81374478316, manufacturer=Panasonic, modelName=BD/DVD Recorder, modelNumber=DMR-BWT740, modelDescription=Panasonic BD/DVD Recorder, manufacturerURL=, modelURL=}

It used the Panasonic-Viera.conf as default but have a very different formats support, as a BlueRay player. We didn't have a Panasonic DMR yet.


Sami32:
@SubJunk Done ;-)

SubJunk:
This looks good and I'm happy for it to be merged. It would be good to add a more generic renderer profile too for DMR-BWT so that it automatically detects next year's ones and previous years' ones too

src/main/external-resources/renderers/Panasonic-DMRBWT740.conf
```diff
@@ -0,0 +1,56 @@
+#----------------------------------------------------------------------------
+# Profile for Panasonic DMR-BWT740.
+# See DefaultRenderer.conf for descriptions of all the available options.
+#
+
+RendererName = Panasonic BWT740
```
SubJunk: With that line being changed we should also change the filename and the other occurrences of "DMR-BWT740" in this file. Otherwise we could make two files: `Panasonic-DMRBWT740.conf` and `Panasonic-DMRBWT.conf`.
They can be identical for now with their configs but it can set it up for future changes. What do you think?

Sami32:
@SubJunk I understand your point, but i confess that i don't like too much make generic when i saw many differences.
But anyway, it'll be better than the Panasonic-Viera default for them.

SubJunk:
Cool, anyway feel free to merge it when you want, it can always be improved later and it is much better than what we have :)

Sami32:
I just become to dislike "common-imaging" that make me mad to try again and again trying to pass the checks !!! just too boring, very annoying...

Since i don't any issue while building with Maven, i guess that a geographical issue or Travis IP bashing ?

src/main/external-resources/renderers/Panasonic-DMRBWT740.conf
```diff
@@ -0,0 +1,56 @@
...
+# manufacturer=Panasonic
+# modelName=BD/DVD Recorder
+# modelNumber=DMR-BWT740
+# modelDescription=Panasonic BD/DVD Recorder
+# manufacturerURL=
+# modelURL=
+# ============================================================================
+# https://www.richersounds.com/static/manuals/DMRBWT740.pdf
+
+UpnpDetailsSearch = DMR-(B|P)WT
```
Sami32: @SubJunk Ok, i guess that i should be more general:
http://www.universalmediaserver.com/forum/viewtopic.php?f=9&t=9173&p=29618#p29618

I didn't even now all these german models...
@Nadahar I'll try to modify it to `DMR-(B|P)` but i don't what is the more clean/correct way to do that...

Nadahar: @Sami32 As long as it's merged, there's no way but to create a new commit.

Sami32: @Nadahar Thansk.
I was suspecting that, so i already pushed one patch.

PR #735, opened 2015-11-19
Additional formats

Nadahar:
This is not ready to merge. I have several things I need to figure out/test related to this PR, but I'm creating the PR now to not have to discuss it on the commits.

PR #1162, opened 2017-01-23
Fix renderers and typos

Sami32:
Some renderer and typos fixes

### DIFF
--- a/src/main/external-resources/UMS.conf
+++ b/src/main/external-resources/UMS.conf
@@ -1129,7 +1129,7 @@ flac_path =
 vlc_path =
 
 ############################################
-## Buffer settings
+## Buffer settings (non-GUI)
 ############################################
 
 # Transcode keep first connection
@@ -1142,11 +1142,11 @@ transcode_keep_first_connection =
 # Transcode block multiple connections
 # ------------------------------------
 # When set to "true" activate the BufferOutputFile.java.
-# Have a Boolean value opposite to that of transcode_keep_first_connection.
+# Have a boolean value opposite to that of transcode_keep_first_connection.
 # Default: false
 transcode_block_multiple_connections =
 
-# Maximum audo buffer size
+# Maximum audio buffer size
 # ------------------------
 # The maximum size (in MB) that UMS should use for buffering audio.
 # Default: 100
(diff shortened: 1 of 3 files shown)


Nadahar:
@Sami32 With "recent audio commit" do you mean #735? If so, how did #735 influence what mimetype should be used for WAV? And, if it's indeed the case, I doubt it only apples to this particular renderer... however I think we need to figure out WHY this change is needed, not just change it.

Sami32:
@Nadahar Nothing to do with this commit, or i will told you.
I suspect that came in 6.5.3 but didn't searched for that, only discover it while testing this PR, so i tested with an other thumbnails PR and it was same.
I didn't see any flaw in the code so i tried with this MIME type and it solved the issue.

Nadahar:
Still, isn't it very likely that this applies to other renderers as well, at least other Panasonics?

Sami32:
No, since i put x-wav MIME type in very few renderers, none was Panasonic.

Nadahar:
@Sami32 But... this renderer is a Panasonic and has ```x-wav``` - it doesn't really matter who set it if it doesn't work...? Have you verified that it works for those that currently have it?

Sami32:
As you know i don't have them.
I get your point, but i'll not search on that as it work and many others bugs recently introduced need to be fixed.
You make think that perhaps you cannot play WAV files on your renderer anymore, do you ? As it could affected the others renderers "indirectly".

Do you want that i change them all ?
My renderer did play any WAV with audio/wav when i tried for the first time, so perhaps that affected only few...or all ?

Sami32:
How Travis could have missed this bug ?

I'm the one that missed it while editing/merging, and only saw now while resolving a conflict in an other PR. _Maven_ did advertised as well.

Nadahar:
@Sami32 I don't know how you manage to commit stage.diff - I've never seen it before. How do you do your commit to get that file committed? You're only supposed to commit the files you choose to commit, that's why I say that TortoiseGit is great for committing as you can see exactly what you're going to commit.

That said, Travis won't "detect" any such things, it will only check if the project can build. From the diff here, it seems that the file is still here though...

Sami32:
You didn't read me: Bug
> How Travis could have missed this bug in setup.nsi ?

Never mind for the other.

Nadahar:
@Sami32 Travils doesn't build for Windows - it's what bit me when I removed a dependency from the POM that was only used during the Windows build at all. I looked for a CI that offers Windows builds, but couldn't find it - I'm guessing that Microsoft wants license fees for the virtual instances used when building.

Configuring a Jenkins server that did Windows builds could solve that, but as long as we don't have that I guess the morale of the story is: If you might have changed anything only applying to Windows, test-build yourself.

Sami32:
Thanks for pointing at that.

PR #1167, opened 2017-01-26
Thai

Sami32:
@Nadahar Thanks :heart:


Nadahar:
I think @valib need to verify that the subtitles part is correct - I haven't verified that at all. The GUI part should be ok, but I can't approve only that part.

src/main/java/net/pms/util/Constants.java
```diff
@@ -11,16 +11,19 @@
 	public static final String CHARSET_ISO_8859_7	= "ISO-8859-7".intern();
 	public static final String CHARSET_ISO_8859_8	= "ISO-8859-8".intern();
 	public static final String CHARSET_ISO_8859_9	= "ISO-8859-9".intern();
+	public static final String CHARSET_ISO_8859_11	= "ISO-8859-11".intern();
```
valib: @Sami32 those constants are used for automatic charset detection by ICU4J. The Thai language is unfortunatelly not supported so adding it here does nothing. See http://userguide.icu-project.org/conversion/detection#TOC-Detected-Encodings

Sami32: @valib Snif snif...
Sadly the only Java libraries that i found have czech or thaï but not the both :wink:

PR #1174, opened 2017-02-07
Fix the recognition issue in this Sony Bravia XD renderer

Sami32:
Correct the bug \\ in the parser that make it unrocognized.
This renderer was already well tested by a user and improved on the forum.

### DIFF
--- a/src/main/external-resources/renderers/Sony-BraviaXD.conf
+++ b/src/main/external-resources/renderers/Sony-BraviaXD.conf
@@ -1,68 +1,76 @@
-#------------------------------------------------------------------------------------
+#---------------------------------------------------------------------------------------------------
 # Profile for Sony Bravia XD 70/75/80/83/85/93/94 series, 4K HDR avec Android TV.
 # See DefaultRenderer.conf for descriptions of all the available options.
 
 RendererName = Sony Bravia XD
 RendererIcon = Sony-BraviaXD.png
 
-# ===================================================================================
+# =================================================================================
 # This renderer has sent the following string/s:
 #
+# User-Agent=Dalvik/2.1.0 (Linux; U; Android 6.0.1; BRAVIA 4K GB Build/MMB29V.S50) #with Chromecast it use a different IP#
 # User-Agent: UPnP/1.0, X-AV-Physical-Unit-Info: pa="BRAVIA KD-65XD7505";
-# User-Agent=Dalvik/2.1.0 (Linux; U; Android 6.0.1; BRAVIA 4K GB Build/MMB29V.S50)
 # X-AV-Client-Info: av=5.0; cn="Sony Corporation"; mn="BRAVIA KD-65XD7505"; mv="3.0";
 # friendlyName=KD-65XD7505
-# udn=uuid:48454d48-01af-4343-ad82-70e21eb9dee1
+# udn=uuid:48454d48-01af-4343-ad82-70e21eb9dee1 *
+# udn=uuid:c2b26bb0-36cb-44de-a38a-b993b2d214a4 *
+# udn=uuid:23456789-1234-1010-8000-104fa8c05b3b ***
 # manufacturer=Sony Corporation
 # modelName=KD-65XD7505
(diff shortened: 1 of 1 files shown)


Nadahar:
@Sami32 Which bug in the parser? I think you've misunderstood how this works. Java replaces [Escape Sequences](http://docs.oracle.com/javase/tutorial/java/data/characters.html) automatically when the file is read, before they "reach UMS". That means that ```\t``` wil be replaced with the ```tab``` character, ```\b``` with backspace, ```n``` with line feed and ```\\``` with ```\```. The thing is that, even though it will work using ```\d``` because ```\d``` isn't an escape character, that's a very bad habbit. It the letter after the ```\``` happens to be one of the "special" characters it will NOT work as expected. It's always safe to use ```\\``` because that will always be replaced with ```\```. It's much easier and safer to say "always used ```\\``` for ```\``` instead of saying: You have to use ```\\``` instead of ```\``` if the next character is t, b, n, r, f ' or ". 

Replacing the ```\\``` with ```\``` is therefore what I'd consider bad practice, and I simply don't believe that this causes a bug. I've attached a picture of how the existing string (```\\```) looks in the debugger after being read from the file:

upnpdetailssearch

Sami32:
That's was not working for the user, that's a good reason isn't it ?

To me, a bad practice, would be ignore the users issues...

2017-02-07 8:01 GMT+01:00 Nadar <notifications@github.com>:

Sami32:
@Nadahar If you feel confident, maybe you can replace it in all the renderers ?

I'll not play on that...

Nadahar:
@Sami32 You're still not listening to what I'm trying to explain. It's not about listening to the users, it's about understanding how this works. If it in fact doesn't work with ```\\``` I'd be very surprised and I'd like to figure out what is happening. I can't see that happening, as I've checked that ```\\``` is replaced with ```\``` when the file is read. Didn't you see the debug screenshot I attached?

Sami32:
@Nadahar You can be funny sometime...Eclipse is like a chineese documentation for me.
But if you tell that it's conform to the java theory and that's work good.

BUT as i'll be the only one doing that, and that i don't know if it will work in the reality, i'll not change it,for now at least.

Which one don't listen to the other ? No offense but i believe what i can saw, touch...

Nadahar:
@Sami32 If you look at the right side of the screen under "Expressions" you can see the value of ```upnpDetails``` when the program is paused at line 1369 in ```RendererConfiguration``` (as is shown on the left side). That is where the matching is done, and the string you can see is as it is when read from the configuration file before it is run through the regex engine.

You can see the variable content printed out, and over there you can see many lines each presenting one character with their codes in square brackets. As you can see, the code for the character in question (number 10) is [```\u005c```](http://www.fileformat.info/info/unicode/char/005c/index.htm). This is directly from memory as UMS is running, and this is when the renderer configuration file has ```\\``` specified.

Nadahar:
@Sami32 Since debugging isn't good enough for you, I just did a test with my Onkyo. I changed the renderer configuration to:
```perl
#UserAgentSearch = Mediabolic-IMHTTP/1.8.225
UpnpDetailsSearch = TX-NR\\d{2}7
```

The actual name is ```TX-NR807```. It is detected like normal.

Sami32:
@Nadahar That worked for me as well, so i have done the change ;-)
An IDE seem useful after all...but each thing in his time.

Though your example was with `BRAVIA KD-\d{2}XD{7,8,9}` i'm not sure to understand why it's displaying `\\` but write `\u005c`. Perhaps because Java convert `\` in `\\` ?
Thank you anyway for your explanation.

Nadahar:
@Sami32 You are correct, Java "translates" ```\``` to ```\\``` when displaying it - but as you can see, the code is for ```\```. It can be a bit confusing some times that they do this, it works better for ```\t```, ```\r``` and ```\n``` which are "invisible" characters. For them it's useful that they are translated so it's easy to see what they are.

Sami32:
@Nadahar Thanks :wink:

PR #1186, opened 2017-02-10
Stuff from findbugs

SubJunk:


Sami32:
@SubJunk  I tested it and didn't see any issue, though i could have tested it not correctly/fully. I just scanned all the media that i can and didn't saw any error or memory hight consumption.

As i don't know if you intended to do that, i like it that way though, but the message saying that the scan could take long time doesn't appear anymore in this PR while clicking on the scan button.

As a side note, when we stop the scan, the icon never return green as before, when @valib did fixed that, so i don't when this "revertion" did occur.

valib:
@SubJunk sorry I approved changes but after retesting I must agree with @Sami32 that there is an issue with scanning shared folders as @Sami32 described.

SubJunk:
Ah yeah that warning removal must have snuck in from 7.0.0

Nadahar:
@SubJunk I've revived this PR as I think it's a shame to let it go to waste. I've resolved conflicts and fixed some further FindBugs stuff in the files that you had already modified.

SubJunk:
Ok, cool. Can you think of any specific tests to do on it?

Nadahar:
@SubJunk Hopefully nothing has changed so I was just thinking of general use testing. Personally I feel least comfortable with the changes in ```FFMpegVideo```, so making sure that video transcoding is unaffected might be a good idea.

PR #1189, opened 2017-02-12
Removed unused code that caused memory bloat

SubJunk:


src/main/java/net/pms/configuration/DeviceConfiguration.java
```diff
@@ -85,7 +85,6 @@ public void inherit(RendererConfiguration ref) throws ConfigurationException {
 		filter = baseConf.filter;
 
 		// Initialize our internal RendererConfiguration vars
-		renderCache = new HashMap<>();
 		sortedHeaderMatcher = ref.sortedHeaderMatcher;
```
Nadahar: @SubJunk Maybe you should remove the comment as well?

SubJunk: @Nadahar thanks for the review. I think the comment is still relevant for the remaining vars `sortedHeaderMatcher` and `loaded`

Nadahar: @SubJunk 😊 

I "read" "Initialize our internal renderer cache". Don't ask me how I managed.... forget it.

SubJunk: Cool, thanks for the approval!

SubJunk:
@Nadahar after testing it seems that this fix isn't enough. It's still good since the code isn't needed anymore but I'll keep looking and see what else is needed

SubJunk:
The bug was introduced again in the global ID commit 7ddcda0
I guess maybe it's storing things as new IDs instead of reusing them, or something. I'll have a look next time I have a chance, but it may not be for a while so any help will be appreciated @UniversalMediaServer/developers

Nadahar:
@SubJunk Even though this doesn't solve the bug, it's nice to clean up that unused code. I think it should be merged for that reason alone.

SubJunk:
Yeah I agree, I'll merge now since we have other branches for the bug now, I was just keeping it around for a place to discuss and work on the issue

Issue #1185, opened 2017-02-10
Memory leak during scan


SubJunk:
Every time a file scan is done (navigation/share tab, button near the bottom) it's not freeing enough memory.

I have reproduced it in latest master (5453878a34d88c91ebfdc0e4d851d50d6b393329) and in 7.0.0 (ef294e4)

I'm not sure how long the bug has existed but in 7.0.0 rescans are triggered often so it makes it more visible than before.

SubJunk:
I made this PR https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1186 initially to see if findbugs would see anything related to this but the bug is the same in that branch

Nadahar:
@SubJunk I assume that caching is enabled when you experience this? Have you tried without caching and confirmed that it is unrelated to caching?

Memory leaks are supposed to be "impossible" to create in Java. That's not true in reality, but the traditional "alloc" but fail to "free" memory leak is. Simply keeping references to objects no longer needed is enough that the GC won't free them, so this, while not technically a memory leak because they are still referenced, will have the same effect. I also suspect that the lack of thread safety can lead to references being "messed up" so that they are either "lost" or are stuck in a state where the GC think they are still in use. I don't know enough about the details of the GC to know for certain, this is just my general thoughts.

What is the message of the ```OutOfMemoryError```? Java Heap Space?

SubJunk:
@Nadahar scanning is only possible when cache is enabled.

Yeah my suspicion is like you said, that we are keeping references to objects that are no longer needed.

I haven't seen an error yet because the memory use keeps going until it freezes

I've been going backwards in time to see when it appeared and I've found a version that doesn't have the memory creep, 4.4.0. I'll keep narrowing down when it first appeared and may be able to find a fix that way instead of profiling.

atamariya:
Few things to check:

- A RootFolder is created for each renderer/browser instance with the corresponding file tree. Do you have multiple renderers?
- MediaLibrary creates a new RealFile for each scan. These are duplicates of the files found during folder scan but held in the memory at the same time.

SubJunk:
Good notes @atamariya 

The first released version it happens in is 5.0.0-a1 so now I'm going to look at the differences between 4.4.0 and 5.0.0-a1 to see what could have caused it.

Nadahar:
[...]

PR #1190, opened 2017-02-12
GlobalIdRepo thread-safety

Nadahar:
This is something I did while looking at #1185. It doesn't solve it, but it can't hurt. The current synchronization of this class is flawed, it's "partially synchronized" and that doesn't really help much.


SubJunk:
Feel free to merge this. If you use the `Squash and merge` button it will remove the merge commit

Nadahar:
@SubJunk Ok, will do. I just rebased it instead 😉

PR #1191, opened 2017-02-12
Global id debug

Nadahar:
This isn't meant to be merged - it's just for debugging. Hands of the ```Merge``` button (@Sami32) 😛 

@SubJunk This is a little "tool" I made to debug #1185. Set logging level to "INFO" and you'll get mostly this output. It will dump the global id's with their ```DLNAResources``` every time one is added. 

I see that every time a folder is scanned, new entries are added. There are multiple entires for each resource. However, it's new ```DLNAResource``` instances. To me that indicates that neither ff11febbe088199f6d27b9cefafa723cae8c08ce nor 7ddcda093e01ac2efa5a30f8a50119f3c77bacd8 is the culprit, they simply expose the underlying problem.

The problem as I see it is simply that when resources are being scanned new instances are being created for the same media file/stream. I've never quite understood how existing resources are supposed to be recycled, but apparantly they aren't. 

I fear that this won't be an easy fix. One one side it partly explains why there aren't more threading issues with the extremely thread-unsafe code in DLNAResource. Every time a new renderer enters a folder, the folder is "scanned" and new resources are generated. That means that two renderers don't actually share the same instance even though they refer to the same underlying media file. The problem is that fixing this also means unleashing a world of pain with all the thread-safety flaws.

As I see it, "fixing" DLNAResource is a daunting task - I'm not even sure if it's doable. I've been thinking about it, and come to the conclution that is might be easier to rewrite it from scratch than to try to untangle everything that goes wrong there. But, I might be wrong, one really don't know before one tries. The problem is that objects are being shared without any regard for concurrency and immutability or synchronization is virtually unheard of. It's not just ```DLNAResource``` itself, it's all the instances it holds. They all need to be either immutable or synchronized. It isn't just to add synchronization either, you generally don't want to hold too many locks at the same time (or, if you hold multiple locks you have to make sure that they are ALWAYS aquired in the same order - which can be very difficult). 

I'm not actually sure how to tackle this, it might be easier to dispose of them after use. That would mean forgetting all about caching them and redoing the global cache. It won't be very efficient, but it might actually be achievable.

### DIFF
--- a/src/main/java/net/pms/dlna/GlobalIdRepo.java
+++ b/src/main/java/net/pms/dlna/GlobalIdRepo.java
@@ -31,23 +31,31 @@ public class GlobalIdRepo {
 	public void add(DLNAResource dlnaResource) {
 		lock.writeLock().lock();
 		try {
-			String id = dlnaResource.getId();
-			if (id != null) {
-				remove(id);
-			}
-			ids.add(new ID(dlnaResource, curGlobalId++));
+			boolean isUniqueResource = true;
+			String resourcePath = dlnaResource.getSystemName();
+//			String id = dlnaResource.getId();
+//			if (id != null) {
+//				remove(id);
+//			}
 
 			LOGGER.info("---------------------------------------------------------------------------------");
 			LOGGER.info("------------------------------- Global ID dump ----------------------------------");
 			LOGGER.info("---------------------------------------------------------------------------------");
 			for (ID element : ids) {
+				if ("RealFile".equals(element.dlnaResource.getClass().getSimpleName()) && resourcePath.equals(element.dlnaResource.getSystemName())) {
+					dlnaResource.setId(element.dlnaResource.getId());
+					isUniqueResource = false;
+				}
 				LOGGER.info(
-					"Id: {}, Name: {}, Class: {}, DisplayName: {}",
-					element.dlnaResource.getId(), element.dlnaResource.getName(), element.dlnaResource.getClass().getSimpleName(), element.dlnaResource.getDisplayName()
+					"Id: {}, Name: {}, Class: {}, DisplayName: {}, Path: {}",
+					element.dlnaResource.getId(), element.dlnaResource.getName(), element.dlnaResource.getClass().getSimpleName(), element.dlnaResource.getDisplayName(), element.dlnaResource.getSystemName()
 				);
 			}
 			LOGGER.info("---------------------------------------------------------------------------------");
 			LOGGER.info("");
+			if (isUniqueResource) {
+				ids.add(new ID(dlnaResource, curGlobalId++));
+			}
 		} finally {
 			lock.writeLock().unlock();
 		}
--- a/src/main/java/net/pms/dlna/GlobalIdRepo.java
+++ b/src/main/java/net/pms/dlna/GlobalIdRepo.java
@@ -53,8 +53,9 @@ public class GlobalIdRepo {
 //					isUniqueResource = false;
 //				}
 				LOGGER.info(
-					"Id: {}, Name: {}, Class: {}, DisplayName: {}, Path: {}",
-					element.dlnaResource.getId(), element.dlnaResource.getName(), element.dlnaResource.getClass().getSimpleName(), element.dlnaResource.getDisplayName(), element.dlnaResource.getSystemName()
+					"Instance: {}, Id: {}, Name: {}, Class: {}, DisplayName: {}, Path: {}",
+					System.identityHashCode(element.dlnaResource), element.dlnaResource.getId(), element.dlnaResource.getName(),
+					element.dlnaResource.getClass().getSimpleName(), element.dlnaResource.getDisplayName(), element.dlnaResource.getSystemName()
 				);
 			}
 			LOGGER.info("---------------------------------------------------------------------------------");
--- a/src/main/java/net/pms/dlna/GlobalIdRepo.java
+++ b/src/main/java/net/pms/dlna/GlobalIdRepo.java
@@ -32,17 +32,23 @@ public class GlobalIdRepo {
 		lock.writeLock().lock();
 		try {
 			boolean isUniqueResource = true;
-			String resourcePath = dlnaResource.getSystemName();
-//			String id = dlnaResource.getId();
-//			if (id != null) {
-//				remove(id);
-//			}
+			String incomingResourcePath = dlnaResource.getSystemName();
+			String incomingClass = dlnaResource.getClass().getSimpleName();
+			String id = dlnaResource.getId();
+			if (id != null) {
+				remove(id);
+			}
+			LOGGER.info(
+				"Incoming Id: {}, Name: {}, Class: {}, DisplayName: {}, Path: {}",
+				dlnaResource.getId(), dlnaResource.getName(), dlnaResource.getClass().getSimpleName(), dlnaResource.getDisplayName(), dlnaResource.getSystemName()
+			);
 
 			LOGGER.info("---------------------------------------------------------------------------------");
 			LOGGER.info("------------------------------- Global ID dump ----------------------------------");
 			LOGGER.info("---------------------------------------------------------------------------------");
 			for (ID element : ids) {
-				if ("RealFile".equals(element.dlnaResource.getClass().getSimpleName()) && resourcePath.equals(element.dlnaResource.getSystemName())) {
+				if (incomingClass.equals(element.dlnaResource.getClass().getSimpleName()) && incomingResourcePath.equals(element.dlnaResource.getSystemName())) {
+					LOGGER.info("already exists, setting id to " + element.dlnaResource.getId());
 					dlnaResource.setId(element.dlnaResource.getId());
 					isUniqueResource = false;
 				}
@@ -54,6 +60,7 @@ public class GlobalIdRepo {
 			LOGGER.info("---------------------------------------------------------------------------------");
 			LOGGER.info("");
 			if (isUniqueResource) {
+				LOGGER.info("adding");
 				ids.add(new ID(dlnaResource, curGlobalId++));
 			}
 		} finally {
--- a/src/main/java/net/pms/dlna/GlobalIdRepo.java
+++ b/src/main/java/net/pms/dlna/GlobalIdRepo.java
@@ -31,9 +31,9 @@ public class GlobalIdRepo {
 	public void add(DLNAResource dlnaResource) {
 		lock.writeLock().lock();
 		try {
-			boolean isUniqueResource = true;
-			String incomingResourcePath = dlnaResource.getSystemName();
-			String incomingClass = dlnaResource.getClass().getSimpleName();
+//			boolean isUniqueResource = true;
+//			String incomingResourcePath = dlnaResource.getSystemName();
+//			String incomingClass = dlnaResource.getClass().getSimpleName();
 			String id = dlnaResource.getId();
 			if (id != null) {
 				remove(id);
@@ -47,11 +47,11 @@ public class GlobalIdRepo {
 			LOGGER.info("------------------------------- Global ID dump ----------------------------------");
 			LOGGER.info("---------------------------------------------------------------------------------");
 			for (ID element : ids) {
-				if (incomingClass.equals(element.dlnaResource.getClass().getSimpleName()) && incomingResourcePath.equals(element.dlnaResource.getSystemName())) {
-					LOGGER.info("already exists, setting id to " + element.dlnaResource.getId());
-					dlnaResource.setId(element.dlnaResource.getId());
-					isUniqueResource = false;
-				}
+//				if (incomingClass.equals(element.dlnaResource.getClass().getSimpleName()) && incomingResourcePath.equals(element.dlnaResource.getSystemName())) {
+//					LOGGER.info("already exists, setting id to " + element.dlnaResource.getId());
+//					dlnaResource.setId(element.dlnaResource.getId());
+//					isUniqueResource = false;
+//				}
 				LOGGER.info(
 					"Id: {}, Name: {}, Class: {}, DisplayName: {}, Path: {}",
 					element.dlnaResource.getId(), element.dlnaResource.getName(), element.dlnaResource.getClass().getSimpleName(), element.dlnaResource.getDisplayName(), element.dlnaResource.getSystemName()
@@ -59,10 +59,10 @@ public class GlobalIdRepo {
 			}
 			LOGGER.info("---------------------------------------------------------------------------------");
 			LOGGER.info("");
-			if (isUniqueResource) {
-				LOGGER.info("adding");
+//			if (isUniqueResource) {
+//				LOGGER.info("adding");
 				ids.add(new ID(dlnaResource, curGlobalId++));
-			}
+//			}
 		} finally {
 			lock.writeLock().unlock();
 		}
--- a/src/main/java/net/pms/dlna/GlobalIdRepo.java
+++ b/src/main/java/net/pms/dlna/GlobalIdRepo.java
@@ -36,9 +36,22 @@ public class GlobalIdRepo {
 				remove(id);
 			}
 			ids.add(new ID(dlnaResource, curGlobalId++));
+
+			LOGGER.info("---------------------------------------------------------------------------------");
+			LOGGER.info("------------------------------- Global ID dump ----------------------------------");
+			LOGGER.info("---------------------------------------------------------------------------------");
+			for (ID element : ids) {
+				LOGGER.info(
+					"Id: {}, Name: {}, Class: {}, DisplayName: {}",
+					element.dlnaResource.getId(), element.dlnaResource.getName(), element.dlnaResource.getClass().getSimpleName(), element.dlnaResource.getDisplayName()
+				);
+			}
+			LOGGER.info("---------------------------------------------------------------------------------");
+			LOGGER.info("");
 		} finally {
 			lock.writeLock().unlock();
 		}
+
 	}
 
 	public DLNAResource get(String id) {


SubJunk:
Thanks for the notes, and I agree with them. I had added similar logging too. I think you're right we should forget about caching them in order to be pragmatic. Maybe if we feel ambitious we can add a better implementation in the future but for now we need to fix the growing memory use bug.

SubJunk:
Another option would be to just stop exposing the issue so much in 7.0.0 by working more on a more specific scanning solution like we already started on. Both approaches aren't complete fixes but the second one may end up with more benefit to the user so I'm actually leaning towards that now.

It's not likely that a user will click to manually scan enough times to run out of memory.

atamariya:
@SubJunk  You can reduce the memory footprint by ensuring that an instance is reused, based on path, while calling GlobalIdRepo.add(). And for future, look for proper cache implementation like EhCache.

SubJunk:
I added a commit here for feedback. It seems to keep the indexes unique but I'm not sure if we are relying on the duplicates anywhere. I'm also not sure whether it's better to reuse the instance or to replace it. Feedback welcome.

Nadahar:
@SubJunk Reusing it would obviously be the best approach were it not for the lack of thread-safety. I'm afraid it can lead to a whole lot of other "strange" bugs, especially when serving multiple renderers.

Nadahar:
@SubJunk Your latest commit seems like it would create a bug. I don't think that's the correct place to make them "unique", there must be a lookup before creating a new DLNAResource instance. What you do now is still creating multiple resources for each media, but you only keep the ID for one of them. The wrong ID will still be sent to the renderer though, but Request/RequestV2 won't be able to find anything but the first ID for that media file and will fail to return a result.

SubJunk:
Feel free to fix/improve

Nadahar:
@SubJunk I haven't verified it, I was just sharing my thoughts about the approach. I don't think it can be done that way.

SubJunk:
I think I will need more time to understand how the global ID repo works, I wasn't involved with it when it was developed. It doesn't seem to be working properly and I'm not sure what the best way to fix it will be. It seems pretty broken to me - during testing it appears that every time I navigate to the same file I get a different ID reference, but maybe that isn't the same ID as the global one.

Nadahar:
@SubJunk You do get a new ID, that's what I pointed out before. But, that's not GlobalID's fault. GlobalID simply assigns a new ID to each ```DLNAResource``` (or subclass) instance. The problem is that new ```DLNAResources``` are being made each time a folder is scanned or accessed, as I said in the first post here:
> The problem as I see it is simply that when resources are being scanned new instances are being created for the same media file/stream. I've never quite understood how existing resources are supposed to be recycled, but apparantly they aren't.

To solve this, the scanning/discovery/resolve must be changed so that it checks if the ```DLNAResource``` already exists before creating a new. While this isn't as easy as it sounds, because you'd have to define some what of defining what is "the same resource" and path isn't enough because they are being cloned... the key to solving this, is to make sure that instead of always creating a new ```DLNAResource``` the existing one should be reused when scanning (either via manual or automatic scan, or via the scan that takes place when you enter a folder). 

It wouldn't be undoable in any way, but what I was trying to say furter up is: ```DLNAResource``` (and its subclasses) have a awful lot of thread-safety issues. I've always wondered how this could cause as few bugs as it does, although it do cause bugs at times. Since I discovered that they are not being reused (I always assumed they were reused), it explains why the lack of thread-safety isn't more of an issue.

If we modify the code so that it starts reusing these, I'm afraid we would get into all sorts of trouble. That would mean that the same instance would be shared much more, with much more potential concurrency issues. In addition, there are pure design-flaws there like having "defualt renderer" and other things that don't belong there at all stored in the ```DLNAResource```. As it is now, this works because they aren't being reused so each renderer (and "browsing instance) has it's "own" resource. When you start reusing them, the information that should never be stored there in the first place will start to create a lot of problems. Different threads will overwrite eachothers information and even if there are no concurrency bugs (by pure luck), the "default renderer" for example can have changed from one time to the next time it's value is queried, or it simply won't be the correct one for that renderer.

That's why I'm saying that I'm not sure it's a good idea to fix this, as it would require fixing a lot of existing problems with the design of ```DLNAResource``` and its subclasses. It should obviously be fixed, but I honestly don't see who would be capable of doing that without creating a lot of new issues and without having to do a LOT of redesign/refactoring.

This is very easy to verify, I'll add some loggin to this PR so you can see that this is what happens.

Nadahar:
@SubJunk I've added "instance identity", a number that is supposed to be unique to each instance. There you can see that they aren't the same, and thus they get new ID numbers assigned.

SharkHunter:
Ok I know I said i resigned but what the heck. UMS/PMS is not as all the other DLNA servers db based. Which IMO a killer feature since it allows for dynamic content which all the others struggles with.  UMS will build a tree per render for what is has discoverd. Th global ID serves as a bridge to enable cross render playback (ie remote control so render A = the remote can tell render B play this resource). Before global ID this was impossible. Now global ID is NOT meant to be a global ID since the same resource will end up in multiple render trees and thus have different global IDs, however each of those Ids can be used for playback. I see you focus a lot on the media library and that is ok but bear in mind that UMS/PMS is "dynamic discovered based". The only way to scan is to do a discover which will cost a lot of memory since we'll build a unique tree. This is nothing that you can "fix", it is how UMS/PMS is designed.  Of course you can resigned but you should know that. 
The reason that DLNAresource etc. don't have thread saftey issues is that each DLNAresource is unique per render. And since most render don't issue gazillions of requests it works (besides I think some coordination takes place i RequestHander). And that also shows that instead of screaming "thread saftey" you should look at how stuff works.

Nadahar:
@SubJunk Basically, the problems are here (each time a constructor is being called without checking if the resource already exists):

```java
DLNAResource() - net.pms.dlna.DLNAResource
	CueFolder(File) - net.pms.dlna.CueFolder
		getPlaylist(String, String, int) : DLNAResource - net.pms.dlna.PlaylistFolder
	DLNAResource(int) - net.pms.dlna.DLNAResource
		Feed(String, String, int) - net.pms.dlna.Feed
			AudiosFeed(String) - net.pms.dlna.AudiosFeed
				addWebFolder(File) : void - net.pms.dlna.RootFolder
			ImagesFeed(String) - net.pms.dlna.ImagesFeed
				addWebFolder(File) : void - net.pms.dlna.RootFolder
				manageItem() : void - net.pms.dlna.ImagesFeed
			VideosFeed(String) - net.pms.dlna.VideosFeed
				addWebFolder(File) : void - net.pms.dlna.RootFolder
		FeedItem(String, String, String, DLNAMediaInfo, int) - net.pms.dlna.FeedItem
			autoMatch(String, String) : DLNAResource - net.pms.dlna.DLNAResource
			manageItem() : void - net.pms.dlna.Feed
			resolveOnce() : void - net.pms.dlna.PlaylistFolder
		WebStream(String, String, String, int) - net.pms.dlna.WebStream
			parse(String, String) : DLNAResource - net.pms.util.UMSUtils.IOList
			setUp() : void - net.pms.util.PlayerUtilTest (3 matches)
			WebAudioStream(String, String, String) - net.pms.dlna.WebAudioStream
				addWebFolder(File) : void - net.pms.dlna.RootFolder
				autoMatch(String, String) : DLNAResource - net.pms.dlna.DLNAResource
				manageItem() : void - net.pms.dlna.AudiosFeed
				resolveOnce() : void - net.pms.dlna.PlaylistFolder
			WebVideoStream(String, String, String) - net.pms.dlna.WebVideoStream
				addWebFolder(File) : void - net.pms.dlna.RootFolder
				autoMatch(String, String) : DLNAResource - net.pms.dlna.DLNAResource
				manageItem() : void - net.pms.dlna.VideosFeed
				resolveOnce() : void - net.pms.dlna.PlaylistFolder
	DVDISOTitle(File, int) - net.pms.dlna.DVDISOTitle
		resolveOnce() : void - net.pms.dlna.DVDISOFile
	MapFile() - net.pms.dlna.MapFile
		RealFile(File, String) - net.pms.dlna.RealFile
			discoverChildren() : void - net.pms.dlna.RootFolder
			getiTunesFolder() : DLNAResource - net.pms.dlna.RootFolder (2 matches)
			parse(String, String) : DLNAResource - net.pms.util.UMSUtils.IOList
			resolveOnce() : void - net.pms.dlna.PlaylistFolder
		RealFile(File) - net.pms.dlna.RealFile
			autoMatch(String, String) : DLNAResource - net.pms.dlna.DLNAResource
			createApertureAlbum(Map<?, ?>, Map<?, ?>, List<?>) : VirtualFolder - net.pms.dlna.RootFolder
			discoverChildren() : void - net.pms.dlna.virtual.MediaLibraryFolder
			doRefreshChildren() : void - net.pms.dlna.virtual.MediaLibraryFolder
			getConfiguredFolders(ArrayList<String>) : List<RealFile> - net.pms.dlna.RootFolder
			getiPhotoFolder() : DLNAResource - net.pms.dlna.RootFolder
			manageFile(File) : void - net.pms.dlna.MapFile
			resolveOnce() : void - net.pms.dlna.CueFolder
			run() : void - net.pms.util.BasicPlayer.Logical.initAutoPlay(...).new Runnable() {...}
			scanDir(File[], DLNAResource) : void - net.pms.dlna.MediaMonitor
			searchFiles(File[], String, DLNAResource, int) : void - net.pms.dlna.FileSearch
			setUp() : void - net.pms.util.PlayerUtilTest (3 matches)
	MapFile(MapFileConfiguration, List<File>) - net.pms.dlna.MapFile
		discoverChildren(String) : void - net.pms.dlna.MapFile
	MapFile(MapFileConfiguration) - net.pms.dlna.MapFile
		analyzeChildren(int) : boolean - net.pms.dlna.MapFile
		getVirtualFolders(ArrayList<String>) : List<DLNAResource> - net.pms.dlna.RootFolder
	PlaylistFolder(File) - net.pms.dlna.PlaylistFolder
		discoverChildren() : void - net.pms.dlna.virtual.MediaLibraryFolder
		doRefreshChildren() : void - net.pms.dlna.virtual.MediaLibraryFolder
		resolveOnce() : void - net.pms.dlna.PlaylistFolder
	PlaylistFolder(String, String, int) - net.pms.dlna.PlaylistFolder
		getPlaylist(String, String, int) : DLNAResource - net.pms.dlna.PlaylistFolder
	PlaySub(String, String, RealFile, String) - net.pms.dlna.PlaySub
	RarredEntry(String, File, String, long) - net.pms.dlna.RarredEntry
		RarredFile(File) - net.pms.dlna.RarredFile
			manageFile(File) : void - net.pms.dlna.MapFile
	RarredFile(File) - net.pms.dlna.RarredFile
		manageFile(File) : void - net.pms.dlna.MapFile
	RootFolder(ArrayList<String>) - net.pms.dlna.RootFolder
		getRoot(String, boolean, HttpExchange) : RootFolder - net.pms.remote.RemoteWeb
		getRootFolder() : RootFolder - net.pms.configuration.RendererConfiguration
		RootFolder() - net.pms.dlna.RootFolder
	SevenZipEntry(File, String, long) - net.pms.dlna.SevenZipEntry
		parse(String, String) : DLNAResource - net.pms.util.UMSUtils.IOList
		SevenZipFile(File) - net.pms.dlna.SevenZipFile
			manageFile(File) : void - net.pms.dlna.MapFile
	SevenZipFile(File) - net.pms.dlna.SevenZipFile
		manageFile(File) : void - net.pms.dlna.MapFile
	VirtualFolder(String, String) - net.pms.dlna.virtual.VirtualFolder
		addAdminFolder() : void - net.pms.dlna.RootFolder (4 matches)
			[constructor] RootFolder() - net.pms.dlna.RootFolder 
			[constructor] RootFolder(ArrayList<String>) - net.pms.dlna.RootFolder 
				getRoot(String, boolean, HttpExchange) : RootFolder - net.pms.remote.RemoteWeb
				getRootFolder() : RootFolder - net.pms.configuration.RendererConfiguration
				RootFolder() - net.pms.dlna.RootFolder
			[callers]
				discoverChildren() : void - net.pms.dlna.RootFolder
		addDynamicPls(DLNAResource) : void - net.pms.dlna.DLNAResource
			[constructor] DLNAResource() - net.pms.dlna.DLNAResource 
			[constructor] DLNAResource(int) - net.pms.dlna.DLNAResource 
				Feed(String, String, int) - net.pms.dlna.Feed
					AudiosFeed(String) - net.pms.dlna.AudiosFeed
						addWebFolder(File) : void - net.pms.dlna.RootFolder
					ImagesFeed(String) - net.pms.dlna.ImagesFeed
						addWebFolder(File) : void - net.pms.dlna.RootFolder
						manageItem() : void - net.pms.dlna.ImagesFeed
					VideosFeed(String) - net.pms.dlna.VideosFeed
						addWebFolder(File) : void - net.pms.dlna.RootFolder
				FeedItem(String, String, String, DLNAMediaInfo, int) - net.pms.dlna.FeedItem
					autoMatch(String, String) : DLNAResource - net.pms.dlna.DLNAResource
					manageItem() : void - net.pms.dlna.Feed
					resolveOnce() : void - net.pms.dlna.PlaylistFolder
				WebStream(String, String, String, int) - net.pms.dlna.WebStream
					parse(String, String) : DLNAResource - net.pms.util.UMSUtils.IOList
					setUp() : void - net.pms.util.PlayerUtilTest (3 matches)
					WebAudioStream(String, String, String) - net.pms.dlna.WebAudioStream
						addWebFolder(File) : void - net.pms.dlna.RootFolder
						autoMatch(String, String) : DLNAResource - net.pms.dlna.DLNAResource
						manageItem() : void - net.pms.dlna.AudiosFeed
						resolveOnce() : void - net.pms.dlna.PlaylistFolder
					WebVideoStream(String, String, String) - net.pms.dlna.WebVideoStream
						addWebFolder(File) : void - net.pms.dlna.RootFolder
						autoMatch(String, String) : DLNAResource - net.pms.dlna.DLNAResource
						manageItem() : void - net.pms.dlna.VideosFeed
						resolveOnce() : void - net.pms.dlna.PlaylistFolder
			[callers]
		addWebFolder(File) : void - net.pms.dlna.RootFolder
		ChapterFileTranscodeVirtualFolder(String, String, int) - net.pms.dlna.ChapterFileTranscodeVirtualFolder
			addChapterFolder(DLNAResource) : void - net.pms.dlna.FileTranscodeVirtualFolder
		CodeEnter(DLNAResource) - net.pms.dlna.CodeEnter
			addChild(DLNAResource, boolean) : void - net.pms.dlna.DLNAResource
			getVideoSettingsFolder() : DLNAResource - net.pms.dlna.RootFolder
		createApertureAlbum(Map<?, ?>, Map<?, ?>, List<?>) : VirtualFolder - net.pms.dlna.RootFolder
		createApertureDlnaLibrary(String) : VirtualFolder - net.pms.dlna.RootFolder
		DVDISOFile(File) - net.pms.dlna.DVDISOFile
			discoverChildren() : void - net.pms.dlna.virtual.MediaLibraryFolder
			doRefreshChildren() : void - net.pms.dlna.virtual.MediaLibraryFolder
			manageFile(File) : void - net.pms.dlna.MapFile
		FileTranscodeVirtualFolder(String, String, boolean) - net.pms.dlna.FileTranscodeVirtualFolder
		FileTranscodeVirtualFolder(String, String) - net.pms.dlna.FileTranscodeVirtualFolder
			addChild(DLNAResource, boolean) : void - net.pms.dlna.DLNAResource
		FolderLimit() - net.pms.dlna.FolderLimit
			discoverChildren() : void - net.pms.dlna.RootFolder
		FolderLimitLevel(int) - net.pms.dlna.FolderLimitLevel
			FolderLimit() - net.pms.dlna.FolderLimit
				discoverChildren() : void - net.pms.dlna.RootFolder
			setStart(DLNAResource) : void - net.pms.dlna.FolderLimit
		getApertureFolder() : DLNAResource - net.pms.dlna.RootFolder
		getiPhotoFolder() : DLNAResource - net.pms.dlna.RootFolder (2 matches)
		getiTunesFolder() : DLNAResource - net.pms.dlna.RootFolder (13 matches)
		getVideoSettingsFolder() : DLNAResource - net.pms.dlna.RootFolder (2 matches)
		init() : void - net.pms.dlna.virtual.MediaLibrary (3 matches)
		MediaLibrary() - net.pms.dlna.virtual.MediaLibrary
			init() : boolean - net.pms.PMS
		MediaLibraryFolder(String, String[], int[]) - net.pms.dlna.virtual.MediaLibraryFolder
			discoverChildren() : void - net.pms.dlna.virtual.MediaLibraryFolder
			doRefreshChildren() : void - net.pms.dlna.virtual.MediaLibraryFolder
			init() : void - net.pms.dlna.virtual.MediaLibrary (11 matches)
			MediaLibraryFolder(String, String, int) - net.pms.dlna.virtual.MediaLibraryFolder
				init() : void - net.pms.dlna.virtual.MediaLibrary (7 matches)
		MediaMonitor(File[]) - net.pms.dlna.MediaMonitor
			discoverChildren() : void - net.pms.dlna.RootFolder
		MonitorEntry(File, MediaMonitor) - net.pms.dlna.MonitorEntry
			scanDir(File[], DLNAResource) : void - net.pms.dlna.MediaMonitor
		Playlist(String, String, int, int) - net.pms.dlna.Playlist
			discoverChildren() : void - net.pms.dlna.RootFolder
			DynamicPlaylist(String, String, int) - net.pms.PMS.DynamicPlaylist
				getDynamicPls() : Playlist - net.pms.PMS
			Playlist(String, String) - net.pms.dlna.Playlist
				getPlaylist(String, String, int) : DLNAResource - net.pms.dlna.PlaylistFolder
			Playlist(String) - net.pms.dlna.Playlist
		Search(SearchObj, String) - net.pms.dlna.Search
			createSearcher(SearchObj, String) : void - net.pms.dlna.SearchFolder
			Search(SearchObj) - net.pms.dlna.Search
		SearchAction(Search, char, String) - net.pms.dlna.SearchAction
			createSearcher(SearchObj, String) : void - net.pms.dlna.SearchFolder (3 matches)
			SearchAction(Search, char) - net.pms.dlna.SearchAction
				createSearcher(SearchObj, String) : void - net.pms.dlna.SearchFolder (2 matches)
		SearchFolder(String, SearchObj) - net.pms.dlna.SearchFolder
			getConfiguredFolders(ArrayList<String>) : List<RealFile> - net.pms.dlna.RootFolder
			SearchFolder(SearchObj) - net.pms.dlna.SearchFolder
				analyzeChildren(int) : boolean - net.pms.dlna.MapFile
		SubSelect() - net.pms.dlna.SubSelect
			getSubSelector(boolean) : SubSelect - net.pms.dlna.DLNAResource
		SubSelFile(DLNAResource) - net.pms.dlna.SubSelFile
			addChild(DLNAResource, boolean) : void - net.pms.dlna.DLNAResource
		TranscodeVirtualFolder(String, PmsConfiguration) - net.pms.dlna.virtual.TranscodeVirtualFolder
			getTranscodeFolder(boolean) : TranscodeVirtualFolder - net.pms.dlna.DLNAResource
		unattachedFolder(String) - net.pms.dlna.DLNAResource.unattachedFolder
			Temp - net.pms.dlna.DLNAResource
	VirtualVideoAction(String, boolean) - net.pms.dlna.virtual.VirtualVideoAction
		addAdminFolder() : void - net.pms.dlna.RootFolder
		addDynamicPls(DLNAResource) : void - net.pms.dlna.DLNAResource
		CodeAction(String, boolean) - net.pms.dlna.CodeEnter.CodeAction
			addCharVVA(String) : void - net.pms.dlna.CodeEnter
			discoverChildren() : void - net.pms.dlna.CodeEnter
		discoverChildren() : void - net.pms.dlna.Playlist (2 matches)
		discoverChildren() : void - net.pms.dlna.RootFolder.addAdminFolder().new VirtualFolder() {...}
		discoverChildren() : void - net.pms.dlna.RootFolder.addAdminFolder().new VirtualFolder() {...}
		discoverChildren() : void - net.pms.dlna.RootFolder.addAdminFolder().new VirtualFolder() {...} (2 matches)
		getVideoSettingsFolder() : DLNAResource - net.pms.dlna.RootFolder (13 matches)
		scanDir(File[], DLNAResource) : void - net.pms.dlna.MediaMonitor
	ZippedEntry(File, String, long) - net.pms.dlna.ZippedEntry
		parse(String, String) : DLNAResource - net.pms.util.UMSUtils.IOList
		ZippedFile(File) - net.pms.dlna.ZippedFile
			manageFile(File) : void - net.pms.dlna.MapFile
	ZippedFile(File) - net.pms.dlna.ZippedFile
		manageFile(File) : void - net.pms.dlna.MapFile
```

Nadahar:
@SharkHunter I don't know if you're addressing me or not. You're basically reiterating what I've said, but in a definsive way. I've said multiple times that GlobalId is not to blame, it's the design. I don't agree with the design, and I fail to see what is so "killer" about it. It's slow, consuming a lot of memory and IO. If all this to be able to be "dynamic" I'm wondering what great benefit I'm not seeing. Do you mean that the great benefit is that it "refreshes" the folder every time you browse it? That could be solved in better ways imo. Is it the ability to add virtual folders "on demand" for misc. functionality, or is it to allow customizing the "resource tree" for each renderer with limiting the number of entries etc? The latter is rather buggy, and both of these could be solved by building a unique tree of references to shared resources for each renderer instead of actually building an unique set of resources for each renderer.

Changing all this is a huge task, and I'm not sure it's worth it - but I wouldn't defend the current design.

When it comes to "screaming thread-safety" I'm pretty sure you're addressing me. I know you think thread-safety are for whimps and that bugs are something to learn to live with, but I simply don't agree. I never considered the possibility that the resources were being recreated each time simply because it's such a waste. But, I've realized it now as a consequence of this "investigation" and that UMS keeps running out of memory for large libraries. That doesn't mean that there aren't concurrency issues however, there are. It's quite common that there are several requests for the same resource in a short periode of time, especially when using HTTPv2. I don't know what causes this, if it's the renderers spamming with requests or some other bug, but that happens quite a lot - and is probably the reason why I get much less bugs when using HTTPv1. So, even though not reusing the resources limits the concurrency issues somewhat, it doesn't eliminate them - and they are still a problem as I see it.

SharkHunter:
@SubJunk - If I understand correct your problem is when you scan all of your files on disc to make a nice database out of it. UMS/PMS is not made for that simple as that. You must build a tree and to do that you must do a render etc. You'll simple run into this sometime sooner or (rater) later if you have a big enough library to scan. Now can this be solved? Of course this is software. But I you'll need to change how resources are added. I would do the scanning outside of the regular discover path.

@Nadahar - you have no idea of what dynamic content is? Try web content. If you focus on disc based content than the discovery process can be much more simplified. Do you think that is this complicated because it's fun?

Nadahar:
@SharkHunter I asked you about what you meant by "dynamic content" for a reason. I'm still in the dark about what kind of web content that is used that makes this a reasonable design. To be able to discover new resources on said web source there will have to be a way to enumerate what's there. I can't see why this enumeration can't alter (add, remove, update) the resources already known instead of just throwing them all out and making new ones.

Actually, I don't think this is complicated at all. I think it's simple: the design is a "lazy" design where the easiest route is chosen, without giving the implications too much thought. That doesn't mean it's easy to change or that the code is easy to read, but that comes down to messy code and years of trying to make functionality that isn't really compatible with the design.

I've never used web sources myself as I have no idea what it's actually useful for. It would be nice if you'd enlighten me, as I must be missing something essential.

I'm not convinced that the ability to lazily resolve web resources is worth "disabling" decent disk based library functions. PMS had some media library (DB based or whatever) functionality that's still here although severly crippled by code added later that has ignored its needs. At what stage was the decision to abandon the media library and give preference to web based "dynamic" content taken?

SharkHunter:
@Nadahar - At no time. But normally people work with what they think is fun and useful. I don't find the medialibrary useful or fun so I didn't work on that. And since I didn't get any major bug reports on that it wasn't working I assumed (just like you assumed stuff) that no one else had any issues with it. The current scheme doesn't do anything to the medialibrary it simply has never been designed to cope with extremely large scans. The per render tree/discover mechanism has been there since start.
This design isn't "lazy" the easy way out would have been to force a scan to build up a sql based datbase where each physical file will get a unique id which then can be fed to renders. And just have a bunch of sql querys as folders. Just like most other DLNA server work. And thy have serious web content problems where you as a user must help the system finding the content (or update some plugin each time a new series i added on SVTPLay for example).

Nadahar:
@SharkHunter With "lazy" I mean the fact that change isn't dealt with and new resources are created instead. I doesn't have to do with whether folders are refreshed when entered or not. I'm not sure I see how fetching the resources from a database is "simpler", it depends on the implementation. If the implementation properly deals with changes, you could simply start a worker when a folder was entered that updated those resources - and you would have the "benefits of both models".

SubJunk:
So maybe the way to fix this and make everyone happy is to make the folder scanning more specific after all

SubJunk:
Ok so I have made it so that we don't have to do a full scan when things are changed in https://github.com/UniversalMediaServer/UniversalMediaServer/pull/942/commits/f41cd6cf411e2afd8838d2d528acf8b06e89989a#diff-397b9769dd54114e75ad412600defe5fR1471

There is one full scan when the program starts, because we don't know if things were changed on the hard drive while UMS was closed, so this gets us to a synchronized state with the hard drive, but from there we don't do it automatically anymore.

I'm still not convinced that this memory issue is really necessary, but at least resolving it isn't urgent anymore.

Nadahar:
@SubJunk I might have a "sort of" solution for this. If we replace use weak references to the ```DLNAResource``` instancesthey will be garbage collected when no other *strong* references to them exists anymore. That means that if no ```RootFolder``` or their children references a given ```DLNAResource``` it will be eligible for garbage collection even though they are referenced here. 

Before they are garbage collected ```WeakReference```s are nulled, so with some adjustment to the methods it should be easily handled if an attempt to get a "dead" ID.

Changing the ```List``` to a ```Map``` might also simplify this - by defining e.g 
```java
HashMap<Integer, WeakReference<DLNAResource>> ids = new HashMap<>();
```
we could get rid of the whole ```ID``` class and use the id a the key and the ```WeakReference<DLNAResource>>``` as the value. 

Using the ```Map``` methods for looking up ID's would also eliminate the need to implement the binary search and keeping the references needed for that. An automatic cleanup could also be performed by looping all entries and removing those holding a ```WeakReference``` whose ```get()``` returns null, although I don't know if that would be worth it. Simply holding the id numbers doesn't cost much.

SubJunk:
Cool solution, I like leaving it to GC if we can. However I'm not sure we will get much utility out of fixing this bug - it existed since the project started and I never heard about someone encountering it until we automated the scanning in 7.0 which we won't be doing.

So while all bugs are good to fix, this is probably quite a low priority compared with other ones.

PR #1192, opened 2017-02-13
Fix 9bcba891a3299f498c00aed3b5abc2d966bd66ff and then some

Nadahar:
This partially reverts 9bcba891a3299f498c00aed3b5abc2d966bd66ff. In addition I've made what I think is a more robust and versatile solution for subtitle colors in the new class ```SubtitleColor```.

@Sami32 Can you test if UMS will start for you with this PR?

@valib Do you think this is an ok solution? I haven't tested if the color parameters are correct for FFmpeg, Mencoder and ASS - I've just used what I found of documentation.

I've assumed that MEncoder use 0 as opaque while FFmpeg and ASS use 0xFF as opaque. If that is wrong, this needs some adjustments.

I'd appreciate if some of you would test this, as I'm not that into how to actually use this.

### DIFF
--- a/src/main/java/net/pms/util/SubtitleColor.java
+++ b/src/main/java/net/pms/util/SubtitleColor.java
@@ -1,0 +1,475 @@
+package net.pms.util;
+
+import java.awt.Color;
+import java.awt.color.ColorSpace;
+import java.util.HashMap;
+import java.util.Locale;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+@SuppressWarnings("serial")
+public class SubtitleColor extends Color {
+
+	private static final Logger LOGGER = LoggerFactory.getLogger(SubtitleColor.class);
+
+	protected static final Pattern hexPattern =
+		Pattern.compile("(?i)^\\s*(?:x'|%|#|0x|\\\\x|%x|\\$|h'|16#|16r|#x|#16r|&h|0h)([0-9a-f]{6}(?:[0-9a-f]{2})?)'?\\s*$");
+	protected static final Pattern decPattern =
+		Pattern.compile("^\\s*(-?\\d{1,10})\\s*$");
+	private static final Map<String, int[]> knownColors = new HashMap<>(140, 1f);
+
+	/**
+	 * @see Color#Color(int)
+	 */
+	public SubtitleColor(int rgb) {
+		super(rgb);
+	}
+
+	/**
+	 * @see Color#Color(int, boolean)
+	 */
+	public SubtitleColor(int rgba, boolean hasalpha) {
+		super(rgba, hasalpha);
+	}
+
+	/**
+	 * @see Color#Color(int, int, int)
+	 */
+	public SubtitleColor(int r, int g, int b) {
+		super(r, g, b);
+	}
+
+	/**
+	 * @see Color#Color(float, float, float)
+	 */
+	public SubtitleColor(float r, float g, float b) {
+		super(r, g, b);
+	}
+
+	/**
+	 * @see Color#Color(ColorSpace, float[], float)
+	 */
+	public SubtitleColor(ColorSpace cspace, float[] components, float alpha) {
+		super(cspace, components, alpha);
+	}
+
+	/**
+	 * @see Color#Color(int, int, int, int)
+	 */
+	public SubtitleColor(int r, int g, int b, int a) {
+		super(r, g, b, a);
+	}
+
+	/**
+	 * @see Color#Color(float, float, float, float)
+	 */
+	public SubtitleColor(float r, float g, float b, float a) {
+		super(r, g, b, a);
+	}
+
+	/**
+	 * Creates a {@link SubtitleColor} instance from a {@link Color} instance.
+	 *
+	 * @param color the {@link Color} to use for the new instance.
+	 */
+	public SubtitleColor(Color color) {
+		super(color.getRed(), color.getGreen(), color.getBlue(), color.getAlpha());
+	}
+
+	/**
+	 * @return A formatted hexadecimal {@link String} with the current value
+	 *         from this {@link SubtitleColor} instance suitable as an argument
+	 *         for FFmpeg.
+	 */
+	public String getFFmpegHexValue() {
+		return getHexValue("0x", "AABBGGRR", null, true, false);
+	}
+
+	/**
+	 * @return A formatted hexadecimal {@link String} with the current value
+	 *         from this {@link SubtitleColor} instance suitable as a ASS v4+
+	 *         styles parameter.
+	 */
+	public String getASSv4StylesHexValue() {
+		return getHexValue("&H", "AABBGGRR", null, true, true);
+	}
+
+	/**
+	 * @return A formatted hexadecimal {@link String} with the current value
+	 *         from this {@link SubtitleColor} instance suitable as a SSA/ASS
+	 *         parameter.
+	 */
+	public String getSSAHexValue() {
+		return getHexValue("&H", "BBGGRR", "&", true, true);
+	}
+
+	/**
+	 * @return A formatted hexadecimal {@link String} with the current value
+	 *         from this {@link SubtitleColor} instance suitable as a SSA/ASS
+	 *         parameter.
+	 */
+	public String getSSAAlphaHexValue() {
+		return getHexValue("&H", "AA", "&", true, true);
+	}
+
+	/**
+	 * @return A formatted hexadecimal {@link String} with the current value
+	 *         from this {@link SubtitleColor} instance suitable as an argument
+	 *         for MEncoder.
+	 */
+	public String getMEncoderHexValue() {
+		return getHexValue(null, "RRGGBBAA", null, true, true);
+	}
+
+	/**
+	 * @return A formatted hexadecimal {@link String} in the form
+	 *         {@code 0xRRGGBBAA}.
+	 */
+	public String get0xRRGGBBAA() {
+		return getHexValue("0x", "RRGGBBAA", null, true, false);
+	}
+
+	/**
+	 * @return A formatted hexadecimal {@link String} in the form
+	 *         {@code 0xRRGGBB}.
+	 */
+	public String get0xRRGGBB() {
+		return getHexValue("0x", "RRGGBB", null, true, true);
+	}
+
+	/**
+	 * Returns a formatted hexadecimal {@link String} with the current value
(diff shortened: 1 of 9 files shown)


Sami32:
@Nadahar What i can tell you for now:
- UMS can run normally
- The RGB color are correctly set, and alpha is displayed.
**- The color is not anymore memorized as before, the log display it, but the interface display white on the next restart.**

TODO: check if the alpha is displaying correctly.

Nadahar:
@Sami32 There was a small detail I had forgotten when saving the parameter. Saving should be ok now, you need to fetch and reset to get the update.

src/main/java/net/pms/util/SubtitleColor.java
```diff
@@ -0,0 +1,466 @@
...
+	public static int[] getNamedColorValues(String colorName) {
+		colorName = colorName.toUpperCase(Locale.ROOT);
+		if (knownColors.containsKey(colorName)) {
+			return knownColors.get(colorName);
+		}
+		return null;
+	}
+
+	static {
+		// The color values are from https://gist.github.com/XiaoxiaoLi/8031146
```
Sami32: This original reference should be better, IMO:
https://ffmpeg.org/ffmpeg-utils.html#toc-Color

Sami32:
@Nadahar This is what i found concerning ASS color specifications:
http://fileformats.wikia.com/wiki/SubStation_Alpha

> Color
>     Color values are expressed in hexadecimal BGR format as &HBBGGRR& or ABGR (with alpha channel) as &HAABBGGRR&. Transparency (alpha) can be expressed as &HAA&. Note that in the alpha channel, 00 is opaque and FF is transparent.

Nadahar:
@Sami32 Great. I've seen it specified both with and without the trailing ```&``` so I'm not sure of that. The whole ASS/SSA thing is a bit of a mess with additions by different people. I've used [this](moodub.free.fr/video/ass-specs.doc) among other sources, and it really doesn't define prefix and suffix. FFmpeg seems to want to prefix but not the suffix..?

It's very useful to know that they use inverted alpha - the question is then - is that true for FFmpeg as well? MEncoder use inverted alpha, so I guess it's pretty likely that FFmpeg does too?

Sami32:
@Nadahar I know this doc. BTW your link is wrong, you should do like me ;)
> \c&H<bbggrr>&	<bbggrr> is a hexadecimal RGB value, but in reverse order. Leading zeroes are not required.
> 					eg. {\c&HFF&}This is pure, full intensity red
> 						   {\c&HFF00&}This is pure, full intensity Green
> 						   {\c&HFF0000&}This is pure, full intensity Blue
> 						   {\c&HFFFFFF&}This is White
> 						   {\c&HA0A0A&}This is dark grey
> 
> 					\1c&Hbbggrr&, \2c&Hbbggrr&, \3c&Hbbggrr&, \4c&Hbbggrr& to set specific colors.
> 					\1a&Haa&, \2a&Haa&, \3a&Haa&, \4a&Haa& to set specific alpha channels.
> 					\alpha defaults to \1a

Nadahar:
@Sami32 > BTW your link is wrong

You should refresh your page more often, I corrected that 5 seconds after I made it ;)

Sami32:
@Nadahar But it was wrong anyway.
I'm not going to click refresh every 5 second...

Sami32:
@Nadahar From the FFmpeg link i posted above:
> It can be the name of a color as defined below (case insensitive match) or a [0x|#]RRGGBB[AA] sequence, possibly followed by @ and a string representing the alpha component.
> 
> The alpha component may be a string composed by "0x" followed by an hexadecimal number or a decimal number between 0.0 and 1.0, which represents the opacity value (‘0x00’ or ‘0.0’ means completely transparent, ‘0xff’ or ‘1.0’ completely opaque). If the alpha component is not specified then ‘0xff’ is assumed.

Nadahar:
> I'm not going to click refresh every 5 second...

I didn't say you had to, I said it was only wrong a very short time after I posted it, and had been correct for a long time when you wrote that. Posting it "your way" wouldn't have helped, because the problem was that Firefox, for some reason unknown to me, hadn't included ```http://``` when I copied the link.

To address what's relevant, FFmpeg does NOT have an inverted alpha. Again I'm confused by the documentation, because I looked at [this](https://ffmpeg.org/ffmpeg-filters.html#subtitles-1):
> To make the subtitles stream from sub.srt appear in transparent green DejaVu Serif, use:
> 
> ```subtitles=sub.srt:force_style='FontName=DejaVu Serif,PrimaryColour=&HAA00FF00'```

Sami32:
@Nadahar Correct, now you get me confuse as well, so i checked their code:

For SRT they use ASS code, it why i guess they use `&H` at the beginning but not the `&` at the end, and the reason why the alpha is also found &H00FF33FF in sSRT.

Nadahar:
@Sami32 So, what's the correct thing to use? I'm completely confused right now.

Nadahar:
It seems to me like both notation is used, ```&HAABBGGRR``` for "styles" and ```&HBBGGRR&``` when inline. What a mess, what is the correct type for us to use? It seems like we're using the "styles" variant since that's where the term "PrimaryColor" is used...?

Nadahar:
@valib @Sami32 I need some guidance to get the formats right for the different uses. I can figure it out myself, but then I have to spend time getting to know this that could be better used on other things I have to do. You both seem to know this much better than me.

Sami32:
So i didn't tested all, but something are not normal. I guess that was here since sometime and that i didn't take attention to it.
- "Choose subtitles color" subtab in the "Transcoding settings" tab is dependant of the computer locale, not from the UMS locale; that let the GUI be able to display in different language at the same time.
- Each time the Transparency or Alpha value is modified the "Green" value (in the RGB tab) is decrease by one.
- "Transparence" (HSL, HSB) and "Alpha" (RGB and CMYK) seem to be inverted value, it's maybe why they have different name ?

Nadahar:
@Sami32 I don't think we should start to fight with the color chooser - it's a Swing component over which we have no control, and it doesn't seem to properly support localization - or even have translations for most languages. See [here](http://stackoverflow.com/questions/7693158/jcolorchooser-and-localization) for some information about this.

Swing seems to have been abandoned - I don't think anything has been fixed there for years. The only option we have is probably to create our own color chooser, or find another ready-made component.

Sami32:
@Nadahar Thanks for the info, i didn't know that. So, i agree that this appearance issue should not be apriority at all.

About FFmpeg use, you are right `PrimaryColour=&HAA0000FF`, "ASS v4 styles", is used (AABBGGRR)
In PrimaryColour the &HAA is the alpha, 00 is fully opaque, and FF is fully transparent.
Note that it is in the opposite order of HTML colors (red would be FF0000 in HTML).

Nadahar:
@Sami32 I know about the color order, BGR and BGRA is actually pretty common, it's the "standard" on Windows. I've updated this with the information I've gotten so far (inverted transparancy as needed). I have no idea how to test how this actually works.

Sami32:
@Nadahar As you've a powerful computer, that's will be easy for you to test ;)

Just chooce a subtitle movie with subtitle that need to be transcoded.
In the transcoding settings tab, select "FFmpeg" and "Use font settings".
After that select "Common transcode settings" then "Subtitles settings".
From then play with the "Styled subtitle" parameters and stream the movie.

src/main/java/net/pms/util/SubtitleColor.java
```diff
@@ -0,0 +1,475 @@
...
+	 * property integer.</li>
+	 * </ul>
+	 * If the parsing fails, {@code null} is returned.
+	 *
+	 * @param subtitleColor the {@link String} to attempt to parse a
+	 *            {@link SubtitleColor} from.
+	 * @return A newly created {@link SubtitleColor} or {@code null} it the
+	 *         parsing fails.
+	 */
+	public static SubtitleColor toSubtitleColor(String subtitleColor) {
```
valib: @Nadahar nice job but there is a current implementation in the https://github.com/UniversalMediaServer/UniversalMediaServer/blob/master/src/main/java/net/pms/util/StringUtil.java#L256 which is almost similar and could be replaced by this. 👍

Nadahar:
@Sami32 @valib Ok, I've managed to "style" some SRT subtitles when burning in. Both color and alpha works correcly for both FFmpeg and MEncoder. But, how do I test the 3. option? Isn't it something with styling subtitles without burning in?

Sami32:
@Nadahar I guess i'm lost. What is the test 3 ?

Nadahar:
@Sami32 I'm not sure, that's why I asked you guys. The code calls for the subtitle color from 3 different places, one of them is in ```SubtitleUtils.getSubtitles()``` and that's the one I don't know how to test. It seems like it "translates" the subtitles before they are sent or something like that.

Sami32:
@Nadahar Ah, i see. I didn't now that one, as i never used. No computer power or hardware for that :wink: 
It's for 3D subtitles, so you'll need to get some if you want test while transcoding them.

Nadahar:
The 3. method seems to only apply to 3D subtitles. I didn't quite understand how it works, but I managed to force the code to run there and it seems to work correct as well.

src/main/java/net/pms/util/SubtitleColor.java
Nadahar: @valib I took a look at that implementation, but it doesn't quite do the same (although almost). It's not used for subtitle color but for color in general and thus returns a ```Color``` instead of a ```SubtitleColor```. It would certainly be possible to merge the two into one, but the implementation looks ok and I think they can stay separate for now.

Nadahar:
Ok, I think this should be good to go then.

valib:
@Nadahar what I am asking for has nothing to do with subtitles. When you made a SubtitleColor class the implementation of `toSubtitleColor` is so general that can be used in the class `getColor` in https://github.com/UniversalMediaServer/UniversalMediaServer/blob/master/src/main/java/net/pms/configuration/RendererConfiguration.java#L360 which is used in the UMS GUI. Maybe the SubtitleColor class could be renamed to `UMSColor` to be more general.

Nadahar:
@valib Yes, I understand that, but the class I made was for subtitle colors specificly. Because of that I though it would be wrong to use if for general colors. The SubtitleColor class has some subtitle specific methods.
```SubtitleColor``` would have to be split into a super and a sub class, where the general stuff was in the super class and the subtitle specific in the sub class. 

The only advantage I could see from doing all that is the ability to use the extended list of color names, but is that really that important?

Nadahar:
@valib It seems like there's [still some issue with subtitle colors](http://www.universalmediaserver.com/forum/viewtopic.php?f=9&t=10295). Can you have a look?

PR #1196, opened 2017-02-15
Revise b235331: backward compatible + less restrictive

skeptical:
Just suggestions following @Nadahar's changes to the API, not necessarily for mergeing :-).

### DIFF
--- a/src/main/java/net/pms/dlna/DLNAMediaSubtitle.java
+++ b/src/main/java/net/pms/dlna/DLNAMediaSubtitle.java
@@ -186,6 +186,11 @@ public class DLNAMediaSubtitle extends DLNAMediaLang implements Cloneable {
 		setFileSubsCharacterSet(forcedLang);
 	}
 
+	// Backward compatibility
+	public void setExternalFile(File externalFile) throws FileNotFoundException {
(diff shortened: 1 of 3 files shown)


Nadahar:
@skeptical I think you and I have different definitions of "avoidable". It's always avoidable, just leave the code as it is - but the price is that nothing will evolve/improve.

I've "countered" this PR with an update of skeptical/jumpy#3. I'm not able to test it though, because there is so much I don't know about configuring jumpy and your file/folder structure is very incompatible with Eclipse - meaning I don't know how to get debugging up and running.

I think it should work though, and if not the concept should still be valid. I think this is a better approach conceptually, even though your take on synchronization om this PR would work for all but external iteration. It's a lot less efficient though since it doesn't allow concurrent reads, but I doubt that matters much in this case.

I don't think we should make API changes for no good reason, although this is a bit of a hopeless situation because the API is basically "everything". Non of us, except maybe you, know what parts are used by plugins and not, thus avoiding to break stuff is very difficult.

I agree with your change to ```DLNAMediaSubtitle``` in this PR, that's an unnecessary API change in my view as well.

PR #1198, opened 2017-02-16
Split SubtitleColor into FormattableColor and SubtitleColor

Nadahar:
I've done what I think @valib asked for in #1192. Is this ok @valib?


Nadahar:
These were so simple, low-risk changes that I don't think there was a reason to test/review more.

PR #1199, opened 2017-02-17
Some fixes in RendererConfiguration

Nadahar:
I'm not sure if this should be merged or not. What happened was that I was dealing with something completely unrelated, and saw [this line](https://github.com/UniversalMediaServer/UniversalMediaServer/blob/7e347199cb9fcb83de40725eeb9e38445807bf77/src/main/java/net/pms/configuration/RendererConfiguration.java#L1623). With the memory issues (#1185) in the back of my mind I thought "maybe we can save some memory by freeing some references here". 

But, when I looked more at it, I'm not sure what the original author meant, how to "delete" rootfolder. All one can do in Java is to dereference it, unless he had some other cleanup procedure in mind. I guess dereferencing it is better than doing nothing, but this obviously has nothing to do with the "core" cause for #1185. Still, cleaning up as much as possible should be a good thing.

The reason I'm reluctant is that ```RendererConfiguration``` is fundamentally full of thread-unsafety. Different threads will be running these methods, and I'm simply afraid that doing anything extra can lead to new and mysterious bugs. The likelihood of this causing issues should still be marginal.

When I were at it, I fixed some of the other bugs FindBugs warned me about. Neither of these are revolutionary, there's some performance fixes, a NPE fix and some other minor stuff.

### DIFF
--- a/src/main/java/net/pms/configuration/RendererConfiguration.java
+++ b/src/main/java/net/pms/configuration/RendererConfiguration.java
@@ -8,7 +8,9 @@ import java.io.File;
 import java.io.IOException;
 import java.io.Reader;
 import java.net.InetAddress;
+import java.nio.charset.StandardCharsets;
 import java.util.*;
+import java.util.Map.Entry;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 import java.util.regex.Matcher;
@@ -33,7 +35,6 @@ import net.pms.util.StringUtil;
 import org.apache.commons.configuration.Configuration;
 import org.apache.commons.configuration.ConfigurationException;
 import org.apache.commons.configuration.PropertiesConfiguration;
-import org.apache.commons.io.Charsets;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang.WordUtils;
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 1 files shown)


Nadahar:
These are mostly findbugs fixes, variable renaming and some logging improvements. It's also a small step towards better cleanup of resources.

I think the risk is really minimal, but some more testing should probably be done..?

Sami32:
I will try it after @valib subs PR, but i'll appreciate any testing advice, as i'll test if the renderer configuration, streaming is working as expected and without any error in the log.

Nadahar:
@Sami32 This is a bit hard to test - the thing I'm a little bit worried about here is how it handles working with multiple renderers. Using e.g two renderers and connecting and disconnecting them several times while browsing each time one is connected is the "best way" I can think of to test it.

The behaviour if it fails is also hard to predict, it could be exceptions thrown in the log, it could be deadlock/freezing and it could be strange results when browsing or unusual memory usage. It could also lead to errors that you wouldn't detect, and would only manifest itself by strange behaviour in seemingly unrelated code. That's why I'm so bent on taking thread-safety seriously, it's very hard to both prove and disprove.

There's no way we can test this "properly", so we have to just do some testing to make sure nothing obvious goes wrong. I think the likelyhood that this causes additional issues is very low though. There are already issues with the things I mentioned above, so if one is detected it would be far from certain that it came from this PR.

PR #1205, opened 2017-02-21
Samsung galaxy note tab

Sami32:
http://www.universalmediaserver.com/forum/viewtopic.php?f=9&t=9213

As i saw that the others phones don't apply the left rotation...so i did the same.


Nadahar:
@Sami32 What left rotation?

Sami32:
The one explained in the read me text, perhaps with other words though...

Nadahar:
@Sami32 Sorry, I don't know what you're referencing, but it's probably not important that I understand since you're probably addressing someone else.

Issue #1123, opened 2017-01-03
Out of Sync issue when using embeded subtitles


n00ble:
Thanks for great software

This may be an old issue. I have basic skills so I was not able to check of duplicate reporting. I tried to find a solution to no avail. I have uploaded the shortest file that does this too.

My wife recently bought an old series "3rd Rock From The Sun". I ripped the CDs so to be able to view them easier on the TV. I used Handbrake 1.0.1 and included the subtitles. I have tried to change the ripping variables to no avail. I have also tried to disable "Keep AC-3 tracks" in "Audio Settings" under "Transcoding Settings", but the out of sync still occurs.

The movie starts with the fast forward and backward options greyed out. Few seconds later it gets like a hiccup and things go out of sync. I tried to run UMS with Subtitles disabled, and the voice is in-sync but it exits the episode like 5 minutes later...

I am using UMS on a WiFi network with a Samsung TV (UA50JU6400).
I have uploaded the shortest file: "UMS Bug Out of Sync 3rd Rock S6d1-15.mp4"

"UMS Bug Out of Sync 3rd Rock S6d1-15.7z"
MD5: f8a738c89e726a28fbcceb87924aedc7
http://ge.tt/7arYU9i2

Please let me know if there is a way to avoid these issues.

Thnx

n00ble:
The same bug is still present in version 6.5.3

n00ble:
The same issue is still present in version 6.6.0

Nadahar:
@n00ble I'm not expert at transcoding issues, but it seems you have to little memory:
```powershell
Muxer frame buffer cannot allocate memory!
```

2.2 GB on Windows 10 is extremely low, you should upgrade to atleast 4 GB or downgrade to a less memory-hungry OS.

You can also try to turn of the "Defer to MEncoder" option for FFmpeg in the transcoding settings so that FFmpeg will be used instead, but I'm not sure anything can work with this little memory.

Sami32:
@n00ble I agree on that:
Deselect the "Defer to MEncoder when transcoding problematic subtitles" option found in the "Transcoding settings" tab in "FFmpeg" video files engines section.

Avoid MEncoder, as it an old encoder far less efficient that FFmpeg, and near to not maintened anymore. So i'll suggest you to use only for CD/DVD discs.
And as you see it give you unecessary errors as well.

I recommend you to force the renderer called "Samsung JU6400", that will fit your renderer much better.

[...]

PR #1171, opened 2017-01-28
Updated JNA to 4.4.0

SubJunk:
Does anyone anticipate an issue with this update?

PR #1208, opened 2017-02-27
Additional memory and CPU logging

Nadahar:
My mistake in #1123 made me think that we should have more info available for debugging. I recalled that @valib had done something like that, and cherry-picked 7a84789419eb3316bd082eccde18061b96524236 from #979/#1091.

I've further modified the logging, and added ```StringUtil.formatBytes()``` to easily format byte values for output.

### DIFF
--- a/src/main/java/net/pms/util/StringUtil.java
+++ b/src/main/java/net/pms/util/StringUtil.java
@@ -64,6 +64,18 @@ public class StringUtil {
 	public static final String SEC_TIME_FORMAT = "%02d:%02d:%02.0f";
 	public static final String DURATION_TIME_FORMAT = "%02d:%02d:%05.2f";
 	public static final String NEWLINE_CHARACTER = System.getProperty("line.separator");
+	public static final long KIBI = 1L << 10;
+	public static final long MEBI = 1L << 20;
+	public static final long GIBI = 1L << 30;
+	public static final long TEBI = 1L << 40;
+	public static final long PEBI = 1L << 50;
+	public static final long EXBI = 1L << 60;
+	public static final long KILO = 1000L;
+	public static final long MEGA = 1000000L;
+	public static final long GIGA = 1000000000L;
+	public static final long TERA = 1000000000000L;
+	public static final long PETA = 1000000000000000L;
+	public static final long EXA  = 1000000000000000000L;
 
 	/**
 	 * Appends "&lt;<u>tag</u> " to the StringBuilder. This is a typical HTML/DIDL/XML tag opening.
@@ -358,7 +370,7 @@ public class StringUtil {
 		Pattern pattern = Pattern.compile("<body>(.*)</body>", Pattern.CASE_INSENSITIVE + Pattern.DOTALL);
 		Matcher matcher = pattern.matcher(html);
 		if (matcher.find()) {
-			return matcher.group(1).replaceAll("\n    ", "").trim().replaceAll("(?i)<br>", "\n").replaceAll("<.*?>","");
+			return matcher.group(1).replaceAll("\n    ", "").trim().replaceAll("(?i)<br>", "\n").replaceAll("<.*?>", "");
 		}
 		throw new IllegalArgumentException("HTML text not as expected, must have <body> section");
 	}
@@ -375,10 +387,10 @@ public class StringUtil {
 	}
 
 	/**
-	 * Escapes {@link org.apache.lucene} special characters with backslash
+	 * Escapes {@link org.apache.lucene} special characters with backslash.
 	 *
-	 * @param s the {@link String} to evaluate
-	 * @return The converted String
+	 * @param s the {@link String} to evaluate.
+	 * @return The converted String.
 	 */
 	@SuppressFBWarnings("SF_SWITCH_NO_DEFAULT")
 	public static String luceneEscape(final String s) {
@@ -415,10 +427,10 @@ public class StringUtil {
 	}
 
 	/**
-	 * Escapes special characters with backslashes for FFmpeg subtitles
+	 * Escapes special characters with backslashes for FFmpeg subtitles.
 	 *
-	 * @param s the {@link String} to evaluate
-	 * @return The converted String
+	 * @param s the {@link String} to evaluate.
+	 * @return The converted String.
 	 */
 	public static String ffmpegEscape(String s) {
 		StringBuilder sb = new StringBuilder();
@@ -447,7 +459,21 @@ public class StringUtil {
 		return sb.toString();
 	}
 
-	public static String prettifyXML(String xml, int indentWidth) throws SAXException, ParserConfigurationException, XPathExpressionException, TransformerException {
+	/**
+	 * Formats a XML string to be easier to read with newlines and indentations.
+	 *
+	 * @param xml the {@link String} to "prettify".
+	 * @param indentWidth the width of one indentation in number of characters.
+	 * @return The "prettified" {@link String}.
+	 * @throws SAXException If a parsing error occurs.
+	 * @throws ParserConfigurationException If a parsing error occurs.
+	 * @throws XPathExpressionException If a parsing error occurs.
+	 * @throws TransformerException If a parsing error occurs.
+	 */
+	public static String prettifyXML(
+		String xml,
+		int indentWidth
+	) throws SAXException, ParserConfigurationException, XPathExpressionException, TransformerException {
 		try {
 			// Turn XML string into a document
 			Document xmlDocument =
@@ -576,5 +602,60 @@ public class StringUtil {
 			sb.append(strings[i]);
 		}
 		return sb.toString();
+	}
+
+	/**
+	 * Formats bytes into a rounded {@link String} representation in either
+	 * binary/power of 2 or SI notation using {@link Locale#ROOT}.
+	 *
+	 * @param bytes the value to format.
+	 * @param binary whether the representation should be binary/power of 2 or
+	 *            SI/metric.
+	 * @return The formatted byte value and unit.
+	 */
+	public static String formatBytes(long bytes, boolean binary) {
+		return formatBytes(bytes, binary, Locale.ROOT);
+	}
+
+	/**
+	 * Formats bytes into a rounded {@link String} representation in either
+	 * binary/power of 2 or SI notation.
+	 *
+	 * @param bytes the value to format.
+	 * @param binary whether the representation should be binary/power of 2 or
+	 *            SI/metric.
+	 * @param locale the {@link Locale} to use when formatting.
+	 * @return The formatted byte value and unit.
+	 */
+	public static String formatBytes(long bytes, boolean binary, Locale locale) {
+		if ((binary && bytes < 1L << 10) || bytes < KILO) {
+			return String.format("%d %s", bytes, bytes == 1L ? "byte" : "bytes");
+		}
+
+		long divisor;
+		String unit;
+		if ((binary && bytes < MEBI) || bytes < MEGA) { // kibi/kilo
+			divisor = binary ? KIBI : KILO;
+			unit = binary ? "KiB" : "kB";
+		} else if ((binary && bytes < GIBI) || bytes < GIGA) { // mebi/mega
+			divisor = binary ? MEBI : MEGA;
+			unit = binary ? "MiB" : "MB";
+		} else if ((binary && bytes < TEBI) || bytes < TERA) { // gibi/giga
+			divisor = binary ? GIBI : GIGA;
+			unit = binary ? "GiB" : "GB";
+		} else if ((binary && bytes < PEBI) || bytes < PETA) { // tebi/tera
+			divisor = binary ? TEBI : TERA;
+			unit = binary ? "TiB" : "TB";
+		} else if ((binary && bytes < EXBI) || bytes < EXA) { // pebi/peta
+			divisor = binary ? PEBI : PETA;
+			unit = binary ? "PiB" : "PB";
+		} else { // exbi/exa
+			divisor = binary ? EXBI : EXA;
+			unit = binary ? "EiB" : "EB";
+		}
+		if (bytes % divisor == 0) {
+			return String.format(locale, "%d %s", bytes / divisor, unit);
+		}
+		return String.format(locale, "%.1f %s", (double) bytes / divisor, unit);
 	}
 }
--- a/pom.xml
+++ b/pom.xml
@@ -74,7 +74,7 @@
 
 		<!-- Skip 4.65-1.05-RC because it doesn't work, author is working on fixing it -->
 		<sevenzip-version>4.65-1.04-RC</sevenzip-version>
-		
+
 		<twelvemonkeys-imageio-version>3.3.2</twelvemonkeys-imageio-version>
 
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
@@ -87,8 +87,25 @@
 		<maven-javadoc-plugin-version>2.10.4</maven-javadoc-plugin-version>
 		<git-commit-id-plugin-version>2.2.2</git-commit-id-plugin-version>
 		<maven-assembly-plugin-version>3.0.0</maven-assembly-plugin-version>
+
+		<!--
+			org.slf4j:slf4j-api is shared with
+				- su.litvak.chromecast:api-v2
+				- fm.last:coverartarchive-api
+				- com.github.oshi:oshi-core
+
+			Check the org.slf4j:slf4j-api version of the other libraries before upgrading
+		-->
 		<logback-version>1.2.3</logback-version>
 		<surefire-version>2.20</surefire-version>
+
+		<!--
+			net.java.dev.jna:jna-platform is shared with
+				- com.github.oshi:oshi-core
+
+			Check the net.java.dev.jna:jna-platform version of the other library before upgrading
+		-->
+		<jna-version>4.4.0</jna-version>
 
 		<!--
 			 use the Windows makensis.exe for Windows builds unless
@@ -133,26 +150,35 @@
 		<dependency>
 			<groupId>commons-configuration</groupId>
 			<artifactId>commons-configuration</artifactId>
+			<!--
+				commons-logging:commons-logging is shared with
+					- commons-httpclient:commons-httpclient
+					- com.github.junrar:junrar
+					- fm.last:coverartarchive-api
+
+				Check the commons-logging:commons-logging version of the other libraries before upgrading
+			-->
 			<version>1.10</version>
-			<exclusions>
-				<!--  no need for commons-logging, as jcl-over-slf4j provides the impl -->
-				<exclusion>
-					<artifactId>commons-logging</artifactId>
-					<groupId>commons-logging</groupId>
-				</exclusion>
-			</exclusions>
 		</dependency>
 		<dependency>
 			<groupId>commons-httpclient</groupId>
 			<artifactId>commons-httpclient</artifactId> <!-- TODO this should be changed to org.apache.httpcomponents which is the follower. -->
 			<version>3.1</version>
 			<exclusions>
-				<!--  no need for commons-logging, as jcl-over-slf4j provides the impl -->
 				<exclusion>
+					<!--
+						commons-logging:commons-logging is shared with
+							- commons-configuration:commons-configuration
+							- com.github.junrar:junrar
+							- fm.last:coverartarchive-api
+
+						The other libraries use a much newer version since commons-httpclient is ancient and abandoned
+					-->
 					<groupId>commons-logging</groupId>
 					<artifactId>commons-logging</artifactId>
 				</exclusion>
 				<exclusion>
+					<!-- fm.last:coverartarchive-api uses a much newer version since commons-httpclient is ancient and abandoned -->
 					<groupId>commons-codec</groupId>
 					<artifactId>commons-codec</artifactId>
 				</exclusion>
@@ -161,6 +187,12 @@
 		<dependency>
 			<groupId>commons-io</groupId>
 			<artifactId>commons-io</artifactId>
+			<!--
+				commons-io:commons-io is shared with
+					- fm.last:coverartarchive-api
+
+				Check the commons-io:commons-io version of the library before upgrading
+			-->
 			<version>2.5</version>
 		</dependency>
 		<dependency>
@@ -211,8 +243,8 @@
 			<artifactId>thumbnailator</artifactId>
 			<version>[0.4, 0.5)</version>
 		</dependency>
-		
-		<!-- 
+
+		<!--
 			If any ImageIO plugins are added or removed, corresponding changes
 		 	must be made in src/main/external-resources/META-INF/services
 		 -->
@@ -290,7 +322,7 @@
 			<groupId>com.drewnoakes</groupId>
 			<artifactId>metadata-extractor</artifactId>
 			<version>2.10.1</version>
-		</dependency>		
+		</dependency>
 		<dependency>
 			<groupId>net.jthink</groupId>
 			<artifactId>jaudiotagger</artifactId>
@@ -309,6 +341,14 @@
 		<dependency>
 			<groupId>com.github.junrar</groupId>
 			<artifactId>junrar</artifactId>
+			<!--
+				commons-logging:commons-logging is shared with
+					- commons-configuration:commons-configuration
+					- commons-httpclient:commons-httpclient
+					- fm.last:coverartarchive-api
+
+				Check the commons-logging:commons-logging version of the other libraries before upgrading
+			-->
 			<version>0.7</version>
 		</dependency>
 
@@ -355,12 +395,12 @@
 		<dependency>
 			<groupId>net.java.dev.jna</groupId>
 			<artifactId>jna</artifactId>
-			<version>4.4.0</version>
+			<version>${jna-version}</version>
 		</dependency>
 		<dependency>
 			<groupId>net.java.dev.jna</groupId>
 			<artifactId>jna-platform</artifactId>
-			<version>4.4.0</version>
+			<version>${jna-version}</version>
 		</dependency>
 		<dependency>
 			<groupId>org.fourthline.cling</groupId>
@@ -395,6 +435,14 @@
 		<dependency>
 			<groupId>su.litvak.chromecast</groupId>
 			<artifactId>api-v2</artifactId>
+			<!--
+				org.slf4j:slf4j-api is shared with
+					- ch.qos.logback:logback-classic
+					- fm.last:coverartarchive-api
+					- com.github.oshi:oshi-core
+
+				Check the org.slf4j:slf4j-api version of the other libraries before upgrading
+			-->
 			<version>0.10.0</version>
 		</dependency>
 		<dependency>
@@ -410,8 +458,29 @@
 		<dependency>
 			<groupId>fm.last</groupId>
 			<artifactId>coverartarchive-api</artifactId>
+			<!--
+				org.slf4j:slf4j-api is shared with
+					- ch.qos.logback:logback-classic
+					- su.litvak.chromecast
+					- com.github.oshi:oshi-core
+
+				Check the org.slf4j:slf4j-api version of the other libraries before upgrading
+
+				commons-logging:commons-logging is shared with
+					- commons-configuration:commons-configuration
+					- commons-httpclient:commons-httpclient
+					- com.github.junrar:junrar
+
+				Check the commons-logging:commons-logging version of the other libraries before upgrading
+
+				commons-io:commons-io is shared with
+					- net.pms:ums
+
+				Check the commons-io:commons-io version before upgrading
+			-->
 			<version>2.1.0</version>
 			<exclusions>
+				<!-- This should never have been a dependency in the first place -->
 				<exclusion>
 					<groupId>com.google.code.findbugs</groupId>
 					<artifactId>jsr305</artifactId>
@@ -425,6 +494,19 @@
 		<dependency>
 			<groupId>com.github.oshi</groupId>
 			<artifactId>oshi-core</artifactId>
+			<!--
+				org.slf4j:slf4j-api is shared with
+					- ch.qos.logback:logback-classic
+					- su.litvak.chromecast
+					- fm.last:coverartarchive-api
+
+				Check the org.slf4j:slf4j-api version of the other libraries before upgrading
+
+				net.java.dev.jna:jna-platform is shared with
+					- net.pms:ums
+
+				Check the net.java.dev.jna:jna-platform version before upgrading
+			-->
 			<version>3.4.2</version>
 			<exclusions>
 				<exclusion>
@@ -539,9 +621,9 @@
 					<version>${surefire-version}</version>
 				</plugin>
 
-				<!-- 
+				<!--
 					This plugin's configuration is used to run crowdin plugin from the command line,
-					it has no influence on the Maven build itself. 
+					it has no influence on the Maven build itself.
 				-->
 				<plugin>
 					<groupId>com.universalmediaserver</groupId>
@@ -618,7 +700,7 @@
 										</artifactId>
 										<versionRange>
 											[0.1,)
-										</versionRange>										
+										</versionRange>
 										<goals>
 											<goal>clean</goal>
 											<goal>deploy</goal>
@@ -1418,7 +1500,7 @@
 								</configuration>
 							</execution>
 							<!--
-								Assemble the UMS files in the right place before 
+								Assemble the UMS files in the right place before
 								moving them to the app Bundle.
 							-->
 							<execution>
@@ -1595,7 +1677,7 @@
 								</configuration>
 							</execution>
 							<!--
-								Assemble the UMS files in the right place before 
+								Assemble the UMS files in the right place before
 								moving them to the app Bundle.
 							-->
 							<execution>
--- a/pom.xml
+++ b/pom.xml
@@ -422,6 +422,17 @@
 				</exclusion>
 			</exclusions>
 		</dependency>
+		<dependency>
+			<groupId>com.github.oshi</groupId>
+			<artifactId>oshi-core</artifactId>
+			<version>3.4.2</version>
+			<exclusions>
+				<exclusion>
+					<groupId>org.slf4j</groupId>
+					<artifactId>slf4j-api</artifactId>
+				</exclusion>
+			</exclusions>
+		</dependency>
 	</dependencies>
 	<build>
 		<defaultGoal>assembly:assembly</defaultGoal>
--- a/pom.xml
+++ b/pom.xml
@@ -431,6 +431,10 @@
 					<groupId>org.slf4j</groupId>
 					<artifactId>slf4j-api</artifactId>
 				</exclusion>
+				<exclusion>
+					<groupId>net.java.dev.jna</groupId>
+					<artifactId>jna-platform</artifactId>
+				</exclusion>
 			</exclusions>
 		</dependency>
 	</dependencies>
--- a/src/main/external-resources/logback.headless.xml
+++ b/src/main/external-resources/logback.headless.xml
@@ -81,6 +81,7 @@
 	<logger name="org.apache" level="WARN" />
 	<logger name="httpclient.wire" level="WARN" />
 	<logger name="fm.last.musicbrainz.coverart" level="WARN" />
+	<logger name="oshi" level="WARN" />
 
 	<!-- Root logger defines the minimum log level for all loggers -->
 	<root level="${rootLevel}">
--- a/src/main/external-resources/logback.xml
+++ b/src/main/external-resources/logback.xml
@@ -90,6 +90,7 @@
 	<logger name="org.apache" level="WARN" />
 	<logger name="httpclient.wire" level="WARN" />
 	<logger name="fm.last.musicbrainz.coverart" level="WARN" />
+	<logger name="oshi" level="WARN" />
 
 	<!-- Root logger defines the minimum log level for all loggers -->
 	<root level="${rootLevel}">
--- a/src/main/java/net/pms/PMS.java
+++ b/src/main/java/net/pms/PMS.java
@@ -69,6 +69,10 @@ import net.pms.newgui.*;
 import net.pms.remote.RemoteWeb;
 import net.pms.update.AutoUpdater;
 import net.pms.util.*;
+import oshi.SystemInfo;
+import oshi.hardware.CentralProcessor;
+import oshi.hardware.GlobalMemory;
+import oshi.hardware.HardwareAbstractionLayer;
 import org.apache.commons.configuration.ConfigurationException;
 import org.apache.commons.configuration.event.ConfigurationEvent;
 import org.apache.commons.configuration.event.ConfigurationListener;
@@ -1414,12 +1418,17 @@ public class PMS {
 	 */
 	private void logSystemInfo() {
 		long memoryInMB = Runtime.getRuntime().maxMemory() / 1048576;
+		HardwareAbstractionLayer hal = new SystemInfo().getHardware();
+		CentralProcessor processor = hal.getProcessor();
+		GlobalMemory memory = hal.getMemory();
 
 		LOGGER.info("Java: " + System.getProperty("java.vm.name") + " " + System.getProperty("java.version") + " " + System.getProperty("sun.arch.data.model") + "-bit" + " by " + System.getProperty("java.vendor"));
 		LOGGER.info("OS: " + System.getProperty("os.name") + " " + getOSBitness() + "-bit " + System.getProperty("os.version"));
-		LOGGER.info("Encoding: " + System.getProperty("file.encoding"));
-		LOGGER.info("Memory: {} MB", memoryInMB);
+		LOGGER.info("Physical memory available: {} MB", memory.getAvailable() / 1048576);
+		LOGGER.info("Memory accessible by Java: {} MB", memoryInMB);
+		LOGGER.info("CPU: " + processor.getName());
 		LOGGER.info("Language: " + WordUtils.capitalize(PMS.getLocale().getDisplayName(Locale.ENGLISH)));
+		LOGGER.info("Encoding: " + System.getProperty("file.encoding"));
 		LOGGER.info("");
 
 		if (Platform.isMac()) {
--- a/src/main/java/net/pms/PMS.java
+++ b/src/main/java/net/pms/PMS.java
@@ -19,6 +19,7 @@
 
 package net.pms;
 
+import static org.apache.commons.lang3.StringUtils.isNotBlank;
 import ch.qos.logback.classic.Level;
 import ch.qos.logback.classic.LoggerContext;
 import com.sun.jna.Platform;
@@ -73,6 +74,7 @@ import oshi.SystemInfo;
 import oshi.hardware.CentralProcessor;
 import oshi.hardware.GlobalMemory;
 import oshi.hardware.HardwareAbstractionLayer;
+import oshi.software.os.OperatingSystem;
 import org.apache.commons.configuration.ConfigurationException;
 import org.apache.commons.configuration.event.ConfigurationEvent;
 import org.apache.commons.configuration.event.ConfigurationListener;
@@ -874,20 +876,17 @@ public class PMS {
 		return mediaLibrary;
 	}
 
-	private SystemUtils createSystemUtils() {
+	private static SystemUtils createSystemUtils() {
 		if (Platform.isWindows()) {
 			return new WinUtils();
-		} else {
-			if (Platform.isMac()) {
-				return new MacSystemUtils();
-			} else {
-				if (Platform.isSolaris()) {
-					return new SolarisUtils();
-				} else {
-					return new BasicSystemUtils();
-				}
-			}
 		}
+		if (Platform.isMac()) {
+			return new MacSystemUtils();
+		}
+		if (Platform.isSolaris()) {
+			return new SolarisUtils();
+		}
+		return new BasicSystemUtils();
 	}
 
 	/**
@@ -1416,19 +1415,50 @@ public class PMS {
 	 * Log system properties identifying Java, the OS and encoding and log
 	 * warnings where appropriate.
 	 */
-	private void logSystemInfo() {
-		long memoryInMB = Runtime.getRuntime().maxMemory() / 1048576;
-		HardwareAbstractionLayer hal = new SystemInfo().getHardware();
-		CentralProcessor processor = hal.getProcessor();
-		GlobalMemory memory = hal.getMemory();
+	private static void logSystemInfo() {
+		long jvmMemory = Runtime.getRuntime().maxMemory();
+		OperatingSystem os = null;
+		CentralProcessor processor = null;
+		GlobalMemory memory = null;
+		try {
+			SystemInfo systemInfo = new SystemInfo();
+			HardwareAbstractionLayer hardware = systemInfo.getHardware();
+			os = systemInfo.getOperatingSystem();
+			processor = hardware.getProcessor();
+			memory = hardware.getMemory();
+		} catch (Error e) {
+			LOGGER.debug("Could not retrieve OS and hardware information: {}", e.getMessage());
+			LOGGER.trace("", e);
+		}
 
-		LOGGER.info("Java: " + System.getProperty("java.vm.name") + " " + System.getProperty("java.version") + " " + System.getProperty("sun.arch.data.model") + "-bit" + " by " + System.getProperty("java.vendor"));
-		LOGGER.info("OS: " + System.getProperty("os.name") + " " + getOSBitness() + "-bit " + System.getProperty("os.version"));
-		LOGGER.info("Physical memory available: {} MB", memory.getAvailable() / 1048576);
-		LOGGER.info("Memory accessible by Java: {} MB", memoryInMB);
-		LOGGER.info("CPU: " + processor.getName());
-		LOGGER.info("Language: " + WordUtils.capitalize(PMS.getLocale().getDisplayName(Locale.ENGLISH)));
-		LOGGER.info("Encoding: " + System.getProperty("file.encoding"));
+		LOGGER.info(
+			"Java: {} {} ({}-bit) by {}",
+			System.getProperty("java.vm.name"),
+			System.getProperty("java.version"),
+			System.getProperty("sun.arch.data.model"),
+			System.getProperty("java.vendor")
+		);
+		if (os != null && isNotBlank(os.toString())) {
+			LOGGER.info("OS: {} {}-bit", os.toString(), getOSBitness());
+		} else {
+			LOGGER.info("OS: {} {}-bit {}", System.getProperty("os.name"), getOSBitness(), System.getProperty("os.version"));
+		}
+		if (processor != null) {
+			LOGGER.info(
+				"CPU: {} with {} cores ({} virtual cores)",
+				processor.getName(),
+				processor.getPhysicalProcessorCount(),
+				processor.getLogicalProcessorCount()
+			);
+		}
+		if (memory != null) {
+			LOGGER.info("Physical Memory: {}", StringUtil.formatBytes(memory.getTotal(), true));
+			LOGGER.info("Free memory: {}", StringUtil.formatBytes(memory.getAvailable(), true));
+
+		}
+		LOGGER.info("Maximum JVM Memory: {}", jvmMemory == Long.MAX_VALUE ? "Unlimited" : StringUtil.formatBytes(jvmMemory, true));
+		LOGGER.info("Language: {}", WordUtils.capitalize(PMS.getLocale().getDisplayName(Locale.ENGLISH)));
+		LOGGER.info("Encoding: {}", System.getProperty("file.encoding"));
 		LOGGER.info("");
 
 		if (Platform.isMac()) {
@@ -1694,9 +1724,8 @@ public class PMS {
 		try {
 			if (locale != null) {
 				return locale;
-			} else {
-				return Locale.getDefault();
 			}
+			return Locale.getDefault();
 		} finally {
 			localeLock.readLock().unlock();
 		}


Sami32:
Still doesn't work, as with Java 7 or 8 version, the Javaw.exe stay only a few second in the TaskBar.
https://github.com/UniversalMediaServer/UniversalMediaServer/pull/979#issuecomment-238706203

Nadahar:
@Sami32 I tried to make sure that an Exception wouldn't kill it. Do you have a log up until it crashes?

Nadahar:
@Sami32 Never mind, I tested on my VM and got the same issue:

```java
Exception in thread "main" java.lang.UnsatisfiedLinkError: Error looking up function 'PdhAddEnglishCounterA': Den angitte prosedyren ble ikke funnet.

	at com.sun.jna.Function.<init>(Function.java:212)
	at com.sun.jna.NativeLibrary.getFunction(NativeLibrary.java:541)
	at com.sun.jna.NativeLibrary.getFunction(NativeLibrary.java:518)
	at com.sun.jna.NativeLibrary.getFunction(NativeLibrary.java:504)
	at com.sun.jna.Library$Handler.invoke(Library.java:220)
	at com.sun.proxy.$Proxy4.PdhAddEnglishCounterA(Unknown Source)
	at oshi.util.platform.windows.PdhUtil.addCounter(PdhUtil.java:66)
	at oshi.hardware.platform.windows.WindowsCentralProcessor.initPdhCounters(WindowsCentralProcessor.java:154)
	at oshi.hardware.platform.windows.WindowsCentralProcessor.<init>(WindowsCentralProcessor.java:101)
	at oshi.hardware.platform.windows.WindowsHardwareAbstractionLayer.getProcessor(WindowsHardwareAbstractionLayer.java:55)
```

Nadahar:
@Sami32 I think I've fixed it - that is, we won't get all the information on WinXP but it won't crash.

Sami32:
I will have shared informations if i had it, but my log was totally empty, and the process was so fastly killed that i wasn't able to take any thread dump.

Also your Windows XP is better than mine :P
Now it start, but it throw an error:
> DEBUG 2017-02-28 02:07:04.453 [main] Could not retrieve OS and hardware information: Error looking up function 'PdhAddEnglishCounterA': La procédure spécifiée est introuvable.
> TRACE 2017-02-28 02:07:04.468 [main] 
> java.lang.UnsatisfiedLinkError: Error looking up function 'PdhAddEnglishCounterA': La procédure spécifiée est introuvable.
> 	at com.sun.jna.Function.<init>(Function.java:212)
> 	at com.sun.jna.NativeLibrary.getFunction(NativeLibrary.java:541)

Nadahar:
@Sami32 Yes - the stack trace will only be logged in trace mode. I could have hidden it, but there can be many reasons why this fails so I think logging it is useful.

This will still be useful for all those not running Windows XP. There is also hope that we will get Windows XP supported in the near future: oshi/oshi#326.

Sami32:
For information, the Redhat OpenJDK 8 version doesn't work for my processor, but perhaps will work on your...

Nadahar:
@Sami32 I hate forced registration, so I haven't downloaded it. Are you saying that it supports Windows XP?

Sami32:
I'm just saying that it throw me a message about my processor being not supported, nothing about Windows version, so we could reasonnabily think that XP should be supported ?

Nadahar:
@Sami32 Yes, it sounds reasonable. I'll see if I can be bothered to make a fake account with a temporary e-mail to get it and test later. It would be great if Java 8 became available for Windows XP again, but it will never work as long as you have to give them your life story to download. If they would make a separate download for just the JRE (without registration) it would be ok.

Sami32:
As i already said, i tried with Java 8 and compiled with Java 8 as well, and did get the same error.

So, i'm not sure if it's because the Java classes used was introduced after Java 8u60 or if it's something else. Anyway, if the next release fix it that should be nice :)

Nadahar:
@Sami32 The error you get from this PR is because [](https://msdn.microsoft.com/en-us/library/windows/desktop/aa372536(v=vs.85).aspx) only exists in Windows Vista and later. However, OSHI has a different code in 3.x that doesn't use this method. The version in this PR is 2.6.1, which is the last with Java 7 support. 

I've tested using oshi/oshi#327 on Windows XP/Java 7, and then it works because that is 3.x code that doesn't use that Windows API method.

Sami32:
Arrrg, i missed that it was the 2.6.1 version in that PR...Though you already told in the OSHI issue, i didn't get it until now...

dbwiddis:
FYI, OSHI 3.5.0 will restore Java7 compatibility and most features will work on XP.  The 3.5.0-SNAPSHOT will work until we get the main release out (sometime in March).

Please test as many features on XP as you can (running the SystemInfoTest from the oshi-json project does this effectively) so we can address them.

Nadahar:
@dbwiddis Thank you very much for a fantastic response! I'm overwhelmed at the speed and efficiency 😉 

Is the SNAPSHOT available at any Maven repo? It's obviously not at central, I checked at the SonaType OSS repo as well but only found 3.4. There's no easy way for us to bundle non-Maven dependencies.

```SystemInfoTest``` from latest master on Windows XP looks pretty good - I've attached ```TRACE``` output here: [SystemInfoTestOutput.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/807316/SystemInfoTestOutput.txt)

There are some ```NumberFormatException```, but I don't know if they pose a problem - or are exclusive to Windows XP.

**Comment by dbwiddis** [comment]
Yes, it's at OSSRH in the snapshots.  Add this to your POM:
```
	<distributionManagement>
		<snapshotRepository>
			<id>sonatype-nexus-snapshots</id>
			<name>Sonatype Nexus Snapshots</name>
			<url>https://oss.sonatype.org/content/repositories/snapshots</url>
		</snapshotRepository>
	</distributionManagement>
```

**Comment by dbwiddis** [comment]
I looked at the trace log -- pretty normal.  The NFE's are failing to parse "unknown" which is the result when WMI doesn't have your table. (Win32_LogicalDisk is Vista+).    So it looks like you're mostly set on WinXP (at least with what's available).

**Comment by dbwiddis** [comment]
Correction on the snapshots... that line is for the deploy plugin.  To fetch them you need to add this to your `~/.m2/settings.xml` file:
Just add this to your ~/.m2/settings.xml:

```
<profiles>
  <profile>
     <id>allow-snapshots</id>
        <activation><activeByDefault>true</activeByDefault></activation>
     <repositories>
       <repository>
         <id>snapshots-repo</id>
         <url>https://oss.sonatype.org/content/repositories/snapshots</url>
         <releases><enabled>false</enabled></releases>
         <snapshots><enabled>true</enabled></snapshots>
       </repository>
     </repositories>
   </profile>
</profiles>
```

**Comment by Nadahar** [comment]
@dbwiddis Thank you. It wouldn't do any good to add the repo to my ```setting.xml``` (it would only work for me), but it worked to put it in out POM as well. I did struggle a bit to find the coordinates, but a peek in your POM sorted it out. I don't understand why I can't find it when searching Nexus - or Google for that matter.
The coordinates in the snapshot repo are:
```xml
	<dependency>
    		<groupId>com.github.oshi</groupId>
    		<artifactId>oshi-core</artifactId>
    		<version>3.5.0-SNAPSHOT</version>
	</dependency>
```

Nadahar:
@dbwiddis When I try ```3.5.0-SNAPSHOT``` I get:
```java
java.lang.NoSuchMethodError: com.sun.jna.platform.win32.OleAuto.VariantClear(Lcom/sun/jna/platform/win32/Variant$VARIANT;)Lcom/sun/jna/platform/win32/WinNT$HRESULT;
	at oshi.util.platform.windows.WmiUtil.enumerateProperties(WmiUtil.java:579)
	at oshi.util.platform.windows.WmiUtil.queryWMI(WmiUtil.java:351)
	at oshi.util.platform.windows.WmiUtil.selectObjectsFrom(WmiUtil.java:293)
	at oshi.software.os.windows.WindowsOSVersionInfoEx.<init>(WindowsOSVersionInfoEx.java:51)
	at oshi.software.os.windows.WindowsOperatingSystem.<init>(WindowsOperatingSystem.java:90)
	at oshi.SystemInfo.getOperatingSystem(SystemInfo.java:92)
```

This doesn't happen when I build myself from your latest ```master```. Is the snapshot updated?

Nadahar:
@Sami32 Does this work for you now?

It does for me:
```logtalk
INFO  16:28:44.781 [main] Java: Java HotSpot(TM) Client VM 1.7.0_80 (32-bit) by Oracle Corporation
INFO  16:28:47.435 [main] OS: Microsoft Windows XP SP3 build 2600 32-bit
INFO  16:28:50.910 [main] CPU: Intel(R) Core(TM)2 Quad CPU    Q9650  @ 3.00GHz with 1 cores (1 virtual cores)
INFO  16:28:51.941 [main] Physical Memory: 3.0 GiB
INFO  16:28:52.282 [main] Free memory: 1.2 GiB
INFO  16:28:52.662 [main] Maximum JVM Memory: 247.5 MiB
INFO  16:28:53.053 [main] Language: Norwegian
INFO  16:28:53.524 [main] Encoding: UTF-8
INFO  16:28:53.984 [main] Language: Norwegian
```

Nadahar:
This PR now depends on #1171

dbwiddis:
I had put the new JNA dependency in the POM but also erred and duplicated some other dependencies which may have broken Maven's parsing to tell you you needed JNA 4.3.

In any case, I think I'm going to cut an OSHI 3.4.1 release today so you don't need to include a snapshot dependency in your own project.

Sami32:
Before depending on JNA 4.3.0, it was compiling with Java 8 but not running, now i'm not even able to compile it. I'll try again.

As this PR is now, it compile and **work fine with Java 7**:
> DEBUG 2017-03-01 20:33:05.906 [main] Reading log_level: "TRACE" (default: "DEBUG")
> DEBUG 2017-03-01 20:33:05.937 [main] Default logfile folder set to: C:\Documents and Settings\All Users\UMS
> DEBUG 2017-03-01 20:33:06.437 [main] LogBack started with configuration file: C:\Program Files\Universal Media Server\logback.xml
> DEBUG 2017-03-01 20:33:06.437 [main] Reading logging_filter_logs_tab: "TRACE" (default: "INFO")
> DEBUG 2017-03-01 20:33:06.453 [main] Reading logging_buffered: true (default: false)
> INFO  2017-03-01 20:33:06.453 [main] Buffered logging turned ON
> DEBUG 2017-03-01 20:33:06.484 [main] Wed Mar 01 20:33:06 CET 2017
> DEBUG 2017-03-01 20:33:06.484 [main] Reading cred.path: "C:\Documents and Settings\All Users\UMS\UMS.cred" (default: "")
> DEBUG 2017-03-01 20:33:07.156 [main] Writing PID: 2524
> DEBUG 2017-03-01 20:33:07.156 [main] 
> INFO  2017-03-01 20:33:07.156 [main] Starting Universal Media Server 6.6.1-SNAPSHOT
> INFO  2017-03-01 20:33:07.156 [main] Based on PS3 Media Server by shagrath, copyright 2008-2014
> INFO  2017-03-01 20:33:07.156 [main] http://www.universalmediaserver.com
> INFO  2017-03-01 20:33:07.156 [main] 
> INFO  2017-03-01 20:33:07.156 [main] Build: eb8d21884 (2017-03-01)
> DEBUG 2017-03-01 20:33:07.671 [main] Connected to ROOT\CIMV2 WMI namespace
> DEBUG 2017-03-01 20:33:07.671 [main] Query: SELECT Version,ProductType,BuildNumber,CSDVersion,SuiteMask FROM Win32_OperatingSystem 
> DEBUG 2017-03-01 20:33:08.343 [main] Initialized OSVersionInfoEx
> DEBUG 2017-03-01 20:33:08.375 [main] Oracle MXBean detected.
> DEBUG 2017-03-01 20:33:08.484 [main] Connected to ROOT\CIMV2 WMI namespace
> DEBUG 2017-03-01 20:33:08.484 [main] Query: SELECT ProcessorID FROM Win32_Processor 
> TRACE 2017-03-01 20:33:08.515 [main] Updating Processor Ticks
> DEBUG 2017-03-01 20:33:08.531 [main] Connected to ROOT\CIMV2 WMI namespace
> DEBUG 2017-03-01 20:33:08.531 [main] Query: SELECT Name,PercentIdleTime,PercentPrivilegedTime,PercentUserTime,PercentInterruptTime,PercentDPCTime FROM Win32_PerfRawData_Counters_ProcessorInformation WHERE NOT Name LIKE "%_Total"
> TRACE 2017-03-01 20:33:08.546 [main] Updating System Ticks
> DEBUG 2017-03-01 20:33:08.546 [main] Connected to ROOT\CIMV2 WMI namespace
> DEBUG 2017-03-01 20:33:08.546 [main] Query: SELECT PercentInterruptTime,PercentDPCTime FROM Win32_PerfRawData_Counters_ProcessorInformation WHERE Name="_Total"
> DEBUG 2017-03-01 20:33:08.562 [main] Initialized Processor
> INFO  2017-03-01 20:33:08.609 [main] Java: Java HotSpot(TM) Client VM 1.8.0_60 (32-bit) by Oracle Corporation
> INFO  2017-03-01 20:33:08.609 [main] OS: Microsoft Windows XP SP3 build 2600 32-bit
> INFO  2017-03-01 20:33:08.609 [main] CPU: AMD Athlon(TM) XP 1700+ with 1 cores (1 virtual cores)
> INFO  2017-03-01 20:33:08.625 [main] Physical Memory: 3.0 GiB
> INFO  2017-03-01 20:33:08.625 [main] Free memory: 2.1 GiB
> INFO  2017-03-01 20:33:08.625 [main] Maximum JVM Memory: 742.4 MiB
> INFO  2017-03-01 20:33:08.640 [main] Language: English (United States)
> INFO  2017-03-01 20:33:08.640 [main] Encoding: UTF-8
> INFO  2017-03-01 20:33:08.640 [main] Language: English (United States)

src/main/java/net/pms/util/StringUtil.java
```diff
@@ -41,6 +41,18 @@
 	public static final String SEC_TIME_FORMAT = "%02d:%02d:%02.0f";
 	public static final String DURATION_TIME_FORMAT = "%02d:%02d:%05.2f";
 	public static final String NEWLINE_CHARACTER = System.getProperty("line.separator");
+	public static final long KIBI = 1L << 10;
```
Sami32: I do not agree with this unit choice.
From a long traditional/conventional use, memory capacities are reported in GB, MB...Informatic and technical literature mainly use that ones.
GiBi...are generally only used as speed/transfer unit. And for "commercial" publicity as it make it easy understandable for everyone.

https://www.reference.com/technology/ram-measured-6b59364967cdb1f4#
http://smallbusiness.chron.com/processor-speed-vs-memory-55329.html

dbwiddis: I agree with that unit choice for memory.  The SI units are only used for Network bandwidth and disk size. IEC units are used everywhere else.  [Ubuntu published a UnitsPolicy in 2010](https://wiki.ubuntu.com/UnitsPolicy) that's generally consistent across Linux distributions.

> Use base-10 for:
> 
> network bandwidth (for example, 6 Mbit/s or 50 kB/s)
> disk sizes (for example, 500 GB hard drive or 4.7 GB DVD)
> Use base-2 for:
> 
> RAM sizes (for example, 2 GiB RAM)
> For file sizes there are two possibilities:
> 
> Show both, base-10 and base-2 (in this order). An example is the Linux kernel: "2930277168 512-byte hardware sectors: (1.50 TB/1.36 TiB)"
> Only show base-10, or give the user the opportunity to decide between base-10 and base-2 (the default must be base-10).

You may see memory marketed as 16 GB, but rest assured that the memory chips actually contain 16 GiB.

Sami32:
Using JNA 4.3.0 modified something in the way UMS handle the network.
Could it come from [that class](https://java-native-access.github.io/jna/4.3.0/javadoc/com/sun/jna/platform/win32/WinRas.RASIPV6ADDR.html) ?
I saw that in our log but didn't tested it yet.

@dbwiddis Thanks for your input.
I will not hide that's not my taste though...I guess i'm too old  ;)

dbwiddis:
@Sami32 You're still using WinXP, so that's part of it. Windows is stubbornly resisting the *nix world conventions.  :)   Personally, I am strongly in favor of using the IEC units when they are accurate, because it leaves zero confusion as to whether the numbers are binary or decimal.  When you have a decimal prefix, you can never be sure thanks to those pesky marketing hand-waves.

src/main/java/net/pms/util/StringUtil.java
Nadahar: @Sami32 We've had these units around since 1995, and standardized since 1999. They've been [adoption by IEC, NIST and ISO](https://en.wikipedia.org/wiki/Binary_prefix#Adoption_by_IEC.2C_NIST_and_ISO). The use of the SI units for both 1000- and 1024-based units has caused enough confusion in the world IMO. It's understandable than the SI units were back in the days what the correct units didn't exist. But to cling to them today, soon 20 years later, makes no sense.

The method handles both notations, but it would be very confusing if we suddenly were to specify RAM in SI units, we would get "too high" values compared to other references and it would cause even more confusion.

Nadahar:
@Sami32 This is working just fine for me:
```logtalk
INFO  22:38:25.075 [main] Java: Java HotSpot(TM) Client VM 1.8.0_101 (32-bit) by Oracle Corporation
INFO  22:38:25.075 [main] OS: Microsoft Windows XP SP3 build 2600 32-bit
INFO  22:38:25.075 [main] CPU: Intel(R) Core(TM)2 Quad CPU    Q9650  @ 3.00GHz with 1 cores (1 virtual cores)
INFO  22:38:25.075 [main] Physical Memory: 3.0 GiB
INFO  22:38:25.075 [main] Free memory: 1.2 GiB
INFO  22:38:25.075 [main] Maximum JVM Memory: 247.5 MiB
INFO  22:38:25.085 [main] Language: Norwegian
INFO  22:38:25.085 [main] Encoding: UTF-8
INFO  22:38:25.085 [main] Language: Norwegian
INFO  22:38:25.085 [main] 
```

Nadahar:
@Sami32 

> Using JNA 4.3.0 modified something in the way UMS handle the network.
> Could it come from that class ?
> I saw that in our log but didn't tested it yet.

I don't understand. What do you mean by "modified something in the way UMS handle the network"? What is different?

Nadahar:
@Sami32 There's one thing I don't understand. I can build fine with Eclipse on Windows XP, but when I try to build with Maven I can't get past the tests because if hangs in ```RendererConfigurationTest``` MediaInfo is called. That is the same with or without this PR. The reason is that the new version of MediaInfo.dll is always downloaded when I build with Maven, so it doesn't help if I try to replace it with 7.81. You must have the same problem, how do you get around that?

With Eclipse I can just replace the MediaInfo.dll with the working version, and all is well - no tests are hanging or failing.

Nadahar:
OS X 10.8.5:
```logtalk
00:10:19 INFO  Java: Java HotSpot(TM) 64-Bit Server VM 1.8.0_60 (64-bit) by Oracle Corporation
00:10:19 INFO  OS: Apple Mac OS X 10.8.5 (Mountain Lion) build 12F2560 64-bit
00:10:19 INFO  CPU: Intel(R) Core(TM)2 Quad CPU    Q9650  @ 3.00GHz with 4 cores (4 virtual cores)
00:10:19 INFO  Physical Memory: 4 GiB
00:10:19 INFO  Free memory: 2.6 GiB
00:10:19 INFO  Maximum JVM Memory: 683 MiB
00:10:19 INFO  Language: English (United States)
00:10:19 INFO  Encoding: UTF-8
00:10:19 INFO  Language: English (United States)
```

Nadahar:
@dbwiddis This one might be something for you:
```java
java.io.IOException: Cannot run program "cpuid": error=2, No such file or directory
	at java.lang.ProcessBuilder.start(ProcessBuilder.java:1047)
	at java.lang.Runtime.exec(Runtime.java:617)
	at java.lang.Runtime.exec(Runtime.java:485)
	at oshi.util.ExecutingCommand.runNative(ExecutingCommand.java:67)
	at oshi.util.ExecutingCommand.runNative(ExecutingCommand.java:53)
	at oshi.hardware.platform.linux.LinuxCentralProcessor.getProcessorID(LinuxCentralProcessor.java:311)
	at oshi.hardware.platform.linux.LinuxCentralProcessor.initVars(LinuxCentralProcessor.java:100)
	at oshi.hardware.platform.linux.LinuxCentralProcessor.<init>(LinuxCentralProcessor.java:54)
	at oshi.hardware.platform.linux.LinuxHardwareAbstractionLayer.getProcessor(LinuxHardwareAbstractionLayer.java:64)
	at net.pms.PMS.logSystemInfo(PMS.java:1403)
	at net.pms.PMS.displayBanner(PMS.java:304)
	at net.pms.PMS.init(PMS.java:430)
	at net.pms.PMS.createInstance(PMS.java:1111)
	at net.pms.PMS.main(PMS.java:1271)
Caused by: java.io.IOException: error=2, No such file or directory
	at java.lang.UNIXProcess.forkAndExec(Native Method)
	at java.lang.UNIXProcess.<init>(UNIXProcess.java:187)
	at java.lang.ProcessImpl.start(ProcessImpl.java:130)
	at java.lang.ProcessBuilder.start(ProcessBuilder.java:1028)
```

It does however manage to retrieve some info, Ubuntu 12.04 LTS:
```logtalk
00:21:22 INFO  Java: Java HotSpot(TM) 64-Bit Server VM 1.7.0_80 (64-bit) by Oracle Corporation
00:21:22 INFO  OS: GNU/Linux Ubuntu 12.04.5 LTS (Precise Pangolin) build 3.13.0-105-generic 64-bit
00:21:22 INFO  CPU: Intel(R) Core(TM)2 Quad CPU    Q9650  @ 3.00GHz with 4 cores (4 virtual cores)
00:21:22 DEBUG Reading file /proc/meminfo
00:21:23 INFO  Physical Memory: 2.0 GiB
00:21:23 DEBUG Reading file /proc/meminfo
00:21:23 INFO  Free memory: 705.4 MiB
00:21:23 INFO  Maximum JVM Memory: 446.5 MiB
00:21:23 INFO  Language: Norwegian
00:21:23 INFO  Encoding: UTF-8
00:21:23 INFO  Language: Norwegian
```

Nadahar:
Antergos:
```logtalk
00:42:07 INFO  Build: ${git.com (${git.commit.time})
00:42:08 INFO  Java: OpenJDK 64-Bit Server VM 1.8.0_112 (64-bit) by Oracle Corporation
00:42:08 INFO  OS: GNU/Linux Antergos Linux 2015.12-ISO-Rolling build 4.8.11-1-ARCH 64-bit
00:42:08 INFO  CPU: Intel(R) Core(TM)2 Quad CPU    Q9650  @ 3.00GHz with 2 cores (2 virtual cores)
00:42:08 INFO  Physical Memory: 1.5 GiB
00:42:08 INFO  Free memory: 218.9 MiB
00:42:08 INFO  Maximum JVM Memory: 363.5 MiB
00:42:08 INFO  Language: Norwegian
00:42:08 INFO  Encoding: UTF-8
00:42:08 INFO  Language: Norwegian
```

Sami32:
@Nadahar Yes, but i get that "freezing" while building with Maven that PR/JNA 4.3.0 using Java 8 only.
With Java 7 or without JNA included, i don't get this "freezing".

About the JNA 4.3.0 network handling modifications: (IPv6)
```logtalk
TRACE 2017-03-02 02:25:22.203 [main] checkNetworkInterface(parent = null, child interfaces = java.net.NetworkInterface$2@1001b64)
TRACE 2017-03-02 02:25:22.203 [main] checking lo, display name: MS TCP Loopback interface
TRACE 2017-03-02 02:25:22.203 [main] available addresses for lo is: [/127.0.0.1, /0:0:0:0:0:0:0:1, /fe80:0:0:0:0:0:0:1%lo]
TRACE 2017-03-02 02:25:22.203 [main] non loopback/ipv4 addresses: []
TRACE 2017-03-02 02:25:22.203 [main] checkNetworkInterface(parent = lo, child interfaces = java.net.NetworkInterface$1subIFs@310d43)
TRACE 2017-03-02 02:25:22.203 [main] checkNetworkInterface(parent = lo) finished
TRACE 2017-03-02 02:25:22.203 [main] sub address for lo is []
TRACE 2017-03-02 02:25:22.203 [main] checking /127.0.0.1 on lo
TRACE 2017-03-02 02:25:22.203 [main] has /127.0.0.1, which is skipped, because loopback=true, ipv6=false
TRACE 2017-03-02 02:25:22.203 [main] checking /0:0:0:0:0:0:0:1 on lo
TRACE 2017-03-02 02:25:22.203 [main] has /0:0:0:0:0:0:0:1, which is skipped, because loopback=true, ipv6=true
TRACE 2017-03-02 02:25:22.203 [main] checking /fe80:0:0:0:0:0:0:1%lo on lo
TRACE 2017-03-02 02:25:22.203 [main] has /fe80:0:0:0:0:0:0:1%lo, which is skipped, because loopback=false, ipv6=true
TRACE 2017-03-02 02:25:22.203 [main] found lo, without valid address
TRACE 2017-03-02 02:25:22.203 [main] checking eth0, display name: ASUSTeK/Broadcom 440x 10/100 Integrated Controller - Miniport d'ordonnancement de paquets
TRACE 2017-03-02 02:25:22.203 [main] available addresses for eth0 is: [/192.168.1.28]
TRACE 2017-03-02 02:25:22.203 [main] non loopback/ipv4 addresses: [/192.168.1.28]
TRACE 2017-03-02 02:25:22.203 [main] checkNetworkInterface(parent = eth0, child interfaces = java.net.NetworkInterface$1subIFs@1d9a273)
TRACE 2017-03-02 02:25:22.203 [main] checkNetworkInterface(parent = eth0) finished
TRACE 2017-03-02 02:25:22.203 [main] sub address for eth0 is []
TRACE 2017-03-02 02:25:22.203 [main] checking /192.168.1.28 on eth0
TRACE 2017-03-02 02:25:22.203 [main] found eth0 -> 192.168.1.28
TRACE 2017-03-02 02:25:22.203 [main] checking tun0, display name: Teredo Tunneling Pseudo-Interface
TRACE 2017-03-02 02:25:22.203 [main] available addresses for tun0 is: [/fe80:0:0:0:0:ffff:ffff:fffd%tun0]
TRACE 2017-03-02 02:25:22.203 [main] non loopback/ipv4 addresses: []
TRACE 2017-03-02 02:25:22.203 [main] checkNetworkInterface(parent = tun0, child interfaces = java.net.NetworkInterface$1subIFs@18ee4de)
TRACE 2017-03-02 02:25:22.203 [main] checkNetworkInterface(parent = tun0) finished
TRACE 2017-03-02 02:25:22.203 [main] sub address for tun0 is []
TRACE 2017-03-02 02:25:22.203 [main] checking /fe80:0:0:0:0:ffff:ffff:fffd%tun0 on tun0
TRACE 2017-03-02 02:25:22.203 [main] has /fe80:0:0:0:0:ffff:ffff:fffd%tun0, which is skipped, because loopback=false, ipv6=true
TRACE 2017-03-02 02:25:22.203 [main] found tun0, without valid address
TRACE 2017-03-02 02:25:22.203 [main] checking tun1, display name: 6to4 Pseudo-Interface
TRACE 2017-03-02 02:25:22.203 [main] available addresses for tun1 is: []
TRACE 2017-03-02 02:25:22.203 [main] non loopback/ipv4 addresses: []
TRACE 2017-03-02 02:25:22.203 [main] checkNetworkInterface(parent = tun1, child interfaces = java.net.NetworkInterface$1subIFs@1f981c6)
TRACE 2017-03-02 02:25:22.203 [main] checkNetworkInterface(parent = tun1) finished
TRACE 2017-03-02 02:25:22.203 [main] sub address for tun1 is []
TRACE 2017-03-02 02:25:22.203 [main] found tun1, without valid address
TRACE 2017-03-02 02:25:22.203 [main] checking tun2, display name: Automatic Tunneling Pseudo-Interface
TRACE 2017-03-02 02:25:22.203 [main] available addresses for tun2 is: []
TRACE 2017-03-02 02:25:22.203 [main] non loopback/ipv4 addresses: []
TRACE 2017-03-02 02:25:22.203 [main] checkNetworkInterface(parent = tun2, child interfaces = java.net.NetworkInterface$1subIFs@111171f)
TRACE 2017-03-02 02:25:22.203 [main] checkNetworkInterface(parent = tun2) finished
TRACE 2017-03-02 02:25:22.203 [main] sub address for tun2 is []
TRACE 2017-03-02 02:25:22.203 [main] found tun2, without valid address
TRACE 2017-03-02 02:25:22.203 [main] checkNetworkInterface(parent = null) finished
DEBUG 2017-03-02 02:25:22.218 [main] Reading network_interface: "eth0" (default: "")
```

Nadahar:
@Sami32 I'm sorry, but that log output doesn't tell me anything. What is different than usual?

dbwiddis:
> java.io.IOException: Cannot run program "cpuid": error=2, No such file or directory
> 	at java.lang.ProcessBuilder.start(ProcessBuilder.java:1047)
> 	at java.lang.Runtime.exec(Runtime.java:617)
> 	at java.lang.Runtime.exec(Runtime.java:485)
> 	at oshi.util.ExecutingCommand.runNative(ExecutingCommand.java:67)

Well, this is odd, because that line 67 is is the `exec()` call below...
```
        try {
            p = Runtime.getRuntime().exec(cmdToRunWithArgs);
        } catch (SecurityException | IOException e) {
            LOG.trace("", e);
            return new ArrayList<>(0);
        }
```

Nadahar:
@dbwiddis Not really, I think what I see in the log simply is the output from ```LOG.trace("", e)```. First I thought an ```Exception``` was being thrown, but as you can see the code continued on and gave the correct output (which it wouldn't if it was interrupted by a throw). Your handling is working correcly, the only "problem" is that your logging is output. I think we can configure the logger to not output logs from your library, but I'm not sure it's necessary. I was using TRACE after all, so one must expect a lot of stuff. The only thing is, if this is expected, you might not want to log the complete stack trace.

dbwiddis:
You're correct, I should not log complete stack traces on expected errors.

Nadahar:
Windows 10:
```logtalk
03:54:48 INFO  Java: Java HotSpot(TM) 64-Bit Server VM 1.8.0_111 (64-bit) by Oracle Corporation
03:54:48 INFO  OS: Microsoft Windows 10 build 14393 64-bit
03:54:48 INFO  CPU: Intel(R) Core(TM)2 Quad CPU    Q9650  @ 3.00GHz with 4 cores (4 virtual cores)
03:54:48 INFO  Physical Memory: 4.0 GiB
03:54:48 INFO  Free memory: 2.0 GiB
03:54:48 INFO  Maximum JVM Memory: 910.5 MiB
03:54:48 INFO  Language: Norwegian
03:54:48 INFO  Encoding: UTF-8
03:54:48 INFO  Language: Norwegian
```

Nadahar:
FYI @dbwiddis, running ```TRACE``` on Antergos gave similar output:
```java
java.io.IOException: Cannot run program "dmidecode": error=2, Ingen slik fil eller filkatalog
	at java.lang.ProcessBuilder.start(ProcessBuilder.java:1048)
	at java.lang.Runtime.exec(Runtime.java:620)
	at java.lang.Runtime.exec(Runtime.java:485)
	at oshi.util.ExecutingCommand.runNative(ExecutingCommand.java:67)
	at oshi.util.ExecutingCommand.runNative(ExecutingCommand.java:53)
	at oshi.hardware.platform.linux.LinuxCentralProcessor.getProcessorID(LinuxCentralProcessor.java:301)
	at oshi.hardware.platform.linux.LinuxCentralProcessor.initVars(LinuxCentralProcessor.java:100)
	at oshi.hardware.platform.linux.LinuxCentralProcessor.<init>(LinuxCentralProcessor.java:54)
	at oshi.hardware.platform.linux.LinuxHardwareAbstractionLayer.getProcessor(LinuxHardwareAbstractionLayer.java:64)
	at net.pms.PMS.logSystemInfo(PMS.java:1403)
	at net.pms.PMS.displayBanner(PMS.java:304)
	at net.pms.PMS.init(PMS.java:430)
	at net.pms.PMS.createInstance(PMS.java:1111)
	at net.pms.PMS.main(PMS.java:1271)
Caused by: java.io.IOException: error=2, Ingen slik fil eller filkatalog
	at java.lang.UNIXProcess.forkAndExec(Native Method)
	at java.lang.UNIXProcess.<init>(UNIXProcess.java:247)
	at java.lang.ProcessImpl.start(ProcessImpl.java:134)
	at java.lang.ProcessBuilder.start(ProcessBuilder.java:1029)
	... 13 common frames omitted
04:11:46 TRACE 
java.io.IOException: Cannot run program "cpuid": error=2, Ingen slik fil eller filkatalog
	at java.lang.ProcessBuilder.start(ProcessBuilder.java:1048)
	at java.lang.Runtime.exec(Runtime.java:620)
	at java.lang.Runtime.exec(Runtime.java:485)
	at oshi.util.ExecutingCommand.runNative(ExecutingCommand.java:67)
	at oshi.util.ExecutingCommand.runNative(ExecutingCommand.java:53)
	at oshi.hardware.platform.linux.LinuxCentralProcessor.getProcessorID(LinuxCentralProcessor.java:311)
	at oshi.hardware.platform.linux.LinuxCentralProcessor.initVars(LinuxCentralProcessor.java:100)
	at oshi.hardware.platform.linux.LinuxCentralProcessor.<init>(LinuxCentralProcessor.java:54)
	at oshi.hardware.platform.linux.LinuxHardwareAbstractionLayer.getProcessor(LinuxHardwareAbstractionLayer.java:64)
	at net.pms.PMS.logSystemInfo(PMS.java:1403)
	at net.pms.PMS.displayBanner(PMS.java:304)
	at net.pms.PMS.init(PMS.java:430)
	at net.pms.PMS.createInstance(PMS.java:1111)
	at net.pms.PMS.main(PMS.java:1271)
Caused by: java.io.IOException: error=2, Ingen slik fil eller filkatalog
	at java.lang.UNIXProcess.forkAndExec(Native Method)
	at java.lang.UNIXProcess.<init>(UNIXProcess.java:247)
	at java.lang.ProcessImpl.start(ProcessImpl.java:134)
	at java.lang.ProcessBuilder.start(ProcessBuilder.java:1029)
	... 13 common frames omitted
```

Nadahar:
Centos 7:
```logtalk
04:40:51 INFO  Java: OpenJDK 64-Bit Server VM 1.7.0_111 (64-bit) by Oracle Corporation
04:40:51 INFO  OS: GNU/Linux CentOS Linux 7 (Core) build 3.10.0-327.36.3.el7.x86_64 64-bit
04:40:51 INFO  CPU: Intel(R) Core(TM)2 Quad CPU    Q9650  @ 3.00GHz with 4 cores (4 virtual cores)
04:40:51 DEBUG Reading file /proc/meminfo
04:40:51 INFO  Physical Memory: 1.5 GiB
04:40:51 INFO  Free memory: 236.9 MiB
04:40:51 INFO  Maximum JVM Memory: 643.8 MiB
04:40:51 INFO  Language: Norwegian
04:40:51 INFO  Encoding: UTF-8
04:40:51 INFO  Language: Norwegian
```

The same log output again:
```java
java.io.IOException: Cannot run program "cpuid": error=2, No such file or directory
	at java.lang.ProcessBuilder.start(ProcessBuilder.java:1047)
	at java.lang.Runtime.exec(Runtime.java:617)
	at java.lang.Runtime.exec(Runtime.java:485)
	at oshi.util.ExecutingCommand.runNative(ExecutingCommand.java:67)
	at oshi.util.ExecutingCommand.runNative(ExecutingCommand.java:53)
	at oshi.hardware.platform.linux.LinuxCentralProcessor.getProcessorID(LinuxCentralProcessor.java:311)
	at oshi.hardware.platform.linux.LinuxCentralProcessor.initVars(LinuxCentralProcessor.java:100)
	at oshi.hardware.platform.linux.LinuxCentralProcessor.<init>(LinuxCentralProcessor.java:54)
	at oshi.hardware.platform.linux.LinuxHardwareAbstractionLayer.getProcessor(LinuxHardwareAbstractionLayer.java:64)
	at net.pms.PMS.logSystemInfo(PMS.java:1403)
	at net.pms.PMS.displayBanner(PMS.java:304)
	at net.pms.PMS.init(PMS.java:430)
	at net.pms.PMS.createInstance(PMS.java:1111)
	at net.pms.PMS.main(PMS.java:1271)
Caused by: java.io.IOException: error=2, No such file or directory
	at java.lang.UNIXProcess.forkAndExec(Native Method)
	at java.lang.UNIXProcess.<init>(UNIXProcess.java:187)
	at java.lang.ProcessImpl.start(ProcessImpl.java:130)
	at java.lang.ProcessBuilder.start(ProcessBuilder.java:1028)
	... 13 common frames omitted
```

Nadahar:
Kubuntu 16.04.1 LTS:
```logtalk
06:35:02 INFO  Java: OpenJDK 64-Bit Server VM 1.8.0_121 (64-bit) by Oracle Corporation
06:35:02 INFO  OS: GNU/Linux Ubuntu 16.04.2 LTS (Xenial Xerus) build 4.4.0-64-generic 64-bit
06:35:02 INFO  CPU: Intel(R) Core(TM)2 Quad CPU    Q9650  @ 3.00GHz with 4 cores (4 virtual cores)
06:35:02 INFO  Physical Memory: 1.5 GiB
06:35:02 INFO  Free memory: 111.7 MiB
06:35:02 INFO  Maximum JVM Memory: 361.6 MiB
06:35:02 INFO  Language: Norwegian
06:35:02 INFO  Encoding: UTF-8
06:35:02 INFO  Language: Norwegian
```

(and)
```java
java.io.IOException: Cannot run program "cpuid": error=2, No such file or directory
	at java.lang.ProcessBuilder.start(ProcessBuilder.java:1048)
	at java.lang.Runtime.exec(Runtime.java:620)
	at java.lang.Runtime.exec(Runtime.java:485)
	at oshi.util.ExecutingCommand.runNative(ExecutingCommand.java:67)
	at oshi.util.ExecutingCommand.runNative(ExecutingCommand.java:53)
	at oshi.hardware.platform.linux.LinuxCentralProcessor.getProcessorID(LinuxCentralProcessor.java:311)
	at oshi.hardware.platform.linux.LinuxCentralProcessor.initVars(LinuxCentralProcessor.java:100)
	at oshi.hardware.platform.linux.LinuxCentralProcessor.<init>(LinuxCentralProcessor.java:54)
	at oshi.hardware.platform.linux.LinuxHardwareAbstractionLayer.getProcessor(LinuxHardwareAbstractionLayer.java:64)
	at net.pms.PMS.logSystemInfo(PMS.java:1403)
	at net.pms.PMS.displayBanner(PMS.java:304)
	at net.pms.PMS.init(PMS.java:430)
	at net.pms.PMS.createInstance(PMS.java:1111)
	at net.pms.PMS.main(PMS.java:1271)
Caused by: java.io.IOException: error=2, No such file or directory
	at java.lang.UNIXProcess.forkAndExec(Native Method)
	at java.lang.UNIXProcess.<init>(UNIXProcess.java:247)
	at java.lang.ProcessImpl.start(ProcessImpl.java:134)
	at java.lang.ProcessBuilder.start(ProcessBuilder.java:1029)
	... 13 common frames omitted
```

Sami32:
https://github.com/oshi/oshi/pull/332

dbwiddis:
I think you meant to reference [oshi/oshi#335](https://github.com/oshi/oshi/pull/335).  I've fixed the stack traces, are they still being seen?

Sami32:
@dbwiddis No, i was refering to [your licence comments](https://github.com/oshi/oshi/pull/332#issuecomment-283591926), but it's nice to know :)

@Nadahar is the one that have around 15 different Linux OS, so i guess he missed time for new tests or didn't saw, like me, that you already fixed it :+1:

dbwiddis:
I didn't see any previous discussion of license on this issue, so I was confused.  And there's a full release without any license issues. But here's the complete status anyway:
* 3.4.0 was released Java 8 compatible without any dependencies other than JNA and SLF4J
* 3.4.1 was released with Java 7 compatibility using threetenbp (BSD3) and streamsupport (GPLv2 with classpath exception).
* 3.4.2 was released within a day of 3.4.1 removing the streamsupport dependency.

Even though I removed streamsupport thinking it was full GPLv2, I later realized it was probably not an issue at all.  The license clause (GPLv2) is precisely the same one used in the JDK itself.  (In fact, some of the source code is precisely the same, thus the identical license.)  If you don't have license issues building against Java 8, you don't have license issues building against Java 7 with some class files lifted from Java 8.    The "Classpath Exception" specifically allows linking as a library and does not impose any upstream license exceptions.

Anyway, the current 3.4.2 has been released and has no license issues.  There have been [a few commits since 3.4.2 ](https://github.com/oshi/oshi/compare/oshi-parent-3.4.2...master) which are in the SNAPSHOT.

Nadahar:
I haven't done any tests as I wasn't aware of the release, and I've been working on other things. I still need to figure out how to get this (that is, JNA 4.3) to compile under Windows XP with MediaInfo. I really don't know what the real problem is there, but it seems there's no other way than to really dig in and figure it out somehow.

I'll try to get some Linux VM tests done, but the stacktraces wasn't really a "problem" in that sense anyway, rater a minor inconvenience.

I had noe idea that GPL had all these issues, but reading about it now was educational I guess. It's a bit complicated with all the companies that keep trying to exploit OSS for profit in any way they can.

dbwiddis:
Yes, learning all this stuff is an education!

GPLv2's main issue is that if you use it, you are required to license your software with it.  Not bad if you're only doing free software, and I suppose that's the intent of people who license with GPLv2 -- they only want their software used in other free projects.  But if you're writing a library that you hope commercial products can use, GPLv2 is very incompatible with commercial products -- even if you simply link to it.

The GPLv2 CE ("Classpath Exeption"), however, allows linking without imposing this license restriction. It's how OpenJDK (derived from source code for Oracle Java SE) is licensed.  Linking to `streamsupport` was no different, legally, than running the code on OpenJDK.

Nadahar:
@dbwiddis My understanding of GPL is that it's supposed to be incompatible with commercial products, and I think that's generally fine for *applications*. You don't want anyone to be able to make money from a product you've developed to be "free for all". For libraries it's a bit different, libraries can often be developed with commercial use in mind - I think of it like commercial developers can still find it useful to share the "tools" they use to make their commercial product - so that everyone doesn't have to waste their time sitting in separate offices doing the same stuff. My understanding is that LGPL is there for such use - but it seems that a lot of libraries also use plain GPL. I guess it depends on what you want to achieve, if you want to give only the "free world" a benefit or simply want to avoid that people waste time doing the same stuff over and over again.

I'm glad I'm not developing a library, or that could quickly become a headache 😉

Nadahar:
Using 3.4.2:

* OS X 10.8.5: No stack trace, all ok
* Ubuntu 12.04 LTS:
```java
05:27:20 DEBUG Reading file /proc/cpuinfo
05:27:20 DEBUG Reading file /proc/cpuinfo
05:27:20 TRACE 
java.io.IOException: Cannot run program "cpuid": error=2, No such file or directory
	at java.lang.ProcessBuilder.start(ProcessBuilder.java:1047)
	at java.lang.Runtime.exec(Runtime.java:617)
	at java.lang.Runtime.exec(Runtime.java:485)
	at oshi.util.ExecutingCommand.runNative(ExecutingCommand.java:67)
	at oshi.util.ExecutingCommand.runNative(ExecutingCommand.java:53)
	at oshi.hardware.platform.linux.LinuxCentralProcessor.getProcessorID(LinuxCentralProcessor.java:311)
	at oshi.hardware.platform.linux.LinuxCentralProcessor.initVars(LinuxCentralProcessor.java:100)
	at oshi.hardware.platform.linux.LinuxCentralProcessor.<init>(LinuxCentralProcessor.java:54)
	at oshi.hardware.platform.linux.LinuxHardwareAbstractionLayer.getProcessor(LinuxHardwareAbstractionLayer.java:64)
	at net.pms.PMS.logSystemInfo(PMS.java:1403)
	at net.pms.PMS.displayBanner(PMS.java:304)
	at net.pms.PMS.init(PMS.java:430)
	at net.pms.PMS.createInstance(PMS.java:1111)
	at net.pms.PMS.main(PMS.java:1271)
Caused by: java.io.IOException: error=2, No such file or directory
	at java.lang.UNIXProcess.forkAndExec(Native Method)
	at java.lang.UNIXProcess.<init>(UNIXProcess.java:187)
	at java.lang.ProcessImpl.start(ProcessImpl.java:130)
	at java.lang.ProcessBuilder.start(ProcessBuilder.java:1028)
	... 13 common frames omitted
05:27:20 TRACE Updating Processor Ticks
05:27:20 DEBUG Reading file /proc/stat
05:27:20 TRACE Updating System Ticks
05:27:20 DEBUG Reading file /proc/stat
05:27:20 DEBUG Initialized Processor
05:27:20 INFO  Java: Java HotSpot(TM) 64-Bit Server VM 1.7.0_80 (64-bit) by Oracle Corporation
05:27:20 INFO  OS: GNU/Linux Ubuntu 12.04.5 LTS (Precise Pangolin) build 3.13.0-105-generic 64-bit
05:27:20 INFO  CPU: Intel(R) Core(TM)2 Quad CPU    Q9650  @ 3.00GHz with 4 cores (4 virtual cores)
05:27:20 DEBUG Reading file /proc/meminfo
05:27:20 INFO  Physical Memory: 2.0 GiB
05:27:20 INFO  Free memory: 704.1 MiB
05:27:20 INFO  Maximum JVM Memory: 446.5 MiB
05:27:20 INFO  Language: Norwegian
05:27:20 INFO  Encoding: UTF-8
05:27:20 INFO  Language: Norwegian
```
* Antergos: Stack trace on ```dmidecode``` and ```cpuid```. 

I'm aborting the testing, it seems I get the same results as last time. 

@dbwiddis Aren't these fixes in 3.4.2? Do I still need to use the snapshot to get them?

Nadahar:
I still can't build this on WinXP as normal. I then tried 
```mvn test -DforkMode=never``` and then it didn't hang. This only applies to WinXP and I really don't understand what is causing this.

@Sami32 Can you see if you can build this if you use ```-DforkMode=never``` too?

**Comment by Sami32** [comment]
@Nadahar Yes, it did.
I tested it with Java 8, as build it with Java 7 don't give me any issue, and JNA 4.3.0 PR, as it's the one that give this issue.

**Comment by dbwiddis** [comment]
The stack trace TRACE logging is not in 3.4.2.  It's in the snapshot.  It is purely cosmetic though.  The failures are expected.

**Comment by Nadahar** [comment]
@dbwiddis Yes, I know and understand. Thank you, I'll see if I test the snapshot on some of the VMs later just to make absolutely sure it works as intended.

**Comment by Nadahar** [comment]
OS X Sierra 10.12:
```powershell
18:20:12 INFO  Java: Java HotSpot(TM) 64-Bit Server VM 1.8.0_121 (64-bit) by Oracle Corporation
18:20:12 INFO  OS: Apple macOS 10.12 (Sierra) build 16A323 64-bit
18:20:12 INFO  CPU: Intel(R) Core(TM)2 Quad CPU    Q9650  @ 3.00GHz with 4 cores (4 virtual cores)
18:20:12 INFO  Physical Memory: 4 GiB
18:20:12 INFO  Free memory: 1.7 GiB
18:20:12 INFO  Maximum JVM Memory: 683 MiB
18:20:12 INFO  Language: English (United States)
18:20:12 INFO  Encoding: UTF-8
18:20:12 INFO  Language: English (United States)
```

Nadahar:
@valib We can't merge this until we figure out the issue with JNA 4.3.0.

Sami32:
@nadahar Actually we could, as it only affecting XP users wishing building it, AFAIU, so very few users.
Everyone should be able to run UMS without issue.

Or did i misunderstood one more time ?

Nadahar:
@Sami32 If we knew that it only affected building it would be one thing - but we don't know that. All we know is that something "mysteriously" hangs under some circumstances, so the chance that the same problem will manifest itself under other circumstances are is relatively high.

Sami32:
@Nadahar That's only what i've experimented myself.

valib:
@Nadahar who changed this PR and removed all what was done and now this PR is only about the  	c1318c9

Nadahar:
@valib I have no idea, but I had it locally and has pushed my local branch. I've also synced #1171 with master and then synced this with #1171.

This PR can't be merged until #1171 is sorted out.

Nadahar:
@valib I still don't understand why you do that. When #1171 is merged I will update this PR. This has to be rebased anyway, as it currently contains commits that's from an older version of #1171.

Nadahar:
I've updated this PR to be mergable (and tiedied a little bit 😊 ). 

Any further thoughts, reviews, opinions about this?

Nadahar:
@valib Now this fails on dependency convergence again since Oshi use JNA 4.3.0 and we use JNA 4.4.0. I really can't understand why we should have that rule.

Nadahar:
I've managed to do some quick tests using ```-Denfocer.skip=true``` to work around the enforcer issue.

Windows 7:
```logtalk
INFO  2017-06-15 01:17:30.718 [main] Java: Java HotSpot(TM) 64-Bit Server VM 1.7.0_80 (64-bit) by Oracle Corporation
INFO  2017-06-15 01:17:30.719 [main] OS: Microsoft Windows 7 SP1 build 7601 64-bit
INFO  2017-06-15 01:17:30.719 [main] CPU: Intel(R) Core(TM) i7-5820K CPU @ 3.30GHz with 6 cores (12 virtual cores)
INFO  2017-06-15 01:17:30.722 [main] Physical Memory: 15.9 GiB
INFO  2017-06-15 01:17:30.722 [main] Free memory: 8.0 GiB
INFO  2017-06-15 01:17:30.722 [main] Maximum JVM Memory: 3.5 GiB
INFO  2017-06-15 01:17:30.726 [main] Language: Norwegian
INFO  2017-06-15 01:17:30.726 [main] Encoding: UTF-8
```

Kubuntu 16.04.1 LTS:
```logtalk
INFO  2017-06-15 01:13:11.638 [main] Java: OpenJDK 64-Bit Server VM 1.8.0_131 (64-bit) by Oracle Corporation
INFO  2017-06-15 01:13:11.639 [main] OS: GNU/Linux Ubuntu 16.04.2 LTS (Xenial Xerus) build 4.4.0-79-generic 64-bit
INFO  2017-06-15 01:13:11.639 [main] CPU: Intel(R) Core(TM)2 Quad CPU    Q9650  @ 3.00GHz with 4 cores (4 virtual cores)
INFO  2017-06-15 01:13:11.665 [main] Physical Memory: 1.5 GiB
INFO  2017-06-15 01:13:11.665 [main] Free memory: 662.1 MiB
INFO  2017-06-15 01:13:11.666 [main] Maximum JVM Memory: 742.4 MiB
INFO  2017-06-15 01:13:11.685 [main] Language: Norwegian
INFO  2017-06-15 01:13:11.685 [main] Encoding: UTF-8
```

macOS Sierra 10.12.5:
```logtalk
INFO  2017-06-15 01:29:14.386 [main] Java: Java HotSpot(TM) 64-Bit Server VM 1.8.0_131 (64-bit) by Oracle Corporation
INFO  2017-06-15 01:29:14.387 [main] OS: Apple macOS 10.12.5 (Sierra) build 16F73 64-bit
INFO  2017-06-15 01:29:14.387 [main] CPU: Intel(R) Core(TM)2 Quad CPU    Q9650  @ 3.00GHz with 4 cores (4 virtual cores)
INFO  2017-06-15 01:29:14.404 [main] Physical Memory: 4 GiB
INFO  2017-06-15 01:29:14.407 [main] Free memory: 1.6 GiB
INFO  2017-06-15 01:29:14.408 [main] Maximum JVM Memory: 683 MiB
INFO  2017-06-15 01:29:14.466 [main] Language: English (United States)
INFO  2017-06-15 01:29:14.467 [main] Encoding: UTF-8
```

Sami32:
OSHI and JNA 4.4.0, not for Java 7 ?
https://github.com/oshi/oshi/issues/351

Nadahar:
@Sami32 Just look a bit further up in this PR: https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1208#issuecomment-285753180

Also, look at the result of my above test:
```logtalk
NFO  2017-06-15 01:17:30.718 [main] Java: Java HotSpot(TM) 64-Bit Server VM 1.7.0_80 (64-bit) by Oracle Corporation
```
This PR isn't supposed to have problems with Java 7 - what is the problem you're experiencing? Remember that until @valib fixes the Enforcer issue, you must build with ```-Denfocer.skip=true```.

Sami32:
@Nadahar Did you read the link ?

> This is a bug in JNA 4.3 and 4.4. It will be fixed in JNA 4.5, which I'll leave this issue open to track until it's complete. In the meantime, if you don't need Java 7, you can use OSHI 3.4.0 to avoid this problem.

Nadahar:
@Sami32 Yes, but I fail to see the relevance for us. We don't use gradle, spingboot or tomcat. As far as I can see, this bug doesn't affect us. What problem are you experiencing?

Sami32:
@Nadahar Who said i experienced issue ? where did you read it ?
I just pointed at this issue in case it could be relevant with our JNA 4.4.0 update.

Nadahar:
@Sami32 I assumed you had an issue since you seemed to claim that there wasn't Java 7 support.

Nadahar:
@valib 👍

Nadahar:
@valib On second thought, I'm not sure what you do in 1e51e2ae13f7e3c1bf8586737a7344d668202db6 is safe. What exactly does it do? If it does make Oshi make "our" version of JNA, it is a bug that will lead to trouble. You can't just mix different versions of JNA, there's a lot of complications there like differences in the JNI code. 

If it makes Oshi use "its" version of JNA and just means that we don't import that dependency, it's perfectly fine - but it that really what happens?

Sami32:
@Nadahar From what i understand, its make OSHI use our JNA version.

Nadahar:
@Sami32 Yes, I suspect that too and that will be a bug.

Nadahar:
@Sami32 Yes, it might or might not give any problems between JNA 4.3.0 and 4.4.0 (because I think they use the same JNI/native binary), but that can also vary by platform. I know that JNA 4.5.0 will have a new native binary though, but that's not all. There's also changes to the API that will/can break things.

Nadahar:
I looked at the dependency problem a bit more, and it seems there is no good solution: https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html

What I'd like is that the dependency used its version of dependencies and we used our. It seems however that this isn't possible. What Maven does/how Java works is to dump every dependency on the classpath. If two dependencies of different versions are included, they will both be on the classpath but Java will only use the one found first. The second version will never be used, and potential bugs will occur.Using the enforcer plugin the check this doesn't solve the problem, but it make you aware of it at least.

What I'd like was if Maven artifacts were "statically linked" to their dependencies, that is that whatever code was needed from one of its dependencies was included in the JAR. That's not how it is made however, which makes dependency handling potentially very tricky (it has the same basic problem as many Linux distributions).

It seems like the only way to "handle" it is to restrict us to use versions of our dependencies which use matching versions of dependencies as us. In this case it would mean to either drop OSHI or to downgrade to JNA 4.3.0 - and doing the same for the logger. That could quickly become very limiting, which is another reason to really limit the number of dependencies used. The more dependencies we have, to more potential conflicts.

It seems to me that *if* we use exclusions, we'll have to make comments in the POM to both the "primary" dependency and the transitive dependency, and that we would have to do manual checks for the versions used in our dependencies before upgrading any of these. It sounds like something that's just asking to fail, but I don't see another "solution".

Any thoughts?

valib:
@Nadahar I agree. On the other hand the higher dependency version should keep the previous version and only enhance it or fix bugs. Obviously of course. When we remove the Enforcer checking and not exclude some plugin versions the Maven will guess which version should be used and in this case the JNA version 4.3.0 will be choosen and version 4.4.0 ommited. So if after testing it is working on all OS's I recommend to merge it. We use only small part of Oshi so there is a great chance that it will work without any issue.

Sami32:
@Nadahar I don' know, and didn't searched, if a better way exist to handle that, but was i think will happen if we don't set exclusion, is that Maven will throw an warning concerning the concurrency of the 2 different JNA versions.

Nadahar:
This is the best I can come up with, it's far from perfect but at least there's a chance to know what to check. What do you think?

Sami32:
Taken from maven report:

```
[[1;36mDEBUG[m] Error retrieving POM of module-dependency: commons-httpclient:commons-httpclient:jar:3.1; Reason: Some problems were encountered while processing the POMs:
[WARNING] 'build.plugins.plugin.version' for org.apache.maven.plugins:maven-surefire-plugin is missing. @ line 218, column 15
[ERROR] 'distributionManagement.status' must not be specified. @ line 252, column 13
```

Nadahar:
@Sami32 How did you manage to get that? I've searched the Travis logs and can't see that anywhere.

Sami32:
Of course, Travis don't use the `-X` verbose mode, andi guess that i'm the only one still testing the old fashion way :P

Nadahar:
@Sami32 I don't know when running Maven in debug mode became the "old fashion way" 😉 

The messages you see isn't related to us at all, they apply to the POM of ```commons-httpclient:commons-httpclient```. ```commons-httpclient``` is abandoned back in 2005 AFAICR and hasn't been updated as Maven rules have changed (I think it's made for Maven 2, which was very different). 

The only solution (not that we need one, as the messages has no conseqence that I'm aware of) is to not use ```commons-httpclient``` and use another (updated) library instead, like for example ```org.apache.httpcomponents:httpclient``` instead. I took a quick look at it actually, because this library causes several of our dependency convergence issues as well, but too much had to be changed. This library is only used for the automatic updater, and as I don't know how "the other side" (the webserver) work/what the rules are or how to test it, I abandoned it. It should be easy enough to do for @SubJunk though.

Sami32:
@Nadahar We ready saw that Maven test was not run by many, and with `-X` variant again less i think, but i could be wrong.

I never said that it was of any consequence, but that it was worst mention it.

I did tried to update it some month ago, but stopped because of my skill limit.
https://github.com/UniversalMediaServer/UniversalMediaServer/compare/httpclient

Nadahar:
@Sami32 Maven is run multiple times every time we push by Travis. There's no reason to do that locally for every little change. I try to remember to run ```mvn test``` before I push something, but it happens that I forget it. But, I almost never run ```mvn package``` because creating the installer is so slow.

Sami32:
@Nadahar He he, welcome to my world :wink:

Sami32:
@Nadahar Doesn't work anymore with Java 7, at least under WinXP 32-bit; i didn't tested with Java 8, but if you need i can do.
I guess that i just test the build and didn't go further than that, as i already tried it before. My bad.

[hs_err_pid2392.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/1092100/hs_err_pid2392.txt)

Nadahar:
@Sami32 😭

Nadahar:
@Sami32 I've retested on my XP VM without error. I tested Java 8 by building with Maven and running ```UMS.exe``` from the ```target``` folder. I tested Java 7 by running from Eclipse as I don't want to get into fixing all paths and stuff to test it with Java 7 without Eclipse.

I can both build and run without error, so please give some more information of exactly what you're doing when it crashes. How long does you logfile come before crash?

Sami32:
@Nadahar It's all i get, as the GUI doesn't appear and the _debug.log_ just reset to zero.

Nadahar:
@Sami32 How do you run it? Are you running the ```.exe``` or are you running ```javaw...```? If you use ```java.exe -jar ums.jar``` you might get logging to the console until it crashes.

Sami32:
With _UMS.exe_

Nadahar:
@Sami32 Try using ```java.exe -jar ums.jar```.

Sami32:
@Nadahar I don't really see the difference but here it is:
[hs_err_pid3064.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/1092361/hs_err_pid3064.txt)
capture_06212017_193744

Nadahar:
@Sami32 What is the output from ```java.exe -version```?

Nadahar:
@Sami32 I assume that the output from java.exe is similar to the above JRE information. I'm using ```build 1.7.0_80-b15``` (the lastest "official") which might explain the difference.

Sami32:
capture_06212017_194637

Nadahar:
@Sami32 Are you 100% sure that this was the JDK that was used when building the Java 7 JAR (that is, ```mvn -v``` displays the same information)?

Sami32:
@Nadahar I don't think that have anything to do with that, but if you need i can reinstall the latest non fixes updated Java version

Sami32:
Yes

Sami32:
No, i build with 7u80

Nadahar:
@Sami32 Ah... try using the same version.

Sami32:
@Nadahar Sama same, as expected by me.
[hs_err_pid3256.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/1092422/hs_err_pid3256.txt)

Nadahar:
@Sami32 I just built with Maven and Java 7 by setting ```JAVA_HOME``` and run it using ```java -jar ums.jar```. Still no crash.

Nadahar:
@Sami32 So, since you're so sure it's not the Java version, what do you think is causing it? CPU incompatible?

Nadahar:
@Sami32 What if you downgrades JNA to 4.3.0 in the POM?

Nadahar:
@Sami32 Another thing you can try is to download [jnidispatch.dll.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/1092443/jnidispatch.dll.txt) and place it somewhere on your computer that you then set ```jna.boot.library.path``` to that file..:
```powershell
java -Djna.boot.library.path="<pathto>\jnidispatch.dll" -jar ums.jar
```

Nadahar:
@Sami32 If you try with the ```jnidispatch.dll``` I compiled, use JNA 4.4.0 as it's compiled with 4.4.0.

Sami32:
Arrrgg :sob: 
OK, i think i'll take break and try later.

Sami32:
@Nadahar With Java 7u80 from oracle:
- No go with JNA 4.4.0 and the separated JNA DLL.
- JNA 4.3.0 :+1: 

With Java 8u60 from Oracle: same same.

It seem to be a 32-bit issue ?
I still get lucky that the 4.3.0 version work for me :wink: 

The bad think, is that i've now to reinstall Java, UMS, my own excecutables and modify my environment variables :cry:

Nadahar:
@Sami32 Why do you have to reinstall anything? You can have many JDKs installed at the same time.

It's not a 32-bit issue, my XP VM is 32 bit as well.

If JNA 4.3.0 works, then this is exactly what I was afraid of using ```<exclude>``` on JNA. The problem is that Oshi is built with JNA 4.3.0 and we use JNA 4.4.0 - which leads to crash when Oshi tries to detect your system information (calling another version of JNA).

Can you confirm that JNA 4.3.0 works for both Java 7 and Java 8?

Nadahar:
https://github.com/oshi/oshi/issues/376

Sami32:
@Nadahar Yes, with JNA 4.3.0 it work fine on Java 7 and Java 8.

To save place on my already near full main hard drive where my OS is.

Nadahar:
@Sami32 I don't know how much you've edited this since I read it, but what I know was that it was my understanding that it worked with JNA 4.4.0 and not JNA 4.3.0. Whatever the case is, let's wait and see if there's any response from oshi.

dbwiddis:
What response are you waiting on?

Nadahar:
@dbwiddis Whether you have any plans to do a release with JNA 4.4.0 as a dependency in the foreseeable future.

dbwiddis:
I had been planning on waiting until JNA 4.5.0 was released because of https://github.com/oshi/oshi/issues/351 but if you need an earlier incremental release I can push one.

Nadahar:
@dbwiddis A minor release would have been great if it's not too much trouble. If not, it seems like we'll have to downgrade to JNA 4.3.0 and then there's a lot of other stuff that need to be retested.

dbwiddis:
Give me a week.

Nadahar:
@dbwiddis Great, thank you 👍

Nadahar:
I'm getting some strange behaviour now that this is merged. This code in ```WindowsCentralProcessor.getSystemCpuLoadTicks()```
```java 
        Map<String, List<String>> irq = WmiUtil.selectStringsFrom(null,
                "Win32_PerfRawData_Counters_ProcessorInformation", "PercentInterruptTime,PercentDPCTime",
                "WHERE Name=\"_Total\"");
```
is taking a long time to complete. I timed it to approximately 91 seconds on my development computer. Do any of you guys have similar problems?

valib:
@Nadahar for me it works without any problem.

Nadahar:
@valib I did for me as well when I tested this earlier and I have no idea what's changed, but as it is now it's quite hopeless. Any user experiencing this will simply think that UMS doesn't start.

Nadahar:
I think I've identified the problem, but I have no idea why it happens. It's the query itself that takes time, not JNA or oshi. Running this from power shell takes just as long:
```powershell
$M1 = Get-WmiObject -Query 'SELECT Name,PercentIdleTime,PercentPrivilegedTime,PercentUserTime,PercentInterruptTime,PercentDPCTime FROM Win32_PerfRawData_Counters_ProcessorInformation WHERE NOT Name LIKE "%_Total"'
```

The question is why, if this is something that only affects my computer and how we should handle it. As it is now, it's very hard for me to debug UMS as it takes about 1,5 minutes every time I start.

Nadahar:
In fact, running simply 
```powershell
$M2 = Get-WmiObject -Query 'SELECT * FROM Win32_PerfRawData_Counters_ProcessorInformation'
```
is just as slow.

Nadahar:
I finally had to bite the bullet and reboot my computer (loads of software/windows to close and reopen). That solved the problem for now, but I still don't know what caused WMI to become so slow. I couldn't really find anything "useful" when searching for it, other people reporting that WMI queries were slow were mostly talking about seconds, not minutes.

I don't know if we can expect this to be a problem and if something should be done or not.

Sami32:
@Nadahar From what i read it affect Windows 7 users.
Do that do the trick for you:

> The CSE policy is for the Client Side Extension used in older Windows like Vista and 2003 clients.  If you don't have any old machines in need to follow the SBS Group Policy then it's fine to disable them.
> 
> Here is the description: http://blog.mpecsinc.ca/2008/12/sbs-2008-group-policy-client-side.html
> 
> I can't find an exact description to what an Extension Policy is for but from my research it seems as that it's related to the CSE policy to apply the network printers.
> 
> If you don't have any older machines in your environment then I wouldn't worry about it. 

In case you still get that issue, perhaps [this link](http://toncigrgin.blogspot.fr/2015/) can help you debug it.

Nadahar:
@Sami32 Thaks for trying to help, but I'm not sure how any of those links are related. The group policy update for older OS'es doesn't apply to Windows 7, and it's for use in Active Directory environments in any case. I also doubt would have to do with why this WMI query is so slow even if it were relevant for this computer/OS.

The last link only explains how to use the queries as far as I can see, nothing about why a query would take > 90 seconds. I can't "test" it any more now as it is quick since I rebooted (first query about 5 seconds, subsequent queries "instantaneous" - which is "normal" behaviour from what I understand).

My worry is, as always, that we have no idea under which circumstances this occurs or how widespread it is. Most users will conclude that UMS doesn't start when it takes almost 2 minutes before anything happens. Even the splash screen is created after this.

Sami32:
@Nadahar It was just a potential diagnose help link, if you ever meet that issue again.

That's said, you could always returm to Windows XP, we doesn't wait more than 50s LOL

Issue #1210, opened 2017-03-02
Video/audio asynchronous on variable frame rate mkv files


owlhuang:
### SUMMARY
Video and audio is slowly drafting away on some variable frame rate mkv files.  They are about 9 seconds apart after a 2hr40min (9600s) video is played.  It appears that mediainfo is giving out incorrect frame rate number (24.099 FPS instead of 23.976 FPS.)  Detail and a proposed fix is as below.

### ENVIRONMENT
- ums 6.6.0 running on a linux box
- renderer is Playstation 3 (with mediainfo on)
- ums engine of choice is mencoder, to render subtitle correctly

### DETAIL
With UMS debug information on, '-ofps 24' is sent to mencoder as parameter.  Yet the video stream is AVC variable frame rate with 23.976 fps.  If manually set '-ofps 24000/1001', the resulting video and audio are perfectly matching to each other.

Digging into UMS codes, the number for '-ofps' in MEncoder (MEncoerVideo.java) is originated from 'Frame rate' in MediaInfo (LibMediaInfoParser.java) around line 119.  

From mkvtoolnix author, MediaInfo is providing 'Original frame rate' from the video stream directly, while 'Frame rate' is calculated as '1/duration' from the container.  The author also explains the duration in a mkv file is NOT the average duration but the most commonly used duration of the frames.
https://github.com/mbunkus/mkvtoolnix/wiki/Wrong-frame-rate-displayed

I believe that explains why MediaInfo is giving out 24.099 instead of 23.976 to MEncoder, for a variable frame rate video stream.

P.S. ffmpeg is giving out a correct 23.98 fps but I would prefer to use MediaInfo for Playstation3 renderer.  It is easier to config the codec/container the renderer supports.

### PROPOSED FIX
I would suggest to use 'Original frame rate' instead of 'Frame rate' for MediaInfo parsing, when it is present.  The codes in LibMediaInfoParser.java would be something like
```
- media.setFrameRate(getFPSValue(MI.Get(video, i, "FrameRate")));
+ value = MI.Get(video, i, "OriginalFrameRate");    // OriginalFrameRate needs to be confirmed as the correct token to retrive the intended value
+ if (value.isEmpty())
+     media.setFrameRate(getFPSValue(MI.Get(video, i, "FrameRate")));
+ media.setFrameRate(getFPSValue(value));
```

### OTHER SOLUTION
[...]

Issue #1209, opened 2017-02-28
Playstation 3 audio changes to static during playback


TimSC:
I am playing back videos to my PS3. All is fine until after a seemingly random time, the audio changes to static noise. I've seen this in ums-6.5.0 and ums-6.6.0. The only way to recover the audio is to restart UMS. I'm running with "TranscodeVideo = MPEGTS-H264-AC3" added to my Sony-PlayStation3.conf (and I can't find any alternative that works).

Before the loss of audio, the playstation reports the audio is Dolby Digital 5.1 48kHz 650kbps. After the glitch, it reports Linear PCM 2.3Mbps.

I've noticed other people posting about similar problems. http://www.universalmediaserver.com/forum/viewtopic.php?t=9051

Trace logs are attached for two instances... I paused about 5-10 sec after the glitch and saved. 
[debug.log.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/809280/debug.log.txt)
[debug2.log.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/809282/debug2.log.txt)

Sami32:
@TimSC Could sen the full [MediaInfo](https://mediaarea.net/en/MediaInfo/Download) log of one of the file giving you issue ?

In the "View" tab select "Text", and in the "Debug" tab select "Advanced mode", then export the file and post it here.

Sami32:
@TimSC :+1: 
I was suspecting something like that.
Could you backup your _ums.jar_ and use that one instead ?
https://we.tl/UA1GkwvZzr

Let's see if it make any improvements with one or two of the video engines.

Could you post the _FFmpeg_ log as well ? Thank you.
`ffmpeg -i s1e15 Grail.mkv -report`

TimSC:
@Sami32 The same problem occurred with the updated ums.jar (6.6.1-snapshot) using both the default ffmpeg engine and mencoder transcoding.

I used the older MediaInfo as it was in my linux distro repository. I can try a later version if that would help. The ffmpeg log seems to be identical to what I posted in the original report. Are you referring to something else?

Sami32:
@TimSC Interesting, and sadly i'm not surprise. It's anyway making things a little more clear for us.

Could you post the new data from this log with the UMS version i sended you ? Thanks.
Normally, you'll see the framerate infos in this log, and they wasn't displayed/used in your anterior log.
```logtalk
[...]

PR #1211, opened 2017-03-03
Prefer FrameRate_Original over FrameRate from MediaInfo (#1210)

Nadahar:
This is an implementation of the suggested fix to MEncoder sync issues in #1210. I haven't tested it. Please test and give feedback.

Thanks to @owlhuang for doing all the work and presenting it in a very easy-to-understand way 👍

### DIFF
--- a/src/main/java/net/pms/dlna/LibMediaInfoParser.java
+++ b/src/main/java/net/pms/dlna/LibMediaInfoParser.java
@@ -139,6 +139,13 @@ public class LibMediaInfoParser {
 							media.setAspectRatioContainer(MI.Get(video, i, "DisplayAspectRatio/String"));
 							media.setAspectRatioVideoTrack(MI.Get(video, i, "DisplayAspectRatio_Original/String"));
 							media.setFrameRate(getFPSValue(MI.Get(video, i, "FrameRate")));
+							// Prefer "FrameRate_Original over FrameRate if it's there because it is more accurate
+							// https://github.com/UniversalMediaServer/UniversalMediaServer/issues/1210
+							value = MI.Get(video, i, "FrameRate_Original");
+							if (isBlank(value)) {
+								value = MI.Get(video, i, "FrameRate");
+							}
+							media.setFrameRate(getFPSValue(value));
 							media.setFrameRateOriginal(MI.Get(video, i, "FrameRate_Original"));
 							media.setFrameRateMode(getFrameRateModeValue(MI.Get(video, i, "FrameRate_Mode")));
 							media.setFrameRateModeRaw(MI.Get(video, i, "FrameRate_Mode"));


Sami32:
That's doesn't look like a bug to me, but a buggy source as we can find a lot on Torrent, and the intrinsec limitation of the Matroska format.

I read users having the opposite issue, they solved their issue by stopping using `Framerate_Original` and using `Framerate` instead.

Though i think that could be safe if it was restricted to H264 video codec only, as the oppositedidn't occured with H264 codec, AFAIK.

Nadahar:
I don't know enough about this to have a viewpoint, you should decide and agree what is correct or not. I simply implemented the suggested solution.

owlhuang:
Good comment, Sami.  It is always good to be careful when we are pulling workaround commit like this.  There is [another source ](http://forum.doom9.org/showthread.php?t=172086)having issue with `FrameRate_Original`.

The thread in Sami's comment has a video, which cannot be played by any players he has, even `FrameRate `is correct and `FrameRate_Original `is not.    The other thread has `FrameRate_Original `as a funny number yet it is the working one.  Re-encoding with `FrameRate `doesn't work for him.  For me, the first one is messing up with frame rate and the result doesn't work on any players, and the second one got a working stream with strange number in `FrameRate_Original`.  I would say both of them are having incorrect encoded streams.

I understand there are lots of incorrect encoded streams/files like the first example, but those are not the target this workaround trying to cover.  I agree that this is not a bug on Universal Media Server.  But this commit is not trying to address buggy sources but a workaround for the "mis-understanding" between MediaInfo and MKV files, which are arguably one of the most commonly used container format.  MediaInfo calculates the `FrameRate `from `1/duration` in the container header, and `FrameRate_Original `from the video stream itself.  Unfortunately, the duration from a MKV container does NOT represent the average duration but the mostly commonly-used duration.  So it is inevitable to have a wrong `FrameRate `if the MKV contains a variable frame rate video stream, for example.  In the other hand, `FrameRate_Original `is from the video stream directly.  It should be the trustful source when any player trying to play that video stream.  In the perfect world, they should be identical.

This is the reason why the `FrameRate_Original `is the preferred number.  If we do have evident that `FrameRate `should be used instead of `FrameRate_Original` in some cases, I could help to take a look and/or add the condition Sami mentioned in the previous comment.

Universal Media Server is a good and convenient software.  It is a pleasure for me to contribute in any way.  Thanks, guys.

owlhuang:
Is there anything I could help to get this patch in?

Nadahar:
@owlhuang It's about making sure that it's the right thing to do. As said before, I don't know enough about this, @Sami32 seems to disagree and neither @valib nor @SubJunk have presented any opinion.

owlhuang:
No problem.  Just wanted to see if anything is on me.  I will wait for @valib or @SubJunk to jump in.  Thanks for the timely response, @Nadahar .

Sami32:
I don't disagree, something should be done, i just don't consider it lightly, as it should affect near all the users, and i don't have a huge video library as @SubJunk for real life test.

Some other issues related to the bitrate are also there, so it could be nice to solve it as well. I didn' yet go deeper though.

owlhuang:
@Sami32, could you share a bit detail and/or reference to the bitrate issue?  If it happens on PlayStation3 or PlayStation4, I could take a look into it.

I understand and agree your concern.  Just want to find a way to move it forward.  I'll try the sources on my side and report back.  If you have other suggestion on what needs to be done before this patch get commited, please let me know.

From the authors of MKV tools and mediainfo, I believe this patch is the right way to go, to workaround the mis-understanding between these two software.  Furthermore, using the frame rate from a video stream (`FrameRate_Original`) should be much more reliable than from a container header (`FrameRate`), IMO.

Sami32:
That will imply that only the renderer having `MediaInfo =true` will be able to get it working, as i wasn't able to get that kind of informations from _FFmpeg_ itself, also with the `- 9 -loglevel 99` parameters.

Nadahar:
@Sami32 That shouldn't be a "problem" as such, many things only work if you use MediaInfo. Are you saying that FFmpeg only reports the average framerate?

Sami32:
@Nadahar Yes.

owlhuang:
Reporting average frame rate is correct, and that's what my fix does.  Current MediaInfo reports `FrameRate` from 1/duration, and that duration is the 'most used' duration in MKV, not average frame rate.  You could read my original post for detail.

For software reports average frame rate, it should work as it currently is.  In another word, with MediaInfo=false, it should work without this fix (and work with this fix.)  It is getting the frame rate from the video stream itself, which is correct (and identical to `FrameRate_Original` with MediaInfo=true.)

Sami32:
FFmpeg report only what is equivalent to MediaInfo "framerate", and nothing about the" Original_framerate" that you are referring to.
But if you find a way to make FFmpeg display it i'm interested.

owlhuang:
@Sami32 , are you saying FFmpeg calculates the frame rate as 1/duration, and the duration number is from the container header, too?

owlhuang:
To confirm ffmpeg's behavior, I just did a quick test with ffmpeg 3.2.4.

Ffmpeg reports frame rate from the video stream, not from the container header.  That would be the same value reported by MediaInfo as `FrameRate_Original`, when it is available.  I can also confirm on that particular file I have, the value 23.98 is correct, and identical to `FrameRate_Original` from MediaInfo, as 23.976.  For the record, MediaInfo report 24.009 as `FrameRate` for that file, which causes the problem.  I believe my fix is doing the right thing. 

In fact, even this fix doesn't fix ffmpeg (mediainfo = false), we should pull in this commit, IMO.  As @Nadahar  said, that should be a problem.  And no, this problem doesn't affect ffmpeg, nor does this fix impact ffmpeg.

From ffmpeg 3.2.4
`Stream #0:0(jpn): Video: h264 (High), yuv420p(progressive), 1280x696, SAR 1:1 DAR 160:87, 23.98 fps, 23.98 tbr, 1k tbn, 47.95 tbc (default)`

From mediainfo 0.7.91
```
Frame rate mode                          : Variable
Frame rate                               : 24.099 FPS
Original frame rate                      : 23.976 FPS
Color space                              : YUV
Chroma subsampling                       : 4:2:0
```

Sami32:
Exactly what i'm saying since the beginning, it confirm that nobody yet did test it extensivelly...
It's not about your particular file only or what @Nadahar said or believe, but about general users concern.
[VideoToolbox_h264_Original_frame_rate.mp4.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/844492/VideoToolbox_h264_Original_frame_rate.mp4.txt)

From FFmpeg N-80899-gf41e37b:
```logtalk
  Duration: 00:00:10.00, start: 0.000000, bitrate: 495 kb/s
    Stream #0:0(und), 1, 1/15360: Video: h264 (High) (avc1 / 0x31637661), yuv420p, 704x576, 492 kb/s, SAR 128:117 DAR 1408:1053, 60 fps, 60 tbr, 15360 tbn, 50 tbc (default)
``` 
From MediaInfo 7.93:
```logtalk
Frame rate mode                        : Constant
Frame rate                             : 60.000 FPS
Original frame rate                    : 25.000 FPS
```

JeromeMartinez:
Please use `FrameRate_Original` only if you need to know the raw stream frame rate metadata when not same as the container frame rate, not for something else (e.g. A/V synchronization, or playback etc) as this line in MediaInfo is only for showing that the container and the raw stream have incoherent frame rate values.
Frame rate info is for information only especially with VFR files, you need to rely on timestamp of each frame for A/V sync.

Jérôme, main developer of MediaInfo.

owlhuang:
Thank you for the response here, Jerome.  Good info.

Universal Media Server uses MediaInfo to decide the playback frame rate when token mediainfo=true.  We would need to pick between `FrameRate_Original `and `FrameRate `in this case.  Yes, I understand for VFR files, the only thing reliable is the timestamp of each frame, and both `FrameRate_Original `and `FrameRate `are for information only.  But for some technical issue with UMS, we need an "average" framerate.  And I believe using the raw stream frame rate is more reliable than the container header, and that happens to be `FrameRate_Original`.  In fact, MKV author already confirms the container header will NOT contain the average frame rate but the most frequently used frame rate (in duration to be exact.)

@Sami32, again I appreciate your effort to make the UMS better for general users and acknowledge your concern.  It's not about my own file or your particular file.  You or me could always find a file, which is not working in one case or the other.  But it doesn't make sense to block a fix just because it doesn't work on that particular file.  This would practically render no fix to get into the code base because we could always find some files not compatible with the new fix.  As you mentioned before, this world would always contain some incorrect encoded files.  But if the fix is obvious doesn't work on most files, I agree we should not pull it in.  But this is NOT the case here.

In fact, your file is a good example to show my fix is correct and comes in handy.  If you read carefully into `dlna/DLNAMediaInfo.java`, you would know that tbc should take the priority for deciding the frame rate.  In your case, it would be (50/2)=25, which is exactly identical to `FrameRate_Original`.  

```
} else if ((token.contains("tbc") || token.contains("tb(c)"))) {
        // A/V sync issues with newest FFmpeg, due to the new tbr/tbn/tbc outputs
        // Priority to tb(c)
        String frameRateDoubleString = token.substring(0, token.indexOf("tb")).trim();
        try {
                if (!frameRateDoubleString.equals(frameRate)) {// tbc taken into account only if different than tbr
                        Double frameRateDouble = Double.parseDouble(frameRateDoubleString);
                        frameRate = String.format(Locale.ENGLISH, "%.2f", frameRateDouble / 2);
                }
        } catch (NumberFormatException nfe) {
                // Could happen if tbc is "1k" or something like that, no big deal
                LOGGER.debug("Could not parse frame rate \"" + frameRateDoubleString + "\"");
        }

} else if ((token.contains("tbr") || token.contains("tb(r)")) && frameRate == null) {
        frameRate = token.substring(0, token.indexOf("tb")).trim();
} else if ((token.contains("fps") || token.contains("fps(r)")) && frameRate == null) { // dvr-ms ?
        frameRate = token.substring(0, token.indexOf("fps")).trim();
```
From Jerome, `FrameRate_Original `only shows up when the information is conflict between the container and the video stream.  In this case, we should pick the frame rate from the video stream, IMO, which happens to be `FrameRate_Original`.

Also we should decide whether a patch is correct by the reasons behind it, not only by our guts feeling that it might impact people.  I provided the background information in my original bug report to support my fix.  Having said that, I respect everyone's guts feeling because it helps to raise the quality bar for the software.  It is always good to be suspicious.  But based on that feeling, we should have a way to prove whether that is right or not, IMO.

Everyone would love to contribute to a good open source project like Universal Media Server,  I am just one of them.  We are all trying to make this software better for the world, and this fix would be one tiny step toward it.  Please considering commit the patch.

Nadahar:
I'm still not sure exactly what is the best way to handle this. As I understand @JeromeMartinez, we should NOT use ```FrameRate_Original```. I'm wondering what we're actually using this for. Are we dictating to FFmpeg what FPS to use? If so, wouldn't it be better to let FFmpeg figure out the correct FPS dynamically? Or do we need to report it as a part of the DLNA specification? If so, we might be doing something wrong as I seem to remember to have seen support for VFR in the specification.

JeromeMartinez:
>  In fact, MKV author already confirms the container header will NOT contain the average frame rate but the most frequently used frame rate (in duration to be exact.)

Just to be clear: the " most frequently used frame duration" (1/frame rate) is not used as is by MediaInfo, just as an hint.
when you have "Frame rate : 24.099 FPS" and "Frame rate mode : Variable", this could be:
- info comes from Matroska tags and this is total frame count divided by total duration (such tags are in some files, this is not mandatory), you should use it
- there is a bug in MediaInfo and this is a false positive of variable frame rate detection and `FrameRate_Original` is the true value for this file (but it could be just a wrong encoder config in another file which is really with variable frame rate), and in that case I suggest you open an issue on [MediaInfo library issue tracker](https://github.com/MediaArea/MediaInfoLib/issues) instead of trying to find a work around which will never work for all files you have.

> We are all trying to make this software better for the world, and this fix would be one tiny step toward it. Please considering commit the patch.

From my point of view, you should first be sure that the other project (here: MediaInfo) is not providing wrong information before trying to find a work around which may be worse with other files, and if MediaInfo is buggy you'll have a bigger step (fix in MediaInfo and better UMS).

owlhuang:
Thanks, all inputs are good.  Let's wait for @SubJunk for this one.

@Sami32 , from the facts we know, this only happens on VFR files.  We could add a check and apply the fix only when `FrameRate_Mode = Variable`.  Do you think it reduce the risk and impact on general users?

@JeromeMartinez , Matroska container uses "most frequently duration" for variable frame rate stream as reference in this [link](https://github.com/mbunkus/mkvtoolnix/wiki/Wrong-frame-rate-displayed).

> With variable frame rate (e.g. if an external timecode file is used) video the default duration is calculated in a totally different way. mkvmerge calculates the duration for all frames according to the timecode file, counts how often each duration occurs and then uses the duration that occurs the most as the new default duration. This may be a value that doesn't even reverse to any of the often-used frame rates that people are used to.

If MediaInfo gets the FrameRate from this number, that would be the reason why 24.099 comes up.  This will be a common case for all VFR MKV files.  

I cannot agree more on 'fix the uptream' if it has problem.  I would be happy, too, if mediainfo would change not to use that 'duration that occurs the most.'  Yet I guess that's the only information in a MKV container about frame rate?  Any suggestion on dealing with this situation?

Nadahar:
I think it's impossible to answer how to best handle this without knowing exactly how we use this information. Since @owlhuang claims that this impacts audio/video sync it points to this being handed to the transcoding engine at some point. But, I did a quick search in the code and couldn't find that this is being done.

Without knowing the exact use of this value, it's impossible to determine how to best deal with this.

JeromeMartinez:
> @JeromeMartinez , Matroska container uses "most frequently duration" for variable frame rate stream as reference in this link.

MediaInfo does not use this item alone, so this is not the discussion here.

> I cannot agree more on 'fix the uptream' if it has problem. I would be happy, too, if mediainfo would change not to use that 'duration that occurs the most.'

I don't think it does, I think the issue is from somewhere else for MediaInfo.

>  Any suggestion on dealing with this situation?

I suggest to be clear about MediaInfo first (if it has a bug, let's kill it), and in all cases I could help only after having reviewed the file.
Please open a ticket with the file (or you can contact me in private if content can not be made public) and we see if MediaInfo can be changed for this issue and/or if MediaInfo can provide a more accurate value (you want the average frame rate in that case, right? Just wondering about how it could be useful for A/V sync as relying on it will not permit A/V sync if you transform the stream in constant frame rate with this average frame rate)

Note that I focus on potential MediaInfo issues (wrong output from MediaInfo), I don't really understand what is done with this value in UMS and this is not my own goal (which is accurate MediaInfo output compared to what is expected from it i.e. here average frame rate of a VFR stream or showing nothing if not possible, instead of a wrong value).

owlhuang:
@Nadahar , are you saying how this frame rate information being used by the transcoding engine?  It is in `encoders/MEncoderVideo.java` around line 1611

```
		String frameRateRatio = media.getValidFps(true);
		String frameRateNumber = media.getValidFps(false);

```
```
		String framerate = (frameRateRatio != null) ? frameRateRatio : "24000/1001"; // where a framerate is required, use the input framerate or 24000/1001
		String ofps = framerate;

```
```
		cmdList.add("-ofps");
		cmdList.add(ofps);

```

owlhuang:
@JeromeMartinez , I could share with you offline.  Please tell me how to do it since it is over 4GB.

For the record, I agree totally on your focus, on MediaInfo.

Nadahar:
@owlhuang From a quick look at the code, it seems the value is only used if using AviSynth with FFmpeg or MEncoder. MEncoder is sort of "deprecated" since the project is more or less abandoned, but we keep it around "as a backup" because it can sometimes transcode content that FFmpeg won't. Why aren't you simply using FFmpeg to transcode?

Also, the code in ```MEncoderVideo``` dealing with frame rate doesn't exactly look very robust to me, it seems like a lot of assumptions are made. I'd think that there are good chances that there's bugs in there.

owlhuang:
@Nadahar , the reason is for subtitles.  I tried ffmpeg and the subtitles are not working correctly.  

I know mencoder is more like a "backup."  And I have to agree with you that the codes is not robust to me, either.  But since we already have a fix for something, why not improving it?

Nadahar:
@owlhuang The point is to not "fix" one thing and break a dusin others. That's why we need to figure out where the problem *really* is. I just looked [here](http://www.mplayerhq.hu/DOCS/man/en/mplayer.1.txt) and found this:
```powershell
       -ofps <fps>
              Specify a frames per second (fps) value  for  the  output  file,
              which  can  be different from that of the source material.  Must
              be  set  for  variable  fps  (ASF,  some  MOV)  and  progressive
              (30000/1001 fps telecined MPEG) files.
```
I'm not sure how to interpret this, but does it say that MEncoder doesn't support VFR? If so, there doesn't seem to be a "correct" solution. @SubJunk this is really your field of expertise, it would be really helpful with your input.

owlhuang:
@Nadahar , it says `-ofps` is mandatory for VFR and progressive files.  And from the result, it has to be the average frame rate, aka total duration divided by the number of frames.  That's why we should use `FrameRate_Original `when container and video stream disagree on frame rate.  `FrameRate `is just an off number for some reasons, which I could work with @JeromeMartinez to find out.

`FrameRate `might be off because

- MediaInfo is using 1/duration from MKV container (from the [link ](https://github.com/mbunkus/mkvtoolnix/wiki/Wrong-frame-rate-displayed)but Jerome disagrees)
- Something is happening to cause MediaInfo reporting a strange value, e.g. 24.099.  We might need to know how MediaInfo calculates this number

Believe me, I don't want to break anything.  Just try to move things forward.  Apologize if I am offending anyone or breaking rules here.

Nadahar:
@owlhuang I don't think you're offending anyone or breaking any rules, I'm just trying to explain why we're reluctant to implement something without fully understanding the consequences.

I know to little about this and the various formats, but it seems to me that given that the target format supports VFR MEncoder shouldn't be used to transcode VFR sources. If we have to convert it to CFR, the average FPS would be the most correct value to use I would think. But, that doesn't explain why audio and video get out of sync, I'd think it was MEncoders responsibility to sync everything to the specified output FPS. Are we actually trying to "work around" a bug in MEncoder here?

owlhuang:
Yes, @Nadahar .  As I said in the original bug report, in addition to fix in UMS,

> 
> OTHER SOLUTION
> 
>     Fix MediaInfo: mediainfo should not use 1/duration for variable frame rate mkv files
>     Fix MEncoder: mencoder should be smart enough to transcode from 23.976 fps to 24 fps while keeping video/audio in sync.
>     Do not use MediaInfo at all: This is more like a workaround.
>

Nadahar:
@owlhuang I still don't get it. If you don't use MediaInfo the parsing will be done by FFmpeg. Does FFmpeg return a "more correct" value in your case? What value do you think MediaInfo should use for VFR MKV files?

If we're trying to make a workaround for a MEncoder bug, why is MediaInfo named as a possible culprit? To make an effective workaround for a bug in MEncoder, we'd need to know exactly what to avoid and then apply a "fix" for that in ```MEncoderVideo```. It would be wrong to change what's stored in ```DLNAMediaInfo``` if that is the true value.

owlhuang:
@Nadahar , ffmpeg reports tbc as 47.95.  UMS divides that by 2.  We will have frame rate as 23.975, which is same as `FrameRate_Original` from media info.

If MediaInfo uses the frame rate directly from the video stream for a VFR MKV file, UMS will behave correctly.  But MediaInfo does not have responsibility to change its information definition specially for UMS, IMO.  I won't name MediaInfo the culprit.  It is just unfortunate how each software interpret frame rate for a VFR file.  

From the experiment I did, MEncoder has difficulty to convert the frame rate to a new value (e.g. from 23.976 to 24.). Video/audio mismatch would be the result.  The fix/workaround here is to use the same frame rate for MEncoder, which eliminates the need of conversion, hence producing a correct result.  In the end, there is no need to do the conversion.  For the record, since the input is a VFR stream, we just keep the same average frame rate.  I understand there is no real frame rate for a VFR stream.

The value in `DLNAMediaInfo` is from `FrameRate` from MediaInfo.  It is incorrect (24 in my case, and 60 in @Sami32 's case and should be fixed, IMO.  With this patch, this value would be identical to what ffmpeg/tbc gives.

Sami32:
@owlhuang What are the subtitles types that give you issues with _FFmpeg_ ? _SubRip_ ?

Your problematic video file should play fine with _Avisynth/FFmpeg_ engine ;)

owlhuang:
I have SubRip and VobSub mostly.  Some are SSA but not a priority.  Last time I check that is years ago.  I would check that again, too.  Thanks, @Sami32 !

owlhuang:
I built a small collection of files and did a quick check with ffmpeg with UMS 6.6.0, renderer is a PlayStation 3.  Result as below
```
avi_xvid_480p_23.976_ac3_5.1_ext_vobsub.mkv           ok, go-to not working, subtitle position too low
avi_xvid_480p_23.976_ac3_5.1_ext_srt_中_big5.avi      cannot be played
avi_xvid_480p_23.976_mp3_2.0.avi                      ok
avi_xvid_480p_23.976_mp3_2.0_ext_srt_chi_utf8.avi     playing, subtitiles disappear after 42.x min
avi_xvid_480p_25_mp3_2.0_ext_srt_ansi.avi             cannot be played
mkv_h264_1080p_23.976_40mbps.mkv                      ok                                                                                                  
mkv_h264_720p_23.976_ac3_5.1_ext_srt_utf8.mkv         ok                                                                                                  
mkv_h264_720p_23.976_dts_5.1_ext_srt_ansi.mkv         cannot be played                                                                                                    
mkv_h264_720p_23.976_dts_5.1_ext_srt_utf8.mkv         cannot be played                                                                                                   
mkv_h264_720p_23.976_dts_5.1_srt_eng.mkv              ok                                                                                                  
mkv_h264_720p_23.976_vfr_ac3_5.1_ssa_chi.mkv          ok if it plays, intermittently failure                                                                                                  
mkv_h264_720p_25_ac3_2.0_srt_eng.mkv                  ok                                                                                                  
mp4_avc1_720p_23.976_aac_2.0.mp4                      ok                                                                                                 
mpg_mpeg1_240p_29.97_mp2_2.0.mpg                      ok
rmvb_rv40_480p_23.976_aac_2.0.rmvb                    ok
```
It looks like I am having problem with non-UTF8 external SRT files, combination of DTS and external SRT files, etc.  There are some minor problem, like go-to not working, weird location for vobsub, and subtitles not working after some time (that same file plays ok with mencoder engine and local VLC.)  Since I didn't dig deep into each issues, the cause might be different than above.  We might have some work to do to for ffmpeg with subtitles.  

For the record, all files played correctly with mencoder.  

Also with the `FrameRate_Original` fix, all files are playing fine, with mencoder.  I don't see any problem.  Without it, the vfr file will have mismatch a/v along the time, which is expected.

I guess I'll stay with mencoder for now.

JeromeMartinez:
@ owlhuang 

> @JeromeMartinez , I could share with you offline. Please tell me how to do it since it is over 4GB.

Please contact me by email (see my profile) for private FTP server access.

> And from the result, it has to be the average frame rate, aka total duration divided by the number of frames. That's why we should use FrameRate_Original when container and video stream disagree on frame rate.

From MediaInfo point of view, I totally disagree: `FrameRate_Original` displays the frame rate from the raw stream, nothing else. so this is not "total duration divided by the number of frames". It works for you, but it may be worse for other people because their file has another problem (e.g. encoded awith bad frame rate metadata and user has "corrected" it with Matroska time stamps, with your patch you would use the bad metadata). The patch works for **your** files but may be worse for others.

> MediaInfo is using 1/duration from MKV container (from the link but Jerome disagrees)

I think the issue raised in the link is corrected on my side, but I may be wrong and I would like to have examples about when the current code in MediaInfo is still an issue.

> But MediaInfo does not have responsibility to change its information definition specially for UMS, IMO.

I am concern about a risk that MediaInfo is just buggy. If it is not buggy, we could find a method for displaying relevant information for UMS (not `FrameRate_Original` which has another purpose).

> We could add a check and apply the fix only when FrameRate_Mode = Variable. Do you think it reduce the risk and impact on general users?

It would greatly reduce the risk IMHO, as `FrameRate_Original` is usually used for container/stream mismatch only with CFR streams, and this is more acceptable as an hotfix and workaround waiting for more info or change from MediaInfo. Note that this is only a workaround in case MediaInfo is buggy, please keep sending file so we fix the source of the issue if it is in MediaInfo.

owlhuang:
Thanks, @JeromeMartinez.

I agree with your points of view.  Just want to speak up my opinion here.

I think the nature of the issue is, when container and video stream have conflicting information, which should we trust.  There seems to be no **good** way to make the decision.  In an ideal world, `FrameRate` and `FrameRate_Original` should be identical.  Container should have identical frame rate as video stream.  But for me, _[encoded with a bad frame rate metadata and user has "corrected" it with Matroska time stamps,]_  is an incorrect encoded file.  Now the question is, is there a "valid" file with correct frame rate in video steam, and an incorrect one as `FrameRate` from MediaInfo?  If the answer is negative, I totally agree we should not apply this patch.  Both cases are just corrupted files.  There is no reason to prefer one over another.  For the record, we already have 2 cases here with `FrameRate_Original` equals to what ffmpeg reports.  But yes, there is a possibility that both are bad files.

By the way, what is the version MediaInfo fix that bug?  I could double check to make sure I am using a working MediaInfo, not chasing an old bug.

Let's find out what is the devil behind this.  I'll send you the file tomorrow.

@Nadahar , @Sami32 , do you agree to add `if(FrameRateMode=Variable) `for this patch as a hotfix?  This should minimize the impact.  I like this idea but you guys are the people to make the final decision.

Nadahar:
@owlhuang I'd like to wait until @JeromeMartinez has analyzed your file to see if any changes will be made to MediaInfo before we decide how we should handle this.

owlhuang:
Guys, I think I got to the bottom of this issue.  In short, we could drop this fix. :)   For MediaInfo, I believe it is working as expected.  I'll add the detail below.  Thanks for all the helps here and appreciate your time.

In short, I guess I gave too much credit to mencoder and it doesn't really do things right.  Detail follows.

Please stay with me for the long detail if you are interested.  Before I send the file to debug, I decide to understand how MediaInfo works with more information.  The number of frames in that video is 231,720, and total video duration is 2:40:15.311, from ffmpeg.  That actually calculates to 24.099 frames/s, which is `FrameRate `reported by MediaInfo!  I guess @JeromeMartinez could be relieved that MediaInfo is not buggy here.  Yet it leaves a puzzle why mencoder works with `FrameRate_Original`.  So I analyze all the frames.  To my surprise, that 2hr40min VFR files actually contains only 2 different frame rates, 23.976 and 29.97.  The first 2hr37min video is encoded in 23.976, and a much smaller second half 3min video is in 29.97.  It becomes clear that mencoder works with `FrameRate_Original `is simply because most of the video is a constant frame rate stream at 23.976.  That's the frame rate ffmpeg reports and also what the mkvinfo reports.  My guess is that it is the frame rate embedded in the video stream meta data (for ffmpeg), and also the default duration in the container header (for mkvinfo.)   So technical speaking, this video file is valid even it is not a common VFR file, which I assume having changing frame rate through out the video.  But no, it doesn't work with average frame rate, either.

Now I am curious about how mencoder handles `-ofps` option.  So I did some experiments on the second half of the video file.  Here is the result.

image

As you can see, mencoder just output the video with the target frame rate, without dropping the frames.  For all frame rate slower than 29.97, we have slower video which are cut at 3:08.  For frame rates higher than 29.97, things become tricky.  I believe mencoder is doing it case by case according to the `-ofps` number.  For all cases, audio plays in correct speed.  I tried to play with mencoder options for IVTC, hoping that would correct the frame rate.  But no `-vf` options work for me.  There is no video as long as I remove `-noskip` options.

With all the background information, the solution becomes much more ambiguous.  There are some options we have, assuming all meta data are correct and valid,

Before that, assume we have 4 different types of VFR files,

**A**. VFR file with varying frame rate along the time, the variant is up and down, and not too far from the average frame rate.
**B**. VFR file with varying frame rate along the time, the variant is up and down extremely.
**C**. VFR file with multi-zone frame rate, with a dominant frame rate zone, like my file
**D**. VFR file with multi-zone frame rate, without any obvious dominant frame rate zone

1. Using (# of frames/total duration) in the container as the frame rate, aka `FrameRate `from MediaInfo.
A is somehow working but no other 3.  This is the current UMS behavior.

2. Using frame rate in the video stream, aka `FrameRate_Original` from MediaInfo.
We need the information how the video stream is encoded.  It is probably working for A and C but not the other two.

3. Using default duration in the MKV container.  It is the mostly used duration in the video.
A and C might work but no guaranteed.  B and D are not working.
@JeromeMartinez , do you think we could have a new field for MKV container?  Like `FrameRate_Default `from (1/default duration in MKV header)?  This might be helpful for others.  Just my own wish.  (or I could help on MediaInfo)

In fact, there is no perfect solution if mencoder outputs a CFR stream without changing the frame rate correctly.  In this case, things will break when the playing frame rate is different than the number mencoder uses.  So the other 2 options are

4. Find a way for mencoder to output a variable frame rate stream, or a way for mencoder to change the VFR stream into a correct CFR one
Maybe it is worth a try with avc/h.264 stream instead of mpeg2.  I might find a time for this

5. Ditch mencoder and go for ffmpeg
We need ffmpeg to output a variable frame rate stream, or convert to a correct CFR stream.  I guess this is the preferable way since UMS is going for ffmpeg for the future.

Whew, it is a long comment.  I hope this clarifies things here.  I, myself, learn a lot on this.  Please kindly comment and provide input if you have any.

Again, thanks for all your time here.  Let's make UMS better. :)

Sami32:
@owlhuang Thank you for your time and experimentation :+1: 

I'll not hide that for me, find fix for the _FFmpeg_ convertion to CFR is the way to go, as it's the most powerful and the default video engine as well.
As CFR was choosen for it's wide and better support among renderers, and the fact that the only target that i see for _MEncoder_ it's to get rid of it :wink: as soon as _FFmpeg_ will improve its playback of DVDs.

Sami32:
@SubJunk Do you have any objection to an _FFmpeg_ upgrade for the next version ?
Or do you still have that "20 mn" issue ?

Many popular formats support improvements, including text subtitles and the recent encoding  framerate (CFR) handling upgrade.
https://git.ffmpeg.org/gitweb/ffmpeg.git/commit/8db301deadfcf113fb274881e65afcbe3e1bd645
https://git.ffmpeg.org/gitweb/ffmpeg.git/commit/205b8fd078e50aa3f7f401646dbb73de1bb35a10

SubJunk:
@Sami32 I'll test the latest version soon to see if that bug is fixed. It's a pretty big one so if it is still there we can't update

JeromeMartinez:
> that 2hr40min VFR files actually contains only 2 different frame rates, 23.976 and 29.97.

Unfortunately it is pretty difficult to detect such case without parsing the whole file (and by default MediaInfo parses only a small part).

> @JeromeMartinez , do you think we could have a new field for MKV container? Like FrameRate_Defaultfrom (1/default duration in MKV header)? This might be helpful for others.

I could provide it, but not sure it would help a lot in your case. Please provide a sample file.

Sami32:
That doesn't seem to be an advised choice, IMO:
https://github.com/UniversalMediaServer/UniversalMediaServer/blob/86502e6a72005cf18055f8eb118f85d2d833a004/src/test/java/net/pms/dlna/LibMediaInfoParserTest.java#L97

Nadahar:
@Sami32 Sorry, I didn't understand what you meant not to be advised - can you please rephrase?

Sami32:
@Nadahar Sorry, i get an internet disconnection issue.
Simply considerate that CBR/VBR is equal to CBR for transcoding with video engines, is a little risky, IMHO.

I guess that @JeromeMartinez have a clear view on how `FrameRateMode` should be handled when such case arise.

JeromeMartinez:
On my side I am a bit lost, as VBR and CBR refer to bitrate mode, and here you talk about frame rate mode (VFR and CFR?), so I let you deal with it ;-).

Sami32:
@JeromeMartinez Sorry if i was not clear. I was talking about:
> FrameRate_Mode            : Frame rate mode (CFR, VFR)
> FrameRate_Mode/String     : Frame rate mode (Constant, Variable)

As we are also using it as parameter for video engines. And i don't think that considerate, as we do now, CFR/VFR as CFR is safe, but i guess that you know better than us how such framerate mode could be reliable ?

JeromeMartinez:
The link above compare "CBR" to frame rate, no sense ("B" in "CBR" means "Bit", not "Frame", comparison is with "CFR"), so I don't understand the comparison. No worries, it may work, just that I don't follow the comparison if it is from MediaInfo

Sami32:
@JeromeMartinez My bad, my mind was again on one other framerate mode bug from our side.

I was trying to say, considerate CBR/VBR as CBR is safe or not ?

JeromeMartinez:
> I was trying to say, considerate CBR/VBR as CBR is safe or not ?

I can not say what is safe or not, it depends of your own goal.
I describe a stream, that's all.
Usually, when there is "VBR / CBR" info in the `BitRate_Mode` field means that the stream is VBR, with a CBR substream (e.g. DTS-HD has a legacy DTS CBR stream and a VBR HD addition interleaved, you can have a CBR stream without re-encoding if you remove the HD addition, else the stream is VBR).
having "CBR / VBR" would be a bug (I don't have such file).

Sami32:
@JeromeMartinez Thank you for making the things clear :-)

SubJunk:
I finally got some time to look into the FFmpeg bug that was keeping us from updating and I can't reproduce it anymore so I will update us to the latest versions.

As for this pull request I will also look through my library and see if I can find any cases where this would be a bad idea. If I can't then we can merge

src/main/java/net/pms/dlna/LibMediaInfoParser.java
```diff
@@ -117,6 +118,12 @@ public synchronized static void parse(DLNAMediaInfo media, InputFile inputFile,
 							media.setAspectRatioContainer(MI.Get(video, i, "DisplayAspectRatio/String"));
 							media.setAspectRatioVideoTrack(MI.Get(video, i, "DisplayAspectRatio_Original/String"));
 							media.setFrameRate(getFPSValue(MI.Get(video, i, "FrameRate")));
+							// Prefer "FrameRate_Original over FrameRate if it's there because it is more accurate
+							// https://github.com/UniversalMediaServer/UniversalMediaServer/issues/1210
+							value = MI.Get(video, i, "FrameRate_Original");
+							if (StringUtils.isBlank(value))
```
SubJunk: Please add curly brackets :)

Nadahar:
@SubJunk It seems it's a bit more complicated. If you read the discussion above you'll see that this is probably worse "in average" than the current solution. If ```FrameRate_Original``` should be used at all, a set of rules for when that is preferrable should be found and applied.

owlhuang:
I agree with @Nadahar .  It would be better to find some rules before applying this patch.

In short, `FrameRate_Original` is from video stream, and `FrameRate` is from the container meta data.  `FrameRate_Original` only shows up when these two numbers different.  But it would be difficult to know which should be the one to use.

```
FrameRate_Original = frame rate from VIDEO STREAM
FrameRate = (# of frames)/(total video duration), both are from CONTAINER META)
```

If @SubJunk is kind to go through his library, and found it does help, I could help to find the rule from those video.

owlhuang:
@JeromeMartinez , the idea of adding `FrameRate_Default` is just a _nice to have_ feature.  It is more like to complete all the information MediaInfo could find.  I don't have a specific idea of how to use it.  I could provide the reference file if you want to add it into MediaInfo.

Sami32:
@SubJunk It'll nice to have your report, in the hope that you get some such videos in your huge library.
That's said, i'm not for merging it now:

@JeromeMartinez is [still waiting](https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1211#issuecomment-286981276) for @owlhuang video:

> @ owlhuang
> 
>    @JeromeMartinez , I could share with you offline. Please tell me how to do it since it is over 4GB.
> 
> Please contact me by email (see my profile) for private FTP server access.

We should wait for his analyse and report on that specific file, IMHO.

owlhuang:
> We should wait for his analyse and report on that specific file, IMHO.

I already analyzed the file and the detail is in the comment.  I will not send the file to @JeromeMartinez unless he wants to add FrameRate_Default or for other purpose.  The analyze explains why this patch works on my files.

> That's "trick" didn't worked in that case:
> #1209 

To be more precise, this patch doesn't help #1209 but it doesn't make it worse, either.  In fact, in #1209, the MediaInfo is missing `FrameRate `and this patch might help the MEncoder logic to get the frame rate from `FrameRate_Original `even it doesn't help to fix the issue.

Sami32:
@owlhuang A more general logging PR is already [done here](https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1215).

src/main/java/net/pms/dlna/LibMediaInfoParser.java
```diff
@@ -14,6 +14,7 @@
 import net.pms.util.ImagesUtil;
 import org.apache.commons.codec.binary.Base64;
 import org.apache.commons.imaging.ImageReadException;
+import org.apache.commons.lang3.StringUtils;
```
Sami32: @Nadahar Sorry, i don't understand why it need to have 2 import for the same class.
Are they all really necessary ?

Nadahar:
If someone want to make a more sophisticated solution, feel free to reopen.

PR #1213, opened 2017-03-05
Update Samsung-Mobile.conf

Sami32:
Add a string ID.
https://deviceatlas.com/blog/samsung-phones-user-agent-strings-list


SubJunk:
Why does this have the `not ready` label? Will it ever be ready?

PR #1215, opened 2017-03-07
Log video bit depth color and fix frameRateMode bug

Sami32:
- [x] Fix an old `frameRateMode` bug ([<2009](https://sourceforge.net/p/mediainfo/discussion/297610/thread/cd4f521f/?SetFreedomCookie)) that could have caused desynchronization issues.
- [x] Log `videoBitDepth` and `frameRateModeRaw`
For  a more easy debugging, as many desynchronization or transcoding issues are reported, and asking MediaInfo log from users could be hazardous and time consuming.
- [x] Refactoring the parsing logger; thanks to @Nadahar advices.

### DIFF
--- a/src/main/java/net/pms/dlna/DLNAMediaAudio.java
+++ b/src/main/java/net/pms/dlna/DLNAMediaAudio.java
@@ -509,40 +509,52 @@ public class DLNAMediaAudio extends DLNAMediaLang implements Cloneable {
 	@Override
 	public String toString() {
 		StringBuilder result = new StringBuilder();
-		result.append("id: ");
-		result.append(getId());
-		result.append(", lang: ");
-		result.append(getLang());
+		if (!getLang().equals("und")) {
+			result.append("Id: ").append(getId());
+			result.append(", Language Code: ").append(getLang());
+		}
 
 		if (isNotBlank(getAudioTrackTitleFromMetadata())) {
-			result.append(", audio track title from metadata: ");
-			result.append(getAudioTrackTitleFromMetadata());
+			if (result.length() > 0) {
+				result.append(", ");
+			}
+			result.append("Audio Track Title From Metadata: ").append(getAudioTrackTitleFromMetadata());
 		}
 
-		result.append(", audio codec: ");
-		result.append(getAudioCodec());
-		result.append(", sample frequency:");
-		result.append(getSampleFrequency());
-
-		if (getAudioProperties() != null && getAudioProperties().getNumberOfChannels() != 0) {
-			result.append(", number of channels: ");
-			result.append(getAudioProperties().getNumberOfChannels());
+		if (result.length() > 0) {
+			result.append(", ");
 		}
+		result.append("Audio Codec: ").append(getAudioCodec());
 
-		result.append(", bits per sample: ");
-		result.append(getBitsperSample());
+		result.append(", Bitrate: ").append(getBitRate());
+		if (getBitsperSample() != 16) {
+			result.append(", Bits per Sample: ").append(getBitsperSample());
+		}
+		if (getAudioProperties() != null) {
+			result.append(", ").append(getAudioProperties());
+		}
 
 		if (isNotBlank(getArtist())) {
-			result.append(", artist: ");
-			result.append(getArtist());
-			result.append(", album: ");
-			result.append(getAlbum());
-			result.append(", song name: ");
-			result.append(getSongname());
-			result.append(", year: ");
-			result.append(getYear());
-			result.append(", track: ");
-			result.append(getTrack());
+			result.append(", Artist: ").append(getArtist());
+		}
+		if (isNotBlank(getAlbum())) {
+			result.append(", Album: ").append(getAlbum());
+		}
+		if (isNotBlank(getSongname())) {
+			result.append(", Track Name: ").append(getSongname());
+		}
+		if (getYear() != 0) {
+			result.append(", Year: ").append(getYear());
+		}
+		if (getTrack() != 0) {
+			result.append(", Track: ").append(getTrack());
+		}
+		if (isNotBlank(getGenre())) {
+			result.append(", Genre: ").append(getGenre());
+		}
+
+		if (isNotBlank(getMuxingModeAudio())) {
+			result.append(", Muxing Mode: ").append(getMuxingModeAudio());
 		}
 
 		return result.toString();
--- a/src/main/java/net/pms/dlna/DLNAMediaInfo.java
+++ b/src/main/java/net/pms/dlna/DLNAMediaInfo.java
@@ -128,6 +128,12 @@ public class DLNAMediaInfo implements Cloneable {
 	private String frameRateMode;
 
 	/**
+	 * The frame rate mode as read from the parser
+	 */
+	private String frameRateModeRaw;
+	private String frameRateOriginal;
+
+	/**
 	 * @deprecated Use standard getter and setter to access this variable.
 	 */
 	@Deprecated
@@ -1662,81 +1668,100 @@ public class DLNAMediaInfo implements Cloneable {
 	@Override
 	public String toString() {
 		StringBuilder result = new StringBuilder();
-		result.append("container: ");
-		result.append(container);
-		result.append(", bitrate: ");
-		result.append(bitrate);
-		result.append(", size: ");
-		result.append(size);
-		if (videoTrackCount > 0) {
-			result.append(", video tracks: ");
-			result.append(videoTrackCount);
-		}
-		if (getAudioTrackCount() > 0) {
-			result.append(", audio tracks: ");
-			result.append(getAudioTrackCount());
-		}
-		if (imageCount > 0) {
-			result.append(", images: ");
-			result.append(imageCount);
-		}
-		if (getSubTrackCount() > 0) {
-			result.append(", subtitle tracks: ");
-			result.append(getSubTrackCount());
-		}
-		result.append(", video codec: ");
-		result.append(codecV);
-		result.append(", duration: ");
-		result.append(getDurationString());
-		result.append(", width: ");
-		result.append(width);
-		result.append(", height: ");
-		result.append(height);
-		result.append(", frame rate: ");
-		result.append(frameRate);
+		result.append("Container: ").append(getContainer().toUpperCase(Locale.ROOT));
+		result.append(", Size: ").append(getSize());
+		if (isVideo()) {
+			result.append(", Video Bitrate: ").append(getBitrate());
+			result.append(", Video Tracks: ").append(getVideoTrackCount());
+			result.append(", Video Codec: ").append(getCodecV());
+			result.append(", Duration: ").append(getDurationString());
+			result.append(", Video Resolution: ").append(getWidth()).append(" x ").append(getHeight());
+			if (isNotBlank(getFrameRate())) {
+				result.append(", Frame Rate: ").append(getFrameRate());
+			}
+			if (isNotBlank(getFrameRateOriginal())) {
+				result.append(", Original Frame Rate: ").append(getFrameRateOriginal());
+			}
+			if (isNotBlank(getFrameRateMode())) {
+				result.append(", Frame Rate Mode: ");
+				result.append(getFrameRateModeRaw());
+				if (isNotBlank(getFrameRateModeRaw())) {
+					result.append(" (").append(getFrameRateModeRaw()).append(")");
+				}
+			} else if (isNotBlank(getFrameRateModeRaw())) {
+				result.append(", Frame Rate Mode Raw: ");
+				result.append(getFrameRateModeRaw());
+			}
+			if (isNotBlank(getMuxingMode())) {
+				result.append(", Muxing Mode: ").append(getMuxingMode());
+			}
+			if (isNotBlank(getMatrixCoefficients())) {
+				result.append(", Matrix Coefficients: ").append(getMatrixCoefficients());
+			}
+			if (isNotBlank(avcLevel)) {
+				result.append(", AVC Level: ").append(getAvcLevel());
+			}
+//			if (isNotBlank(getHevcLevel())) {
+//				result.append(", HEVC Level: ");
+//				result.append(getHevcLevel());
+			if (getVideoBitDepth() != 8) {
+				result.append(", Video Bit Depth: ").append(getVideoBitDepth());
+			}
+			if (isNotBlank(getFileTitleFromMetadata())) {
+				result.append(", File Title from Metadata: ").append(getFileTitleFromMetadata());
+			}
+			if (isNotBlank(getVideoTrackTitleFromMetadata())) {
+				result.append(", Video Track Title from Metadata: ").append(getVideoTrackTitleFromMetadata());
+			}
 
-		if (thumb != null) {
-			result.append(", thumb size: ");
-			result.append(thumb.length);
-		}
-		if (isNotBlank(muxingMode)) {
-			result.append(", muxing mode: ");
-			result.append(muxingMode);
-		}
+			if (getAudioTrackCount() > 0) {
+				appendAudioTracks(result);
+			}
 
-		result.append(", mime type: ");
-		result.append(mimeType);
+			if (hasSubtitles()) {
+				appendSubtitleTracks(result);
+			}
 
-		if (isNotBlank(matrixCoefficients)) {
-			result.append(", matrix coefficients: ");
-			result.append(matrixCoefficients);
+		} else if (getAudioTrackCount() > 0) {
+			result.append(", Bitrate: ").append(getBitrate());
+			result.append(", Duration: ").append(getDurationString());
+			appendAudioTracks(result);
 		}
-
-		if (isNotBlank(avcLevel)) {
-			result.append(", avc level: ");
-			result.append(avcLevel);
-		}
-
-		if (isNotBlank(fileTitleFromMetadata)) {
-			result.append(", file title from metadata: ");
-			result.append(fileTitleFromMetadata);
-		}
-		if (isNotBlank(videoTrackTitleFromMetadata)) {
-			result.append(", video track title from metadata: ");
-			result.append(videoTrackTitleFromMetadata);
-		}
-
-		for (DLNAMediaAudio audio : audioTracks) {
-			result.append("\n\tAudio track ");
-			result.append(audio.toString());
+		if (getImageCount() > 0) {
+			if (getImageCount() > 1) {
+				result.append(", Images: ").append(getImageCount());
+			}
+			result.append(", Image Width: ").append(getWidth());
+			result.append(", Image Height: ").append(getHeight());
 		}
 
-		for (DLNAMediaSubtitle sub : subtitleTracks) {
-			result.append("\n\tSubtitle track ");
-			result.append(sub.toString());
+		if (getThumb() != null) {
+			result.append(", Thumb Size: ").append(getThumb().length);
 		}
+
+		result.append(", Mime Type: ").append(getMimeType());
 
 		return result.toString();
+	}
+
+	public void appendAudioTracks(StringBuilder sb) {
+		sb.append(", Audio Tracks: ").append(getAudioTrackCount());
+		for (DLNAMediaAudio audio : audioTracks) {
+			if (!audio.equals(audioTracks.get(0))) {
+				sb.append(",");
+			}
+			sb.append(" [").append(audio).append("]");
+		}
+	}
+
+	public void appendSubtitleTracks(StringBuilder sb) {
+		sb.append(", Subtitle Tracks: ").append(getSubTrackCount());
+		for (DLNAMediaSubtitle subtitleTrack : subtitleTracks) {
+			if (!subtitleTrack.equals(subtitleTracks.get(0))) {
+				sb.append(",");
+			}
+			sb.append(" [").append(subtitleTrack).append("]");
+		}
 	}
 
 	public InputStream getThumbnailInputStream() {
@@ -2071,6 +2096,20 @@ public class DLNAMediaInfo implements Cloneable {
 	}
 
 	/**
+	 * @return the frameRateOriginal
+	 */
+	public String getFrameRateOriginal() {
+		return frameRateOriginal;
+	}
+
+	/**
+	 * @param frameRateOriginal the frameRateOriginal to set
+	 */
+	public void setFrameRateOriginal(String frameRateOriginal) {
+		this.frameRateOriginal = frameRateOriginal;
+	}
+
+	/**
 	 * @return the frameRateMode
 	 * @since 1.55.0
 	 */
@@ -2084,6 +2123,20 @@ public class DLNAMediaInfo implements Cloneable {
 	 */
 	public void setFrameRateMode(String frameRateMode) {
 		this.frameRateMode = frameRateMode;
+	}
+
+	/**
+	 * @return The unaltered frame rate mode
+	 */
+	public String getFrameRateModeRaw() {
+		return frameRateModeRaw;
+	}
+
+	/**
+	 * @param frameRateModeRaw the unaltered frame rate mode to set
+	 */
+	public void setFrameRateModeRaw(String frameRateModeRaw) {
+		this.frameRateModeRaw = frameRateModeRaw;
 	}
 
 	/**
--- a/src/main/java/net/pms/formats/v2/AudioProperties.java
+++ b/src/main/java/net/pms/formats/v2/AudioProperties.java
@@ -219,4 +219,16 @@ public class AudioProperties {
 			return result;
 		}
 	}
+
+	@Override
+	public String toString() {
+		StringBuilder result = new StringBuilder();
+		result.append("Channel(s): ").append(getNumberOfChannels());
+		result.append(", Sample Frequency: ").append(getSampleFrequency()).append(" Hz");
+		if (getAudioDelay() != 0) {
+			result.append(", Delay: ").append(getAudioDelay());
+		}
+
+		return result.toString();
+	}
 }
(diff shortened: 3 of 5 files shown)


Nadahar:
@Sami32 I can't see that there's a variable named ```BitDepth```. If it were, it would be against convention (it should be called ```bitDepth```.

Nadahar:
@Sami32 It should also be moved within the ```if (videoTrackCount > 0)``` block as it only makes sense for video. The text should also be ```video bit depth```. It should also be added that this can be misleading, as it's defaulted to ```8``` - so ```8``` will be reported if nothing else is parsed.

Sami32:
@Nadahar Thanks for your review ;)

> It should also be moved within the `if (videoTrackCount > 0)` block as it only makes sense for video. The text should also be video bit depth.

Though i agree with you, i didn't do as i feel that many other values will need as well, so i need more feedback on that matter.
```
result.append(", frame rate: ");
result.append(frameRate);
```
Do you agree if i move that one as well ?

Nadahar:
@Sami32 I'd need to look at the code to see how it's used, but it sounds reasonable that ```frameRate``` is only relevant for video content. 

You should use ```getVideoBitDepth()``` instead of calling the variable directly. The reason for this has to do with design and robustness and isn't really important in this case, but why not do it right when you do it? This is more important when accessing other classes, but it can be useful even internally if you for example want to apply some logic to the returned value. Using the getter will allow code in the getter to do just this, but any calls "skipping" the getter won't get the benefit, so the code is simply more maintainable when using the getters.

Nadahar:
@Sami32 This is improving a lot 👍 But, you might want to use ```isVideo()```, ```isAudio()``` and ```isImage()``` instead of using the track count. The reason for this is that some resources have multiple types of tracks, most videos for example also has audio tracks - but the information stored will be for the video. Audio can have images as well (embedded cover art), but again e.g width and height won't represent that image. 

The logic in the methods above tries to determine what type of media file it actually is, and it probably makes more sense to produce the log output according to that.

src/main/java/net/pms/dlna/DLNAMediaInfo.java
```diff
@@ -1649,72 +1649,72 @@ public String toString() {
 		result.append(bitrate);
 		result.append(", size: ");
 		result.append(size);
-		if (videoTrackCount > 0) {
+		if (isVideo) {
```
Nadahar: @Sami32 You need to use ```isVideo()``` as this is a method.

Nadahar:
@Sami32 It might be best to list audio tracks if there are any under ```isVideo()``` and images under ```isAudio()``` as well.

Sami32:
@Nadahar I agree, it was on my TODO list, but i don't know yet how to proceed for Audio track so i'll check at that later.

src/main/java/net/pms/dlna/DLNAMediaInfo.java
```diff
@@ -1661,9 +1661,14 @@ public String toString() {
 			result.append(", height: ");
 			result.append(getHeight());
 			result.append(", frame rate: ");
+			if (isNotBlank(frameRate)) {
 			result.append(getFrameRate());
+			} else { result.append("variable"):
+			}
```
Nadahar: This should be formatted like this:
```java
		} else {
			result.append("variable"):
		}
		if (videoBitDepth != 8) {
```

src/main/java/net/pms/dlna/DLNAMediaInfo.java
```diff
@@ -1661,9 +1661,14 @@ public String toString() {
...
+			if (videoBitDepth != 8) {
 			result.append(", video bit depth: ");
 			result.append(getVideoBitDepth());
```
Nadahar: These two lines should be indented.

src/main/java/net/pms/dlna/DLNAMediaInfo.java
Sami32: I was sure you'll say that ;)

About `(videoBitDepth != 8)`i'm not sure if that the best way to code it, though many in UMS code; Should the use of `equals()` method be preferred ?

Nadahar: @Sami32 No, videoBitDepth is an ```int```. Jav differentiates between "primary data types" and "classes". Class-based types use ```equals()```, ```compareTo()``` etc while primary data types use operators like ```==```, ```!=``` etc. Trying to use ```equals()``` there would give a compile error.

Sami32:
@Nadahar I leave it to your about your images change request, as i'm not sure what do you want and that anyway you will be better than me at doing that.
IYou can commit in this PR or do an other PR later, i really don't mind.

Nadahar:
@Sami32 Please inform when you're finished with this so that it can be reviewed.

Sami32:
@Nadahar Thanks you :-)

For now i'm stuck with this:
https://github.com/UniversalMediaServer/UniversalMediaServer/blob/d4bf5d47307919d577d7206db9bfa8ff34e2a6e5/src/main/java/net/pms/dlna/DLNAMediaAudio.java#L513

How can i correctly implement something like `isVideo()` ? Any idea ?

Nadahar:
@Sami32 I don't understand what you're trying to do. ```DLNAMediaAudio``` represents an audio track, so ```isVideo()``` isn't relevant. If your goal is to only print language and id for audio tracks belonging to video files, that must be handled in the caller (the loop that calls ```toString()```).

But, I don't know why id shouldn't always be printed..? And language could simply be printed if it has a value, e.g. ```if (!StringUtils.isBlank(getLang())) {```

Sami32:
@Nadahar Thanks for helping :+1: 

If you care to see the id of your MP3, i can leave it for you, but personnally i don't see any use of it, and i need place for useful debugging informations for transcoding/desynchronization issues.

Nadahar:
@Sami32 I have made some suggestions for you, but I can't push them while you keep overwriting the branch. Can you stop that for a little bit, so I can push them?

Nadahar:
@Sami32 Why don't you want to print thumbnail size for images? It's useful to see if there's a thumbnail there as well.

Sami32:
@Nadahar OK ;)

Sure, i did at first, and after i changed my mind. I'll get it back for the images.

Nadahar:
Ok, I'll let you know as soon I've pushed.

Nadahar:
I've pushed them now - I'm not saying you have to use all of these changes - it's mere suggestions and ideas, and some slightly different ways to solve some things.

Sami32:
@Nadahar Good job :+1: 

Thank you :-)

Sami32:
@Nadahar LGTM
I think we are done with that PR ?

src/main/java/net/pms/dlna/DLNAMediaAudio.java
```diff
@@ -509,40 +510,63 @@ public String getAudioCodec() {
...
+			}
+			if (getTrack() != 0) {
+				result.append(", Track: ").append(getTrack());
+			}
+			if (StringUtils.isNotBlank(getGenre())) {
+				result.append(", Genre: ").append(getGenre());
+			}
+		}
+
+		if (StringUtils.isNotBlank(getMuxingModeAudio())) {
```
Sami32: @Nadahar What is the difference between `StringUtils.isNotBlank` and `isNotBlank` ?

Nadahar: @Sami32 Nothing. ```isNotBlank()``` isn't a ```DLNAMediaAudio``` (or parents) method, so for that to be a valid method, you need to import it staticly like this:
```
import static org.apache.commons.lang3.StringUtils.isNotBlank;
```
This is done in this case, but I never use that myself so I didn't think of it. The only reason to do this import is so that you can write ```isNotBlank()``` instead of ```StringUtils.isNotBlank()```, so I simply usually qualify them with ```<class>.<method>``` instead of doing static imports. 

If you use methods from a particular class a LOT in a given code, I guess static import makes sense. In short, in this case, the difference is none 😉

Nadahar:
I can do a review when you are completely finished.

Sami32:
@Nadahar So let's say that i don't plan to modify anything in this PR anymore.

Please, feel free to review it when you get time.

src/main/java/net/pms/dlna/DLNAMediaAudio.java
```diff
@@ -509,40 +509,63 @@ public String getAudioCodec() {
 	@Override
 	public String toString() {
 		StringBuilder result = new StringBuilder();
-		result.append("id: ");
-		result.append(getId());
-		result.append(", lang: ");
-		result.append(getLang());
+		if (!getLang().equals("und")) {
+			result.append("Id: ").append(getId());
```
Nadahar: @Sami32 Are you sure that checking if language is "und" is better than checking if id != 0? As i see it, "und" is also a setting - ```toString()``` is supposed to reveal everything relevant about the given instance. If id == 0 it means that no language is set - while "und" is also as setting as I see it. 

I also see that you removed the language name - not a bit problem for me, but some of the codes arn't that obvious (like ```iw``` for hebrew for example).

src/main/java/net/pms/dlna/DLNAMediaAudio.java
```diff
@@ -509,40 +509,63 @@ public String getAudioCodec() {
...
+			result.append(", ");
+		}
+		result.append("Audio Codec: ").append(getAudioCodec());
+		if (getSampleFrequency() != null) {
+			result.append(", Sample Frequency: ").append(getSampleFrequency());
+		}
+		if (getBitRate() != 0) {
+			result.append(", Bitrate: ").append(getBitRate());
+		}
+		if (getBitsperSample() != 0 && getBitsperSample() != 16 ) {
```
Nadahar: We shouldn't exclude both 0 and 16 from the result. This field initialized to 0, so only 0 should be excluded - if it's set to 16 it has been set to that, even if you consider that "standard".

src/main/java/net/pms/dlna/DLNAMediaAudio.java
```diff
@@ -509,40 +509,63 @@ public String getAudioCodec() {
...
+			result.append(", Bits per Sample: ").append(getBitsperSample());
 		}
 
-		result.append(", bits per sample: ");
-		result.append(getBitsperSample());
+		if (getAudioProperties() != null) {
+			result.append(", ").append(getAudioProperties());
```
Nadahar: @Sami32 Why did you remove the ```[ ]``` around the audio properties? I think that is a good way to indicate that this isn't a print of this instance, but of a instance helt by this instance.

src/main/java/net/pms/dlna/DLNAMediaInfo.java
```diff
@@ -1643,81 +1649,112 @@ public boolean isLossless(String codecA) {
...
-
-		if (isNotBlank(avcLevel)) {
-			result.append(", avc level: ");
-			result.append(avcLevel);
+		if (getImageCount() > 0) {
+			if (getImageCount() > 1) {
+				result.append(", Images: ").append(getImageCount());
+			}
+			result.append(", Image Width: ").append(getWidth());
+			if (getHeight() != 0) {
```
Nadahar: I disagree with this - DNL is extremely rare, and I've implemented it in Metadata-extractor so that in a future version it will still be parsed with height. I'd say that it would be very useful to get 0 as height, as that's very likely to be a bug.

src/main/java/net/pms/formats/v2/AudioProperties.java
```diff
@@ -219,4 +219,15 @@ public static int getSampleFrequencyFromLibMediaInfo(String mediaInfoValue) {
...
 	}
+
+	@Override
+	public String toString() {
+		StringBuilder result = new StringBuilder();
+		result.append("Channel(s): ").append(getNumberOfChannels());
+		if (getAudioDelay() != 0) {
+			result.append(", Delay: ").append(getAudioDelay());
+		}
+
```
Nadahar: @Sami32 You removed ```result.append(", Sample Frequency: ").append(getSampleFrequency()).append(" Hz");```` - I disagree. Remember that even though you already have the sample frequency from ```DLNAMediaAudio``` they are separate values and one can be used one place and another another place. We should see both values when debugging, if they are different that would be a bug.

src/main/java/net/pms/dlna/DLNAMediaAudio.java
Sami32: @Nadahar Yes, believe me, i well tested it, and i see that you commit withdrawed all the language code with id=0 from the log, and believe me it was many testing files.
"und" is only set by our code, when no language is found in the audio file, so i didn't want factice informations for debugging...

About the extra language code meaning, i don't need it for debugging purpose, the id in the file is suffisant to check in the other part of the log, IMHO.

src/main/java/net/pms/dlna/DLNAMediaAudio.java
Sami32: @Nadahar I just followed your own logic !

As your didn't want `videoBitDepth` to be displayed because it was set by default to 8.
`BitperSample` is be set by default to 16.

Did i missed something ? or your logic have changed since ?

src/main/java/net/pms/dlna/DLNAMediaAudio.java
Sami32: @Nadahar Sorry, not my taste at all...no easy readable for me.

src/main/java/net/pms/dlna/DLNAMediaInfo.java
Sami32: @Nadahar No problem, i'll leave as before.

src/main/java/net/pms/dlna/DLNAMediaAudio.java
Nadahar: This is the "standard Java" way of indicating that it's an "inner instance". Some of the fields are duplicates of those found on ```DLNAMediaAudio``` which increases the need to separate which is which.

src/main/java/net/pms/dlna/DLNAMediaInfo.java
Sami32: Done

src/main/java/net/pms/dlna/DLNAMediaAudio.java
Sami32: Only if i agree with your last comment.

src/main/java/net/pms/dlna/DLNAMediaAudio.java
Nadahar: @Sami32 I see when I check the code that the field isn't used quite the way I thought. It's quite a bad design as far as I can see, so I'm sure there are places in the code that "forgets" to set the id - so I agree.

src/main/java/net/pms/formats/v2/AudioProperties.java
Sami32: @Nadahar I will check that.

As thought that i asked you about that and didn't get answer, but maybe i didn't really send my PR annotation, perhaps they has been only pending and deleted later by me...

src/main/java/net/pms/dlna/DLNAMediaAudio.java
Nadahar: @Sami32 My logic is the same - ```videoBitDepth``` is initialized to 8 (it will always be 8 if not set). ```bitsperSample``` is initialized to 0, so 0 should be excluded.

Sami32: @Nadahar Where did you saw that ?
```
public DLNAMediaAudio() {
		setBitsperSample(16);
}
```

src/main/java/net/pms/formats/v2/AudioProperties.java
Nadahar: @Sami32 The thing is that someone did a half-ass implementation of ```AudioProperties```. Honestly I don't quite understand the need to have them as their own class, but if you move them they should be *moved* not *duplicated*. Whoever did it was probably afraid to break something or concerned with backwards compatibility for plugins - in any case, the situation now is a mess - as some properties are found two places and we have no control over which version is used by code. My guess is that some code use one and other another, which is a "bug-waiting-to-happen". But, I don't want to clean up such issues, as it always ends up in a huge argument about not changing ancient code that some plugin might rely on. As such, we should log both so we can at least see when they are different.

src/main/java/net/pms/dlna/DLNAMediaAudio.java
Nadahar: @Sami32 Sorry, I checked for a normal initialization - I didn't dream of that someone might initilalize this in the constructor. You are right then. That probably means we should exclude 16 but not 0.

Sami32: @Nadahar I added 0 later because of an MediaInfo bug with some ASF, WMA and WMV files.
https://github.com/MediaArea/MediaInfo/issues/139

src/main/java/net/pms/dlna/DLNAMediaAudio.java
Nadahar: @Sami32 I didn't understand what you meant with that...

src/main/java/net/pms/dlna/DLNAMediaAudio.java
Nadahar: @Sami32 Remember that these are used for debug printing. You shouldn't exclude a value because there's a bug in MediaInfo - that's exactly the kind of information it's useful to see.

src/main/java/net/pms/dlna/DLNAMediaAudio.java
Sami32: @Nadahar Just that if the sample frequencies are really the same, as i suspect, i will not agree change it, as it's the only possible information that can be double logged.
Of course, if they can be different, i'll change as you suggested.

src/main/java/net/pms/dlna/DLNAMediaAudio.java
Sami32: @Nadahar Yeaah, i sticked on it for some days trying to found a way to get the correct informations ! It was not as i expected as well ;-)

src/main/java/net/pms/dlna/DLNAMediaAudio.java
Sami32: @Nadahar I guess that we don't share debugging view.

As they are MediaInfo issues, i don't really see the use of loosing space, or overcharging the log with wrong informations, when the only fact that they are not displayed in the log just mean exactly the same thing without overcharging the log.

That's said it's only my point of view

src/main/java/net/pms/formats/v2/AudioProperties.java
Sami32: @Nadahar As far i understood the code, the sample frequency is only a duplicated information, so no need to put it in our parsing.

I guess you're right for the explaination, and i don't find it was a good way to do that, whatever was they purpose...

Though, if you really want an other sample frequency log information, i can implemented `SamplingRate/String` and all your wishes will be exhausted.

But we need to agree on that before i do.

Sami32: Or we can also not log them at all, and just duplicate the audio "time delay" information ?

src/main/java/net/pms/dlna/DLNAMediaAudio.java
Nadahar: @Sami32 I don't think you quite understand ```toString()```. It's not only used in the log that you think of, it's used automatically to "print" a class instance. It's a standard Java method that's used to generate a "string representation of an object". It will be used under many different circumstances, for example when using a debugger. The person logging at this output might be looking for something completely different than what you're usually looking for. 

If something is due to a bug in MediaInfo is irrelevant, this is the primary way to view an instance as a string, which means that you need to get all the relevant information out. How are we for example supposed to know that MediaInfo (or something else, a bug in our code) sets this to 0 when 0 is hidden? It could be the reason why something doesn't work, and hiding it might make that much more difficult to detect. Ideally I don't like hiding default values either, but because these classes are "over crowded" with information leaving out default/blank/null values makes some sense.

src/main/java/net/pms/dlna/DLNAMediaAudio.java
Nadahar: Again, this comes down to how ```toString()``` works and what it's supposed to do. You only thinking of one specific situation where you are looking for certain information in the log. That's not what it's for, it is also used in other circumstances like while debugging. ```DLNAMediaAudio``` and ```AudioProperties``` need to be able to represent themselves also, they are not always "used together". You must think of ```AudioProperites``` as an object instance, and ```DLNAMediaAudio``` as another. One ```AudioProperties``` instance can be "linked" to ```DLNAMediaAudio```, in which case we choose to print the content of ```AudioProperties``` together with ```DLNAMediaAudio``` - but ```AudioProperites``` still needs to be able to be presented alone. With the same logic, hiding information in ```DLNAMediaInfo``` because this information also exists in ```AudioProperties``` is wrong - the information is still existing in ```DLNAMediaInfo``` and although they might be the same most of the time (or always), you're not actually printing the same information. They can be different, and if they are we probably have a problem.

src/main/java/net/pms/formats/v2/AudioProperties.java
Nadahar: @Sami32 The same as above, stop thinking about them only from the point of view of the log, ```toString()``` need to generate a string representation of the object. As long as this information is stored double, it should be logged double.

Nadahar:
@Sami32 ??

Nadahar:
@Sami32 Why did you close this?

Sami32:
@nadahar i just resign. I'm not interested anymore.

Nadahar:
@Sami32 Because I don't agree with you on every point?

Sami32:
@Nadahar Because i cannot attempt to your quality standard, so i better use my time for for useful things.

src/main/java/net/pms/dlna/DLNAMediaAudio.java
```diff
@@ -525,22 +525,14 @@ public String toString() {
...
-		}
-		if (getBitsperSample() != 0 && getBitsperSample() != 16 ) {
+		result.append(", Sample Frequency: ").append(getSampleFrequency());
+		result.append(", Bitrate: ").append(getBitRate());
+		if (getBitsperSample() != 16 ) {
 			result.append(", Bits per Sample: ").append(getBitsperSample());
 		}
 
-		if (getAudioProperties() != null) {
-			result.append(", ").append(getAudioProperties());
```
Sami32: @Nadahar I cannot remember if it was my null or 0 condition  or your ?

Nadahar: @Sami32 I'm not sure what you're asking... "null" is correct for checking if there are any ```AudioProperties```.

Sami32: @Nadahar Who write that `if (getAudioProperties() != null) {`, you or me ?
Because i deleted it.

Nadahar: @Sami32 I don't remember, probably me.

Sami32: @Nadahar Since i was the one that make many of them, and you complained, i just deleted them.
Now i really don't bother, so do you want this `null` and the next `0` condition test or not ?

Nadahar: @Sami32 I'm sorry but I don't understand what you're asking. If you're going to print ```AudioProperties``` you need to check that they aren't ```null``` (or face a potential NPE). If you don't want to print them, then just skip the test.

Nadahar:
@Sami32 If you just remove lines 528, 536, 537, 538, 539 (those that are commented out) from ```DLNAMediaAudio``` I'll approve this.

Nadahar:
@Sami32 Let me know when you're done..

Sami32:
Thank you very much, i guess that costed you an harm, as we say here :wink: 

I love it like that, but i wish display the container in upper case:
`append(getContainer()).toUpperCase()`
Do you any objection to it ?

Nadahar:
@Sami32 No, just use ```.toUpperCase(Locale.ROOT)``` to make it locale independent.

Sami32:
Thanks, i didn't know/think about that :+1:

Sami32:
But it give me an compilation error when i set the local language as you suggested. Any idea ?

Nadahar:
Yes, use Eclipse 😛 

You probably have to import Locale, but I never do it manually as Eclipse just adds imports as needed.

Sami32:
:stuck_out_tongue: 
I thought that  `import java.util.*` would have done the job...but it appear that i was wrong.

Nadahar:
Don't do that in any case, you don't want to include more code than you have to.

Sami32:
You misunderstood me, it's already in the code.

And whatever nothing work...so forget about locale.

Nadahar:
You've placed ```.toUpperCase(Locale.ROOT)``` wrong. Move it one parenthes to the left...

Sami32:
Thanks, i missed it :-(

Sami32:
It's done for me.

Nadahar:
@Sami32 I squashed my "Suggested changes", it was never meant to be a separate commit - I just made it separate to not overwrite your changes.

I found one small thing now too that I suspect didn't work like you intended to. It should also be squashed before merge, I just wanted to show you the change. Basicly, album, song name, year and track would only be shown is artist wasn't blank. I doubt that's what you wanted, so moved them out of the "if artist" block.

Sami32:
Thanks :)
Do you think that the video / audio codecs should be in upper case as well ?
Some of the audio already are, so they are not very coherent.

Nadahar:
@Sami32 You shouldn't ask me that, I don't think the case should be changed in ```toString()``` as it's supposed to show what's actually stored in the instance (as you already know). By changing the case in the debug output, you're really masking the fact that the case is inconsistent. So, generally my answer would be no 😉, but do as you want.

Nadahar:
@Sami32 The problem is that the ```toString()``` method is "special" in Java, it's what is automatically called whenever an instance should be made into a string (text). That's why any kind of debugging, errors etc will have this output too.

Nadahar:
I squashed the "fix" commit.

PR #1217, opened 2017-03-08
Fix for the VieraST60.conf

Sami32:
From the log of this user that have a ST60 not correctly recognized:
http://www.universalmediaserver.com/forum/download/file.php?id=3743

### DIFF
--- a/src/main/external-resources/renderers/Panasonic-VieraST60.conf
+++ b/src/main/external-resources/renderers/Panasonic-VieraST60.conf
@@ -12,11 +12,18 @@ RendererIcon = panasonic.png
 # User-Agent: Panasonic MIL DLNA CP UPnP/1.0 DLNADOC/1.50
 # X-PANASONIC-Registration: VklFUkEgU1Q2MCBTZXJpZXMAAAAAAAAAAAA=
 # friendlyName=VIERA ST60 Series
+#
+# X-PANASONIC-DMP-Profile=""
+# X-PANASONIC-Registration: VklFUkEgU1Q2MAAAAAAAAAAAAAAAAAAAAAA=
+# friendlyName=VIERA ST60
+# udn=uuid:4D454930-0100-1000-8001-8CC1217131A6
+# manufacturer=Panasonic
+# modelName=Panasonic VIErA
 # modelNumber=ST60
 # ============================================================================
 #
 
-UpnpDetailsSearch = VIERA ST60 Series
+UpnpDetailsSearch = VIERA ST60
 LoadingPriority = 2
 
 SeekByTime = exclusive


src/main/external-resources/renderers/Panasonic-VieraST60.conf
```diff
@@ -12,11 +12,18 @@ RendererIcon = panasonic.png
 # User-Agent: Panasonic MIL DLNA CP UPnP/1.0 DLNADOC/1.50
 # X-PANASONIC-Registration: VklFUkEgU1Q2MCBTZXJpZXMAAAAAAAAAAAA=
 # friendlyName=VIERA ST60 Series
+#
+# # X-PANASONIC-DMP-Profile=""
+# X-PANASONIC-Registration: VklFUkEgU1Q2MAAAAAAAAAAAAAAAAAAAAAA=
```
Nadahar: Isn't this an unique code per unit and irrelevant in this context?

src/main/external-resources/renderers/Panasonic-VieraST60.conf
```diff
@@ -12,11 +12,18 @@ RendererIcon = panasonic.png
 # User-Agent: Panasonic MIL DLNA CP UPnP/1.0 DLNADOC/1.50
 # X-PANASONIC-Registration: VklFUkEgU1Q2MCBTZXJpZXMAAAAAAAAAAAA=
 # friendlyName=VIERA ST60 Series
+#
+# # X-PANASONIC-DMP-Profile=""
```
Nadahar: Why the double comment symbol, and why include an empty profile?

src/main/external-resources/renderers/Panasonic-VieraST60.conf
Sami32: It's not a unique code.
Every informations sended, as `X-PANASONIC-DMP-Profile=""`, could be pertinent for later differentiation use, so writing it in the renderer is safe. Not every one can remember everything and running after log.

src/main/external-resources/renderers/Panasonic-VieraST60.conf
Sami32: Copy paste typo...

src/main/external-resources/renderers/Panasonic-VieraST60.conf
Nadahar: @Sami32 Sorry, I don't understand what you mean by "later differentiation use". I never thought it was unsafe, I simply wondered if it was useful.

Sami32: Yes, it can be useful.

Sami32:
@SubJunk Thanks for your approval :)

PR #1180, opened 2017-02-07
Fix AAC

Sami32:
Not appropriate anymore. See https://trac.ffmpeg.org/wiki/Encode/AAC

PR #1220, opened 2017-03-11
Fix a logger bug

Sami32:


Sami32:
Thanks.

Like me, as i rememberer report this already and that someone fixed it...but with my bird's memory ;)
That's said, it's the same for [that one](https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1180), as it was fixed before, so i guess that a partial revert did the trick.

PR #1224, opened 2017-03-12
Findbugs and Checkstyle

Sami32:
**Not to be merged !**

For testing purpose only


Nadahar:
@Sami32 FindBugs is already active, the problem is simply that nobody use the report. You have to run ```mvn site``` to generate the reports.

PR #1226, opened 2017-03-15
Manage sleep mode

Nadahar:
I really didn't want to do this as I've got too much unfinished business elsewhere. However, it seemed easier to just throw an implementation together than to keep explaining what I mean.

This is just a draft, the following needs to be done:
- [ ] Verify that the start and stop calls are done in all the right places
- [x] Implement this for OS X
- [x] Fine tune the changed GUI layout
- [ ] Testing

Feedback on the approach is appreciated though.


Nadahar:
It seems that the Windows API function ```SetThreadExecutionState``` works on a per-thread basis. That means this implementation (or the current) won't work. It needs to be only one thread actually calling this function, so a redesign is needed.

Nadahar:
I've updated the implementation to use a worker thread to make all OS calls.

Nadahar:
I've moved to calls to ```startPlaying()``` and ```stopPlaying()``` to the ```StartStopListenerDelegate``` in the hope that it has more consistant calls. This must be checked thorougly though, as otherwise this will fail.

Nadahar:
I'm also having second thoughts about using "away mode". "Away mode" will allow the computer to go "partially" to sleep but keep the application alive. The question is what happens when "away mode" is disabled if the computer has already entered "away mode". If it will "wake up" then, all is OK - but if that mean entering "full" sleep UMS won't be able to respond to subsequent requests. That could mean that one could watch one video, and as soon as it's finished (before one can start the next) the computer will go to sleep.

This must be tested to determine what actually happens.

Nadahar:
I was trying to test this on Windows XP, but it turns out that Windows XP doesn't think that the virtual hardware (PSU, GPU etc) support sleep mode so it's impossible to get the VM to sleep at all (independent from UMS). As a result, I can't test it and I think you'll have to do that @Sami32.

Sami32:
@Nadahar I can try tomorrow.
But to avoid misunderstanding, i'll appreciate any guidance ;)
How do you want i test it exactly ? Stream a video more than 20mn ? try with 2 renderers ? (TV + web interface)

Nadahar:
@Sami32 Good think you asked... No, this isn't a finished solution at all yet - what I wanted to test is whether the control commands work for Windows XP. You don't have to play anything, in the ```General``` tab there's a new setting for prevent sleep mode. If you configure your computer to enter sleep mode after... 1 minute - and then start UMS and select "Never" - confirm that it goes to sleep after a little while. If that works, wake it up again and change it to "Running". Wait and confirm that it doesn't go to sleep.

Sami32:
@Nadahar Sorry, as you know with me it's only issues :wink: 
Whatever option i choose, it go sleep.

Did you get it working with Windows 7/10 ?

Nadahar:
@Sami32 I haven't tested it - the reason I wanted to test it for XP is because I had a suspicion that ```ES_AWAYMODE_REQUIRED``` isn't supported by Windows XP. According to Microsoft's documentation it is supported "from XP and above", but I also found some information that in Windows Vista you have to explicitly configure power management to allow "away mode". That tells me that this was probably a new feature for Vista..

I think the best thing to do is to forget all about "away mode" and use ```ES_SYSTEM_REQUIRED``` instead. If you change ```ES_AWAYMODE_REQUIRED``` to ```ES_SYSTEM_REQUIRED``` in ```WinUtils.disableGoToSleep()```, does that make it work?

Sami32:
Same results.

Perhaps you should "forget" WinXP if it's make your work too cumbersome.

Nadahar:
@Sami32 I'm relatively sure that ```ES_SYSTEM_REQUIRED``` work for Windows XP, so that makes me suspect that something else is wrong. Did you see in the log what commands where sent (search for "sleep")?

Sami32:
Sorry, it seem that i tested the old one. Here the log:
[sleep_awake.log.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/849494/sleep_awake.log.txt)

I'll test the other build later, as for now i'm the that sleep ;)

scriptorron:
I am impressed! You did a lot of work while I was sleeping!

For my opinion `ES_AWAYMODE_REQUIRED` would fit better but if it is not supported by all OS versions a `ES_SYSTEM_REQUIRED` alone will also do a good job. As far as I know the only difference is that the away mode switches off the monitor output. I can do the same with the power button on my monitor.

According to the Microsoft docu you have to set `ES_SYSTEM_REQUIRED` together with `ES_AWAYMODE_REQUIRED`:
> Away mode does not affect the sleep idle timer; to prevent the system from entering sleep when the timer expires, an application must also set the ES_SYSTEM_REQUIRED value.

Nadahar asked what happens when leaving the "continuous" mode: May Windows go to sleep immediately because the idle timer already has expired? I have the same concerns. Maybe the best way is to retrigger the idle timer before leaving the "continuous" mode:
1. disable sleep: `ES_SYSTEM_REQUIRED | ES_CONTINUOUS`
2. do some time consuming tasks (like content streaming)
3. retrigger idle timer: `ES_SYSTEM_REQUIRED`
4. reenable sleep: `ES_CONTINUOUS`

(In my code I made the steps 3 and 4 in the wrong order.)

In my code I called the disable/reenable functions in the `RequestHandlerV2.java`. But in fact the handling of the network requests is not a long task. It will be good enough to retrigger the idle timer there.

Nadahar:
I just tested "away mode" on Windows 10 and it works as expected.

@scriptorron 
> According to the Microsoft docu you have to set ES_SYSTEM_REQUIRED together with ES_AWAYMODE_REQUIRED

No, that is the whole point with "away mode". It looks like the computer goes to sleep, but it doesn't. It will still process requests. I tested this on Windows 10 by simply waiting for it to go to sleep (setting it to 1 minute) and then trying to browse UMS with Kodi. UMS answers as expected, even though the computer appears to be sleeping.

@Sami32 When you tested yesterday, did you test this? The computer is supposed to seem to fall asleep, but UMS should still work.

ExSport:
What I remember is that away mode is limited to 32bit OS and requires bios and driver support.
I am sure for Intel NUC there is a requirement to install away mode driver available in download section so I suppose counting with away support is bad direction 😎

Nadahar:
I've just found another problem. It probably varies how renderers handle this, but Kodi at least will buffer the whole file when playing a music file. From UMS' point of view the whole file is "streamed" as quickly as the network will transfer, and then the playback stops. This takes only a few seconds, while the song will keep playing for several minutes. This means that UMS will allow sleep way too soon. It's probably not a big problem though, because most people will have the sleep timer significantly higher, but it depends on how much the renderer will cache. As far as UMS is concerned the playback is finished when the transfer is complete.

Nadahar:
@ExSport There is no mention of these requirements in [Microsoft's description](https://msdn.microsoft.com/en-us/library/windows/desktop/aa373208(v=vs.85).aspx). Can you give some backing references for this?

Nadahar:
I've removed ```ES_AWAYMODE_REQUIRED``` from the equation and gone back to ```ES_SYSTEM_REQUIRED```.

ExSport:
Hi @Nadahar 
I can't find direct link on Intel NUC site but also don't remember details :))
This link helped me to remember littlebit:
https://www.sevenforums.com/general-discussion/32086-how-disable-away-mode.html#post1158579
I had problems in Device Manager with missing away mode driver in Windows 7 x64 and found only 32bit driver on Intel site if I remember well. So maybe it is supported in any OS but for x64 systems there is no default driver or downloadable driver available? Also I found this away mode in bios with mentioning "XP" like it is supported on WinXP only or it is only misleading old info...
Maybe for more recent OS Intel is using ReadyMode instead of AwayModeXP:
https://communities.intel.com/thread/78669
If I will find time I will try to debug it with powercfg:
https://www.eightforums.com/tutorials/50040-power-options-add-remove-allow-away-mode-policy.html
https://www.eightforums.com/tutorials/32947-power-availability-requests-manage-windows.html

Nadahar:
@ExSport Forget it, see my previous comments. I think using away mode is going to be way too much trouble.

Nadahar:
Except for the above mentioned problem with renderer caced media, this seems to work reasonably well on a Windows 10 VM where I finally got sleep mode to work.

It turns out that some drivers will prevent sleep mode (completely unrelated to UMS) in Windows. In my case it was the [network driver ```\FileSystem\srvnet```](https://social.technet.microsoft.com/Forums/windows/en-US/bce158e2-d057-4761-b715-5f68b8002fc2/not-sleeping-due-to-filesystemsrvnet?forum=w7itproui).

To check what is preventing sleepmode, run ```powercfg /REQUESTS``` (doesn't work in Windows XP). There you will see if UMS (Java) or any other driver, service or process that prevents Windows from entering sleep mode.

I hade to use ```powercfg /REQUESTSOVERRIDE DRIVER srvnet SYSTEM``` to tell Windows to ignore the faulty driver and still enter sleep mode.

This will need to be tested thoroughly if it is to ever be merged. The critical point is if the ```startPlaying()``` ```stopPlayer()``` events gets called correctly, also when e.g errors happen. It's almost bound to be some errors there. Even though one could argue that they're not related to this PR, they will affect this PR badly.

Sami32:
@Nadahar No i didn't, but now i can confirm that UMS is still reliable when the screen "sleep".

About away mode, yes and not ;-)
At first when i implemented it, on your recommandation, i understood the Microsoft information as you:
https://msdn.microsoft.com/en-us/library/windows/desktop/aa373208(v=vs.85).aspx

After you remark, i see many post like that:
https://github.com/aybe/Windows-API-Code-Pack-1.1/blob/master/source/WindowsAPICodePack/Core/PowerManagement/ExecutionState.cs

I thought that perhaps we understood it wrongly, but in fact the answer is more nuanced: Yes and not ;-)

> Away mode is a new feature included in Update Rollup 2 for Microsoft® Windows® XP Media Center Edition 2005 that allows the computer to appear off to the user while it continues to perform tasks that do not require user input. This paper provides OEMs information on installing, uninstalling, and configuring Away mode on computers running Update Rollup 2 for Microsoft Windows XP Media Center Edition 2005. It also includes hardware and software requirements for implementing Away mode functionality.
> This information applies for Microsoft Windows XP Media Center Edition 2005.

From the same official source in 2007:
> Away Mode is a Windows Vista™ feature that is designed for media PC scenarios. Away Mode enables the Windows Vista PC to appear off to the user, when the system is actually operational and processing background media tasks such as recording television content or streaming media to other devices. This paper documents the operation of Away Mode, configuration of power policies, and best practices for OEMs deploying systems with Away Mode functionality and ISVs developing media software for Away Mode.
> This information applies for the Windows Vista operating system.

http://download.microsoft.com/download/a/f/7/af7777e5-7dcd-4800-8a0a-b18336565f5b/awaymode.doc

So your choice seem reasonnable to me, but i have one question:
Is it possible to keep it for the other Windows OS ?

Nadahar:
@Sami32 We could always make it an option or use the Windows version to "autodetect" it, but it needs to be especially enabled by the user also for later Windows versions. I can already imagine all the support threads where we have to explain this over and over again. It's simply not worth it, and I think the difference from "SYSTEM_REQUIRED" is minimal. The screensaver and screen shutdown will function as normal in any case, so the difference is probably limited to what the status LEDs on the computer shows. It's just not worth the extra hazzle IMO.

Sami32:
@Nadahar I just thought that it will be possible to audetect the Windows OS and automatically trigger the right sleep mode.
If it's not possible, as i understood you, then i perfectly understand the support threads part :wink:

Nadahar:
@Sami32 It's possible to autodetect the OS, but we can't autodetect whether "away mode" is enabled in the OS - and it's disabled by default in newer Windows version including Windows 10. You have to start regedit to enable it..... it will just be an endless source of problems.

Sami32:
@Nadahar OK, i see. So i'm not interested in support threads as well ;)

valib:
@Nadahar it works only when video or audio is playing. It is not working for browsing or image show and the PC is stopped. That is why I checked the renderer requests in my PR #959 not only if the media is playing. Anyway good job.

Nadahar:
@valib It's relying on ```StartStopListenerDelegate``` so it won't react to browsing. Viewing images however, should reset the timer. We could reset the timer also from browse requests, but I'm not sure it's such a good idea. Renderers often live their own life, and I wouldn't be surprised if some of them would make browse requests from time to time to "refresh". My idea is that people doesn't have the sleep timer set so short that it will time out during browsing, even though I've also experienced that when testing (because I have a 1 minute timeout).

valib:
@Nadahar it is not only about the browsing. When users would like only watch images than the PC will stop without warning.

Nadahar:
@valib I'll have to look into that then, it must be a bug in the calls to ```StartStopListenerDelegate```. As said above, it's supposed to call ```SystemUtils.resetSleepTimer()``` every time an image is opened. These reset messages will appear in the log, are they missing?

src/main/java/net/pms/external/StartStopListenerDelegate.java
```diff
@@ -35,13 +37,17 @@ public synchronized void start(DLNAResource dlna) {
 		if (!started && ext != null && (ext.isVideo() || ext.isAudio())) {
```
valib: @Nadahar the problem is here.

valib:
@Nadahar after retesting I see that the reset messages are in the log. It seems that it is working.

Sami32:
@Nadahar That's weird, i tried many times, but get the same result as i told you: it go sleep whatever i choose "Running" or "Never" mode.
[Nadahar.log.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/852663/Nadahar.log.txt)

Nadahar:
@Sami32 Define "go to sleep". It's supposed to turn of the screen anyway, but if you choose "Never" UMS will not respond if you try to contact it with a renderer while if you choose "Running" it should respond.

Nadahar:
@UniversalMediaServer/developers I think this is ready from "my side". I've implemented a SleepManager that handles this in a single worker thread both for Windows and OS X. I took a quick look at how to implement this in Linux as well, but that was an endeavour that makes the OS X implementation seem small and easy. 

This evolved into something much larger than I ever envisioned (why does this always happen to me.. 😉), because I had to implement so much of the "core" OS X API to be able to do the simple things I needed. Much of this PR should ideally be submitted to JNA instead, I'll see if I get around to doing that later. This PR is now over 12000 lines of new code, most of it being JNA mappings. 

The changes to existing UMS code is minimal. I've "moved" everything sleep related out of the ```SystemUtils``` interface and into the new "SleepManager" class.

The one thing I haven't given much attention is to verify that the calls to ```startPlaying()``` and ```stopPlaying()``` and ```postponeSleep()``` is done under the correct circumstances. This PR is relying on ```StartStopListenerDelegate``` for that, so if that is bugged so is the sleep handling.

Use of ```SleepManager``` is really simple: Regardless of platform and confiuration settings, call ```startPlaying()``` when a stream/transcode starts and call ```stopPlaying()``` when the playback stops. ```SleepManager``` will count the number of calls and will disable sleep mode accordingly. Supported platforms and how sleep management is configured is handled by ```SleepManager```, so all the rest of UMS' code has to do is to make these calls as appropriate.

As long as every playback makes sure to call start and stop one, this should work properly. If not, it will fail to either prevent sleep mode or allow it again as it should. In addition, there is ```postponeSleep()```. This only resets the idle timer in the OS, so that the "idle time" starts counting from 0. This is currently only used for images, which doesn't have "start" and "stop" events. That means that as long as the user browse to the next image within the idle sleep time, the UMS computer will stay awake.

It would be nice if someone would test if this triggers at it should, but if noone can be bothered I doubt it's any worse than the current implementation.

Nadahar:
@UniversalMediaServer/developers There doesn't seem to be much interest in this. The whole reason I made this was because I had the impression that this was something that needed fixing, it wasn't something I had any particular interest in doing.

Even if nobody can be bothered to test this, I very much doubt it makes this any worse than it currently is.

Sami32:
@Nadahar Patience...wait until monday.

Nadahar:
@valib @SubJunk Don't you have any interest in getting this merged? @Sami32 has said that he feels that this is "too big" for him to be able to review, so if neither of you are going to react this is just going to be forgotten. It doesn't seem like anyone can be bothered to test it either (it takes time to test sleep mode behaviour). I have tested it in an "articial" setting with 1 minute sleep mode idle timer both on Windows and OS X, and it seems to be working as it should, although I can't guarantee that ```StartStopListenerDelegate``` is called appropriately from all code that should. I'm 99% sure it will be an improvement though, as the current implementation is known to get it wrong.

I understand that it's a lot to review, but as it's mostly new code that's not interacting very much with existing code I'd say that the risk is low. In the worst case, users could always disable "sleep mode prevention" in the configuration and none of this should affect them. 

As I see it, the part if this PR that is difficult to review (and some times understand) is the JNA stuff. There's really only one way to properly review this without diving really deep into both JNA and the OS X system libraries. I've done tests on each of the mappings I actually use (and many of the others as well), but there are some mappings that I haven't quite known how to test, but as long as they're not used they won't do any harm anyway. I'd recommend that you simply skip entierly or just browse quickly through eveything ``` src/main/java/net/pms/util/jna.*```. Excluding these should make this PR relatively easy to review. 

The classes in ``` src/main/java/net/pms/util/jna``` and sub packages should ideally be submitted to JNA and not be a part of UMS, as this code is general in nature and provides access to much of the OS X [Core Foundation](https://developer.apple.com/documentation/corefoundation) and [IOKit](https://developer.apple.com/documentation/iokit) frameworks that can be used for a whole lot of other things than managing idle sleep. If this turns out to work well and I can find it within myself to put in the work, I'd like to submit this code to JNA itself at some stage, in which case this code can later be removed from UMS.

SubJunk:
@Nadahar I'll have a look when I get a chance, hopefully this weekend

Nadahar:
@matthiasblaesing I'm considering to create a PR at JNA for what can be found in this PR under ```src/main/java/net/pms/util/jna``` and sub packages. Doing so will be some job though, so I'm a bit undecided if it's worth it. If you would take a quick look at the code (ignore anything in this PR that isn't under ```src/main/java/net/pms/util/jna```) and see if it would be something that you'd be interested in, it would make it easier for me to decide.

I know that I'd probably have to do some changes. I'm very much a fan of type-safety, so I've created a lot of types and tried to "emulate" type-safety to the best of my ability. I'm not sure if I've always found the best way to do it though, or how it fits into the general JNA "convention". A lot of mapping that could just be ```Pointer``` has some custom type assigned instead, and I've made generic integer enum mappings. I expect to having to change some of this and I would have to get some help/advice on what would be the best strategy in some of those cases.

That said, I've mapped a lot of OS X "system" calls in "Core Foundation" and "IOKit", and I've spent a lot of time generating Javadocs based on Apple's documentation, so it would be a shame not to offer this to others.

There are also some mapping that I'm not sure how to test, I've created far more mappings then I've actually used so I don't always know in what context they're used. I also find it a bit hard to "test" the mappings, there's no point in testing the native method itself (I'll assume that Apple has taken care of that), so in a way I think that simply making a successful call using all the parameters is about as good a test I can think of. We'd probably need to discuss this as well.

What I want initially though, is just a feedback on whether it's something you think would I should do or not.

SubJunk:
@Nadahar I just tested this and it works well for me. I'm doing a code review now and if I don't see any red flags I'll merge it

src/main/java/net/pms/configuration/PmsConfiguration.java
```diff
@@ -242,7 +243,8 @@
 	protected static final String KEY_PLUGIN_DIRECTORY = "plugins";
 	protected static final String KEY_PLUGIN_PURGE_ACTION = "plugin_purge";
 	protected static final String KEY_PRETTIFY_FILENAMES = "prettify_filenames";
-	protected static final String KEY_PREVENTS_SLEEP = "prevents_sleep_mode";
+	protected static final String KEY_OLD_PREVENTS_SLEEP = "prevents_sleep_mode";
```
SubJunk: We can probably deprecate this

src/main/resources/i18n/messages.properties
```diff
@@ -363,6 +362,11 @@ NetworkTab.71=Server name:
 NetworkTab.72=Append profile name
 NetworkTab.73=<html><strong>Default:</strong> False<br><strong>Notes:</strong> When this is enabled, the UMS profile name (by default, the hostname of your computer) is appended to the server name when displayed on the renderer.</html>
 NetworkTab.74=Enable splash screen
+NetworkTab.PreventSleepLabel=Prevent sleep:
```
SubJunk: Cool, I like this more than numbers

src/main/java/net/pms/configuration/PmsConfiguration.java
```diff
@@ -2720,12 +2722,29 @@ public void setIpFilter(String value) {
...
+		configuration.setProperty(KEY_PREVENT_SLEEP, value.getValue());
+		PMS.get().getSleepManager().setMode(value);
+	}
+
+	public PreventSleepMode getPreventSleep() {
+		PreventSleepMode sleepMode = null;
+		String value = getString(KEY_PREVENT_SLEEP, null);
+		if (value == null && configuration.containsKey(KEY_OLD_PREVENTS_SLEEP)) {
+			// Backwards compatibility
+				sleepMode =
```
SubJunk: Formatting here is a bit weird, is it a convention to use an extra indent with a multi-line ternary? No need to change it but just curious

Nadahar: @SubJunk I'm not aware of any convention for that, I just try to find a way to make the code readable. I'm not very happy with the way it looks either, but all the other things I've tried have looked worse in my opinion.

Some quick googling shows that some others use this form as well, but there doesn't seem to be much consensus. The Eclipse formatter uses this formatting:
```java
sleepMode = getBoolean(KEY_OLD_PREVENTS_SLEEP, true) 
	? PreventSleepMode.PLAYBACK
	: PreventSleepMode.NEVER;
```
Do you think that is better/more readable? While I find it pretty clean, I think putting the ```?``` on the next line can make it easy to miss that it *is* a ternary in the first place.

I like to use descriptive identifiers - the downside is that lines tend to get long, so I've had to split ternaries in several places and would prefer if we had an "agreed upon" way to format them.

src/main/java/net/pms/configuration/PmsConfiguration.java
Nadahar: @SubJunk Do you mean just using the ```@Deprecated``` notation on the field?

src/main/java/net/pms/configuration/PmsConfiguration.java
SubJunk: @Nadahar to me the most readable and standard way is to use `if...else`:

```
if (getBoolean(KEY_OLD_PREVENTS_SLEEP, true)) {
	sleepMode = PreventSleepMode.PLAYBACK;
} else {
	sleepMode = PreventSleepMode.NEVER;
}
```

Since then the code is more like most human languages. So if we are picking a convention I would choose that.
I do like how ternaries save repetition and space though, they are more efficient.
The change you made in the new PR is an improvement but I agree with your comment about the question mark.

Personally I use `if...else` for multi-line, and ternary is single-line, but that's just me personally.

Nadahar: @SubJunk I agree that if the ternaries get too complex it's preferrable to switch to ```if...else``` instead, but in this case it's actually very basic and I find terniaries *more* readable when the complexity is low (not to mention that the can save a lot of hazzle when used inline on method arguments). The only think that makes this line "too long" is the descriptive identifiers combined with the level of indentation. Those things also change while you work with the code, so it might have fitted just fine when you first wrote it.

This one could still be on one line actually, I think the line length was 119 if it was all on one line. Even though many still treats the 80 characters width as if it were "set in stone", I think it's oulived its usefullness a long time ago. It does stem from the pre-GUI era where a screen was typically 80 characters wide, which isn't really relevant anymore.

I can certainly change this one to ```if...else```, but what I'm really after is how to split too long lines with terniary expressions when that need arise.

Sami32:
@Nadahar Why do `jna.macos.kernreturn` library is not skipped on Windows ?

Nadahar:
@Sami32 It's not calling any (OS X) native code, it's just type mappings that will work on any platform.

Sami32:
@Nadahar Thanks. So, you can merge it LOL

src/main/java/net/pms/configuration/PmsConfiguration.java
SubJunk: @Nadahar yeah this would work on one line. Maybe our rule should be to use one line for ternaries and if it gets longer use `if...else`, instead of making it more complicated than that.

Nadahar: @SubJunk I don't think that's a good solution. As said above, I often rearrange/more code blocks, add exception handlers or otherwise insert code blocks which leads to indentation after I wrote the code itself. I also sometimes write the code using shorter identifier names and then change them to more descriptive names when I'm done. Having to actually change the code (from ternary to ```if...else``` as a result of what level of indentaton the code has at any given moment doesn't seem very practical or reasonable to me. 

There's also another case where I think ternaries is completely superior: If you make a complex method call where let's say you have 4 parameters the can be one thing or another that varies depending on circumstances. The handle that with ```if...else``` you either have to write 16 different method calls using with different parameters, or you have to use temporary variables that you set using ```if...else``` and then use the temporary variables in the call. Using ternaries in that situation not only saves a bunch of code, it also makes it much more readable in my opinion. It much easier when reading the code to see what leads to what because you don't have to "follow" the different temporary variables to figure out what the result is, or study all the different conditions to understand which method call is actually being performed. It allows you to "isolate" the evaluation for each parameter in an easily comprehensible way.

Thus I think we need to have a way to wrap ternary expressions when it's needed. I'm not saying we should use ternaries "all the time", I'm just saying that some times it's the best choice (this case isn't the best example) and when that's the case we need to agree how to do it.

SubJunk: @Nadahar I'm not sure I understand the issue with indentation, I haven't had any issues with that personally. I also don't know how multi-line ternaries can prevent temporary variables in that case but it would be good to see a code example for it.

I've been learning the Go language lately and they actually prescribe formatting and there's a built-in formatting tool in the library, it's a relief to not have to make decisions on this stuff. Oracle used to have a guide but they seem to have stopped updating it. Anyway maybe we can find some code standard to follow and stick to that. Unfortunately the Google style guide doesn't mention multi-line ternaries https://google.github.io/styleguide/javaguide.html

My opinion on the best format is:

```
sleepMode = getBoolean(KEY_OLD_PREVENTS_SLEEP, true) ?
	PreventSleepMode.PLAYBACK :
	PreventSleepMode.NEVER;
```

Once we agree on the format, I think we should record it in a file in the project. Maybe we can say something like:

```
UMS Style Guide:
Things not specified in this guide should use the Google Style Guide at https://google.github.io/styleguide/javaguide.html

Ternaries:
(the style we decide on)
```

We can also consider making a linter config for our styles to make it easier to remember if the list grows.

Nadahar: @Sami32 The indentation issue depends on how you work I guess. I "refactor" the code quite a lot while I work with it as I discover that the code would be better organized another way. I then move blocks of code around, which can mean that the code ends up being at a "higher level of indentation" then it was when I wrote it, which - together with renaming the identifiers might mean that something that fit nicely on one line at one stage suddenly doesn't. I don't want to have to rewrite/change working code just because the line length changes because of such reorganizations.

CheckStyle has both Oracle's and Google's styles by default, neither fits very well with our current "style". I used one of them as a starting poing for the current CheckStyle configuration, but I don't remember which. 

I know very little about golang, but I did a little bit of reading now and it seems that the language is very limited (most even remotely "advanced" concepts are excluded). That makes it very easy to enforce a style 😉 

I'm not saying that there's a need for multi-line ternaries every day, but when the need is there I'd like to know how to format it - especially since I've had a hard time finding a format I'm "satisfied with". 

When it comes to preventing temporary variables, I'll try to give a short pseudo example:

```java
    type tempArg1;
    if (condition1) {
        tempArg1 = value1;
    } else {
        tempArg1 = value2;
    }

    type tempArg2;
    if (condition2) {
        tempArg2 = value3;
    } else {
        tempArg1 = value4;
    }

    method(tempArg1, tmpArg2);
```

Using ternaries this could be expressed much easier and cleaner:
```java
    method(condition1 ? value1 : value2, condition2 ? value3 : value4);
```

The above is using just two "conditional" arguments. As the number of "conditional" arguments grow, the difference grows exponentially.

Nadahar: @SubJunk It seems that we agree on what is the "least worst" way to format multi-line ternaries though. I have no idea why I put ```sleepMode =``` on a separate line, I probably changed that code some times before the result was final, so that it might have made sense at one stage.

SubJunk: @Nadahar cool, I'm glad we agree on that. Yes Go is purposely simple to enable quick adoption - as I understand it was developed by creators of C as a successor to it. I like some of its syntax. I often learn new languages briefly. In particular I like that enforced formatting style and I like that its functions can return more than one value at once unlike Java (you can do it in Java but there's a piece of abstraction like wrapping them in an object/array). It fits with a pattern I'm familiar with at my work which is to often deal with two variables at once in callbacks/promises/awaits, where one is the optional value and one is the optional error.
Anyway there are advantages and disadvantages to all languages whether OO, procedural, functional, etc.

Nadahar: @SubJunk I was created at some people at Google that wanted something quicker/easier than C. In their quest to make it simple, they left our a whole lot of useful stuff which means the possibility to reuse code is severly limited. Inheritance, overloading, generics are all non-existant for example, basic concepts that's crucial to make "useful" code as I see it. I don't want to keep writing *almost* the same code over and over again.

I like that you can return multiple values, that's a fundamental flaw in Java. You can return multiple objects by returning an array of ```Object```, but then you screw type-safety completely. Even though that's a serious shortcoming, it doesn't outweigh all the fundamental stuff missing from golang IMO. It seems to me that what it's primarily replacing is Python code, and Python isn't a "proper language" to me anyway (although you can do a lot of hackish stuff there easily). 

The one thing I *really* like with golang is that it compiles proper, native binaries without any dependencies. That is what *really* sucks with Java. There does exist such a compiler for Java though, but it's commercial. The day the open source community gets around to writing such a compiler for Java, things will improve a lot. Good bye JVMs, bytecode, JAR files, classpaths and all the rest. Distribute the executable and it can do it's stuff on its own - the only *right* way to do it.

Sami32:
@Nadahar Do `window_extended_state` is still used ? for what for ?

valib:
@Sami32 what has `window_extended_state` common with this PR? I introduced this value for saving the UMS window state because the `width` and `height` window value in that case doesn't correspond with the screen size when window is maximized.

Sami32:
@valib Nothing, just that it is not documented anywhere. So i was wondering about, and searched answer without having at look at the code. And i wished add it in our documentation.

`0` is the default value that save the window state ? As i saw that in the log. And 1 is the other value meaning to not save ?

Thanks for explaining ;-)

valib:
@Sami32 it is a value which is not mentioned to be changed by users. It is only a value which the UMS is stored in the .conf to use it when the UMS is initialized.

Sami32:
OK, thanks for having clarified it :+1:

PR #1227, opened 2017-03-16
Preventwindowsfromsleep fix

scriptorron:
This lets Windows count how often sleep is disabled and reenabled. You can check the effect by counting the lines with " ES_CONTINUOUS" and "|ES_CONTINUOUS" in the debug log:

grep "|ES_CONTINUOUS" debug.log|wc
grep " ES_CONTINUOUS" debug.log|wc

(should be the same when all network requests are finished)

and by looking at the outputs of 
powercfg -requests 
(run this as admin on Windows command line)
(there must be no "java" behind "SYSTEM")

One issue is still not solved:
Streaming of large content seems to run in a parallel thread (for instance I see ffmpeg in the log) which is not between a disable and a reenable! There is still work needed to avoid sleep during streaming!

This is not a final solution!

### DIFF
--- a/src/main/java/net/pms/io/WinUtils.java
+++ b/src/main/java/net/pms/io/WinUtils.java
@@ -31,7 +31,6 @@ import net.pms.PMS;
 import net.pms.configuration.PmsConfiguration;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import java.util.concurrent.atomic.AtomicInteger;
 
 /**
  * Contains the Windows specific native functionality. Do not try to instantiate on Linux/MacOS X !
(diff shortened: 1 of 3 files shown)


scriptorron:
I have just seen that I pulled the wrong version. I will try to close this and try it again...

Nadahar:
@scriptorron This PR will be updated if you update your branch - as long as it's not closed.

Just push the version you want to your branch ```PreventWindowsFromSleep_fix``` and this PR will reflect that. You might have to use ```--force``` to push if some of the commits are changed.

scriptorron:
@Nadahar, Thank you! I am still learning git. I use the GUI provided by GitHub and sometimes it is not clear to me what happens when I press a button. Sorry for the circumstances!

Nadahar:
@scriptorron No problem. The web GUI is cumbersome and very limited, so I understand that it can be confusing 😉

Nadahar:
I'd recommend that we merge #1226 instead of this, as it's a more complete solution also handling OS X.

valib:
Merged  #1226 instead

PR #1229, opened 2017-03-21
Samsung 9 series

Sami32:
Following some users request on the forum.

Tested by this user:
http://www.universalmediaserver.com/forum/posting.php?mode=reply&f=9&t=9249


Sami32:
@SubJunk Thank you :-)

PR #1238, opened 2017-03-28
MediaMonkey renderer

Sami32:
@Nadahar I guess that you use MediaMonkey ?
While you'll have some time, nothing hurry, if you, or anyone else, could try it, it should be better.

As a side note, only MediaMonkey 4 support videos.


Nadahar:
I've been using MediaInfo for testing, audio mostly. I don't quite like the UI in MediaMonkey, it's a bit difficult to know what's happening. I'll see if I can do some testing, although I'm not sure if I have MediaMonekey 4.

Sami32:
@Nadahar Whatever you've will better than nothing.
The version is 4.1.5 is out since 22 october 2014, so i guess that you have it,
if you dont have the latest version, at least the 4.1.13 version from June 29th 2016,  ALAC and video will not be supported.
http://www.mediamonkey.com/news/category/releases/

As a side note, i didn't know that the latest version have Chromecast support
http://www.mediamonkey.com/wiki/index.php/What%27s_New_in_V4

Nadahar:
@Sami32 I had 4.1.14 - upgrading to 4.1.15 now.

Sami32:
@Nadahar If you get an AC3 issue, it should be solved in that developpment build:
http://www.mediamonkey.com/beta/MediaMonkey_4.1.16.1831.exe

Nadahar:
@Sami32 I just tested this, and it was properly recognized. But, that was about the only thing that worked.

No transcoded content will play, neither audio or video. I only got AVI files to play of the video files (and maybe MP4..). ASF didn't work, MKV probably would have worked but it complained that I had no codec installed (which I do) - but that's not UMS' problem.

Of audio files, only mp3, flac, some wav, wma and m4a/aac played.

The new version I installed has a major issue in that the "UPnP part" hangs quite often (but the GUI doesn't), so it's hard to tell if something doesn't play because it doesn't work or because it's just hanging. I only see that it's hanging if I try to quit, as it won't and I have to kill it. After restarting, a file that didn't work the last time might work, but if I click on something that doesn't work it's a good chance that it will hang again..

I think it's impossible to get this to work properly because of MediaMonkey's buggy nature. But, getting transcoding to work would sure help a lot.

[debug.zip](https://github.com/UniversalMediaServer/UniversalMediaServer/files/875996/debug.zip)

Sami32:
@Nadahar Joker ;)
I didn't go farter than seen that your are using an old FFmpeg version (2015) that support only experimental AAC and since with withdraw it from the UMS code, you are getting FFmpeg transcoding issue (errors) because of unsupported `-c:a aac`
Try with a recent FFmpeg version and check if the transcoding work correctly.

I didn't tested on DLNA, as i will need to configurate MediaMonkey before.

Nadahar:
@Sami32 AAC was one of the few things that actually worked though...

Nadahar:
@Sami32 I've updated FFmpeg (with the result that FFmpeg crashed as well). Nothing else improved. I've tried to set ```TranscodeAudio``` to LPCM, WAV and MP3. Nothing works, so there must be something on the DLNA level that makes them unplayable. It might be that our bugs collide with their bugs to make a complete chaos. Since audio transcoding generally works for other renderers, I don't know if this is worth spending time on though.

[debug.zip](https://github.com/UniversalMediaServer/UniversalMediaServer/files/876907/debug.zip)

Sami32:
I'll try that later, but no UMS on the compatibility list:
http://www.mediamonkey.com/wiki/index.php/Compatible_UPnP/DLNA_Devices

Nadahar:
@Sami32 I guess that goes both ways, we've never mentioned MediaMonkey as "compatible" either. But, that's irrelevant - the whole point with DLNA is that if every party stick to DLNA, they can all talk to eachother without knowing anything about each other.

Sami32:
@Nadahar Do you know why you get many Global ID report like this ?
```
TRACE 2017-03-28 21:43:57.826 [New I/O worker #8] net.pms.network.RequestHandlerV2 HTTP: upnp/control/content_directory / 0-0
DEBUG 2017-03-28 21:43:57.827 [New I/O worker #8] net.pms.dlna.GlobalIdRepo GlobalIdRepo: id not found: -1
DEBUG 2017-03-28 21:43:57.827 [New I/O worker #8] net.pms.dlna.GlobalIdRepo GlobalIdRepo: id not found: -1
DEBUG 2017-03-28 21:43:57.827 [New I/O worker #8] net.pms.dlna.DLNAResource Bad id at3p_sample3 found in path
```

Nadahar:
@Sami32 No, I'd have to look into the details of the global ID handling. The only thing I think I remember is that the ID is parsed from the URL that the renderer calls. You should look at the URL to see if you see anything strange. UMS doesn't actually use the "filename" part of the URL to find the resource when the renderer is asking for one, it only looks at the ID specified just before the "filename" in the URL. My memory could be wrong though.

Sami32:
@Nadahar It seem that your encounter the subtitles text displaying issue as well while testing this PR, is that correct ?

I only know that something is not right, but don't saw anything wrong as i don't have any skill on that.

Nadahar:
@Sami32 Nothing with transcoding played at all - so I wouldn't know if I had any subtitle issues. 

When it comes to global id I don't know how it works either. I process so much information dealing with all kind of different things that there's no way I can remember such details. I do remember that I looked at it not that long ago relating to the memory leak, but I only looked at how they were stored (and never free'd), not how they were used.

SubJunk:
Should we close this PR since it seems stalled?

Sami32:
@SubJunk I never got the time to test it again.

PR #1239, opened 2017-03-31
Small audio related fixes

Nadahar:
This is a bit of a mixed PR, but I just couldn't let those musicbranz issues be when I found them. The main purpose was to avoid parsing ADPCM and DSD with MediaInfo as they aren't supported and the parsing fails. Their ```container``` and ```audio codec``` will be set based on ```Format``` which is in reality simply based on file extension, so this is far from ideal - but still better than nothing.

MusicBranz has made a couple of changes since the implementation, so I've done a couple of small fixes to make the parsing succeed:
* The field ```year``` used to contain 4 digits with the year only, but now if suddenly contains a full date. I've simply concenated the string to keep the first 4 as they seem containt the "year" part of the date.
* Even though the API documentation doesn't say so, it looks like they've changed the default format from XML to JSON. We use an XML parser, which fails horribly when fed JSON so I've added an extra parameter to explicitly request XML.

### DIFF
--- a/src/main/java/net/pms/dlna/DLNAMediaInfo.java
+++ b/src/main/java/net/pms/dlna/DLNAMediaInfo.java
@@ -28,6 +28,7 @@ import net.pms.configuration.PmsConfiguration;
 import net.pms.configuration.RendererConfiguration;
 import net.pms.formats.AudioAsVideo;
 import net.pms.formats.Format;
+import net.pms.formats.Format.Identifier;
 import net.pms.formats.v2.SubtitleType;
 import net.pms.io.OutputParams;
 import net.pms.io.ProcessWrapperImpl;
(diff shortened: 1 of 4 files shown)


src/main/java/net/pms/dlna/DLNAMediaInfo.java
```diff
@@ -843,7 +846,16 @@ public void parse(InputFile inputFile, Format ext, int type, boolean thumbOnly,
...
+					// Set container for formats that the normal parsing fails to do from Format
+					if (StringUtils.isBlank(container) && ext != null) {
+						if (ext.getIdentifier() == Identifier.ADPCM) {
+							audio.setCodecA(FormatConfiguration.ADPCM);
+						} else if (ext.getIdentifier() == Identifier.DSD) {
+							audio.setCodecA(FormatConfiguration.DSDAudio);
+						}
+					}
+
+					if (StringUtils.isNotBlank(audio.getSongname())) {
```
Sami32: Why did you deleted the presence test  (`length() > 0`)for `getSongname()` ?

Nadahar: @Sami32 ```StringUtils.isNotBlank``` does this and more:
```
Checks if a CharSequence is not empty (""), not null and not whitespace only.

 StringUtils.isNotBlank(null)      = false
 StringUtils.isNotBlank("")        = false
 StringUtils.isNotBlank(" ")       = false
 StringUtils.isNotBlank("bob")     = true
 StringUtils.isNotBlank("  bob  ") = true
```

Sami32: Oups, sorry, i should have mind absence at that time.

Nadahar:
@Sami32 Nope.

Issue #1170, opened 2017-01-27
will not be streamed because the H.264 level is unknown


Sami32:
@SubJunk I was referred to that issue:
http://www.universalmediaserver.com/forum/download/file.php?id=3626
http://www.universalmediaserver.com/forum/download/file.php?id=3546
http://www.universalmediaserver.com/forum/download/file.php?id=3519
And many other since 6.5.2 version. It seem that Samsung users are more affected, or they report more ?

> TRACE 2017-01-22 15:29:23.041 [New I/O worker #1] net.pms.dlna.DLNAResource File "Batman v Superman -  Dawn of Justice..mp4" will not be streamed because the **H.264 level is unknown**.

SubJunk:
Cool, that gives the info I need and it does seem we have a bug here. I'm looking into it now. Thanks!

SubJunk:
I got one of the files in those logs and it gave me the level correctly when we scanned it:

```
Parsing results for file "Chicago.PD.S04E10.HDTV.x264-KILLERS.mkv": container: mkv, bitrate: 808625, size: 256001892, video tracks: 1, audio tracks: 1, video codec: h264, duration: 00:42:12.00, width: 720, height: 404, frame rate: , mime type: video/x-matroska, matrix coefficients: BT.709, avc level: 3.1
	Audio track id: 0, lang: und, audio codec: AAC, sample frequency:48000, number of channels: 2, bits per sample: 16
```

when we played it:

```
Starting transcode/remux of Chicago.PD.S04E10.HDTV.x264-KILLERS.mkv with media info: container: mkv, bitrate: 808625, size: 256001892, video tracks: 1, audio tracks: 1, video codec: h264, duration: 00:42:12.00, width: 720, height: 404, frame rate: , thumb size: 23569, mime type: video/ogg, matrix coefficients: BT.709, avc level: 3.1
	Audio track id: 0, lang: und, audio codec: AAC, sample frequency:48000, number of channels: 2, bits per sample: 16
```

and then I restarted UMS and played it again to make sure it reads the level correctly from the database and it did:

```
[...]

PR #1242, opened 2017-04-01
A better Video Coding Level implementation

Sami32:
- [x] Implement the H.265 level renderer limitation
- [x] Correct a [bug](https://github.com/UniversalMediaServer/UniversalMediaServer/issues/1170) in the AVC level parsing
- [x] Log the HEVC level

The only purpose of that PR is to log this information, as some renderer, like [Samsung 9 series](http://www.samsung.com/us/support/answer/ANS00049952/), have a H.265 limitation as well, and that the video could be correctly triggered to transcoding if they need to be.

I leave a potential implementation in the video engines to some more skilled ones.


src/main/java/net/pms/dlna/LibMediaInfoParser.java
```diff
@@ -678,13 +688,39 @@ public static String getAvcLevel(String value) {
 		// High@L3.0
 		// High@L4.0
 		// High@L4.1
-		final String avcLevel = substringAfterLast(lowerCase(value), "@l");
+		// // Value can look like "Stereo High@L4.1 / High@L4.1".
+		if (isNotBlank(substringBetween(lowerCase(value), "@l" , " /"))) {
+			final String avcLevel = substringBetween(lowerCase(value), "@l" , " /");
+		} else {
```
Sami32: It's seem that it's not the correct way of doing that ?

src/main/java/net/pms/dlna/DLNAMediaInfo.java
```diff
@@ -2403,18 +2445,38 @@ public int getAvcAsInt() {
...
+	}
 
 	public void setH264Profile(String s) {
 		synchronized (h264ProfileLock) {
 			h264Profile = s;
 		}
 	}
+	
+	public void setH265Profile(String s) {
+		synchronized (h265ProfileLock) {
```
Sami32: Is the reuse of `String s` correct in that case ?

Nadahar: @Sami32 You're not reusing it as it's separate methods. ```s``` only exists in that method's scope, so there's no conflict. 

Using synchronization can however be a bit tricky, as you have to look at the context from which the method is called to make sure you don't create deadlocks.

src/main/java/net/pms/dlna/LibMediaInfoParser.java
Nadahar: @Sami32 ```final``` means it can only be set once. However, the bigger problem here is scope. ```avcLevel``` will only exist in the scope of that if-block, so it won't exist as soon as the code leaves that.

Sami32:
I think that it ready to be review now.

src/main/java/net/pms/dlna/LibMediaInfoParser.java
```diff
@@ -156,6 +156,9 @@ public synchronized static void parse(DLNAMediaInfo media, InputFile inputFile,
 						if (!value.isEmpty() && media.getCodecV() != null && media.getCodecV().equals(FormatConfiguration.H264)) {
 							media.setAvcLevel(getAvcLevel(value));
 						}
+						if (isNotBlank(value) && media.getCodecV() != null && media.getCodecV().equals(FormatConfiguration.H265)) {
```
Nadahar: If this is read from the same MediaInfo value, there's no reason to store twice and have twice the locks etc.

Sami32: Sorry, i don't undertsand what you mean: `store twice and have twice the locks etc` ?

Sami32: isNotBlank(value) && media.getCodecV() != null

Sami32: Oh, you mean all the lock part that i copy past ?
I don't really know, as i just copy past and adapt, but you should be right.
But i'm not sure i'll make thing worst if i try to delete some part.

Nadahar: @Sami32 That's ok, the important thing is that ```media.getCodecV() != null``` comes before ```media.getCodecV().equals...```.

What I was trying to say is that it seemed strange to take the same information from MediaInfo, stored in ```value``` and store in two different places. But I see now that ```getAvcLevel()``` and ```getHvecLevel()``` produces different outcome, so I guess it's ok. Still, using one field to store this would mean you didn't have to expand the database etc to store this information twice.

Nadahar: I'd say that storing the "raw" value from ```Format_Profile``` in ```DLNAMediaInfo``` and the database would be better. The parsing of the value should be done in ```DLNAMediaInfo.getAvcLevel()``` and ```DLNAMediaInfo.getHevcLevel()``` instead. Parsing this will be so quick that doing it each time won't be noticable, and the benefit is that every ```DLNAMediaInfo``` instance kept in memory (which can be very many) and each database record will be smaller. It will also be enough with one set of locks.

This is where I feel like showing is easier than explaining, but I won't do that this time...

Sami32:
Ok, now i'm frustrated :-( as i cannot do as you explained, so feel free to commit and merge, as i really don't bother as long as it's done; but if you can make it optimal that'll be better.
Just delete my added attempt to do it if you need.

Sami32:
I cannot do better...

Nadahar:
@Sami32 I takes a bit of time for me too to come up with something, especially since part of the H264 level code was very... unstructured. I've tried to clean it up a bit, fixing a couple of bugs on the way. 

This is completely untested, and I honestly don't know enough about this to even know how to test it properly. You can use this as an inspiration or to build on if you like. You can find it in branch [h265Example](https://github.com/UniversalMediaServer/UniversalMediaServer/tree/h265Example).

Sami32:
@Nadahar Thank you very much for trying to be nice, but i know that i take you 10 minutes with only one finger.
That's make me cry :cry: 

AFAIK, 4.1 AVC profile is wide spreaded around, but some Philips have a 4.0 limitation, for PS4 it's 4.2, for Chromecast 5.2, and for Samsung the profiles support depend on the video resolution as well...

You can push it if you want, as i'm only upset after myself for not being able to do it more optimally, nothing else. When i try to follow you remark i get stuck at the profile gstion for the 2 video coding profiles. It seem that you solved it nicely with enum classes for each one.

I don't understand why you use Pattern and Matcher, as i read that they make the code unnecessary slow ?

Great work by the way :+1:  you added a better limitation level as well, for many renderers that don't stick to the 4.1 limitation.

Note: I'll look at it again later, when i'll come down on this.

Nadahar:
@Sami32 

I did spend 2-3 couple of hours on this, so it didn't take me 10 minutes 😉 If I were to use the GitHub editor or Notepad, I don't know how many hours it would take me, you'll have to take the step into an IDE to cut down on the time it takes, there's no way around it... the whole purpose of an IDE is to give you the information and "shorcuts" you need easily available...

I've just rebased it on latest master, and resolving the conflicts in ```DLNAMediaDatabase``` wasn't fun so I had to squash the two commits into one (or I would have to a lot more job since I reversed much of the changes done in ```DLNAMediaDatabase``` in my commit). This means that this won't randomly fail at Travis-CI anymore 😉 

The enum classes is mostly for type-safety, the "structural change" I made was that instead of splitting the information gotten from the media (format_profile) in 4 (h264 level, h264 profile, h265 level and h265 profile) and storing those 4 separately, I just store the single value from which all the others can be deducted. When one of them is needed, it's just a matter of applying an algorithm that will take a few nanoseconds to transform it into the information you need. It will be more calculation this way, but less I/O. Since CPU is cheap and IO is expensive these days (I'm not talking money, but the fact that IO is generally MUCH slower than CPU resources so that most systems are "IO bound"). If the calculations were heavy (like scaling an image), calculating them once and storing the information separatly would make more sense. In this case there was an added benefit from doing it this way because of thread-safety. Too many locks can be very hard to manage correcly, and by storing the information once we get away with one lock instead of four. Locks can also cost CPU cycles because they impose memory barriers (a way of "synchronizing" the memory between CPU cores), even though this is a smaller problem than many people seem to believe as long as the locks aren't "heavily contested" - and these locks definitly aren't. In any case, getting rid of some locks are always great.

Regex (Pattern and Matcher) has some cost, but it's power is unbeatable. By using those I don't have to write a lot of code that I otherwise would have to if I wanted the same functionality. As always you have to weigh the cost against the benefits. These regex'es means that the parsing can accept great variation in the input syntax and still manage to find the information we're after.

I'll try to explain (I'm setting an option when creating the pattern making it case-insensitive):
```regex
^\s*(?:[^@\s]+@)?(?:L|LEVEL)?\s*([\db]+(?:\.\d+|,\d+)?)(?:@\S*)?\s*$
```
Step by step:
* ```^\s*``` - From the beginning of the string, there can be zero to unlimited number of whitespace (space, tab etc)
* ```(?:``` - Start of non-capturing group
* ```[^@\s]+``` - One or more characters that is NOT ```@``` and NOT whitespace
* ```@```- Exactly one ```@```
* ```)?``` - End of non-capturing group - ```?``` makes the group optional, so that it will match both if there are some text ending with a ```@``` or not.
* ```(?:``` - Start of a new non-capturing group
* ```L|LEVEL``` - The content can be either ```L``` or ```LEVEL```
* ```)?``` - End of optional group, which means that the prefix L or LEVEL can either be there or not.
* ```\s*``` - Zero or more whitespace
* ```(``` - Start of capturing group (since there's no ```?``` after the closing ```)```
* ```[\db]+``` - One or more characters that's either a digit or the letter ```b```
* ```(?:``` - Start of a new non-capturing group
* ```\.\d+|,\d+``` - There can either be a ```.``` and one or more digits or a ```,``` and one or more digits
* ```)?``` - End of optional group
* ```)``` - End of capturing group
* ```(?:``` - Start of a new non-capturing group
* ```@\S*``` - Exactly one ```@``` and then zero or more NON-whitespace characters
* ```)?``` - End of non-capturing group which captures any ```@xxxx``` after the number
* ```\s*$``` - Zero or more whitespace until the end of the string.

By doing this, I ensure that to get a match, the string must follow the rules stated above. In addition I get access to what was captured in the "capture group", that is - just the "number" (digits or ```b```, optionally ```.``` or ```,``` followed by more digits). The next thing I do is to replace ```,``` with ```.```, which means I'm ready to match the content of the capturing group to one of the pre-defined values (```1.0```, ```1b```, ```4.1``` etc).

This means that you can throw a lot of different syntax on this, and it will manage to make sense of it. ```Level3.2```, ```l2```, ```lo@5,1```, ```3@you``` and ```ali@level 4.2@baba``` will all be parsed to valid levels.

Accomplishing the same using ```if's``` would require much more code and multiple checks, so I'm not so sure using regex is so ineffective after all. But sure, if one made an effort to optimize code to do the same, one could probably shave off a few nanoseconds - but the cost would be that the code would be hard to read/maintain and it would take a long time to write.

I don't use regex to check things that can be easily and quickly checked using simpler means, but one have to be aware of that other methods might call regex internally. ```replaceAll()```, which is used a lot in UMS uses regex, the same does for example ```String.matches()```. Using them is just as expensive as using regex directly. Another advantage with regex (that was of no use here, but can be some times) is that you can "compile" the regex once, and then use it to match again and again (given that the same regex applies). 

There are some things still missing here:
* The fact that the "default 4.1 limitation" is gone. 
* I've made a comment in ```LibMediaInfoParser``` where I didn't understand what the code was supposed to do. That needs figuring out.
* Implementation in the encoders for other levels/H265. I don't really know how this ties together, but this is from what I can see used for two things. 1) Do transcode when needed 2) To regulate the output format from the transcoding process
* All this "level 4.1 validation" stuff looks strange to me, I'm not sure that code does what it's supposed to. It would also be useful to expand this to be valid for more than just H.264 4.1.
* Testing

Sami32:
Thank you for the explaination :+1: 
So that mean that all the profiles similar to "Main 4:2:2@L" or "Stereo High@L" will not be be parsed correctly ?

I admit that when i saw "Resolve conflits", i just stay far aways...
I asked about the Matcher because i have read that using that Java native function is faster and less CPU use.

I prefer leave it to the developpers having skill in video engines for the implementation.
As i don't have skill, i don't try to go in a work if i know i cannot finish it. I can only afford littles and easy tasks.

Nadahar:
@Sami32 Their level will not be parsed - this has nothing to do with parsing the profile. How could one parse level from that?

I can take a look at the conflicts..

Sami32:
I guess that i should been less lazzy in the copy/paste :wink:
 "Main 4:2:2@L5.1" or "Stereo High@L4.2"
My remark concern the space in the profile, before the level.

Nadahar:
@Sami32 Maybe it was copy/haste?

Space will make it not parse yes, as none of the examples I saw had space. It's easy to modify the regex tough - or those valid specifications?

Nadahar:
@Sami32 I put "my commit" first and your second. Unfortunately there were so many conflicts between them that when I removed everything from the second commit that conflicted with the first, there's only the change to ```DefaultRenderer.conf``` left. I've made a backup of your original commit locally though, so I can push it if you want it.

In any case, it's resolved and up to date now.

Sami32:
It's why i did leave them in the code i write. For informing about that.

100% correct. Do you want some samples ?
https://en.wikipedia.org/wiki/High_Efficiency_Video_Coding
http://blog.mediacoderhq.com/h264-profiles-and-levels/
[Google Book](https://books.google.fr/books?id=R6vBBAAAQBAJ&pg=PA719&lpg=PA719&dq=profile+%22main+10%22+OR+%22main+4:2:2%22+%22high+10%22+OR+%22high+4:2:2%22&source=bl&ots=6Yx3OmLGyn&sig=AOl3nH7blb4hW0TVBIzee0Vznfs&hl=en&sa=X&ved=0ahUKEwjb-Zu29ZLTAhUCCBoKHbuOCrsQ6AEINTAH#v=onepage&q=profile%20%22main%2010%22%20OR%20%22main%204%3A2%3A2%22%20%22high%2010%22%20OR%20%22high%204%3A2%3A2%22&f=false)

Sami32:
Thank you very much :+1: :smile: 
Don't worry, i don't mind for the order or the name.

Nadahar:
@Sami32 Sorry, that was my fault - I forgot to include the examples and didn't see that they could have spaces in them. The regexes is not updated to :

H264:
```regex
^\s*(?:\w[^@]*@)?(?:L|LEVEL)?\s*([\db]+(?:\.\d+|,\d+)?)(?:@\S.*\S)?\s*$
```
H265:
```regex
^\s*(?:\w[^@]*@)?(?:L|LEVEL)?\s*(\d+(?:\.\d+|,\d+)?)(?:@\S.*\S)?\s*$
```

src/main/external-resources/renderers/Samsung-9series.conf
```diff
@@ -33,8 +33,8 @@ TranscodeAudio = WAV
 #MaxVideoWidth = 3840
 #MaxVideoHeight = 2160
 SupportedVideoBitDepths = 8,10
-H264Level41Limited = true
-#H265Level41Limited = true
+H264LevelLimit = 4.1
```
Sami32: In fact, it's depending of the resolution, 4.1 for HD and 5.1 level for UHD.

Since our DLNA is implementation is incorrect, most of the Samsung are not able to transcode without "hipcup", i choosen to set it to false before, but now 5.1 should be prefereable.

Sami32:
Why LEVEL ? is that a standard specification ?

src/test/java/net/pms/dlna/DLNAMediaInfoTest.java
```diff
@@ -0,0 +1,109 @@
+/*
+ * PS3 Media Server, for streaming any medias to your PS3.
+ * Copyright (C) 2012  I. Sokolov
```
Sami32: Old copy/haste ;)

Sami32:
In my code i leaved this other level sample as well: `Stereo High@L4.1 / High@L4.1`

And this case is not correctly handled.

src/main/external-resources/renderers/Samsung-9series.conf
Nadahar: @Sami32 I expected you to fix those details, I just replaced it with a setting correspondig to the old one.

src/test/java/net/pms/dlna/DLNAMediaInfoTest.java
Nadahar: @Sami32 Hehe, yes and no. I took an existing test class and duplicated and modified because I didn't want to have to set up a test class "from scratch" in case I forgot somthing.

Nadahar:
@Sami32 

> In my code i leaved this other AVC level sample as well: Stereo High@L4.1 / High@L4.1
> 
> And this case is not correctly handled here.

Both of those are handled individually. Are you saying that they should be parsed as one?

Nadahar:
> Why LEVEL ? is that in a standard specification ?

This isn't only designed to interpret the specification values, but user configuration values as well. Remember that it's case sensitive, which means that "Level 5.1", "level4.0" etc will work.

src/main/external-resources/renderers/Samsung-9series.conf
Sami32: Sure, i prefer explain now, as i know you'll ask me later why i did that ;)

Nadahar:
@Sami32 The case-insensitivity flag is set in Java, outside of the regex-string. Therefore, when you test the regex outside UMS, add ```(?i)``` in front to get the correct result, e.g:
```regex
(?i)^\s*(?:\w[^@]*@)?(?:L|LEVEL)?\s*([\db]+(?:\.\d+|,\d+)?)(?:@\S.*\S)?\s*$
```

Sami32:
No, i only mean that testing it on _Regex101_ website give me:
> Your regular expression does not match the subject string.

"Case sensitive" is not easily understandable to me, because i need to translate it to "not sensitive" or "unsensitive" for my french brain.

Sami32:
Sorry, that doesn't work either. Do that really work for you ?

Nadahar:
@Sami32 
Using this exact regex:
```regex
(?i)^\s*(?:\w[^@]*@)?(?:L|LEVEL)?\s*([\db]+(?:\.\d+|,\d+)?)(?:@\S.*\S)?\s*$
```
with this exact string: ```High@L1B``` works for me.

Sami32:
But that's work:
`(?:\w[^@]*@)?(?:L|LEVEL)?\s*([\db]+(?:\.\d+|,\d+)?)(?:@\S.*\S)?\s*$`

Actually, more i withdraw from you Regex, more it work ;)

Nadahar:
@Sami32 The exact regex and subject I pasted above works for me on regex101. There are small differences between the different flavours of regex. I use the perl/pcre flavour, which I believe matches Java's regex best of the choices there.

Sami32:
I don't know what to say, `Stereo High@L4.1 / High@L4.1` doesn't worked for me.
Anyway, if it's work for you that should be fine.

Nadahar:
@Sami32 The reason ```Stereo High@L4.1 / High@L4.1``` works when removing the start, is that you remove the requirement for the start of the string. That means it will match whatever is before the ```@``` - even if it's a complete book. It opens up for a lot of false positives.

I thought your said that ```Stereo High@L4.1 / High@L4.1``` wasn't meant to be parsed, are you saying it should be? It's a very strange specification imo.

Sami32:
I mean that the AVC level should be parsed, the first one is suffisant.

That part of the specification is related to 3D videos. There was only introduced in 2015.

Nadahar:
@Sami32 Parsing that "properly" will take a lot of redesign. The quick fix is to remove the ```$``` in the end, it means that anything can follow and that it's the first one that's parsed:
```regex
(?i)^\s*(?:\w[^@]*@)?(?:L|LEVEL)?\s*([\db]+(?:\.\d+|,\d+)?)(?:@\S.*\S)?\s*
```

Sami32:
That's say, these double entries are the MediaInfo way to indicate the main profile and the fallback one that are interleaved, imho.

Nadahar:
```regex
(?i)^\s*(?:\w[^@]*@)?(?:L|LEVEL)?\s*([\db]+(?:\.\d+|,\d+)?)(?:@\S.*\S)?\s*(?:/|$)
```
Makes it a little bit safer, but still works (use the python flavor).

src/main/java/net/pms/dlna/LibMediaInfoParser.java
```diff
@@ -175,8 +175,8 @@ public synchronized static void parse(DLNAMediaInfo media, InputFile inputFile,
 						}
 
 						value = MI.Get(video, i, "Format_Profile");
-						if (!value.isEmpty() && media.getCodecV() != null && media.getCodecV().equals(FormatConfiguration.H264)) {
-							media.setAvcLevel(getAvcLevel(value));
+						if (media.getCodecV() != null && isNotBlank(value)) {
```
Sami32: Calling that "avcLevel" when all the video profiles (mpeg-4 visual, 3gp...) was logged without distinction, was of no mean to me...

Nadahar: @Sami32 I didn't understand that - it there something wrong with the code?

src/main/java/net/pms/dlna/LibMediaInfoParser.java
```diff
@@ -645,9 +645,10 @@ private static void getFormat(StreamType streamType, DLNAMediaInfo media, DLNAMe
 			format = FormatConfiguration.BMP;
 		} else if (value.equals("tiff")) {
 			format = FormatConfiguration.TIFF;
+		// Value can look like "Advanced@L1", "Complex@L2", "MP@LL" or "Simple@L3" with VC-1 or MPEG-4 Visual at this point.
 		} else if (containsIgnoreCase(value, "@l") && streamType == StreamType.Video) {
-			media.setAvcLevel(getAvcLevel(value));
-			media.setH264Profile(getAvcProfile(value));
+		//TODO: Is this really needed, it's already set earlier..?
```
Sami32: Nice catch :+1: 
It's look good to be deleted to me.

Sami32:
I never noticed, or searched, that a Python version existed. Thanks for the tips, though i'm not able to handle the diffrence except that your code work on that one and not on the default one.
Is that the more similar to Java Regex use ?

src/main/java/net/pms/dlna/LibMediaInfoParser.java
Sami32: No.
I just tried to say that your appellation is more correct than the old one having nothing to do with AVC specific profiles, as all the video profiles are logged without any AVC filter.

Nadahar:
@Sami32 Non of them are exactly as Java, but "my regex tester" uses the pcre regex which has been equal to Java's regex in all regex'es I've made until now. But, I know there are some differences.

When it comes to regex101 it accounts for the qualificators. Java use ```"``` to qualify strings (and regex'es), which means that any ```"``` must be escaped with a ```\```. Perl however has a specific qualifier for regex, which is ```/```. Since I added a ```/``` to the regex it doesn't work unless it's escaped, but since we're using it in Java that doesn't apply. Therefore I used the Python flavor instead, which uses ```'``` as qualifier, and thus it works. Here is a match in Perl, as you can see ```/``` is used instead of ```"```:
```perl
if($mystring =~ m/World/)
```

"My" regex tester doesn't care about all that crap, it only processes the regex itself and there is no qualifier that needs escaping, which means it's easier to get them to work there. When "moving" a regex into Java however, all ```\``` must be escaped to ```\\``` because ```\``` is the escape character for both Java and regex.

Nadahar:
When it comes to regex syntax, there's no easy answer to what's the difference between the flavors. My experience is that, unless you go really advanced, why mostly behave the same. The only thing I've stopped using with Java is named groups. I've usually used named groups in other languages as it's more robust, but in Java I have to rely on the group number to extract the correct group. With named groups you can give the groups "names", so the only group here could be called ```level```. Then you would ask for the result of the ```level``` group instead of asking for group ```1``` as I do here. It makes no big difference, but using names is a bit "safer" when editing complex regex'es where groups can appear or disappear which will reindex the group numbers.

If you're interested, there are some information about the differences [here](http://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html#jcc) and [here](https://en.wikipedia.org/wiki/Comparison_of_regular_expression_engines#Language_features). I'm pretty sure you're not *that* interested though 😉

src/main/java/net/pms/dlna/LibMediaInfoParser.java
Sami32: Since no format filter is applied, as i did, and that the profile is only used for the database, AFAIK, using only that "@" will log again more video profiles.

What is really the purpose of the video profile ?

Nadahar: > Since no format filter is applied, as i did, and that the profile is only used for the database, AFAIK, using only that "@" will log again more video profiles.

Sorry, I don't understand.

> What is really the purpose of the video profile ?

The video profile is used to parse H264 and H265 level and profile for now. I imagine it can be used for other things as well in the future.

src/main/java/net/pms/dlna/DLNAMediaInfo.java
```diff
@@ -1743,12 +1751,9 @@ public String toString() {
 			if (isNotBlank(getMatrixCoefficients())) {
 				result.append(", Matrix Coefficients: ").append(getMatrixCoefficients());
 			}
-			if (isNotBlank(avcLevel)) {
-				result.append(", AVC Level: ").append(getAvcLevel());
+			if (isNotBlank(getVideoFormatProfile())) {
+				result.append(", Video Format Profile: ").append(getVideoFormatProfile());
```
Sami32: You appelation totally confuse me...
_Video Coding Level_ should be more important to log than the profile, that is the first part before the @

Nadahar: @Sami32 Maybe the name is confusing, but I used the term from MediaInfo which is ```format_profile```. It's the whole string, not just the profile or just the level. It's the complete string as returned from MediaInfo, e.g ```High@l4.1```.

src/main/java/net/pms/dlna/LibMediaInfoParser.java
Sami32: I guess one of us didn't understand correctly. It's look like you seing it as the whole string ? and me i see the "profile" as the first part before @ (@L in the case of that parsing)

Nadahar: @Sami32 That was the core of my change, instead of storing H264Level, H264Profile, H265Level and H265Profile (having to add yet more fields for other standards if needed) I simply store the unparsed string as returned from MediaInfo. Then, when one of the values above is needed, they are parsed from this string. I called it "format profile" because that's the name MediaInfo use, and I really didn't know a more suitable name. It's more general that when you think of, which is the "profile" part of H264 or H265.

src/main/java/net/pms/dlna/DLNAMediaInfo.java
Sami32: If it's the whole string that's make sens to me, but it's not what i saw, so i need to check it again.

src/main/java/net/pms/dlna/LibMediaInfoParser.java
Sami32: My bad, i misunderstood the condition string with the videoformatprofile value...i gues that mean that it's time to go sleep ;)

Nadahar:
@Sami32 For future reference, your original commit is 00002d07f19bb5b6c0d6e06dd358a36220c6a331.

Nadahar:
@Sami32 I've done some small updates based on our discussion last night.

Sami32:
https://github.com/UniversalMediaServer/UniversalMediaServer/issues/1170

Sami32:
Thanks. I never read anything related to that Github policy, but that seem logic though.

Nadahar:
@Sami32 

> Thanks. I never read anything related to that Github policy, but that seem logic though.

Now I don't understand - did we discuss a GitHub policy?

Sami32:
Yes ;)
Sorry, i should take habit to develop more for a better understanding.

Nadahar:
I think I read about the garbage collection on GitHub somewhere, but I can't find it again. It seems that the default TTL for an orphaned commit in git is 90 days. I can't find now anywhere about what GitHub use, but they might use the default..? When GC is run this is checked, and old, orphaned commits are cleared. It seems [GitHub run GC pretty much every day](https://twitter.com/githubhelp/status/387926738161774592?lang=fr),

Sami32:
Thanks for sharing. i checked in their help section and didn't found anything related about the period or "cleaning" process. Though that seem logic.

Sami32:
@Sami32
Wishing log some more video parameters, i discovered that it was only partially implemented in video engines, and also discovered that the video coding "profile" implementation didn't work at all. That's said it has not been used until now.
Anyway, since i intended to only log it, it should be done only after the new level and profile refactoring.

Nadahar:
@Sami32 Are you talking to yourself? 😉 

In any case, what happened to this? Did the H265 stuff we made ever get merged? I'm sorry, there are so many different "half done" projects now that I don't remember.

Sami32:
Yes, because i'm afraid to forget it.
And if i don't set my name, someone could think that i'm addressing to him ;)

No, it's in standby. I plan to work again on it, as the video issues already reported are probably linked to the incorrect/incomplet video level parsing ("unknow" level). 3D videos are also partially affected.

Since you've made a refactoring branch i didn't worked on that, but i'll need to check the point that you leave "suspension" and continue from your branch.
In fact it was when i was searching at your branch, that i'm still searching ;), that i worked on the subject that make me link here, and work on an other user issue met on the way.

SubJunk:
Can this be closed or will there be more work on it?

Nadahar:
@SubJunk There are some good improvements in here so I wouldn't delete it.

SubJunk:
Cool, is there anything I can do to help it progress?

Nadahar:
@SubJunk I don't remember exactly why or where it stranded, but you could study the approach and give a feedback on that. I think somebody will have to rebase it on latest master and resolve the conflicts if @Sami32 is going to be able to continue.

src/main/java/net/pms/dlna/DLNAMediaInfo.java
```diff
@@ -2438,48 +2443,74 @@ public void setReferenceFrameCount(byte referenceFrameCount) {
...
+	 * @return The AVC/H.264 profile for video stream or {@code null} if not
+	 *         parsed/relevant.
+	 */
 	public String getH264Profile() {
-		synchronized (h264ProfileLock) {
-			return h264Profile;
+		String profile = substringBefore(lowerCase(getVideoFormatProfile(), Locale.ROOT), "@l");
+		if (isNotBlank(profile)) {
+			return profile;
+		} else {
```
SubJunk: Just a formatting note, there's no need for the `else` here, or on line 2511, since it will return beforehand

src/main/java/net/pms/configuration/RendererConfiguration.java
```diff
@@ -1251,10 +1254,17 @@ public boolean isTranscodeAudioTo441() {
 	}
 
 	/**
-	 * @return whether to transcode H.264 video if it exceeds level 4.1
+	 * @return The H.264 level limit, if any, or {@code null}.
 	 */
-	public boolean isH264Level41Limited() {
-		return getBoolean(H264_L41_LIMITED, true);
+	public H264Level getH264LevelLimit() {
+		return H264Level.typeOf(getString(H264_LEVEL_LIMIT, null));
```
SubJunk: Maybe it should default to level 4.1?

SubJunk:
Ok, I have left some notes. Overall it looks like a good change

src/main/java/net/pms/dlna/DLNAMediaDatabase.java
```diff
@@ -63,15 +63,15 @@
...
+	private final String latestVersion = "9";
 
 	// Database column sizes
 	private final int SIZE_CODECV = 32;
 	private final int SIZE_FRAMERATE = 32;
 	private final int SIZE_ASPECTRATIO_DVDISO = 32;
 	private final int SIZE_ASPECTRATIO_CONTAINER = 5;
 	private final int SIZE_ASPECTRATIO_VIDEOTRACK = 5;
-	private final int SIZE_AVC_LEVEL = 3;
+	private final int SIZE_VIDEO_FORMAT_PROFILE = 40;
```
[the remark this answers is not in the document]
Nadahar: @SubJunk There seems to be a general misunderstanding of VARCHAR in our implementation. The old CHAR implementation allocates and use the space assigned to it. The is not the case for VARCHAR. These are allocated outside the main database table and only a pointer placed in the table. As such, they only use the space they need. The only point of restricting their size is to make sure they never get longer than that number. 

Thus, I prefer to define them too large and hopefully never have to deal with trucated strings.

valib:
@Sami32 why did you closed this PR? Will you continue on it? I rebased it to the latest master to be ready to merge after some requested changes asked by @SubJunk.

src/main/java/net/pms/configuration/RendererConfiguration.java
valib: @SubJunk agree that will be better to set the default value to level 4.1 and in this case the implemented logic in the DLNAResource line  976 must be changed to
 `} else if (media.isH264() && renderer.getH264LevelLimit() != H264Level.L4_1) {` 
and for all renderers where previously was set `H264Level41Limited = false` and this setting was deleted in this PR should be added `H264LevelLimit = 5.2`.

Nadahar: I'm not sure about this. Even though the old renderers were often limited to 4.1, is that really the case these days? I don't like to define defaults that isn't "logical", if no limit is defined for a given renderer it's most logical to me that there actually isn't a limit for that renderer.

src/main/external-resources/renderers/Sony-PlayStationVita.conf
```diff
@@ -21,7 +21,7 @@ SeekByTime = exclusive
 DLNALocalizationRequired = true
 TranscodeVideo = MPEGTS-H264-AAC
 TranscodeAudio = WAV
-H264Level41Limited = true
+H264LevelLimit = 4.1
```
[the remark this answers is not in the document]
Nadahar: @SubJunk Since this was introduced in 6.5.x (?) and the default set to 4.1, I assume that was without actually checking the capabilities of the renderers. If my assumption is correct, I see no reason to "preserve" this arbitrary value. As you can see from #1418 this has already created problems for users, and for every report I assume there are hundreds that doesn't report it. In my opinion, it would be better to define those limitations that has actually been verified and leave the rest undefined.

src/main/java/net/pms/configuration/RendererConfiguration.java
valib: @Nadahar I am not sure either but for me this is the safest way how to implement this PR. Users are using the UMS with the restriction to level 4.1 for the long time without complaining except users with renderers where was set `H264Level41Limited = false`. What we offer in this PR is only a different way how to decide which video will be streamed or transcoded.

Sami32:
@valib I don't remember, perhaps a bad mood...
Not right now, as i'm working on other stuff.

I think remember that something needed to be verified and i saw one that could be done more effectively.
This idea come from users having reported video level bugs:
https://github.com/UniversalMediaServer/UniversalMediaServer/issues/1170

src/main/java/net/pms/dlna/DLNAMediaDatabase.java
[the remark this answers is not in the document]
SubJunk: Oh cool, thanks for letting me know, I didn't know that

src/main/java/net/pms/configuration/RendererConfiguration.java
SubJunk: Maybe this isn't the case anymore but AFAIK most renderers support 4.1 nicely. As @ExSport said in https://github.com/UniversalMediaServer/UniversalMediaServer/issues/1417#issuecomment-346393516 it can cause real problems the other way too (and it did) so either way we will inconvenience some users.
I think 4.1 is a safer default than blank though because it's better to have transcoded video than no video.

valib: @SubJunk by the way we use the function `isVideoWithinH264LevelLimits` mentioned by @ExSport only in FFmpegVideo, MencoderVideo and TsMuxeRVideo for recognition if the video is correct to be muxed not during media parsing to distinguish if the media will be streamed or transcoded. I tried to use it also for parsing but it slows down the process because it calls the function `getAnnexBFrameHeader` where the FFmpeg is used to obtain the `h264_annexB` header and it takes long time.

Nadahar: @SubJunk I think it's better if playback fails than if transcoding is forced when not needed. If it fails, we have a chance to find out (because they post that it doesn't work) and correct the renderer configuration. If it transcodes when not needed, it's unlikely that the user will report it, and they will just be annoyed by the delay, the lack of seeking, the lack of audio and subtitles choices on the renderer and the quality loss. 

It's fine that most renderers support 4.1 - the problem is that many can take more.

valib: @Nadahar I didn't look at it from this angle and must say that you are right.

SubJunk: I don't agree with that, in fact our top competitor Plex aggressively transcodes on many renderers - they don't even have many specific renderer profiles. I think if a user has broken playback they will usually just uninstall us and go find something that works. Sometimes we get users reporting these things but that's not something to rely on.

Anyway I'm outvoted so go ahead :)

SharkHunter: I essentially only logged on to reduce my mail spam by unwatching, but couldn'thelp myself. This must be one of the stupidest things I've ever heard when working with software design. As @SubJunk  says normally if playback fails the user will just say "oh this doen't work". He won't write a bug report, since he will consider the program broken (failing to perform it's main task). He is more likely to write a report if seeking fails since that is a feature.

Nadahar: @SubJunk It's nothing that I feel that strongly about, I was just explaining my stance and why I did like I did. I've seen many examples on people being annoyed by needless transcoding. Some will probably turn away in either case, which is why this should be set correctly - not rely on the default.

One thing that also weighs in for me and that probably only is a "me thing" is that I feel it is "logically wrong" to set a default that is just a "random guess". I value "quality of information", so I like to make sure that whatever information I have is reasonably accurate, and prefer knowing that I don't know over using a guess that seems like real information to me. I realize that this is very likely to be something that is just important to me, and that the users won't care about this at all.

When it comes to Plex I don't really look at what they are doing. I don't consider Plex a "competitor" since it's payware. I tried the free version for a little bit a long time ago, and it annoyed the hell out of me with all it's lack of configuration/decisions being made for you. I know that for all the Apple users out there, this is appealing - but I think it's important to note what most companies seems to miss today as they rush to copy Apple - that this isn't the whole market. It's the most ignorant part of the market that appreciates having no choices, as they don't know how to make the decisions and rely 100% on the default behaviour. However, given how "difficult" UMS/PMS is and has always been, I doubt that's our current user base. If you just cater to the ignorant you will alienate the informed. That said, this is a general view of mine, that isn't particularly relevant to this decision, but it explains why I never look to what Plex does. For all I care, they cater to a different user base.

@SharkHunter Can I do anything to help you unwatch? I honestly can't wait to get rid of your mostly stupid and useless comments. I have a very hard time believing that this is one of the stupidest things you've ever heard, as you most often say more stupid things yourself.

Your argument doesn't really hold water - the default will "alienate" some users whatever choice is made. You're right that most users won't report, but the question is what will affect most users and how bad. If most renderers can play a higher level so that a lot of users will be impacted by needless transcoding and all the negative aspects with that, I'd say the average user experience is worse limiting to 4.1. If most renderers can't play a higher level, the situation is the opposite. In addition, my argument was that for those few that do report issues, it's much harder to understand what's wrong if the default is too low - so the chance that we will ever know about it and be able to correct it is much smaller.

SubJunk: My assumption is that most users don't even know what transcoding is. They are just installing a program that serves video to their device. If they don't see video on the device then our program is broken. If they see it, it works.
Our advanced users are capable of seeing that there is transcoding happening and maybe they don't like hearing the fan in the computer make more noise than usual or slower seeking, but as advanced users they can be the ones who submit the better renderer configs to us to reduce the transcoding.

So from my perspective we will make the most people happy by succeeding at playing video more often.

About the comment "given how difficult UMS/PMS is", you're right but it has always been my goal to make it easier for new users and we have made good progress on that. I hope we make more so that we can still offer the advanced features to opt-in to while "just working" for our Plex-lovers :)

SubJunk:
This was a good change

Nadahar:
@SubJunk Many good things have been made that has simply stranded because you don't respond/ignore them.

This was actually made 1 year ago now, and I sure don't remember anything of this now. I doubt @Sami32 does either. It shouldn't be hard to fix and merge this, but I suspect you'll have to do that yourself now.

SubJunk:
If you scroll up a bit you'll see I did respond to this and even did a code review last year...

Nadahar:
@SubJunk We made this between April 1 and April 8 2017. Your first comment here was 30 July 2017. I know that you've reviewed it since then, and that's all fine and well. I can't speak for @Sami32 and don't know exactly why he chose not to do anything further with that.

My point is that where there's no reaction or input if this is even wanted functionality for 4 months people have moved on. I don't know about @Sami32, but the way I work I have everything related to a certain piece of code/PR in my head when I work with it, and can easily make changes or answer questions about it. When some time has passed, this has been replaced by something else I've done and I no longer know why decisions were made or if there is something that still needs to be done or handled in some particular way.

Refreshing my memory so that I'm up-to-date on a months old PR means that I basically have to study my own code and "relearn" information which I already knew. I've had to do that over and over again here because interest/feedback is lacking, and I've come to a point where I'm not willing to do that anymore, when the reason simply is laziness from another person.

If this had been handled in a timely manner, that is within a couple of weeks after completion, I'm sure this would have been merged a long time ago. Now we're in the situation where somebody has to do the job of regaining overview over the changes, and I'm simply saying that I won't be me. Considering how silent @Sami32 has been on this, I assume that he feels the same way.

SubJunk:
I appreciate your frustration but I'm not here to, nor am I interested in, micromanaging the code. You guys have the ability to review and merge your own branches and I think you did back then too, and I encourage it for things that don't change huge parts of behavior of the program.

I'm not always around here but I have at least done my best to give you guys the ability to have autonomy, which is a hard thing to do for me because I love this project!

I'm not the perfect project leader but I think we have a good program, there have been some nice reviews coming in for the v7 release too. I think we should be proud of what we have accomplished since you guys started contributing code to UMS.

PR #1243, opened 2017-04-02
Improve DIDL-Lite protocolInfo for transcoded audio

Nadahar:
This is in no way a "proper" fix, it would require a huge refactoring. Much information is reported wrong, this removes information that we don't know (instread of sending 0 or blank which is explicitly against the DLNA standard). The information we send should be correct, or the information shouldn't be there.

It also fixes sampling rate and the number of channels sent for transcoded audio so that it more closely resembles the actual data. 

The biggest hurdle is the size - the estimation of size has some issues. In any case, while this doesn't eliminate sending false info, it greatly reduces the false info sent to the renderer.

### DIFF
--- a/src/main/java/net/pms/dlna/DLNAResource.java
+++ b/src/main/java/net/pms/dlna/DLNAResource.java
@@ -2476,45 +2476,62 @@ public abstract class DLNAResource extends HTTPResource implements Cloneable, Ru
 					}
 				} else if (getFormat() != null && getFormat().isAudio()) {
 					if (media != null && media.isMediaparsed()) {
-						addAttribute(sb, "bitrate", media.getBitrate());
-						if (media.getDuration() != null) {
+						if (media.getBitrate() > 0) {
+							addAttribute(sb, "bitrate", media.getBitrate());
+						}
+						if (media.getDuration() != null && media.getDuration().doubleValue() != 0.0) {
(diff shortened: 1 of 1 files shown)


Sami32:
That's said, adding `DLNA-ORG_CI=1` flag to transcoded audio should be even better ?

Sami32:
If you really want the correct duration / size you can do something like that:
`FFmpeg -y -i input -vn -ar 48000 -ac 2 NUL`   (or `/dev/null` for Mac and Linux OS)
And parse the duration and size from the output.

Nadahar:
@Sami32 I know that, but we can't do parsing at that stage (this is done every time a browse request is made). The way to fix this properly is to fix the parsing, so that for example FFmpeg is used to fill in values not parsed by MediaInfo - but that's a bigger task. The easiest (for us) would be if MediaInfo parsed it 😉 

There's no point in using the CI flag when we only offer one version. The whole point with the CI flag is to indicate which versions are transcoded and which aren't (like I've done for images in #1061). Telling the renderer that "this is transcoded so choose another version if you can" is meaningless when the transcoded version is the only one made available.

Sami32:
It's happen to deprecated format or very rarely used ones so to be honest i don't really bother to go further than was you've just done.
Until i meet an other issue, of course :wink: 

Thank for clarifying the CI flag, as i didn't understood it that way.

PR #1245, opened 2017-04-02
NPE fix

Sami32:
I don't know how i missed it :-(
With .act and .dsf files.


src/main/java/net/pms/dlna/DLNAMediaAudio.java
```diff
@@ -509,7 +509,7 @@ public String getAudioCodec() {
 	@Override
 	public String toString() {
 		StringBuilder result = new StringBuilder();
-		if (!getLang().equals("und")) {
+		if (!getLang().equals("und") && getLang() != null) {
```
Nadahar: @Sami32 You have to move the null check first, or you will still get NPE:

```java
if (getLang() != null && !getLang().equals("und")) {
```
They are checked in that order...

PR #1250, opened 2017-04-05
Add missing DIDL-Lite subtitle namespaces

Nadahar:
The DIDL-Lite XML is invalid when streaming subtitles because the ```pv``` and ```sec``` namespaces aren't declared. See comments in d22486b0439ba5d3ee66b005adb498513cebe026.  This will hopefully fix it, but is untested by me.

I added a small, unrelated commit to this PR instead of creating a new PR for it. Removing this repository should be perfectly safe as GSON is available through Maven central. For every repository we can remove, the builds become faster.


Sami32:
Thank you :+1:

PR #1253, opened 2017-04-08
Improve English in doc index

dandv:


Nadahar:
@dandv Thanks, I don't think anyone has paid any attention to those help files for many years.

PR #1254, opened 2017-04-08
Improve English in docs -> introduction

dandv:
These docs should be reviewed for out-of-date references, e.g.

* USB 1.1
* no mention of Plex (instead the discontinued MediaTomb is mentioned)
* wireless LANs being unable to stream full-HD videos

### DIFF
--- a/src/main/external-resources/documentation/introduction.html
+++ b/src/main/external-resources/documentation/introduction.html
@@ -61,7 +61,7 @@
 	<h2>Alternatives</h2>
 
 	<p>There are several streaming servers available which can be used
-		to communicate with a your viewing device. Here is a small selection:</p>
+		to communicate with viewing devices. Here is a small selection:</p>
 	<ul>
 		<li><a
 			href="http://www.microsoft.com/windows/windowsmedia/player/11/default.aspx">Windows
@@ -120,15 +120,15 @@
 		for more information.
 	</p>
 
-	<p>The most stunning features of UMS are transcoding and muxing.</p>
+	<p>The most important features of UMS are transcoding and muxing.</p>
 
 
 	<h2>What are transcoding and muxing?</h2>
 
 	<p>
-		If the video codec is not supported UMS can convert the video stream
-		to a device compatible codec. For example the PS3 is compatible with
-		MPEG2. This procedure is called <a
+		If the video codec is not supported by your viewing device, UMS can convert
+		the video stream to a device-compatible codec. For example, the PS3 is
+		compatible with MPEG2. This procedure is called <a
 			href="http://en.wikipedia.org/wiki/Transcoding">transcoding</a> and
 		it costs a lot of CPU power.
 	</p>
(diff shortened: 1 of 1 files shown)


src/main/external-resources/documentation/introduction.html
```diff
@@ -120,15 +120,15 @@
...
+	<p>The most important features of UMS are transcoding and muxing.</p>
 
 
 	<h2>What are transcoding and muxing?</h2>
 
 	<p>
-		If the video codec is not supported UMS can convert the video stream
-		to a device compatible codec. For example the PS3 is compatible with
-		MPEG2. This procedure is called <a
+		If the video codec is not supported by your viewing device, UMS can convert
```
Nadahar: "viewing device" is called renderer in DLNA context, and we use the same term everywhere else, so it should just as well be that here. This applies to audio meda as well for example.

src/main/external-resources/documentation/introduction.html
```diff
@@ -120,15 +120,15 @@
 		for more information.
 	</p>
 
-	<p>The most stunning features of UMS are transcoding and muxing.</p>
+	<p>The most important features of UMS are transcoding and muxing.</p>
```
Nadahar: 😊

src/main/external-resources/documentation/introduction.html
```diff
@@ -61,7 +61,7 @@
 	<h2>Alternatives</h2>
 
 	<p>There are several streaming servers available which can be used
-		to communicate with a your viewing device. Here is a small selection:</p>
+		to communicate with viewing devices. Here is a small selection:</p>
 	<ul>
```
Nadahar: Same as below, should be renderer "with your renderer"

src/main/external-resources/documentation/introduction.html
```diff
@@ -152,19 +152,19 @@
...
+		sufficient. When transcoding a full HD resolution video, an up-to-date
+		dual core processor from Intel or AMD running at at least 2.6 GHz is
 		necessary.</p>
 
-	<p>It is also necessary to have enough network bandwidth available
-		when streaming HD videos. It is nearly impossible to stream a full HD
-		video via WLAN, even when restricting network bandwidth and setting
+	<p>You should also have enough network bandwidth available when streaming HD
+		videos. It is nearly impossible to stream a full HD video via
+		wireless LAN, even when restricting network bandwidth and setting
```
Nadahar: This might not be entirely true anymore, any input @UniversalMediaServer/developers? Whouldn't it be more correct to say that it can be problematic to stream a full HD vidio via Wi-Fi, depending on your Wi-Fi standard and coverage.

src/main/external-resources/documentation/introduction.html
```diff
@@ -152,19 +152,19 @@
...
 		video quality very low.</p>
 
 	<p>You should also have enough free RAM, which means about 512 MB.
-		There is also some space needed on your systems hard drive for
-		buffering videos. 500 MB free disk space is enough. If you want to
-		stream media files from an external hard drive please consider that an
+		There is also some space needed on your hard drive for
+		buffering videos. 500 MB of free disk space is enough. If you want to
+		stream media files from an external hard drive please consider that a
 		USB 1.1 connection is too slow for videos.</p>
```
Nadahar: This should probably be USB 2.0

src/main/external-resources/documentation/introduction.html
Sami32: @Nadahar I totally agree, and it's not @SubJunk that will say the opposite :wink: 

@dandv That documentation part should be modified.

Nadahar:
@SubJunk I think you'll need to take some decisions here or this will probably just die out. After all, you are the only one that's properly familiar with English here.

src/main/external-resources/documentation/introduction.html
SubJunk: Yes that is outdated now, I stream HD on an AC network easily and it's not even the fastest kind of AC connection

src/main/external-resources/documentation/introduction.html
SubJunk: I've streamed videos from USB 2.0 nicely before so that bit seems accurate - it is only really USB 1x that usually wouldn't be capable of it

PR #1259, opened 2017-04-10
Legacy RealAudio parsing

Nadahar:
It turns out that neither MediaInfo nor JAudioTagger can parse the earliest RA formats. They have a very confusing versioning scheme where RealAudio 1.0 is RA version 3 and RealAudio 2.0 is RA version 4 and 5. Don't ask me how they managed to think that this was good versioning practice, but because we had trouble parsing these I figured I could quickly "throw together" a parser since the file formats seemed very simple to parse. It turns out it wasn't quite that simple because there's very little information available about these, and the little I found wasn't entirely accurate. 

In any case, I think this is working properly now so since I've already made it, we might as well use it..?

### DIFF
--- a/src/main/java/net/pms/util/AudioUtils.java
+++ b/src/main/java/net/pms/util/AudioUtils.java
@@ -19,22 +19,42 @@
  */
 package net.pms.util;
 
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.nio.channels.FileChannel;
+import java.nio.channels.ReadableByteChannel;
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+import java.util.Locale;
+import net.pms.PMS;
+import net.pms.configuration.FormatConfiguration;
 import net.pms.dlna.DLNAMediaAudio;
+import net.pms.dlna.DLNAMediaInfo;
+import net.pms.dlna.DLNAThumbnail;
+import net.pms.image.ImageFormat;
+import net.pms.image.ImagesUtil.ScaleType;
+import org.apache.commons.lang3.StringUtils;
 import org.jaudiotagger.tag.FieldKey;
 import org.jaudiotagger.tag.Tag;
+import org.jaudiotagger.tag.id3.ID3v1Tag;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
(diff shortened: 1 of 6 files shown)


SubJunk:
I haven't tested it but the code looks good to me

Sami32:
@Nadahar You are a damned killing machine :wink: 
I just tested it and it work great.

Though 2 minor issues occur:
- no MIME types (null)
- "realaudio" as audio codec in the log. I think you will agree that "RealAudio" should be more correct.

**Note**: That make think that still some human part stay in you :smile: 
Take it for what it is: a joke.

src/main/java/net/pms/util/AudioUtils.java
```diff
@@ -101,10 +121,254 @@ public static String getLPCMChannelMappingForMencoder(DLNAMediaAudio audioTrack)
...
+				if (version == 4) {
+					buffer.position(buffer.get() + buffer.position()); // skip interleaver id
+					fourCC = new byte[buffer.get()];
+					buffer.get(fourCC);
+				} else {
+					buffer.getFloat(); // skip deinterlace id
+					fourCC = new byte[4];
+					buffer.get(fourCC);
+				}
+				String fourCCString = new String(fourCC, StandardCharsets.US_ASCII).toLowerCase(Locale.ROOT);
```
Sami32: This is why it not displayed "RealAudio".
Why did you set the `toLowerCase` ?

Nadahar: Wrong, the reason why it's displayed in lower case is this: a4f837958f623c178d0132c3d24509e54f87c371

What I set to lower case is the "fourCC" code. The reason for doing that is to be sure that I get a hit in the following switch statement independent of the case used in the fourCC code in the file.

Sami32: OK, but i was right for the `toLowerCase` at least ? :flushed:

Nadahar:
@Sami32 I saw the missing mime type when I tested, but I though that it would be done later (in ```postParse()``` so I didn't care about it. However, I forgot to ever call ```postParse()``` 😊 

This should take care of it: https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1259/commits/dc3b4b45aedcb3876c87f5215f6f25054319ee71#diff-6f9c57a374990c14f75e43d3dd15b5deR388

PR #1266, opened 2017-04-20
Set reporting for maven site plugin.

valib:
The latest Maven version 3.5 reports wrong configuration of the maven site plugins.

### DIFF
--- a/CheckStyle.xml
+++ b/CheckStyle.xml
@@ -7,18 +7,20 @@
 <!--
     Checkstyle-Configuration: UMS
     Description:
-Checkstyle configuration that checks the sun coding conventions.
+Checkstyle configuration for Universal Media Server.
 -->
 <module name="Checker">
   <property name="severity" value="warning"/>
-  <property name="fileExtensions" value="java, properties, xml"/>
+  <property name="fileExtensions" value="java, xml"/>
+  <module name="SuppressWarningsFilter" />
   <module name="TreeWalker">
+    <property name="tabWidth" value="4"/>
     <module name="RegexpSinglelineJava">
       <property name="format" value="^\t*(?! \*) "/>
       <property name="message" value="Indent must use tab characters"/>
       <property name="ignoreComments" value="true"/>
     </module>
-    <property name="tabWidth" value="4"/>
+	<module name="SuppressWarningsHolder" />
     <module name="JavadocMethod">
       <property name="scope" value="protected"/>
     </module>
@@ -29,6 +31,7 @@ Checkstyle configuration that checks the sun coding conventions.
       <property name="scope" value="protected"/>
     </module>
     <module name="JavadocStyle">
+      <property name="endOfSentenceFormat" value="([.?!:][ \t\n\r\f&lt;])|([.?!:]$)"/>
       <property name="tokens" value="INTERFACE_DEF,CLASS_DEF,METHOD_DEF,CTOR_DEF"/>
     </module>
     <module name="ConstantName"/>
@@ -118,7 +121,6 @@ Checkstyle configuration that checks the sun coding conventions.
     <module name="InterfaceIsType"/>
     <module name="VisibilityModifier">
       <property name="protectedAllowed" value="true"/>
-      <property name="allowPublicFinalFields" value="true"/>
       <property name="allowPublicImmutableFields" value="true"/>
     </module>
     <module name="ArrayTypeStyle"/>
--- a/pom.xml
+++ b/pom.xml
@@ -34,11 +34,6 @@
 	<url>http://www.universalmediaserver.com/</url>
 	<inceptionYear>2012</inceptionYear>
 
-	<!-- this is not required, but is recommended for plugin version checks -->
-	<prerequisites>
-		<maven>2.2.1</maven>
-	</prerequisites>
-
 	<organization>
 		<name>Universal Media Server</name>
 		<url>http://www.universalmediaserver.com/</url>
@@ -668,6 +663,10 @@
 						<configuration>
 							<rules>
 								<dependencyConvergence/>
+								<!-- this is not required, but is recommended for plugin version checks -->
+								<requireMavenVersion>
+									<version>3.0.0</version>
+								</requireMavenVersion>
 							</rules>
 						</configuration>
 						<goals>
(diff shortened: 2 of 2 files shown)


Sami32:
Why disable Checkstyle plugin ?

valib:
@Sami32 it produses very long html file about 18 MB because it checks not only the Java and xml files but also the properties with the translations. I actually changed it and use a slightly modified the CheckStyle.xml introduced by @Nadahar in the #1266 and now the Checkstyle plugin is working well.

pom.xml
```diff
@@ -895,6 +780,131 @@
...
+				<groupId>org.codehaus.mojo</groupId>
+				<artifactId>findbugs-maven-plugin</artifactId>
+				<version>3.0.4</version>														
+			</plugin>
+
+			<!-- Checkstyle report -->
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-checkstyle-plugin</artifactId>
+				<version>2.17</version>
```
Sami32: For that specific case, is that `${mavenCheckstylePluginVersion}` instead of a nominative version, will make the POM more easily maintenable without having any secondary negative effects ?

Though perhaps it could be extended to all report plugins without negative consequences ?

Sami32:
@valib Please, don't take it personally.
I just tried to clean it, hoping that i didn't mess up things, to avoid unnecessaty merge while conflicts appear.
That's said, as i'm not too familiar with Git, it'll better if you have a look at it :flushed: 
I checked 2 times but never sure...

Nice job by the way :+1:  though i only review the POM for now.

CheckStyle.xml
```diff
@@ -118,7 +118,6 @@ Checkstyle configuration that checks the sun coding conventions.
     <module name="InterfaceIsType"/>
     <module name="VisibilityModifier">
       <property name="protectedAllowed" value="true"/>
-      <property name="allowPublicFinalFields" value="true"/>
       <property name="allowPublicImmutableFields" value="true"/>
```
Nadahar: @valib Why did you remove this?

CheckStyle.xml
```diff
@@ -118,7 +118,6 @@ Checkstyle configuration that checks the sun coding conventions.
     <module name="InterfaceIsType"/>
     <module name="VisibilityModifier">
       <property name="protectedAllowed" value="true"/>
-      <property name="allowPublicFinalFields" value="true"/>
```
Nadahar: @valib Why did you remove this?

valib: @Nadahar because Maven doesn't know that and causes error..

Nadahar:
I just gave this a test and it seems to work, but: How do you run it? I was assuming it was meant to run with ```mvn site```, and it does - but it takes ages. It took me over 5 minutes to run. The reason seems to be that the build (and in particular the antrun) is launched 4-5 times, at one stage it even builds the installer. 

I see that this PR doesn't cause this as master does the same, but I can't understand that this is supposed to happen. It can't be a required to build multiple times to generate the reports..?

I think this might be due to more problems in our POM, and it would be nice if that could be fixed as well.

valib:
@Nadahar I have no idea why is build repeated during the site report generation. It is doing that long time ago.

Nadahar:
@valib I'm pretty sure there is something wrong with lifecycles and goals. I haven't figured out exactly what yet, [this](http://stackoverflow.com/a/2066145) and [this](http://stackoverflow.com/a/35610377) give some helpful commands to try to figure it out.

```powershell
mvn help:describe -Dcmd=package
```

```powershell
mvn fr.jcgay.maven.plugins:buildplan-maven-plugin:list
```

```powershell
mvn fr.jcgay.maven.plugins:buildplan-maven-plugin:list-phase
```

```powershell
mvn fr.jcgay.maven.plugins:buildplan-maven-plugin:list-plugin
```

```powershell
mvn fr.jcgay.maven.plugins:buildplan-maven-plugin:list-plugin -Dbuildplan.tasks=site
```

valib:
@Nadahar unfortunatelly I am not the one who can solve it.

Sami32:
And what about that one ?
capture_04222017_022549

Sami32:
For Eclipse users:
```logtalk
[WARNING] The POM for org.eclipse.m2e:lifecycle-mapping:jar:1.0.0 is missing, no dependency information available
[WARNING] Failed to retrieve plugin descriptor for org.eclipse.m2e:lifecycle-mapping:1.0.0: Plugin org.eclipse.m2e:lifecycle-mapping:1.0.0 or one of its dependencies could not be resolved: Failure to find org.eclipse.m2e:lifecycle-mapping:jar:1.0.0 in https://oss.sonatype.org/content/repositories/snapshots was cached in the local repository, resolution will not be reattempted until the update interval of ossrh has elapsed or updates are forced
```

valib:
@SubJunk it is finally up to you to decide which PR will be merged. There are a lot of PRs who are waiting to be merged and there is nobody who takes care about them.

Sami32:
@valib Nice improvements :+1:

_Maven_ still complain about prerequisites, also after that you had implemented `maven-enforcer-plugin`.
https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1266?_pjax=%23js-repo-pjax-container#issuecomment-296333280

Nadahar:
@valib It's impossible to approve when you don't say when you're done adding to a PR. If you tell me when you're done and rebase/squash some of the commits, I can make a review.

Nadahar:
@valib Thanks for organizing it before merge 👍

Nadahar:
@valib It seems like [some problems](http://www.universalmediaserver.com/forum/viewtopic.php?f=14&t=9588) have arisen from this PR - I can't be sure it's because of this, but it seems likely. Will you take a look?

valib:
@Nadahar thanks for info. I recommended some solutions to user.

Nadahar:
@valib As far as I can see the reason the build fails [in this thread](http://www.universalmediaserver.com/forum/viewtopic.php?f=14&t=9588) is that you've enabled the ```dependencyConvergence``` rule in f2c1099134cc16484a1cec956877bdc71efde357. Although I don't know why this fails only on some configurations, I don't understand why we'd want to enforce that. We have no control over which dependencies our dependencies use, why would we want the build to fail if some dependencies use different versions of the same dependency? 

https://maven.apache.org/enforcer/enforcer-rules/dependencyConvergence.html

Why did you add this rule?

PR #1268, opened 2017-04-28
Adjustments to Improve English in docs -> introduction (#1254)

Nadahar:
This PR is so we can make some adjustments to #1254 before merging this into master.


SubJunk:
That change to the wireless part looks good. Maybe it can also give a rough idea with something like:

```
Wi-Fi standards prior to IEEE 802.11n will usually struggle with HD streaming, though this is heavily dependent on levels of compression and codec choices. In general, IEEE 802.11ac networks are capable of HD video streaming if the signal at the receiving end is strong.
```

Any input on that is welcome

Nadahar:
@SubJunk Sure, I just tried to generalize it. I installed AC at my parents, and with two walls and 6-7 meters between the AP and the NIC there was no speed difference from N. The reason AFAICR is that AC only works at 5GHz, and that low wavelength signals lose their strength very quickly. The 5GHz signal got so weak because of those two walls that the NIC switched to 2.4 GHz (which is "N").

I'm sure AC is fast if within the same room or if the walls are very thin/easily penetrable, but that means it's very hard to give any general guidelines for what content can travel over what standard. All AC equipment also supports B/G/N and will downgrade as needed, so it can be very hard for people to actually know what they are getting. The marketing typically doesn't mention this, and it can be hard to find any "real" information about network speed on a computer.

I think there is so much hype and so many factors when it comes to WiFi speeds that I consider it impossible to make any assumptions of actual bandwidth, so I simply wanted to generalize as much as possible. If you think it's "safe to say" anything more specific, just add a commit to this PR.

SubJunk:
I agree the speed varies a lot so it's hard to say. Does anyone else have an opinion on it? @UniversalMediaServer/developers

Nadahar:
@SubJunk I don't think anyone else has much of a view, remember that you're the only native English speaker among us. Just change it to whatever you like, it's not important to me if it says this or that, I just explained why I wrote what I wrote..

PR #1269, opened 2017-04-29
Removed unused maven plugins and update plugins version

valib:
It saves almost 1 MB of the UMS instalation size.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -87,7 +87,7 @@
 		<maven-javadoc-plugin-version>2.10.4</maven-javadoc-plugin-version>
 		<git-commit-id-plugin-version>2.2.2</git-commit-id-plugin-version>
 		<maven-assembly-plugin-version>3.0.0</maven-assembly-plugin-version>
-		<logback-version>1.1.8</logback-version>
+		<logback-version>1.1.11</logback-version>
 
 		<!--
 			 use the Windows makensis.exe for Windows builds unless
@@ -362,11 +362,6 @@
 			<version>4.2.2</version>
 		</dependency>
 		<dependency>
-			<groupId>org.jdom</groupId>
-			<artifactId>jdom2</artifactId>
-			<version>2.0.6</version>
-		</dependency>
-		<dependency>
 			<groupId>org.fourthline.cling</groupId>
 			<artifactId>cling-core</artifactId>
 			<version>2.1.1</version>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -377,21 +372,6 @@
 			<version>2.1.1</version>
 		</dependency>
 		<dependency>
-			<groupId>org.seamless</groupId>
-			<artifactId>seamless-http</artifactId>
-			<version>1.1.1</version>
-		</dependency>
-		<dependency>
-			<groupId>org.seamless</groupId>
```
Nadahar: This isn't unused for long, I'm currently using it in a PR I'm working on.

valib: @Nadahar actually it is not used. If you use it in your PR than it should be added with your commit or you can modify this PR.

Nadahar: @valib Many of the classes used by Cling is from Seamless, it's by the same author and Seamless is the his "common code" used both in Cling and other projects. Debugging Cling will also get harder without access to this source code, so I think they should stay. To save space in the compiled binary, using the shade plugin would be a better solution.

Nadahar:
@valib I reversed the seamless removals and upgraded LogBack and related code.

valib:
@Nadahar great. Tested and it is working.

Nadahar:
@valib I see that there's some [major changes](http://site.icu-project.org/download/59) in ```icu4j``` version 59, and I haven't looked into what consequences that would have for UMS:

> ICU 59 upgrades to emoji 5.0 data, together with segmentation and bidi updates from Unicode 10 beta. The Java code for number formatting has been completely rewritten for reliability and performance. There is also a new case mapping API for styled text, and a technology preview of enhanced language matching.
> 
> The source code repository has been reorganized, creating a combined trunk with icu4c and icu4j (and tools) folders. (#12800)
> 
> There are major changes for ICU4C that require changes in projects using ICU.
> 

I didn't approve this because I don't know how to verify how or if this affects UMS, even though the biggest changes are for ```icu4c``` big refactorings can often create new bugs.

I'll approve this PR with the given reservation: You'll take responsibility for checking that we're not affected by the ```icu4j``` changes.

valib:
@Nadahar we use only a tiny part of the ICU4J as I introduced in #988 and there is also a way how to made a jar which include only what we need and reduce the UMS size but I personally have no idea how to do that.

Nadahar:
@valib Ok, than I guess ICU4J isn't much of a risk to upgrade.

PR #1282, opened 2017-05-19
Update renderers selection to notify users to restart server

valib:


Nadahar:
@valib If this makes enabling renderers only requiring a "server restart", we should change it so that it changes the restart server icon. Also, I don't think we need to say both things, "server restart" should suffice as application restart is implicit then.

valib:
@Nadahar good idea.

valib:
@Nadahar is this acceptable?

Nadahar:
@valib I tested it and it didn't work properly. It only notified the users if some renderers were selected, not if none or all were. It also didn't care if the list had actually been changed, as long as "OK" was pressed.

I think I've fixed it, but it should be tested before merge.

valib:
@Nadahar thanks for your input.

Sami32:
I just saw that changing the default renderer, was not affected by that change: a restart was not "asked"/suggested.
Maybe it's expected to be like this because we force it, but i'm wondering...

Nadahar:
@Sami32 @valib I think changing the default renderer was forgotten.

Issue #1283, opened 2017-05-23
NPE in newgui.StatusTab


ezpuzz:
I'm receiving the following stacktrace when connecting with BubbleUPNP.
UMS is running on Windows 10 / Java 7

```
TRACE 2017-05-23 14:17:00.843 [New I/O worker #7] Received on socket: User-Agent: BubbleUPnP UPnP/1.1
TRACE 2017-05-23 14:17:00.843 [New I/O worker #6] Received on socket: Connection: Keep-Alive
TRACE 2017-05-23 14:17:00.843 [New I/O worker #7] Received on socket: Accept-Encoding: gzip,deflate
TRACE 2017-05-23 14:17:00.843 [New I/O worker #6] Received on socket: User-Agent: BubbleUPnP UPnP/1.1
TRACE 2017-05-23 14:17:00.843 [New I/O worker #6] Received on socket: Accept-Encoding: gzip,deflate
DEBUG 2017-05-23 14:17:00.843 [New I/O worker #7] Reading MediaInfo: true (default: false)
DEBUG 2017-05-23 14:17:00.844 [New I/O worker #6] Reading MediaInfo: true (default: false)
DEBUG 2017-05-23 14:17:00.844 [New I/O worker #7] Adding status button for Unknown renderer
DEBUG 2017-05-23 14:17:00.844 [New I/O worker #6] Purging renderer Unknown renderer as invalid
DEBUG 2017-05-23 14:17:00.845 [AWT-EventQueue-0] Deleting renderer Unknown renderer
DEBUG 2017-05-23 14:17:00.849 [New I/O worker #7] Marking renderer "Unknown renderer" at /192.168.1.4 as unrecognized
DEBUG 2017-05-23 14:17:00.849 [New I/O worker #6] Adding status button for Unknown renderer
TRACE 2017-05-23 14:17:00.850 [New I/O worker #7] Using default media renderer: Unknown renderer
INFO  2017-05-23 14:17:00.850 [New I/O worker #7] Media renderer was not recognized. Possible identifying HTTP headers:User-Agent: BubbleUPnP UPnP/1.1
DEBUG 2017-05-23 14:17:00.860 [New I/O worker #6] Marking renderer "Unknown renderer" at /192.168.1.4 as unrecognized
DEBUG 2017-05-23 14:17:00.860 [New I/O worker #7] Adding status button for Unknown renderer
[...]

PR #1284, opened 2017-05-24
Fix NPE in RendererItem constructor if address is null (#1283)

Nadahar:
This is a simple NPE fix for a rare situation that still can occur - see #1283.


Nadahar:
@Sami32 Ty 👍

PR #1285, opened 2017-05-24
Handle audio in generic containers

Nadahar:
This is an issue I've wanted to fix for a long time. This lets UMS, in a "general way", handle the fact that certain container formats can be either video or audio media depending only on the content of the container.

There's also a couple of bugfixes and formatting in here.


src/main/java/net/pms/dlna/DLNAMediaInfo.java
```diff
@@ -86,6 +88,29 @@
...
+	static {
+		Map<String, String> mutableAudioOrVideoContainers = new HashMap<String, String>();
+
+		// Map container formats to their "audio variant".
+		mutableAudioOrVideoContainers.put(FormatConfiguration.MP4, FormatConfiguration.M4A);
+		mutableAudioOrVideoContainers.put(FormatConfiguration.MKV, FormatConfiguration.MKA);
+		mutableAudioOrVideoContainers.put(FormatConfiguration.OGG, FormatConfiguration.OGA);
+		mutableAudioOrVideoContainers.put(FormatConfiguration.RM, FormatConfiguration.RA);
+		mutableAudioOrVideoContainers.put(FormatConfiguration.MPEG1, FormatConfiguration.MPA);
+		mutableAudioOrVideoContainers.put(FormatConfiguration.MPEG2, FormatConfiguration.MPA);
```
Sami32: I never told you, but perhaps you already saw it in my code: WEBM audio has been forgotten.

src/test/java/net/pms/configuration/FormatConfigurationRegressionTest0.java
```diff
@@ -10,966 +10,699 @@
...
 
-    java.lang.String str0 = net.pms.configuration.FormatConfiguration.AAC;
+  @Test
+  public void testAACMain() throws Throwable {
+    java.lang.String str0 = net.pms.configuration.FormatConfiguration.AAC_MAIN;
 
     // Regression assertion (captures the current behavior of the code)
-    org.junit.Assert.assertTrue("'" + str0 + "' != '" + "aac"+ "'", str0.equals("aac"));
-
+    org.junit.Assert.assertTrue("'" + str0 + "' != '" + "aac-main"+ "'", str0.equals("aac-main"));
```
Sami32: Honestly, i never say it, but i never understood that point. What use for ?
For me it's just look as a dead appendice, as MediaInfo don't recognize it ? So that will be never used. Though i can be wrong.

Sami32:
Though i don't think that you SIPRO and ATRAC implementation is efficient, correct.
I could push a PR later if you don't mind ?

src/test/java/net/pms/configuration/FormatConfigurationRegressionTest0.java
Nadahar: @Sami32 You're so focused on MediaInfo, I'm not. I think more general and doesn't limit the logic to one specific parser. That said, I don't remember why I added the "main" profile and it might not be relevant.

Nadahar:
@Sami32 I think we should try to avoid merging stuff we know is wrong, it would be better if you pushed a commit to this branch.

src/test/java/net/pms/configuration/FormatConfigurationRegressionTest0.java
Sami32: @Nadahar Are you searching trouble ? Or it's just an other way to say that you're always right and when not we need to report to the first article ?
Maybe because we call it `LibMediaInfoParser.java` for a reason ?
But if you prefer i can refer to GSpot lookup tables ?

Sami32: You seem take it personnally the most little comment that is not a typo. That's said, that could be only my incorrect perception.

Nadahar: @Sami32 
> Are you searching trouble ? Or it's just an other way to say that you're always right and when not we need to report to the first article ?

I don't understand this except that it's something about me "always having to be right".

> Maybe because we call it LibMediaInfoParser.java for a reason ?

I can't see that I've used ```aac-main``` in ```LibMediaInfoParser``` at all.

> But if you prefer i can refer to GSpot lookup tables ?

I don't know what I've done this time, but it's obvious that you want to "get back at me" for something. It would be easier if you'd tell me what that is.

> You seem take it personnally the most little comment that is not a typo, and still it's limit... That's said, that could be only my incorrect perception.

This I also get that is some kind of "revenge", but I don't understand what you mean. In particular ```and still it's limit...``` is Greek to me.

Sami32: @Nadahar Nothing about revenche.
> You're so focused on _MediaInfo_

I didn't see any basement to your remark, so it need to be something else ?

Sami32: > I can't see that I've used aac-main in LibMediaInfoParser at all.

Sarcastically, i'll say that you're not suffisantly focused on _MediaInfo_, as `aac-main` will stay a dead part in the code :P
That's said, you can keep it if you like, i don't bother, it was just a review to be sure that you was aware of...

Sami32:
> it would be better if you pushed a commit to this branch.

If i didn't it's because it was not pushed yet, and you normally answer such remark by: i'm working on it's not yet reviewable
So i prefer wait, and still, i don't feel confident to push anything into your PR.

src/test/java/net/pms/configuration/FormatConfigurationRegressionTest0.java
Nadahar: @Sami32 When I make something I don't make it thinking like that. I separate concepts that I think shouldn't be reliant on each other. MediaInfo is on possible parser, FFmpeg is another, for audio files so is JAudioTagger. I made a parser for Real Audio, and we might find another external or make another internal parser in the future. 

In my experience software that doesn't assume a lot about what other parts is doing work best. The parser should do it's part: parse at much and as accurate as it can. Other code should do its part, in this situation handle what the parsers might output.

Don't forget that there's many renderer configurations that doesn't even use MediaInfo. I think that blindly focusing on what MediaInfo outputs is a mistake.

But, as I said above, I don't remember my reason for including ```aac-main``` now, as it's been too long since I did it. It might have been that I saw that it could be a parsing result (maybe RealAudio can have it?), it could be that I thought it should be there for "logical reasons" or it could be that I followed some line of thought that turns out to be irrelevant. The problem is that I don't remember.

I didn't mean to say that it has to stay, I'm actually unsure as I can't remember the reason for including it, I just wanted to say that what MediaInfo can parse or not isn't the deciding factor for me. It seems to be for you, since you use the fact that MediaInfo can't parse it as an argument for why it will never be used.

Nadahar:
@Sami32 The problem for me has been when I have a lot of changes locally and there's pushed something that is in conflict with what I've been doing (not "git conflict" - but actually doing something a different way). When we communicate about it there's no problem. I currently doesn't have any unpushed work in this branch, I will have to adress some of the things you've pointed out above, but I haven't started that yet and I can't really see how that would conflict with you pushing fixes to SIPRO and ATRAC anyway - so please do.

src/main/java/net/pms/dlna/DLNAMediaInfo.java
Nadahar: Fixed, I think. I don't know if it's correctly parsed etc, I've just added it to the "dual container" handling.

Nadahar:
@Sami32 Could you either push or tell me what to change for SIPRO and ATRAC? I don't remember, there's too much stuff going through my head...

Sami32:
I was not in the mood, and when i tried i discovered MediaInfo error that i reported and fixed, so on...time go but it's OK i'll do, if you can wait.

For now only ATRAC part is done.

A flaw in your code will need to be solved, but i'll report it later.

Sami32:
Of course, you can remove, merge or modify it.
It just my very modest suggestion, contribution.

A i planned to implement ACELP, well supported by renderers also if not specified (Your included ;) as if i remember well you've a WAV with it inside) and it not only related to RM/RA.
I leave the Sipro for the 2 that i don't know about, and don't have any samples.

Sami32:
The serious issue that i saw, from my point of view, is that some audio files are still recognized as videos.

src/main/java/net/pms/dlna/LibMediaInfoParser.java
```diff
@@ -628,11 +632,10 @@ private static void getFormat(StreamType streamType, DLNAMediaInfo media, DLNAMe
...
 		} else if (value.contains("mlp")) {
 			format = FormatConfiguration.MLP;
+		} else if (value.equals("openmg")) {
+			format = FormatConfiguration.ATRAC;
 		} else if (value.contains("atrac")) {
 			format = FormatConfiguration.ATRAC;
-			if (media.getCodecV() == null && !media.getContainer().equals(FormatConfiguration.RM)) {
-				media.setContainer(FormatConfiguration.ATRAC);
-			}
 		} else if (value.equals("jpeg")) {
```
Nadahar: @Sami32 Are you sure it's safe to remove that? It's probably there for a reason.

Sami32: Actually you did that because you wanted separated the "pure" Atrac (Sony) from the RealMedia ones. But other formats can have Atrac as well, so "openmg" do that nicely, without RM limitation.
Sony Atrac are handled by "openmg".
I've many RM or Sony Atrac tests files if you want test. I already test that in 2 of my PRs and it has always worked fine.

Do you have some idea where it can fail ?

Sami32: I hesitate to do at first, but i just verify, and you can change the "contains" by "startWith" for "atrac"

Nadahar: @Sami32 The above code isn't mine at all, I just added "RM": https://github.com/UniversalMediaServer/UniversalMediaServer/commit/81680c3f398a3d3fd032e1987f92a011c2ffc063#diff-6ecbf84dbeada135aeff5865473cf43eR330

What this does is to set the *container* to ATRAC if the *audio codec* is ATRAC. That breaks video files with ATRAC audio (if that exists) and it breaks the RM files. I don't remember if there's a "pure" ATRAC file format, but if it is removing this will no longer set their container correcly.

Sami32: Sorry, i thought it was your code.
But that doesn't change the fact that it was not correctly handle, as other format, as WAV, having Atrac codec was considerate as Atrac container !
Anyway, do as you like. I'mm report error later, when you'll have merged :P

Nadahar: @Sami32 I think you misunderstood me. I'm not saying we shouldn't change it, I'm just saying that we must make sure that changing it doesn't break anything.

I'm not really sure if ATRAC is only an audio codec or also a file format, but if it's also a file format how will the correct "container" be set without this code?

Sami32: Yes, you're, as you've .oma and .aa3 test files...

This code care for the Atrac container:
```
} else if (value.equals("openmg")) {
+			format = FormatConfiguration.ATRAC;
```

Sami32:
While trying to implement SSR profile, until he refactor the AAC part as he told me, i saw that the only way for now, as for your "Main" profile, is to use the Matroska flag; so, yes, it'll be only limited to that case, but can easily be extended later.
So, if you decide to keep it, you should, IMHO, add their detection ("`a_aac/mpeg2/main`" and "`a_aac/mpeg4/main`") in _LibMediaInfoParser.java_

Nadahar:
@Sami32 I've already removed the "main" profile as I couldn't remember why I added it and couldn't find an obvious reason to have it there. That said, I don't think it's widely supported. As I understand it it's more "complicated" than the LC profile, and it's like the "first version" of LTP. Support is said to be very limited and probably only in old AAC implementations.

I think it's strange as well, that it should be the "basic" AAC profile, but it seems that is LC in "reality".

src/main/java/net/pms/dlna/LibMediaInfoParser.java
Nadahar: @Sami32 Does that set the *container*? Looking in MediaInfo it seems it does. I guess you're right then, it should be ok.

Sami32: Sorry, but when you do that, i all the time think you are playing with my nerves and that make me feel...
I've difficulty to understand that you cannot see what a simple guy as me see as evidence.

Nadahar: @Sami32 Please think of that I do so many different things that I cannot remember all the details. Or, maybe it's just early Alzheimer's - I don't know, but often when I see a case, it's like I "see it for the first time". I don't remember much from I wrote this PR for example, I've written over 20000 lines of unrelated code in the meanwhile.

Nadahar:
@Sami32 Can I have this file ```F:!MEDIA_TEST_FILES\AUDIO\AAC-ALAC-ALS-SLS-SSR\MP4-SLS\sls2100_aot02_048_16.mp4``` so I can see what goes wrong?

Sami32:
Sure.
Sorry, i thought that you did understood that was not this specific file only but all the audio files having more than one track (MP4, OGA, MKA...) that are not recognized as audio by your code.
[sls8100_pce_aot02_048_16.mp4.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/1032159/sls8100_pce_aot02_048_16.mp4.txt)

Nadahar:
@Sami32 No, I didn't see that. But, I can explain why that happens. The "definition" I use for it to be an audio files is that it only contains *one* stream and that stream is an audio track.

I've seen that definition other places as well. Would an audio player even handle a multi track file?

Sami32:
Yes, you've an official, by standard, example.
That's said, it's also the purpose of MKA, i suspect that it's the same for OGA but i'll need to check at their specifactions.
https://en.wikipedia.org/wiki/MPEG-4_SLS
> MPEG-4 SLS allows for having both a lossy layer and a lossless correction layer similar to Wavpack Hybrid, OptimFROG DualStream and DTS-HD Master Audio, allowing for backwards compatibility to MPEG AAC-compliant bitstreams. MPEG-4 SLS can also work without a lossy layer (a.k.a. "SLS Non-Core"), in which case it will not be backwards compatible,[4] Lossy compression of files is necessary for files that need to be streamed to the Internet or played in devices with limited storage.

Nadahar:
@Sami32 Are you sure this means different *streams*? They refer to it as "layers", I interpret that like the two are embedded within the same stream..?

Sami32:
For SLS, it's how it happen, though usually we only have layer without a different stream.
The other kind is called nocore and they have only one stream.

Nadahar:
@Sami32 I've started implementing SLS support (it's parsed correctly now as far as I can see). There are some headaches though:
* The "renderer support" matching fails if the renderer doesn't support both AAC-LC and SLS. That's incorrect, the whole point of this format is that it can play as long as the renderer supports AAC-LC if I understand it correctly?
* Transcoding with FFmpeg fails as we probably need to use some special parameters to transcode two streams into one.

Sami32:
Hehe, i'll not complain as it's a part of my PR, so it'll have more chance to get merged here ;P
- Yes, you're right it's the only purpose, it's why the "core" is the first stream to be offered
- :wink: I agree on the AAC core use. But i think that you could be surprised to see how fast _FFmpeg_ can be if they want to ;) Maybe from august or november...

Sami32:
But i don't see where you handled multi streams MKA and OGA ?

Nadahar:
@Sami32 
> 😉 I agree on the AAC core use. But i think that you could be surprised to see how fast FFmpeg can be if they want to ;) Maybe from august or november...

That doesn't help much when we don't upgrade our FFmpeg version. Do you know the FFmpeg parameters to transcode only the first stream? As it is now the transcoded output isn't even playable by Kodi (which generally plays almost anything).

> But i don't see where you handled multi streams MKA and OGA ?

I didn't. Is there a corresponding case there where two audio streams "play as one" for MKA or OGA? I only found that this was specified for MP4, WavPack etc doesn't use a generic container format so there's no such case there.

Sami32:
I guess that you do like me, you update it yourself ?
For _FFmpeg_, we use` -map 0:a:2` for example to use the third audio stream. Many way to write it, but [-map](https://trac.ffmpeg.org/wiki/AudioChannelManipulation) command line is the key.
For audio only file we can just use  [-map 0:0](https://trac.ffmpeg.org/wiki/Map)

Not "play at once", but MKA and OGA can have plenty of audio streams, and as is your code now, no one will be played. That's said, many player just forget about that possibility:
User having OGG radio will be the most affected:

> There is a known problem when importing an OGG file that inludes multiple OGG encoded audio streams, that Audacity can only access the first logical stream in such a file. Multiple stream OGG files are often found where the file has been recorded or created from a streaming radio source, and typically, each song will be an individual stream.

Nadahar:
@Sami32 

> Not "play at once", but MKA and OGA can have plenty of audio streams, and as is your code now, no one will be played. That's said, many media players just forget about that possibility:
Users having OGG radio will be the most affected:

I understand that they can have multiple audio streams, but I think it's better if those are sent as video. Many video players allow you to choose audio stream, while I don't think that's a "concept" when it comes to audio. I see how it could be useful to have for example different languages in one audio file, but as long as there's no standard for this I think the probability that the user can actually play them is greatest if sent as video.

Sami32:
Oh i see, i didn't think about that, but from what i read many players just stuck on it.
So i think you're right, it should better for support, but for now, from what i saw, the MIME type is incorrectly in such case (the default one: video/mpeg)

Sami32:
OK forget about, i think it's because the SLS format was unknown...My bad.

Sami32:
Just tell me when you're done.

src/main/java/net/pms/configuration/FormatConfiguration.java
```diff
@@ -96,17 +100,24 @@
...
 	public static final String PNM = "pnm";
 	public static final String PSD = "psd";
 	public static final String QDESIGN = "qdmc";
+	/** This is the RealAudio file format, not one of the codecs */
 	public static final String RA = "ra";
 	public static final String RAW = "raw";
-	public static final String REALAUDIO_LOSSLESS = "ralf";
+	public static final String REALAUDIO_14_4 = "ra14.4";
+	public static final String REALAUDIO_28_8 = "ra28.8";
+	public static final String RALF = "ralf";
```
Sami32: The order should be readable if set after RA. It's just minor formatting though.

src/main/java/net/pms/dlna/DLNAMediaAudio.java
```diff
@@ -407,7 +449,7 @@ public boolean isNonPCMEncodedAudio() {
 	public boolean isLossless() {
 		return getCodecA() != null &&
 			(
-				isAIFF() || isALAC() || isFLAC() || isMLP() ||
+				isAIFF() || isALAC() || isALS() || isFLAC() || isMLP() ||
```
Sami32: `isSLS()` should be added.

src/test/java/net/pms/configuration/FormatConfigurationRegressionTest0.java
```diff
@@ -10,966 +10,707 @@
...
     // Regression assertion (captures the current behavior of the code)
     org.junit.Assert.assertTrue("'" + str0 + "' != '" + "dtshd"+ "'", str0.equals("dtshd"));
-
   }
 
   @Test
   public void test80() throws Throwable {
-
-    if (debug) { System.out.format("%n%s%n","RegressionTest0.test80"); }
-
```
Sami32: What was the use of this string ? to log it when debug level was used ?

src/test/java/net/pms/configuration/FormatConfigurationRegressionTest0.java
```diff
@@ -10,966 +10,707 @@
...
+  public void testALS() throws Throwable {
+    java.lang.String str0 = net.pms.configuration.FormatConfiguration.ALS;
 
+    // Regression assertion (captures the current behavior of the code)
+    org.junit.Assert.assertTrue("'" + str0 + "' != '" + "als"+ "'", str0.equals("als"));
   }
 
+  @Test
+  public void testOGGAudio() throws Throwable {
+    java.lang.String str0 = net.pms.configuration.FormatConfiguration.OGA;
```
Sami32: You mean` testOGA()` ?

src/test/java/net/pms/formats/FormatsRegressionTest0.java
```diff
@@ -11,86 +11,49 @@
...
-
-		if (debug) {
-			System.out.format("%n%s%n", "FormatsRegressionTest0.test06");
-		}
-
 		net.pms.formats.MKV mKV0 = new net.pms.formats.MKV();
 		net.pms.formats.audio.RA rA1 = new net.pms.formats.audio.RA();
-		mKV0.setSecondaryFormat((net.pms.formats.Format) rA1);
-
+		mKV0.setSecondaryFormat(rA1);
```
Sami32: If you don't mind explain me, at least tried ;)
What that is supposed do ?

**Note**: Don't worry, i'll not ask every time, but if i understand only this test it's already good :P

src/main/external-resources/renderers/DefaultRenderer.conf
```diff
@@ -618,7 +621,8 @@ CharMap =
 #    qdmc     QDesign Music Codec (RaveSound)
 #    ralf     RealAudio Lossless Format                     audio/vnd.rn-realaudio
 #    rm       RealMedia (RMVB)                              application/vnd.rn-realmedia-vbr
-#    shn      Shorten
+#    shn      Shorten                                       audio/x-shorten
+#    sipro    Sipro Lab Telecom
```
Sami32: Should be added:
`#    sls     Scalable to Lossless                       audio/mp4                          Also known as MPEG-4 SLS`

src/main/java/net/pms/dlna/DLNAMediaInfo.java
```diff
@@ -324,6 +350,18 @@ public boolean hasAudio() {
...
+	/**
+	 * @return {@code true} is this {@link DLNAMediaInfo} instance has two audio
+	 *         tracks where the first has codec AAC-LC and the second has codec
+	 *         SLS, {@code false} otherwise.
+	 */
+	public boolean isSLS() {
+		if (audioTracks.size() != 2) {
+			return false;
+		}
+		return audioTracks.get(0).isAACLC() && audioTracks.get(1).isSLS();
```
Sami32: It's just an informative precision. **Your code don't need to be modified**, it's look fine to me.

> SLS is MPEG-4's hybrid lossless audio codec. It uses a standard MPEG-4 GA core layer. Valid cores include AAC-LC, AAC Scalable (without LTP), ER AAC LC, ER AAC Scalable, and ER BSAC.

> The HD-AAC Profile is a limited subset of MPEG-4 SLS in an effort to gain commercial adoption. HD-AAC limits SLS to an AAC-LC core and BPGC with LEMC entropy coding.

https://wiki.multimedia.cx/index.php/MPEG-4_SLS

src/test/java/net/pms/configuration/FormatConfigurationRegressionTest0.java
Nadahar: @Sami32 It was never used. I used Randoop to create those tests, and if creates these, but ```debug``` is hardcoded to "false". The idea is probably to enable ```debug``` if you want a more verbose output, but:
1) We have no such system for tests.
2) The wouldn't be very informative anyway

I just removed them as a general "cleanup", I should have done it when I made the tests but didn't for some reason.

src/test/java/net/pms/formats/FormatsRegressionTest0.java
Nadahar: @Sami32 I can't. It's not easy using automated stuff and trying to clean it up manually. I don't know why Randoop made this, there's no test (assertXX) there. The only thing it actually "tests" is that no exception is thrown if you set RA as a seconday format of MKV. 

Randoop doesn't know anything about the data, it just "documents" how the current code is working. Even if setting RA as a secondary format for MKV is meaningless to us, our code allows it without throwing an exception and thus this has been generated.

src/test/java/net/pms/configuration/FormatConfigurationRegressionTest0.java
Sami32: How do you use Randoop ? it's a standalone application or an Eclipse plugin ?

Nadahar: @Sami32 Standalone, but I'm not sure I'm going to use it much more. "Cleaning up" the tests it makes and make it into something useful is as much job as writing them from scratch, the last time I used it I regretted doing so.

[Randoop](https://randoop.github.io/randoop/)

Sami32:
You can squash my commit, as it has never been intented to be alone.

src/test/java/net/pms/configuration/FormatConfigurationRegressionTest0.java
Sami32: As i never listen of it before you told me, i've searched and saw that exist as a [plugin](https://github.com/randoop/randoop-eclipse-plugin) and a standalone version existed, so i wondered which one you used.

Nadahar:
@Sami32 Thanks, but the SLS implementation doesn't really work because of the "supported" matching and the transcoding.

src/test/java/net/pms/configuration/FormatConfigurationRegressionTest0.java
Nadahar: @Sami32 I my memory serves me right, the plugin was outdated and abandoned and it didn't even work when I tested it.

Sami32:
I don't see it as a big issue, i see it as waiting for a _FFmpeg_ "upgrade".
But if it's annoying you too much, you can just comment the "support" lines and it will not recognized any more and hidden.

Nadahar:
@Sami32 I think I've fixed SLS support now - it would be nice if you too could test both streaming and transcoding.

Sami32:
You shouldn't have asked this to me :wink: 
The parsing is fine, and most of the playing.

But as you gone so far as the RAR and ZIP, in that particular case, it's not recognised correctly (MPEG2-TS) and not played. Honestly i don't really mind about but you're the one that asked me for testing it ;)

I have a Regex question for you ;)
Is `(?!als|sls).+` the optimal writing for exclude these format for the renderer supported ones ?

Sami32:
I don't think it's come from your code at all, but since your modified that part it was worst a remark. It's why i told that i really don't mind about.

As always, i was not clear as i thought...
`Supported = f:aac|mp4       m:audio/mp4                a:(?!alac|als|sls).+`

Nadahar:
@Sami32 

> Do you use that Eclipse plugin?

> As always, i was not clear as i thought...
> Supported = f:aac|mp4 m:audio/mp4 a:(?!als|sls).+
> 
I don't know if that will work because I'm not exactly how these regex'es are applied. Given that the form ```a:(?!alac).+``` seems to work, ```a:(?!als|sls).+``` should work as well. Personally I'd prefer to use "positive matching" rather than negative though, that is: ```a:aac-lc|he-aac``` instead.

Yes, you should read [this](http://www.universalmediaserver.com/forum/viewtopic.php?f=16&t=6019).

Sami32:
For you, positive match is fine, but as added much more profile for your renderer negative match is better for most of us ;)

When i tried to see your link, that said me "You are not authorized to read this forum", so i guess that @SubJunk blocked me the access to that specific thread.

Nadahar:
@Sami32 
> When i tried to see your link, that said me "You are not authorized to read this forum", so i guess that @SubJunk blocked me the access to that specific thread.

I doubt he's blocked you, but he probably hasn't given you "Developer" status on the forum. @SubJunk Could you make @Sami32 a member of the developers group?

Sami32:
That could be, as i thought it was only color or appareance difference, i never cared about that and enjoyed my difference and freedom :-)
I didn't know that some thread have a restricted access.

Nadahar:
@Sami32 

> For you, positive match is fine, but as added much more profiles (5 all together) for your renderer, negative match is better for most of us ;)

The reason I prefer it is because I don't "trust" that using negative lookahead will always work as expected. Actually, I don't quite understand how using ```a:(?!alac).+``` can work. Lookahead is only valid for the specific position under evaluation: The above  doesn't mean "alac" can't be anywhere later, it only means that it can't be where we are now. If you have a string ```alac``` and you test position 0, it won't be a match. What I don't understand is why it won't be a match when the position is then moved to 1. Seen from position 1, the string would be ```lac```, which doesn't match the negative pattern ```alac``` and since the rest of the match is ```.*``` which means "anything", the regex should match for position 1, 2 and 3.
If it does work, it's probably because ```match()``` is used instead of ```find()```. The ```match()``` method breaks the normal regex pattern in that it requires the whole string to match, effectively making the regex ```^a:(?!alac).+$```, but if that's the case then the positive matches would have to match exactly. It's probably easier to look at the code than to speculate, but it's even easier to say "test it" 😉

Sami32:
Thank for the explaination. I just asked to you because Regex is your cup of tea and perhaps you'll have told me that a more efficient Regex command line was possible.
I already tested it, and it worked fine until now. Yes, for me testing is much faster and sure that looking at the code :wink: 
Actually, i thought same as you before adding all theseAAC profiles and see how long became the supported line....

Sami32:
Not that really matter, as it's not popular at all, but i just saw that we totally missed the ASF format as audio or video.
I just saw that today while making some Sony renderers for users.

Nadahar:
@Sami32 Is ASF really used for as an audio format?

Sami32:
Why i would say that if it wasn't ? Look at the Sony Bravia X8505B specifications.

But i checked other Sony Bravia documentation as well, and saw that what they call ASF audio is in fact WMA. They real like create confusion...
In conclusion, forget about my comment.

Nadahar:
@Sami32 As I understand it, WMV and WMA use and older version of the ASF standard. That means that while it's safe to say that WMA is ASF, it isn't safe to say to audio-only ASF is WMA.

The ASF standard hasn't been updated since 2004 though, so it looks like the standard won't get any further updates. The only audio codecs registered in the standard are these:

Codec name | Format tag | Notes
-------------|-------------|-------
Windows|Media Audio|0x161|Versions 7, 8, and 9 Series
Windows Media Audio 9 Professional|0x162|9 Series
Windows Media Audio 9 Lossless|0x163|9 Series
GSM-AMR|0x7A21|Fixed bit rate, no SID
GSM-AMR|0x7A22|Variable bit rate, including SID

I really don't know what to make of it and how we should "define" ASF-audio.

Sami32:
> In conclusion, forget about my comment above.

What about that ?
Me, i know: nothing, just leave it as it's  :wink: 
It's not even marketing, just an confusing notation,  that is useless in our case.
The only positive fact, is that it make me remember that many renderer don't support WMA 10 and that we didn't implemented such profile.

Sami32:
@Nadahar I post this off topic here, but i will delete it later.
I'm sure that you know how to make it set `wma10`, because it seem like it won't work like that:
`getFormat(audio, media, currentAudioTrack, "wma10", file);`

Note: This getter look more as a setter to me.

Nadahar:
@Sami32 I'm sorry, but I don't understand what you're asking. You're right that it's more a setter than a getter, but it's not a setter either because it doesn't store anything in it's own instance. Instead it stores the result in the second argument (media). However you put it, it's a mess of a method - and as you know I think the parsing should be done completely different.

I don't understand what you want to achieve. The ```getFormat()``` method in ```LibMediaInfoParser``` (which I assume you're referring to) is meant only for parsing MediaInfo output. If you want to set the audio codec to "wma10" you can just call ```currentAudioTrack.setCodecA()```.

Nadahar:
I've decided not to do anything about RAR and ZIP, their parsing is completely wrong and I'm surprised that any of it work at all. SLS support doesn't even "make the list" of what needs to be fixed with those.

As there have been no other objections to this, I think this is ready to merge.

Sami32:
Thanks for your answer.

Do `currentAudioTrack.setCodecA("wma10")` will be usable for LibMediaInfoParser parsing ? I didn't thought that it will, it's why i tried to get in `getFormat()`

Nadahar:
@Sami32 No - I don't understand what you're trying to do. Why do you want to "set" something for LibMediaInfoParser parsing? It's simply parsing the output from MediaInfo. I think you need to explain a bit more about what you're trying to do.

Sami32:
@Nadahar OK, sorry.
I don't want add the MediaInfo "CodecID_Description" field in the parsed data, as it could give a total mess in the parsing result.
I just need it to identify the WMA 10 at the "main" level and set "wma10" manually as codec ID.
See: https://github.com/UniversalMediaServer/UniversalMediaServer/compare/wma10

Nadahar:
@Sami32 You're almost doing it correctly. I would just let ```LibMediaInfoParser``` parse "as normal", and when that's done, correct the audio codec. 

Don't remove the line:
```java
getFormat(audio, media, currentAudioTrack, MI.Get(audio, i, "CodecID"), file);
```
Instead, let it do it's thing and *then* you check ```CodecID_Description``` and correct the audio codec if necessary:
```java
value = MI.Get(audio, i, "CodecID_Description");
if (isNotBlank(value) && value.startsWith("Windows Media Audio 10")) {
    currentAudioTrack.setCodecA(FormatConfiguration.WMA10);
}
```

Sami32:
Thank you for having at least tried ;-)

That doesn't worked, but it was worst a try.

Nadahar:
@Sami32 It will work, but I wrote this without an IDE and without compiling so there might be some missing details. What is the error you get?

Sami32:
@Nadahar Sorry, i didn't saw your answer when i first edited my previous post. You're too fast :wink:

src/main/java/net/pms/dlna/DLNAMediaInfo.java
```diff
@@ -86,6 +89,33 @@
...
+	 */
+	protected static final Map<String, AudioVariantInfo> audioOrVideoContainers;
+
+	static {
+		Map<String, AudioVariantInfo> mutableAudioOrVideoContainers = new HashMap<String, AudioVariantInfo>();
+
+		// Map container formats to their "audio variant".
+		mutableAudioOrVideoContainers.put(FormatConfiguration.MP4, new AudioVariantInfo(new M4A(), FormatConfiguration.M4A));
+		mutableAudioOrVideoContainers.put(FormatConfiguration.MKV, new AudioVariantInfo(new MKA(), FormatConfiguration.MKA));
+		mutableAudioOrVideoContainers.put(FormatConfiguration.OGG, new AudioVariantInfo(new OGG(), FormatConfiguration.OGA));
```
Sami32: I didn't saw that at first. Here also `AudioVariantInfo(new OGG()` seem need to be replaced by `AudioVariantInfo(new OGA()`

Nadahar: @Sami32 It was changed during the process, and this was forgotten 😢

Sami32: @Nadahar Arrrg...i totally forgotten to told you about Flash audio, also known as FLA or F4A.
As i've done it in some_AUDIO_fixes, i did forget. My bad :(

Nadahar: @Sami32 That can be done later, it will be additional functionality. What I don't want is for PRs the remove existing functionality, so I think adding Flash in a later PR is fine.

PR #1286, opened 2017-05-24
Sony Bravia XE add

Sami32:
As related by this user on our forum:
http://www.universalmediaserver.com/forum/viewtopic.php?f=5&t=9678

Documentation:
https://www.sony.fr/electronics/televiseurs/XE9005-Series/specifications
https://www.sony.co.uk/electronics/support/res/manuals/4690/46909281M.pdf

### DIFF
--- a/src/main/external-resources/renderers/Sony-BraviaXD.conf
+++ b/src/main/external-resources/renderers/Sony-BraviaXD.conf
@@ -2,7 +2,7 @@
 # Profile for Sony Bravia XD 70/75/80/83/85/93/94 series, 4K HDR avec Android TV.
 # See DefaultRenderer.conf for descriptions of all the available options.
 
-RendererName = Sony Bravia XD
+RendererName = Sony Bravia XD/XE
 RendererIcon = Sony-BraviaXD.png
 
 # =================================================================================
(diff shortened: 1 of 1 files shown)


SubJunk:
What's `chromecast_extension`?

Sami32:
Since many users reported that they have connection issue without this option, and that they ask and search where is their correct _ums.conf_ file to modify, i thought it was much easier for users to set it directly in their renderer, as it will work same as if that was done in _ums.conf_ anyway.

Because when i read that this user make 3H, i suspect that some other users just give up ?

Nadahar:
We shouldn't "mix the cards" by trying to patch a bug by making another one in my opinion. The Cromecast implementation is bugged for several reasons, and unless someone that actually has a Chromecast for testing cleans up this code we should disable Chromecast by default. I don't even know if the current implementation does much good for those that have one, and this buggy implementation certainly can cause issues for those who don't. Disabling it in the renderer configuration is definitly wrong, as the problem isn't related to that particular renderer. 

It won't work as intended either. UMS doesn't read that particular renderer's configuration when determining if ```ChromecastMgr``` should be started. It's done in ```PMS.init()``` using the "global" configuration.

Sami32:
Corrected.

@Nadahar I understood that you said me that the parameters should be handled in whatever configuration file they was, so after all, it's seem that it's not all the parameters ?

Nadahar:
@Sami32 No, it's more complicated than that. All parameters are "supported" in the renderer configuration file as well, but you have to think of how this works. It only applies to parameters that's relevant for a particular renderer. The chromecast "service" starts up without being related to any particular renderer. Even if you turn it off in that particular renderer configuration, it's not turned off in all the other renderer configurations. How should UMS know to listen to that particular renderer configuration file?

Sami32:
@Nadahar So it was fine, as i intended to disable only for this kind of renderers having issue with it activated, avoiding the total desactivation for the other Chromacast able renderers that users can have and work fine (I've serious doubt about "fine" ;))

Nadahar:
@Sami32 No, as I tried to explain, it doesn't work like that. UMS either runs the chromecast service or not. That's not "tied" to a renderer, so setting that in a renderer configuration will have no effect at all.

I think we should disable it in UMS.conf by default because it is so buggy and will probably create problems not only for this renderer.

Sami32:
@Nadahar I see, so i agree that it should disable by default, as it's not the first user on the forum having issue with their Chromecast renderer, and it did take me some time to found why they cannot make UMS working with it.

PR #1288, opened 2017-05-30
Fix Cinepak issue introduced since MediaInfo 0.7.93

Sami32:
See https://github.com/MediaArea/MediaInfoLib/commit/b45b4e178c51add8653447cae616172a1c3b71f1
Should be merged before the next release that will use _MediaInfo_ 0.7.96.

### DIFF
--- a/src/main/java/net/pms/configuration/FormatConfiguration.java
+++ b/src/main/java/net/pms/configuration/FormatConfiguration.java
@@ -58,7 +58,7 @@ public class FormatConfiguration {
 	public static final String AU = "au";
 	public static final String AVI = "avi";
 	public static final String BMP = "bmp";
-	public static final String CINEPACK = "cvid";
+	public static final String CINEPAK = "cvid";
 	public static final String COOK = "cook";
 	public static final String CUR = "cur";
 	public static final String DIVX = "divx";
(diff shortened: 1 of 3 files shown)


Sami32:
I don't see the point of your modification ???

New MediaInfo is used by this new version only, that's all.
Do i need to add all the old MediaInfo codecs support as well ?

Nadahar:
@Sami32 Compatibility should always be preferred if possible without serious implications in my opinion. It this case it's quite easy to see how it could be useful:
* On Windows XP where one is limited to using an older version of MediaInfo
* On Linux where MediaInfo isn't bundled and the version is controlled by the system, not by us.
* On any platform if a certain feature is broken or removed in recent versions of MediaInfo.

There's no "cost" in supporting both versions, it will only improve compatibility. If there are "old versions" of keywords that we use, I think we should support them as well, and shouldn't assume things we don't know (like the version of an external library being used).

Sami32:
> On Windows XP where one is limited to using an older version of MediaInfo

This has been fixed many month ago (0.7.91), XP users can use MediaInfo without issues anymore.
> On any platform if a certain feature is broken or removed in recent versions of MediaInfo.

That'll not happen in that case.
> On Linux where MediaInfo isn't bundled and the version is controlled by the system, not by us

:+1: I didn't thought about Linux :confounded: 
You're, that doesn't cost, but i appreciate to understand why, this i know that you can understand :wink:

Sami32:
:+1:

PR #1289, opened 2017-05-30
Update ffmpeg_formats.txt

Sami32:
From _FFmpeg_ recent version: (It has not updated since 6 years already)
`$ ffmpeg.exe -formats`
https://www.ffmpeg.org/ffmpeg-formats.html

Though i didn't checked how we parse it yet, so i'm not sure if the coma in the first row will be correctly handled.
https://github.com/UniversalMediaServer/UniversalMediaServer/blob/2bc07700532dfd131b5d74608dbdde0c86d48775/src/main/java/net/pms/util/CodecUtil.java#L50

**EDIT**: AFAIU, it should be fine. I reformatted the 3 or 4 lines having comma.
Originally i was searching for that: `codecs.add("iso")` and the default `"mpeg1"`[ type setting](https://github.com/UniversalMediaServer/UniversalMediaServer/blob/50ecf1217100a23d92eefc37d273c1e13ac2d1d0/src/main/java/net/pms/encoders/MEncoderVideo.java#L2533-L2563).

### DIFF
--- a/src/main/resources/ffmpeg_formats.txt
+++ b/src/main/resources/ffmpeg_formats.txt
@@ -114,7 +114,7 @@ hash            Hash testing
 hds             HDS Muxer
 hevc            raw HEVC video
 hls             Apple HTTP Live Streaming
-hls,applehttp   Apple HTTP Live Streaming
+applehttp       Apple HTTP Live Streaming
 hnm             Cryo HNM v4
 ico             Microsoft Windows ICO
 idcin           id Cinematic
@@ -150,7 +150,7 @@ lvf             LVF
 lxf             VR native stream (LXF)
 m4v             raw MPEG-4 video
 matroska        Matroska
-matroska,webm   Matroska / WebM
+webm            WebM
 md5             MD5 testing
 mgsts           Metal Gear Solid: The Twin Snakes
 microdvd        MicroDVD subtitle format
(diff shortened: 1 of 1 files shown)


Nadahar:
@Sami32 Despite the name ```ffmpeg_formats.txt```, this is used only for MEncoder. Remember that MEncoder and FFmpeg was very close once upon a time with many common developers. This is used exclusively for evaluating the content "Codec-specific parameters" in the Mencoder configuration dialog.

It's hard for me to understand how exactly this works, as the configuration is evaluated using [Beanshell Interpreter](http://www.beanshell.org/javadoc/bsh/Interpreter.html) which is an "almost Java" scripting language from what I can understand, but I can see that if compares the values of the "script" "in Codec-specific parameters" with the values of the ```DLNAMediaInfo``` and ```DLNAMediaAudio``` instances in question, and if there's a match that specific MEncoder option is added to the transcoding command line.

It seems to me like the list of codecs is used to "filter" the values gotten from ```container```, ```vcodec``` and ```acodec```, so that values found there that's not in the list is ignored and is impossible to use in the script. I'm not sure I understand the purpose of this, why it couldn't simply be a string match between whatever is in one of these fields against the content of the script. In any case, updating the file will probably break some existing scripts. 

If you've found a particular bug/issue because some valid value isn't in the list, I think it would be safer to simply add that or those values than to update it as you do here. As said above, it really has nothing to do with what FFmpeg accepts, it has to do with what values we use for ```container```, ```vcodec``` and ```acoder``` when using this MEncoder transcoding engine "script".

The parser is very "stupid" and will not handle comma. It will simply add anything on each line that comes before either space (```" "```) or newline. That means it will add ```mp4,mp3``` as the "name" of one codec.

Sami32:
Thanks for reviewing :)
I only saw that it was used by MEncoder and didn't go further, but now that i now how limited is our MEncoder support, that doesn't make any sens anymore, IMO.

PR #1296, opened 2017-06-10
Fix Sleep Manager NPE on Linux

Nadahar:
It seems that I forgot to test ```SleepManager``` on Linux, and that ```isPreventSleepSupported()``` wasn't checked every place it needed, making any media playback crash on Linux when ```startPlaying()``` is called.

A workaround is setting ```prevent_sleep = never``` in ```UMS.conf```, but since ```prevent_sleep = playback``` is the default, this is a very bad bug that will affect as good as all Linux installations.

I'm very sorry.

---
This change is [Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/1296)

### DIFF
--- a/src/main/java/net/pms/util/SleepManager.java
+++ b/src/main/java/net/pms/util/SleepManager.java
@@ -73,7 +73,7 @@ public class SleepManager {
 	 */
 	public synchronized void startPlaying() {
 		playingCount++;
-		if (!sleepPrevented && mode == PreventSleepMode.PLAYBACK) {
+		if (isPreventSleepSupported() && !sleepPrevented && mode == PreventSleepMode.PLAYBACK) {
 			preventSleep();
 		}
 	}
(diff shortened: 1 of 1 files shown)


SubJunk:
@Nadahar no problem, we have all made mistakes like that

Sami32:
I should add that we are all more responsable than you as reviewers, or not reviewing, your code.

Nadahar:
@Sami32 I don't know if it would be fair to expect a reviewer to notice that. Testing didn't reveal it either because none of us tested this on Linux. 

What bugs me most is that fact that I had this on my (in memory) "to-do list" earlier, but forgot it somewhere along the way. When I read the bug report log I suddenly remembered it again.. It wasn't a "bug" as such, it was simply that I forgot to do it - which means it shouldn't have happened 😞

Sami32:
@Nadahar Don't blame yourself. As we said: "A toute chose, malheur est bon"
Meaning that without that, we wouldn't have fixed 3GA and OGG by now, and more users would have more issues.
And if you said that are also your mistakes, i can answer you that without your work we'll not have nice improvements as well :wink:

PR #1297, opened 2017-06-10
Add wma10 profile.

Sami32:
Some renderers, doesn't support [WMA 10 Pro](https://msdn.microsoft.com/en-us/library/windows/desktop/gg153556(v=vs.85).aspx), so it'll make them to be transcoded and playable at their full resolution.

I tested it without any issue.

**EDIT**:
PHILIPS PFL-L7605H/12
PHILIPS PUS6561/12
SAMSUNG BD-F5100-ZF
Sony SMP-N100
Sony BDP-590S
Sony STR-DN840
TOSHIBA_39L4300UM
Toshiba L74xx
...

https://en.wikipedia.org/wiki/Windows_Media_Audio

> All versions of WMA released since version 9.0 - namely 9.1, 9.2 and 10 - have been backwards compatible with the original v9 decoder and are therefore not considered separate codecs. The sole exception to this is the WMA 10 Professional codec whose Low Bit Rate (LBR) mode is only backwards compatible with the older WMA Professional decoders at half sampling rate (similar to how HE-AAC is backwards compatible with AAC-LC). Full fidelity decoding of WMA 10 Professional LBR bitstreams requires a WMA version 10 or newer decoder.

### DIFF
--- a/src/main/external-resources/renderers/DefaultRenderer.conf
+++ b/src/main/external-resources/renderers/DefaultRenderer.conf
@@ -640,8 +640,13 @@ CharMap =
 #    vp9      Google VP9                                    video/webm
 #    wavpack  WavPack                                       audio/wavpack
 #    wav      Waveform Audio File Format                    audio/x-wav
+#    weba     Webm Audio                                    audio/webm
 #    webm     WebM Video                                    video/webm
 #    wma      Windows Media Audio                           audio/x-ms-wma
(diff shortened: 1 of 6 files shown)


Sami32:
It seem that PS3 as well.
http://manuals.playstation.net/document/en/ps3/current/video/filetypes.html

Note: Added a missing WMA profile (160).
https://wiki.multimedia.cx/index.php/Windows_Media_Audio_9

Sami32:
Thanks :smile:

PR #1299, opened 2017-06-10
New6.7.0version

valib:
@SubJunk I recommend to reset master to this branch and rewrite the official release 6.7.0.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,4 +1,4 @@
-6.7.0 - 2017-06-??
+6.7.0 - 2017-06-10
 
 	General:
 		Added a button for the web interface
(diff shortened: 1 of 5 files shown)


SubJunk:
@valib the way we have addressed critical bugs like this in the past is to do a release like 6.7.0.1. That's less confusing for people who have already downloaded it (which is a lot already)

valib:
@SubJunk OK. I am deleting this branch.

Sami32:
@SubJunk If you do that, don't forget to update MediaInfo to 0.7.96 as well.
Because TS, MTS, M2TS issues (freezing) will arise with the 0.7.94...

Nadahar:
I agree that once something is released, it's "frozen forever". That's the general "rule" used by most software, so that's what people expect.

PR #1300, opened 2017-06-10
Formatting #1226

Nadahar:
@SubJunk I've addressed the formatting comments you had in your review for #1226. I'm not sure if this is a better way to format multi-line ternaries - it's merely a suggestion.


src/main/java/net/pms/configuration/PmsConfiguration.java
```diff
@@ -243,6 +243,7 @@
 	protected static final String KEY_PLUGIN_DIRECTORY = "plugins";
 	protected static final String KEY_PLUGIN_PURGE_ACTION = "plugin_purge";
 	protected static final String KEY_PRETTIFY_FILENAMES = "prettify_filenames";
+	@Deprecated
```
SubJunk: Cool, yeah that is what I meant, also I wouldn't rename it to `OLD` anymore since that's what the deprecated tag is there for

Nadahar:
@SubJunk I assume this is how you want it?

SubJunk:
Looks great, thanks!

Issue #1302, opened 2017-06-12
Correct DLNA Flag?


onon765trb:
Thank you for the good app.

DLNA.ORG_OP=11 and ORG_CI=0 are used in transcoding even when using SeekByTime= true

For example,
`TRACE 2017-06-12 16:23:46.384 [HTTPv2 Request Worker 2] Network debugger: "mkv.mkv" http-get:*:video/mpeg:DLNA.ORG_PN=AVC_TS_HP_HD_AAC;DLNA.ORG_OP=11 size=100000000000 duration=0:18:19 http://xxx.xxx.xxx.xxx:10045/get/27/mkv.mkv_transcoded_to.ts`
`ContentFeatures.DLNA.ORG: DLNA.ORG_PN=AVC_TS_HP_HD_AAC;DLNA.ORG_OP=11;DLNA.ORG_CI=0;DLNA.ORG_FLAGS=01700000000000000000000000000000`
This is a part of the attached log.
On my understanding, DLNA.ORG_OP=10;DLNA.ORG_CI=1 is correct.
My misunderstanding?

This question arises from impossible seeking when using TranscodeVideo= MPEGTS-H264-AAC.
But possible when using TranscodeVideo= MPEGPS-MPEG2-AC3, SeekByTime= true, TranscodedVideoFileSize= -1, ChunkedTransfer= True.

[ums_dbg_MPEGTS-H264-AAC.zip](https://github.com/UniversalMediaServer/UniversalMediaServer/files/1067886/ums_dbg_MPEGTS-H264-AAC.zip)

Nadahar:
@onon765trb DLNA has never been properly implemented in UMS or it's predecessor. There's loads of inconsistencies like this. When it comes to the CI flag, it doesn't really matter as long as there's just one ```<res>``` source to choose from. The CI flag is supposed to aid renderers in chossing the "best" version of a given media when multiple versions are offered with the assumption that CI == 0 sources will be chosen if supported by the renderer. UMS only offer one source per media file, so it's irrelevant although technically wrong.

The seek flag on the other hand is a real problem if set wrong. Without verifying the facts here, you might have found a bug.

Sami32:
@onon765trb In fact in our code, we call it exclusive. (that was done for Panasonic renderer ;))
You need to add this line in the renderer:
`SeekByTime = exclusive`

onon765trb:
Thank you for your replies.
[...]

PR #1307, opened 2017-06-15
OP_flag_better_comments

Sami32:
To avoid the users confusion:
https://github.com/UniversalMediaServer/UniversalMediaServer/issues/1302

### DIFF
--- a/src/main/external-resources/renderers/DefaultRenderer.conf
+++ b/src/main/external-resources/renderers/DefaultRenderer.conf
@@ -159,12 +159,12 @@ Image =
 #
 #     true:
 #
-#         The renderer supports seek-by-time.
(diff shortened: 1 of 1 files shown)


Sami32:
:smile:

Sami32:
@Nadahar Do you know if Linux and Mac OS users can use _**update.jar**_ file as the Windows users ?
Thanks.

Note: i'll delete our comments later.

Nadahar:
@Sami32 Linux users can, but I'm not so sure about macOS. In fact I doubt it will work on macOS when run as an "application", and I also doubt that most macOS users would know how to find the correct location to put ```update.jar```. 

The thing is that both the Windows ```.exe``` file and the Linux ```.sh``` script looks for ```update.jar``` and use it if it's there. I doubt that "appbundler" does that, but I can't say 100% sure without digging into the code.

PR #1308, opened 2017-06-15
Set MediaInfo characterset to UTF-8 for non-Windows platforms (#355)

### LINKED ISSUE #355: Filenames with non-ASCII character cannot be parsed on OS X 10
Files with filenames with non-ASCII characters cannot be parsed, that is, the log shows "The file {} could not be parsed. It will be hidden", where {} is the pathname. There is nothing special about the video file itself, as I can rename the file using only ASCII characters, and UMS will show and play it normally.

This was tested on OS X 10.9 using UMS 4.0.2 (Java 8).

Nadahar:
I've done some quick tests on Windows 7, macOS 10.12.5 and Kubuntu 16.04 LTS and this actually seems to fix #355. I've also requested [user testing](http://www.universalmediaserver.com/forum/viewtopic.php?p=31530#p31530).

Tests and feedback are welcome.

### DIFF
--- a/src/main/java/net/pms/dlna/MediaInfo.java
+++ b/src/main/java/net/pms/dlna/MediaInfo.java
@@ -29,7 +29,9 @@ package net.pms.dlna;
 
 import com.sun.jna.*;
 import java.lang.reflect.Method;
-import static java.util.Collections.singletonMap;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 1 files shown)


Nadahar:
I'm a bit surprised that there's so little interest in this one, it's a bug that's been there for a very long time (and have had quite a few support threads). I have already tested it on Windows, Linux and OS X, so there shouldn't be too much testing needed. I'd just like some more (Windows) testing simply using MediaInfo "normally" to make sure nothing is broken. When it comes to Linux and OS X it is broken as it is today, so I don't think we need a lot of testing there.

Nadahar:
I just got [positive user feedback](http://www.universalmediaserver.com/forum/viewtopic.php?f=9&t=10007&p=31676#p31676), so I think I'll just merge this.

Nadahar:
@Sami32 Thank you btw. 👍

SubJunk:
Nice one

PR #1311, opened 2017-06-18
Update CONTRIBUTING.md

Sami32:
@subjunk As i don't know if you've some reason to not telling it, i'll appreciate your feedback.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,7 +11,7 @@ Often in order to help diagnose and fix your issue we need your "debug informati
 7.  Save the zip file to a location you will remember
 8.  Attach the zip file as described [below](#Attachments), with a link to the file in your issue description
 
-If it is a new bug that wasn't there in a previous version, we will often request debug info from the last version that did not have the bug, too.
+If it is a new bug that wasn't there in a previous version, we will often request debug info from the last version that did not have the bug, too.  
 So if that is the case, please repeat the above steps for the last version without the bug.
 
 Thanks for your efforts in helping us make UMS better!
@@ -24,7 +24,10 @@ For more information, please see STYLEGUIDE.md
 
 # <a name="Attachments"></a>Attachments
 
-GitHub only allows image attachments. Other attachments, for example zipped logs, can be posted in the <a href="http://www.universalmediaserver.com/forum/viewtopic.php?f=14&t=1656">Github attachments thread</a> in the UMS forum.
-Remember to include a link to the attachment in your post here.
+GitHub only allows a few attachment extensions: .png, .gif, .jpg, .docx, .pptx, .xlsx, .txt, .pdf, .zip and .gz.  
+Other attachments can be posted in the <a href="http://www.universalmediaserver.com/forum/viewtopic.php?f=14&t=1656">Github attachments thread</a> in the UMS forum.
+Remember to include a link to the attachment in your post here.  
+They can also be attached here if you add a fake supported extension (.txt, .jpg, etc.).
 
-If you prefer to upload your attachment elsewhere, please avoid sites with restricted access or captcha if possible.
+
+If you prefer to upload your attachment elsewhere, please avoid sites with restricted access, captcha or ads.


CONTRIBUTING.md
```diff
@@ -24,7 +24,9 @@ For more information, please see STYLEGUIDE.md
 
 # <a name="Attachments"></a>Attachments
 
-GitHub only allows image attachments. Other attachments, for example zipped logs, can be posted in the <a href="http://www.universalmediaserver.com/forum/viewtopic.php?f=14&t=1656">Github attachments thread</a> in the UMS forum.
+GitHub only allows few attachments formats. Other attachments, for example zipped logs, can be posted in the <a href="http://www.universalmediaserver.com/forum/viewtopic.php?f=14&t=1656">Github attachments thread</a> in the UMS forum.
```
Nadahar: "GitHub only allows **a** few attachments **extensions**: ```.png```, ```.gif```, ```.jpg```, ```.docx```, ```.pptx```, ```.xlsx```, ```.txt```, ```.pdf```, ```.zip``` and ```.gz```. (GitHub doesn't check the format at all). Reference: https://help.github.com/articles/file-attachments-on-issues-and-pull-requests/

Remove ```, for example zipped logs```.

Sami32: As everyone can see it by themself i didn't intended to write them all, but leave only the ZIP example.
I hope that you're sure about "a few" as for me that look strange ;)

Nadahar: @Sami32 I am sure, without "a" the sentence is wrong.

CONTRIBUTING.md
```diff
@@ -24,7 +24,10 @@ For more information, please see STYLEGUIDE.md
 
 # <a name="Attachments"></a>Attachments
 
-GitHub only allows image attachments. Other attachments, for example zipped logs, can be posted in the <a href="http://www.universalmediaserver.com/forum/viewtopic.php?f=14&t=1656">Github attachments thread</a> in the UMS forum.
+GitHub only allows a few attachments extensions: .png, .gif, .jpg, .docx, .pptx, .xlsx, .txt, .pdf, .zip and .gz.
```
Nadahar: attachment without "s"

src/main/external-resources/UMS.conf
```diff
@@ -196,7 +196,7 @@ http_engine_v2 =
 # Prevent system idle sleep
 # -------------------------
 # If your UMS host has a power-saving mode, use this option to prevent the
-# system from hibernating or sleeping. Currently only Windows and macOS are
+# system from hibernating or sleeping. Currently only Windows and Mac OS are
 # supported. There are three modes:
```
Nadahar: @Sami32 Wrong: https://en.wikipedia.org/wiki/MacOS

CONTRIBUTING.md
Sami32: I've copy pas you ;)

src/main/external-resources/UMS.conf
Sami32: Not wrong. That's said i never saw "your" wrinting before, sorry to not be a Mac fan :package: 
As i guess that you prefer the other, weird to me, variant, i'll revert it.

I'll go to sleep smarter ;-)

CONTRIBUTING.md
Nadahar: Hehe, and I copy hasted your text and added the "a" and replaced "formats" 😉

src/main/external-resources/UMS.conf
Nadahar: It is "wrong" in that it isn't the official name. Don't blame me that aPple keeps changing the name, the current name is ```macOS```.

CONTRIBUTING.md
Sami32: And me, added the word "few" in the already existing text :smile:

src/main/external-resources/UMS.conf
Sami32: It's too much for a dinosaur like me LOL
I will stick with Mac OS, but now i leaned that they changed name, as the political parties in my country when things go wrong.

Nadahar: @Sami32 The current OS was never called ```Mac OS```. The OS made by Apple was called ```Mac OS```, but they abandoned it with version 9 - and that OS has little in common with the current ```macOS```. The current OS, which is a "heavily modified version of BSD", was first called ```Mac OS X```. In 2012 it was renamed to ```OS X``` - which was it's name until they renamed it again in 2016 to ```macOS```. 

It's always a sure sign that there's no real innovation when companies constantly keep renaming their product.

CONTRIBUTING.md
```diff
@@ -11,7 +11,7 @@ Often in order to help diagnose and fix your issue we need your "debug informati
 7.  Save the zip file to a location you will remember
 8.  Attach the zip file as described [below](#Attachments), with a link to the file in your issue description
 
-If it is a new bug that wasn't there in a previous version, we will often request debug info from the last version that did not have the bug, too.
+If it is a new bug that wasn't there in a previous version, we will often request debug info from the last version that did not have the bug, too.
```
SubJunk: Trailing spaces

CONTRIBUTING.md
```diff
@@ -24,7 +24,10 @@ For more information, please see STYLEGUIDE.md
 
 # <a name="Attachments"></a>Attachments
 
-GitHub only allows image attachments. Other attachments, for example zipped logs, can be posted in the <a href="http://www.universalmediaserver.com/forum/viewtopic.php?f=14&t=1656">Github attachments thread</a> in the UMS forum.
-Remember to include a link to the attachment in your post here.
+GitHub only allows a few attachment extensions: .png, .gif, .jpg, .docx, .pptx, .xlsx, .txt, .pdf, .zip and .gz.
```
SubJunk: Maybe we can keep it more general so that we don't have to update it every time GitHub updates its formats. Maybe "GitHub supports attaching some files like images and archives."?

CONTRIBUTING.md
Sami32: No ! That's MarkDown formatting, so in purpose.

CONTRIBUTING.md
Nadahar: @Sami32 Thit was my suggestion (https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1311#discussion_r122598408), and the reason to list them is the amount of confusion this system causes. I had to look for a while to even figure out what extensions was allowed. A lot of image formats isn't allowed, and the same goes for archives. Even 7zip, which is very popular, isn't allowed - so it's a bit hard for anyone to guess from a general description.

Sami32: @SubJunk I don't find easily the information concerning the supported files formats, do you ?
I do not expect that people want search that much for such link:
https://github.com/blog/2061-attach-files-to-comments

So i don't think delete this comment should be the thing to do, and how many times did they changed they support ? the last time they added was in october 2015, and as they just add and not withdraw formats support, AFAIK, i think that should be kept as it is, IMO.

PR #1315, opened 2017-06-20
Clean and fixes some renderers

Sami32:


Sami32:
Thank you for your approval :+1: 

If the user give feed back, i'll prefer wait for it, as its special AVI coming from camera use specific codecs and low frame rate, that i suspect FFmpeg having trouble to handle it; the easy and correct way is to stream it directly to his renderer.
But if i can get a sample and MediaInfo log from him, i prefer wait for it a little bit, and give him this renderer to test after ;)
As discovering what FFmpeg do not natively handle correctly could be useful, for an option fix or a bug report.

Too many renderers still have this transcoding string that give users more issues than it solve, IMO.

Nadahar:
@Sami32 Just wait until you feel it's as correct as it can be. The ```TranscodeExtensions``` string is from when MediaInfo wasn't used, it obviously shouldn't be there when the ```supported``` lines are there as well.

Sami32:
At first i thought like you, but in fact, it is used also when `MediaInfo = true`.
I only advantage is to be able to handle formats that are not supported by MediaInfo.
http://www.universalmediaserver.com/forum/download/file.php?id=3900

My purpose is only to make such videos, as i guess that he is not the only one having family AVI videos taken from camera or phone, be able to play without hassle for users.

Nadahar:
@Sami32 I was a bit unclear about ```TranscodeExtensions``` - it can be used (since I fixed it a while ago) to fill in when MediaInfo can't. But, in most cases it shouldn't - when I fixed it so that ```TranscodeExtensions``` worked even if ```MediaInfo = true``` some renderers can actually have been "broken", as this parameter used to be ignored when ```MediaInfo = true```. I didn't dream of that profiles were "converted" to use MediaInfo without removing ```TranscodeExtensions```.

Sami32:
Right.

But as you saw, users themself make their own renderer that ugly way...
It seem that 12 renderer, so 11 eleven less this one, are affected.

Nadahar:
@Sami32 *Someone™* should write a guide about how to make a renderer configuration, where the use of the ```supported``` lines is explained and the most commonly needed options are also mentioned - in addition to explaining that there's help text explaining the options in ```DefaultRenderer.conf```. If that was done, I'm pretty sure it work much better.

Sami32:
@Nadahar I was naively thinkiing that our FAQ and Wiki was suffisant.
I could check at it later.

Sami32:
I just tried to make quick "fixes".

IMO, that will only make things better, in the way that as it defer to what is already defined in the supported lines, so we can slowly upgrade it on users report.
The images will not be transcoded by default as well.

Slowly these lines should disappear. With the help of the new log, thanks to your good work, we'll be more easily able to do that.

Nadahar:
@Sami32 I don't think that code will apply for images anyway, as images follows a completely different logic. But, it's nice to remove them anyway, because the fact that they are there indicates to the user that they can be specified that way.

When doing these "quick" fixes, just make sure not to be "too quick". That's what I often end up doing when I'm trying to be "quick" 😉

Sami32:
@Nadahar Sadly to me too :wink: 
An also if i'm not doing it quickly...

I'll wait next week.

PR #1316, opened 2017-06-20
Old code cleaning

Sami32:
@SubJunk Could you confirm that you renderer display the thumbnails correctly without these 2 lines ?

### DIFF
--- a/src/main/external-resources/renderers/Panasonic-VieraAS650.conf
+++ b/src/main/external-resources/renderers/Panasonic-VieraAS650.conf
@@ -17,7 +17,7 @@ RendererIcon = Panasonic-VieraAS650.png
 
 UserAgentAdditionalHeader = X-PANASONIC-Registration
 UserAgentAdditionalHeaderSearch = REFOVFYAAAAAAAAAAAAAAAA
-UpnpDetailsSearch = \d{2}AS650
+UpnpDetailsSearch = \\d{2}AS650
 LoadingPriority = 2
 
 SeekByTime = exclusive
@@ -29,8 +29,6 @@ TranscodeFastStart = true
 KeepAspectRatioTranscoding = true
 RescaleByRenderer = false
 SendDateMetadata = false
-ThumbnailWidth = 320
-ThumbnailHeight = 236
 MediaInfo = true
 
 # Supported video formats:
--- a/src/main/external-resources/renderers/Panasonic-VieraAS650.conf
+++ b/src/main/external-resources/renderers/Panasonic-VieraAS650.conf
@@ -17,7 +17,7 @@ RendererIcon = Panasonic-VieraAS650.png
 
 UserAgentAdditionalHeader = X-PANASONIC-Registration
 UserAgentAdditionalHeaderSearch = REFOVFYAAAAAAAAAAAAAAAA
-UpnpDetailsSearch = \\d{2}AS650
+UpnpDetailsSearch = \d{2}AS650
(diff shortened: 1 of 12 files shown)


Nadahar:
I "approve" of the removal of the thumbnail properties 👍 I can't approve the PR as a whole as I can't evaluate the other changes. I would rename the PR to "Renderer configuration cleaning" or similar, since there's no code changed here at all.

Sami32:
I just avoided the tricky "_Xbox one_" and the @valib renderer, as asked by him if i remember well.

src/main/external-resources/renderers/Microsoft-XboxOne.conf
```diff
@@ -24,10 +24,10 @@ LoadingPriority = 1
 
 SeekByTime = exclusive
 TranscodeVideo = MPEGTS-H264-AC3
-MediaInfo = true
 ThumbnailWidth = 256
 ThumbnailHeight = 256
 HalveBitrate = true
+MediaInfo = true
```
Nadahar: @Sami32 If you're not making a change to this file, it's better to exclude the file from the commit IMO. It will make reading the history for the file easier. In this case though, you should remove the Thumbnail properties which will lead to the file being included anyway.

Nadahar:
@Sami32 Actually, when checking the code I see that the ```Thumbnail...``` properties are still being used when generating thumbnails with FFmpeg. 

I really thought I'd removed all references to these, but that means that we should figure out what to do about this a bit more.

Sami32:
Yes, i saw that 1 or 2 week ago, and forget to report it to you.

src/main/external-resources/renderers/Panasonic-VieraAS650.conf
```diff
@@ -17,7 +17,7 @@ RendererIcon = Panasonic-VieraAS650.png
 
 UserAgentAdditionalHeader = X-PANASONIC-Registration
 UserAgentAdditionalHeaderSearch = REFOVFYAAAAAAAAAAAAAAAA
-UpnpDetailsSearch = \d{2}AS650
+UpnpDetailsSearch = \\d{2}AS650
```
SubJunk: Why?

Sami32: Because of Jave rules, as explained below: (we've already tested it in an other commit before)
https://docs.oracle.com/javase/tutorial/java/data/characters.html
Or [here](https://books.google.fr/books?id=fu5HtixRje8C&pg=PA327&lpg=PA327&dq=java+%5C%5C+double+each+backslash+character&source=bl&ots=nKMB9x2mvA&sig=aWPE004azB5ZDMnWEo2Oh0Kkeo4&hl=en&sa=X&ved=0ahUKEwi9xrm2qqTVAhVnIcAKHVbRAqcQ6AEIQjAF#v=onepage&q=java%20%5C%5C%20double%20each%20backslash%20character&f=false) and [there](http://www.javaworld.com/article/3188545/learn-java/java-101-regular-expressions-in-java-part-1.html).

SubJunk: @Sami32 I wrote a new unit test for this detection which shows it works either way (`\d` or `\\d`) https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1316/commits/85651a905055c6c7426544b607291203208b1962#diff-dc6893ca8f996b0b646af9054ea477f2

PR #1317, opened 2017-06-20
Conversion of 3D video to 2D 

valib:
using renderer setting `Output3DFormat = ml` or `Output3DFormat = mr`


src/main/java/net/pms/encoders/FFMpegVideo.java
```diff
@@ -266,14 +266,20 @@ public FFMpegVideo(PmsConfiguration configuration) {
 			filterChain.add(overrideVF);
 		}
 
-		// Convert 3D video to the other output 3D format
+		// Convert 3D video to the other output 3D format or to 2D using "Output3DFormat = ml" or "Output3DFormat = mr" in the renderer conf
+		String stereoLayout = null;
+		if (media.get3DLayout() != null) {
+			stereoLayout = media.get3DLayout().toString().toLowerCase();
```
Nadahar: @valib You should use ```.toLowerCase(Locale.ROOT)``` on all ```toLower``` or ```toUpper``` conversions that you want to use "English" rules instead of localized rules. The best example of why is the Turkish locale where ```i``` and ```I``` are different letters and won't be changed as you expect if using localized rules.

src/main/java/net/pms/dlna/DLNAMediaInfo.java
```diff
@@ -2888,8 +2888,8 @@ public void setStereoscopy(String stereoscopy) {
 		SBSL,
 		SBSR,
 		HSBSL,
-		OUL,
-		OUR,
+		ABL,
+		ABR,
```
Nadahar: @valib I have no idea what these codes mean. Shouldn't ```OUx``` be used anymore/isn't it valid?

valib: @Nadahar yes it is `OUx` but the FFmpeg filters accept the following options:
```
abl       above-below (left eye above, right eye below) 
abr       above-below (right eye above, left eye below) 
```
and I use them and not convert `OUx` to this convention.

src/main/java/net/pms/dlna/DLNAMediaInfo.java
```diff
@@ -2888,8 +2888,8 @@ public void setStereoscopy(String stereoscopy) {
...
 		HOUL,
```
Sami32: @valib `HOUL` doesn't exist anymore, isn't it ?
https://ffmpeg.org/ffmpeg-filters.html#stereo3d

src/main/java/net/pms/dlna/DLNAMediaInfo.java
```diff
@@ -2888,8 +2888,8 @@ public void setStereoscopy(String stereoscopy) {
 		SBSL,
 		SBSR,
 		HSBSL,
```
Sami32: @valib `HSBSL` do not exist anymore, is that correct ?
https://github.com/FFmpeg/FFmpeg/blob/415f907ce8dcca87c9e7cfdc954b92df399d3d80/libavfilter/vf_stereo3d.c

src/main/java/net/pms/dlna/DLNAMediaInfo.java
valib: @Sami32 you are right I will also change it. The same apply to your comment above.

src/main/java/net/pms/dlna/DLNAMediaInfo.java
```diff
@@ -2916,11 +2918,11 @@ public Mode3D get3DLayout() {
 			case "oulf":
```
Sami32: @valib All the `case "oulf"` should be renamed, is that correct ?
Sorry, i saw that only later.

src/main/java/net/pms/dlna/DLNAMediaInfo.java
```diff
@@ -2916,11 +2918,11 @@ public Mode3D get3DLayout() {
...
 			case "top-bottom (left eye first)":
 				isAnaglyph = false;
-				return Mode3D.OUL;
+				return Mode3D.ABL;
 			case "ourf":
```
Sami32: @valib All the `case "ourf"` should be renamed, isn't it ?
Yes, i know it's just a little "refreshing" / "updating", but since you're on it i thought that it was worst mention it

**P.S.** Don't worry, i shouldn't review it more :wink:

valib: @Sami32 no because this is the value which is given by MediaInfo. The same apply for `oulf`.

Sami32: @valib Sorry, but i didn't saw that value in MediaInfo code source. They seem doesn't exist anymore.
I don't get it with the latest MediaInfo.

Are you able to get / find these values ?

Nadahar: Remember to support older versions of MediaInfo as well

Sami32: Sorry, it has never been a MediaInfo string since the beginning AFAIK:
https://github.com/MediaArea/MediaInfoLib/commit/11ca3e72b2f1fad071a0510e6134bf35dd30442b

valib: @Sami32 @Nadahar my mistake. Those values stand from [here]( https://github.com/UniversalMediaServer/UniversalMediaServer/blob/master/src/main/java/net/pms/dlna/LibMediaInfoParser.java#L352). It is the recognition of the 3D layout from the video filename.

Sami32: @valib Thank you for pointing at this, as i totally missed it.
OK, so we've choose a weird name, as we should have chosen a name similar to FFmpeg nomenclature, IMO.

Sami32:
@valib I will approve it, if you clean that merge commit and squash a little :wink: 
Not that i can't do it myself, but i'm afraid to offend you with such things so i prefer ask you.

src/main/java/net/pms/dlna/DLNAMediaInfo.java
valib: @Sami32 I am not sure what is correct. When you check the naming convention which users use for the 3D videos e.g. on the YouTube it is different to the FFmpeg. What is right?

Sami32: @valib I wasn't aware about Youtube naming convention.
So, to me it looked that as most of the time we use FFmpeg naming convention we should stick to it for a more easy understanding / updating.

That's said, i cannot said that one is more right than the other. But i really don't know where to search for having accurate and updated informations about Youtube naming convention, as we'll need it if they change it. So i hope that i'm the only one in that case.

As i said i'll approve your PR as it is anyways.

PR #1319, opened 2017-06-24
Fix Hebrew Charset detection and add the test for it

valib:
It could fix the issue reported [here](http://www.universalmediaserver.com/forum/viewtopic.php?f=9&t=10238)


Sami32:
@valib Please, could you rebase it on the latest master that i approve it ?

PR #1322, opened 2017-06-25
Show Splash Screen as early as possible

valib:
This could partially solve the problem introduced by @Nadahar in #1208 and show Splash Screen before the oshi is called and could make users wait and not think that UMS doesn't start.

### DIFF
--- a/src/main/java/net/pms/PMS.java
+++ b/src/main/java/net/pms/PMS.java
@@ -415,6 +415,12 @@ public class PMS {
 	 * @throws Exception
 	 */
 	private boolean init() throws Exception {
+		// Splash
+		Splash splash = null;
+		if (!isHeadless()) {
+			splash = new Splash(configuration);
+		}
+
 		// Show the language selection dialog before displayBanner();
 		if (
 			!isHeadless() &&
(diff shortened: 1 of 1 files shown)


SubJunk:
Looks good to me, merge if you want

Nadahar:
This was merged before I had the time to review it. It has a couple of problems as I see it:

* The splash screen is off by default, so it won't make a difference for most users
* The splash screen isn't hidden when the language selection window or the startup wizard is run, which looks buggy to me.

I'm going to do a little PR on the oshi logging anyway (to make it possible to disable if it creates a lot of problems on for some users) and some other small things, so I can add those changes in there.

valib:
@Nadahar I am aware of that but it is very simple to hide it during the language selection window or the startup wizard and show it after that. Also to set the splash screen on by default is only about our decision. So @UniversalMediaServer/developers please vote for that sugestion.

Nadahar:
@valib I know it's easy, I just wanted to make that discussion before it was merged 😉 

Anyway, I've started my other "tweaks" so I've already made 27eb46f4e5d047539e077dfb305b24a695c14a7d.

Nadahar:
I vote for turning it on by default, as otherwise it won't help solve the problem with the extra wait time.

valib:
@Nadahar I agree with your vote and the code you made in 27eb46f4e5d047539e077dfb305b24a695c14a7d was also my original solution. I slightly updated it here #1324

Issue #1181, opened 2017-02-08
Support CustomFFmpegOptions in FFmpeg Audio


onon765trb:
I understand trascoding with tsmuxer also use the setting of FFmpeg Audio. I don't read the source of UMS, sorry.

`CustomFFmpegOptions=` isn't used other than FFmpeg Video transcoding. FFmpeg has many useful audio filters.
For example, soxr is the change of sampling rate, which doen't occur audio delay, inserting silence, squeeze, expand, dynaudnorm avoids a sudden loud sound, etc.

Thank for reading my issues and many comments.

Nadahar:
I think this could be worth it if anyone is up for implementing it.

onon765trb:
What build can I test?

Sami32:
@onon765trb That one:
https://we.tl/VvlZOCoKLZ

onon765trb:
Thank you.

onon765trb:
It works well!

SubJunk:
Done in https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1323

Issue #1179, opened 2017-02-07
Support AudioToolBox encoding


onon765trb:
I request support AudioToolBox encoding.
AudioToolBox is hardware audio tool and outputs high quality audio. It can be used through ffmpeg.
`ffmpeg -i input -c:a aac_at output`
But FfmpegAudio of UniversalMediaServer can't set it. `CustomFFmpegOptions=` doesn't override the setting of FfmpegAudio of UniversalMediaServer.

 /Applications/Universal\ Media\ Server_v8.app/Contents/Resources/osx/ffmpeg  -codecs | grep _at
ffmpeg version N-80408-g39c0b22-tessus Copyright (c) 2000-2016 the FFmpeg developers
  built with Apple LLVM version 6.0 (clang-600.0.57) (based on LLVM 3.5svn)
  configuration: --cc=/usr/bin/clang --prefix=/opt/ffmpeg --as=yasm --extra-version=tessus --enable-avisynth --enable-fontconfig --enable-gpl --enable-libass --enable-libbluray --enable-libfreetype --enable-libgsm --enable-libmodplug --enable-libmp3lame --enable-libopencore-amrnb --enable-libopencore-amrwb --enable-libopus --enable-libschroedinger --enable-libsnappy --enable-libsoxr --enable-libspeex --enable-libtheora --enable-libvidstab --enable-libvo-amrwbenc --enable-libvorbis --enable-libvpx --enable-libwavpack --enable-libx264 --enable-libx265 --enable-libxavs --enable-libxvid --enable-libzmq --enable-version3 --disable-ffplay --disable-indev=qtkit --disable-indev=x11grab_xcb
  libavutil      55. 24.100 / 55. 24.100
  libavcodec     57. 44.100 / 57. 44.100
  libavformat    57. 37.101 / 57. 37.101
  libavdevice    57.  0.101 / 57.  0.101
  libavfilter     6. 46.100 /  6. 46.100
  libswscale      4.  1.100 /  4.  1.100
  libswresample   2.  0.101 /  2.  0.101
  libpostproc    54.  0.100 / 54.  0.100
 DEA.L. aac                  AAC (Advanced Audio Coding) (decoders: aac aac_fixed aac_at ) (encoders: aac aac_at )
 DEA.L. ac3                  ATSC A/52A (AC-3) (decoders: ac3 ac3_fixed ac3_at ) (encoders: ac3 ac3_fixed )
[...]

PR #1323, opened 2017-06-26
Add custom FFmpeg Audio options

Sami32:
As requested here: https://github.com/UniversalMediaServer/UniversalMediaServer/issues/1181 https://github.com/UniversalMediaServer/UniversalMediaServer/issues/1179


onon765trb:
GJ!!!

Sami32:
@ler0y In your case using adding these lines to your renderer will reduce your transcoding for WAV with a sample rate < 48000:
```logtalk
TranscodeAudio = WAV
CustomAudioFFmpegOptions = -af aformat=sample_fmts=u8|s16:sample_rates=48000|96000
```
Fine tune it to your renderer support. Ex. 96000 if it's DTS is supported in WAV.

**P.S.** You can also use the `-apre` option, more powerful, but it also ask more modifications from the user side.

src/main/external-resources/renderers/DefaultRenderer.conf
```diff
@@ -375,6 +375,11 @@ ThumbnailHeight =
 # Default: false
 ChunkedTransfer = 
 
+# Overrides the FFmpeg Audio options in UMS for this renderer.
+# An empty value means that the default UMS settings will be used.
+# Default: ""
+CustomAudioFFmpegOptions =
```
SubJunk: It would be good to rename this to `CustomFFmpegAudioOptions` to match our other conventions

src/main/java/net/pms/configuration/RendererConfiguration.java
```diff
@@ -2262,6 +2263,10 @@ public int getAutoPlayTmo() {
 		return getInt(AUTO_PLAY_TMO, 5000);
 	}
 
+	public String getCustomAudioFFmpegOptions() {
```
SubJunk: Same here

src/main/java/net/pms/encoders/FFmpegAudio.java
```diff
@@ -142,7 +144,7 @@ public synchronized ProcessWrapper launchTranscode(
 		configuration = (DeviceConfiguration)params.mediaRenderer;
 		final String filename = dlna.getFileName();
 		params.maxBufferSize = configuration.getMaxAudioBuffer();
-		params.waitbeforestart = 2000;
+		params.waitbeforestart = 1;
```
SubJunk: Why change this?

src/main/java/net/pms/encoders/FFmpegAudio.java
```diff
@@ -200,30 +202,53 @@ public synchronized ProcessWrapper launchTranscode(
 			cmdList.add("" + params.timeend);
 		}
 
+		String customAudioFFmpegOptions = params.mediaRenderer.getCustomAudioFFmpegOptions();
+
+		// Add audio options (-af, -filter_complex, -ab, -ar, -ac, -c:a, -f, -apre, -fpre, -pre, etc.)
```
SubJunk: This comment could probably be put into `DefaultRenderer.conf` too

src/main/java/net/pms/encoders/FFmpegAudio.java
```diff
@@ -247,6 +272,44 @@ public synchronized ProcessWrapper launchTranscode(
 		return pw;
 	}
 
+	/**
```
SubJunk: Since these are already in `FFmpegVideo` which this extends, we shouldn't need to add them again unless they are different from those ones, in which case the `@override` flag should be present

src/main/java/net/pms/encoders/FFmpegAudio.java
Sami32: No waiting delay while playing music.
Tested by us without any issues since some months, also with my very old CPU.
If i remember correctly, we are talking about 100MB to be buffereized in 2s, it seem out of age, also for me.

src/main/external-resources/renderers/DefaultRenderer.conf
Sami32: I agree. I should admit that i saw it afterward and was too lazzy to change it.

src/main/java/net/pms/encoders/FFmpegAudio.java
Sami32: Damned Github, i cannot see anything except` /**`

Please, feel free to push a commit that i can understand what you mean, for an eventual next time.

src/main/java/net/pms/encoders/FFmpegAudio.java
SubJunk: Ok

src/main/java/net/pms/encoders/FFmpegAudio.java
SubJunk: @Sami32 I just checked and those `parseOptions` methods are identical to the ones in `FFmpegVideo` except for comment changes in the second one, so it would be better to remove those from this file and if you want, make the comment changes in `FFmpegVideo`. Otherwise it's just writing the same code twice which also means bugfixes need to get written twice, etc.

Sami32: I agree that doesn't use the same code twice will be good.

But, just to clear, when i do the changes asked and that errors arise, i just stop brutally to be interested...
So, or you push a commit, or this PR will wait until my mood become better...

SubJunk: @Sami32 "when i do the changes asked" you didn't do the changes asked. I said to add `@override` *if* the methods are different otherwise we shouldn't need to add them, and they are not different :)

SurfaceS:
fix #1181

SubJunk (approved):

SubJunk:
Wow, 7 years! Great work @SurfaceS and @Sami32 😆

PR #1324, opened 2017-06-26
Update splash screen behaviour

valib:
Based on @Nadahar input.


valib:
I prefer the PR #1327

Issue #1325, opened 2017-06-27
Bravia XBR renderer config update request


lopopolo:
It appears the XBR config has not been touched since 2014. I don't know exactly what I'm looking for, but [this spec sheet](https://docs.sony.com/release//specs/XBR65X900E_mksp.pdf) indicates that the newer XBRs at least support mkv:VP8 and webm for video, JPEG images, and FLAC audio.

<img width="244" alt="screen shot 2017-06-26 at 8 33 57 pm" src="https://user-images.githubusercontent.com/860434/27569867-d883d720-5aae-11e7-9824-76722844bfba.png">

lopopolo:
I have a FLAC, VP8 video in an MKV, and VP8 video in a webm.

lopopolo:
Also, this TV is a 4K TV. If the older XBRs are not, does that warrant creating a renderer profile at higher priority for the E series? The suggestion in the pull request that i manually uncomment the max resolution configs seems user hostile.

SubJunk:
Yes it would be good to create an E series profile. If you can provide us with logs we can make an attempt at that

lopopolo:
What do I need to do to capture these logs?

Nadahar:
@lopopolo Follow [these steps](http://www.universalmediaserver.com/forum/viewtopic.php?f=9&t=556). Under "reproduce bugs" try to play the file(s) in question. Attach the resulting zip file in a comment here.

lopopolo:
I've tried my best to pick out the relevant logs. My network has 6 renderers so I wanted to spare you the noise.

My test folder contains the following:

```console
$ ls -1a
./
../
.DS_Store
Kanye West Ft_ DJ Khaled -- WAY TOO COLD.flac
vp8.mkv
vp8.webm
vp9.mkv

```

The XBR-65X900E does not natively support the VP9 codec; however, attempting to play vp9.mkv yields these logs, which indicate that the transcode has failed:

```
DEBUG 2017-06-30 14:47:12.619 [HTTPv2 Request Worker 16] Switching rendering context to 'XBR-65X900E Blue Ivy [FFmpeg]' from 'Chrome [FFmpeg]'
[...]

PR #1326, opened 2017-06-27
Update Sony-BraviaXBR.conf

Sami32:
As requested by this user [here](https://github.com/UniversalMediaServer/UniversalMediaServer/issues/1325).

### DIFF
--- a/src/main/external-resources/renderers/Sony-BraviaXBR.conf
+++ b/src/main/external-resources/renderers/Sony-BraviaXBR.conf
@@ -24,6 +24,10 @@ UpnpDetailsSearch = Sony , XBR-[0-9][0-9]
 SeekByTime = true
 DLNALocalizationRequired = true
 TranscodeAudio = WAV
+#MaxVideoWidth = 3840
+#MaxVideoHeight = 2160
+#SupportedVideoBitDepths = 8,10
+#H264Level41Limited = false
 MuxNonMod4Resolution = true
 CreateDLNATreeFaster = true
 MediaParserV2_ThumbnailGeneration = true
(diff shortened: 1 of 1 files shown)


src/main/external-resources/renderers/Sony-BraviaXBR.conf
```diff
@@ -24,6 +24,10 @@ UpnpDetailsSearch = Sony , XBR-[0-9][0-9]
 SeekByTime = true
 DLNALocalizationRequired = true
 TranscodeAudio = WAV
+#MaxVideoWidth = 3840
```
SubJunk: Why add these?

Sami32: Because i get bored to repeat myself, as more and more users with new renderers ask for that on the forum.
So now i just need to ask them to uncomment these lines in their renderer configuration file.
Anyway, they are just the reality of their renderer, so better add them, or someone will need to read their renderer specification again and again...

If you remark is about the comment sign, you're the reason. As i understood that you don't want set it by default. No problem, i can understand.

Nadahar:
@Sami32 It has to be possible to ask you something without you taking it personally. If this were my PR, I wouldn't have felt bad by this question at all but simply explained the reason. It's natural that anyone that doesn't know your reasons for doing things will ask why, especially when it's not explained/described. That doesn't mean that they're criticising what you've done, it simply means that they don't know/understand why.

Sami32:
@Nadahar What i understood is that he request that i delete these lines, so i just prefered close it and move on, leaving more place for others PRs.
Yes, i never say that i was perfect, sometimes suceptible, sometimes stubborn :stuck_out_tongue:

Nadahar:
@Sami32 What I understood was simply that he asked for an explanation, and didn't want the PR to be merged until it was discussed further.

Sami32:
No need anymore

PR #1327, opened 2017-06-28
Memory logging adjustments

Nadahar:
This PR does some adjustments to #1208 based on past-merge experience. It basically "reverts" the logging done in the ```main``` thread to a slightly tweaked version of what it used to be.

All the OSHI logging has been moved into a new class ```SystemInformation```. This class has both utility methods for gathering and logging this information and is a ```Thread``` subclass that will gather and log the same information in a separate thread if instantiated and started. In addition, I've made the system information logging configurable with:
* Never log system information
* Log system information only when log level is ```TRACE```
* Always log system information

I've set the default to ```Trace only```. The benefits of all this are:
* UMS doesn't need to gather and log this information during normal use, when it has no value
* Gathering the information can take considerable time on some platforms. When run in a separate thread nothing will wait for this, and the results will be logged when they are available.
* It is possible to disable this completely if it turns out to be installations where the JNA calls leads to problems/crashes.

I've also added some fixes, and the splash screen is on by default.

---
This change is [Reviewable](https://reviewable.io/reviews/universalmediaserver/universalmediaserver/1327)

### DIFF
--- a/src/main/java/net/pms/dlna/DLNAMediaInfo.java
+++ b/src/main/java/net/pms/dlna/DLNAMediaInfo.java
@@ -112,6 +112,7 @@ public class DLNAMediaInfo implements Cloneable {
 		mutableAudioOrVideoContainers.put(FormatConfiguration.THREEGPP2, new AudioVariantInfo(new THREEG2A(), FormatConfiguration.THREEGA));
 		// XXX WEBM Audio is NOT MKA, but it will have to stay this way until WEBM Audio is implemented.
 		mutableAudioOrVideoContainers.put(FormatConfiguration.WEBM, new AudioVariantInfo(new MKA(), FormatConfiguration.WEBA));
+		mutableAudioOrVideoContainers.put(FormatConfiguration.WMV, new AudioVariantInfo(new WMA(), FormatConfiguration.WMA));
 
 		audioOrVideoContainers = Collections.unmodifiableMap(mutableAudioOrVideoContainers);
 	}
--- a/src/main/java/net/pms/newgui/Splash.java
+++ b/src/main/java/net/pms/newgui/Splash.java
@@ -36,8 +36,6 @@ import org.slf4j.LoggerFactory;
 public class Splash extends JFrame implements MouseListener {
 	private static final long serialVersionUID = 2357524127613134620L;
 	private static final Logger LOGGER = LoggerFactory.getLogger(Splash.class);
-	private JLabel imglabel;
-	private ImageIcon img;
 	private PmsConfiguration configuration;
 
 	/**
@@ -53,8 +51,8 @@ public class Splash extends JFrame implements MouseListener {
 			return;
 		}
 
-		img = new ImageIcon(getClass().getResource("/resources/images/splash.png"));
-		imglabel = new JLabel(img);
+		ImageIcon img = new ImageIcon(getClass().getResource("/resources/images/splash.png"));
+		JLabel imglabel = new JLabel(img);
 		imglabel.setBounds(0, 0, img.getIconWidth(), img.getIconHeight());
 		setSize(imglabel.getWidth(), imglabel.getHeight());
 		setUndecorated(true);
@@ -63,6 +61,8 @@ public class Splash extends JFrame implements MouseListener {
 		setLayout(null);
 		add(imglabel);
 		imglabel.addMouseListener(this);
+		img = new ImageIcon(getClass().getResource("/resources/images/icon-32.png"));
+		setIconImage(img.getImage());
 		if (System.getProperty("console") == null) {
 			setVisible(true);
 		}
--- a/src/main/external-resources/UMS.conf
+++ b/src/main/external-resources/UMS.conf
@@ -1517,7 +1517,7 @@ search_recurse_depth=
 log_level = 
 
 # Logfile name
-# ---------------------
+# ------------
 # The name of the standard log file. This is just the file name, 
 # no path information must be given here. The location of the log
 # file is decided by rules that varies by platform and some other
@@ -1526,7 +1526,7 @@ log_level =
 logging_logfile_name =
 
 # Logging Buffered
-# ---------------------
+# ----------------
 # By default the log file(s) are flushed/written to disk after each new log 
 # event. SSD drives can be written to a fixed number of times only, and logging
 # can be a great source of wear on such drives. Buffered logging makes UMS 
@@ -1543,7 +1543,7 @@ logging_logfile_name =
 logging_buffered =
 
 # Logging Filter Console
-# ---------------------
+# ----------------------
 # A separate filter for the console that filters after the root filter have
 # done it's filtering. It's not possible to get back things filtered out by
 # Log Level above, but you can further filter out more before it's written to
@@ -1553,7 +1553,7 @@ logging_buffered =
 logging_filter_console =
 
 # Logging Filter Log Tab
-# ---------------------
+# ----------------------
 # A separate filter for the log window in the "Logs" tab in the GUI that
 # filters after the root filter have done it's filtering. It's not possible to
 # get back things filtered out by Log Level above, but you can further filter
@@ -1563,12 +1563,30 @@ logging_filter_console =
 logging_filter_logs_tab =
 
 # Logging Logs Tab Linebuffer
-# ---------------------
+# ---------------------------
 # Specifies how many lines the log windows on the "Logs" tab in the GUI will
 # store before deleting the oldest entries. This will decide the length of the
 # scrollable area.
 # Default: 1000
 logging_logs_tab_linebuffer = 
+
+# Log System Information
+# ----------------------
+# Specifies if or when to log information about the OS, CPU and memory. This
+# is useful for diagnosing problems, but isn't stricly necessary during normal
+# use. On some platforms, acquiring the information can take some time,
+# although that doesn't mean that it uses a lot of resources. There are three
+# modes:
+#
+# - Never: Never log system information. This is useful it gathering of the
+#   information leads to an error or other problem).
+# - Trace only: Only log system information if the log level is "TRACE". This
+#   is the default which only triggers gathering of system information when
+#   the log level commonly used for diagnostics is activated.
+# - Always: Always log system information.
+#
+# Default: Trace only
+log_system_info = 
 
 #---< Syslog settings >----------------
 #
@@ -1581,7 +1599,7 @@ logging_logs_tab_linebuffer =
 # want to acquire in-depth knowledge about LogBack configuration. 
 
 # Logging Use Syslog
-# ---------------------
+# ------------------
 # Activates syslog logging and disables file logging. That enables you to 
 # send all logging to a differnt computer running a syslog server completely
 # removing all local disk I/O from logging, or to collect logs in a centralized
@@ -1592,7 +1610,7 @@ logging_logs_tab_linebuffer =
 logging_use_syslog = 
 
 # Logging Syslog Host
-# ---------------------
+# -------------------
 # Host name or IP address for the syslog server. Use "localhost" for logging 
 # to a syslog server on the same computer. The name must be resolvable (via
 # DNS, WINS, HOSTS etc.) or be an IP address for syslog logging to be activated.
@@ -1600,13 +1618,14 @@ logging_use_syslog =
 logging_syslog_host =
 
 # Logging Syslog Port
-# ---------------------
+# -------------------
 # UDP port to use for syslog logging. This should normally be left at the 
 # default value.
 # Default: 514
 logging_syslog_port =
 
 # Logging Syslog Facility
+# -----------------------
 # Syslog facility can be specified for filtering purposes, but is limited to
 # one of the following values: AUTH, AUTHPRIV, DAEMON, CRON, FTP, LPR, KERN,
 # MAIL, NEWS, SYSLOG, USER, UUCP, LOCAL1, LOCAL2, LOCAL3, LOCAL4, LOCAL5,
--- a/src/main/java/net/pms/PMS.java
+++ b/src/main/java/net/pms/PMS.java
@@ -19,7 +19,6 @@
 
 package net.pms;
 
-import static org.apache.commons.lang3.StringUtils.isNotBlank;
 import ch.qos.logback.classic.Level;
 import ch.qos.logback.classic.LoggerContext;
 import com.sun.jna.Platform;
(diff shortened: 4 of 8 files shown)


src/main/java/net/pms/newgui/Splash.java
```diff
@@ -63,6 +61,8 @@ public Splash(PmsConfiguration config) {
 		setLayout(null);
 		add(imglabel);
 		imglabel.addMouseListener(this);
+		img = new ImageIcon(getClass().getResource("/resources/images/icon-32.png"));
+		setIconImage(img.getImage());
```
valib: @Nadahar where and when the frame icon is displayed?

Nadahar: @valib On the "task bar" in Windows for example, or of you use ```Alt + Tab``` to switch foreground application., I'm sure it's used in similar situations for other OS'es.

valib: @Nadahar ahh I see. I was loking for the icon in the indicative area of the main panel and missed the main window icon which before your change has shown an Java icon. Good job 👍 
I recommend to merge this PR and I will delete the #1324

Sami32:
@Nadahar I remember you complain about the thread number incorrectness concerning your CPU.
But i don't see any use of OSHI core informations, `getPhysicalProcessorCount()` and `getLogicalProcessorCount()`, into the thread number variable of the video or audio engines, `getNumberOfCpuCores()`, so i'm wondering if i missed something or that you've some separated plan ?

Nadahar:
@Sami32 I'm not sure I understand what you mean. Are you referring the the "thread number" in the transcode settings? It might be something I've forgotten now, but I think that is correct. I think what my argument was about was the terms used, and that threads, cores and virtual cores were mixed together. There's really no connection between the number of threads and the number or (virtual) cores, except that to use all cores you need to have at least one thread per core as one thread can only run in the context of one core.

I hope we can use the OSHI information for more things later, but I think it's premature to do so now. We need to get some more experience with it first, and to see if it generally works well under all circumstances or not.

src/main/java/net/pms/dlna/DLNAMediaInfo.java
```diff
@@ -112,6 +112,7 @@
 		mutableAudioOrVideoContainers.put(FormatConfiguration.THREEGPP2, new AudioVariantInfo(new THREEG2A(), FormatConfiguration.THREEGA));
 		// XXX WEBM Audio is NOT MKA, but it will have to stay this way until WEBM Audio is implemented.
 		mutableAudioOrVideoContainers.put(FormatConfiguration.WEBM, new AudioVariantInfo(new MKA(), FormatConfiguration.WEBA));
+		mutableAudioOrVideoContainers.put(FormatConfiguration.WMV, new AudioVariantInfo(new WMA(), FormatConfiguration.WMA));
```
Sami32: Why not, but it's only look as a quick patch to me.
As such files should be ASF specifications, that have audio and video MIME type, `audio/x-ms-asf`, and not WMV.
Apparently WP will not play such files as WMA or WMV:
https://github.com/videolan/vlc/blob/master/share/vlc.desktop.mimetypes#L26
https://docs.microsoft.com/en-us/windows/uwp/audio-video-camera/supported-codecs
Note: ASF is totally broken, not working, or has never worked at all, but it's not a big deal to fix it, at least by the way i did.

That was my 2 cents.

Sami32:
@Nadahar I'm not sure i understood you. But if your core numbers are correctly displayed in the GUI now it's great, as you're the only having reported that issue. Me with my only one core CPU, evetything is fine LOL

src/main/java/net/pms/dlna/DLNAMediaInfo.java
Nadahar: @Sami32 I haven't seen how you fixed it, so I don't know, but I disagree that this is "a quick patch". This was something that was forgotten in #1285 and should have been a part of that PR. The ASF problem is "a separate issue" as far as I can see. The "confusion" is that WM use a previous version of ASF as its container structure, so in a sense a WM file is also an ASF file. The opposite isn't true though. If we don't throw this confusion into mix, the above is, as far as I can see, correct. UMS parse WM as WMV "by default", and if the initial parsing of the format is correct (that is, doesn't parse ASF as WM), then this should work fine.

The ASF problem should be addressed, but my intention here was just to fix what was forgotten in #1285.

Nadahar:
@Sami32 The main confusion is that many people don't understand the difference between cores and threads. One core can run an unlimited number of threads, so there is no direct correlation between the two.

When it comes to if the current implementation is working, I'll have to test it on my development computer (which I'm not currently in front of). 

@valib Thank you for the approval. It seems like you and me are the only ones that has "voted" on the splash default, so it seems like it is unanimous consent for changing this.

I'll wait to see if @Sami32 has any objections before I proceed. As this PR is a "fix" only, it shouldn't be very controversial.

Sami32:
@Nadahar You can merge it if you want. I didn't need to approve as @valib already done.
I never read any report similar to your, but it'll not hurt and it's easy to disable from the GUI, so it's OK.

src/main/java/net/pms/dlna/DLNAMediaInfo.java
Sami32: AFAIK, WMV is only video/x-ms-wmv and no audio only.
So it shouldn't be considered as such.
Only ASF can be.

Nadahar:
@Sami32 Reviews and approvals aren't only there to be able to merge. I will approve a PR even if it's already approved by others if I have the time, but most of the time PRs are merged as soon as they are approved.

What report of mine are you referring to?

Sami32:
@Nadahar That your UMS GUI display an incorrect number of cores.

Nadahar:
@Sami32 I tested it now, and I see that you are correct now. It does display the wrong number of cores, but I guess it isn't too much a problem. I have 6 cores with [hyper-threading](https://en.wikipedia.org/wiki/Hyper-threading) which leads UMS to report that I have 12 cores. That isn't correct, but it's semantics really. Hyper-threading is some Intel trick that somehow make 1 core "behave as" 2. There's no or very little performance gain, so I'm still not sure how hyper-threading is useful. As far as I understand, Intel i3's are dual-core with hyper-threading, and will be reported as having 4 cores by UMS. Likewise, Intel i7's are normally quad-cores with hyper-threading, and will be reported as having 8 cores.

That said, it doesn't have any negative effect that I know of. I means that when transcoding, my computer will use multiple threads per core - which isn't really ideal from a performance point of view, but I doubt the difference is big enough to actually matter. What bothers me more, is to be a part of "contributing" to dis-information, because there is so much wrong information out there as it is. 

Ideally the information from OSHI could be used for this. When it comes to what actually gives the best performance (using 6 or 12 threads in my case), only a test could tell. It also seems that the "-threads" parameter for FFmpeg is [a lot more complicated](https://stackoverflow.com/a/41010102/363791) so I don't know what would be the best approach to this. It's probably not what we're currently doing.

Regarding the 91 seconds I've said several times that it got a lot quicker after a restart. However, it seems like the time is slowly increasing the longer the computer has been running, but I haven't really tested this. It "feels" like it takes longer now than it did right after the reboot. In any case, it takes several seconds in many cases, and there's no good reason why the user should have to wait for that so I think doing it in a separate thread makes sense regardless of why my Windows installation behaves as it does.

valib:
@Sami32  we can simply change the TrTab2.24 to something like 

> CPU threads to use when enabled for engine: (you seem to have %d physical or virtual cores)

 to meet your requirement.

Nadahar:
I did some quick, unscientific, transcoding test usings different values for ```-threads```. The values vary slightly from one run to another, so I've made "estimated averages" for the speed values.

Video conversion | Audio conversion | ```-threads``` value | Transcode speed
----------------- | -------------------| ------------------ | ------------------
wmv3 -> mpeg2 | wmav2 -> mp2 | not specified | 37.8x
wmv3 -> mpeg2 | wmav2 -> mp2 | 1 | 28x
wmv3 -> mpeg2 | wmav2 -> mp2 | 2 | 33.8x
wmv3 -> mpeg2 | wmav2 -> mp2 | 3 | 36.6x
wmv3 -> mpeg2 | wmav2 -> mp2 | 4 | 37.9x
wmv3 -> mpeg2 | wmav2 -> mp2 | 6 | 37.8x
wmv3 -> mpeg2 | wmav2 -> mp2 | 8 | 37.5x
wmv3 -> mpeg2 | wmav2 -> mp2 | 12 | 38.2x
wmv3 -> mpeg2 | wmav2 -> mp2 | 40 | 34.1x
wmv3 -> mpeg2 | wmav2 -> mp2 | 1000 | 2.57x
wmv3 -> h264 | wmav2 -> vorbis | not specified |  13.6x
wmv3 -> h264 | wmav2 -> vorbis | 1 |  2.26x
wmv3 -> h264 | wmav2 -> vorbis | 2 |  4.8x
wmv3 -> h264 | wmav2 -> vorbis | 4 |  8.11x
wmv3 -> h264 | wmav2 -> vorbis | 6 |  9.89x
wmv3 -> h264 | wmav2 -> vorbis | 8 |  11x
wmv3 -> h264 | wmav2 -> vorbis | 12 |  13.1x
wmv3 -> h264 | wmav2 -> vorbis | 40 |  13.8x
wmv3 -> h264 | wmav2 -> vorbis | 100 |  13.2x
wmv3 -> h264 | wmav2 -> vorbis | 1000 |  7.29x

Based on my very unscientific tests it seems like the best thing to do would be to not specify the number of threads at all and let FFmpeg figure it out. But, it might give other results for other conversions and other hardware/platforms.

Nadahar:
@valib Haven't we had that discussion before? It wouldn't be any more correct that way, "physical *or* virtual". It would probably be better to say just "logical cores". In any case, it doesn't really matter that much.

Nadahar:
It seems that the discussion isn't about this PR, so I'll merge it.

valib:
@Nadahar agree with you that it is not so important. About that discussion I am not sure because I am quite old men and you don't anticipate I remember everything. 😃

src/main/java/net/pms/dlna/DLNAMediaInfo.java
Nadahar: @Sami32 As with all the "audio or video containers" UMS parse it as video first, and then change it to audio later if the content turns out to be only a single audio track. WMV and WMA aren't "formats", Windows Media is. Windows Media can be either audio/video or just audio (or probably just video as well). The point is that as long as it *is* a Windows Media (WM) format and it only contains one audio track, we should treat it as audio media (and thus WMA). The file extension isn't really relevant in that context.

If, like you say, ASF is parsed as WMV, that is the real problem - as that's simply not always the case. I have very few ASF files and MediaInfo claims that all of them is WM. I don't know what's really the case, a non-WM ASF file would be needed to test where it goes wrong.

Nadahar:
@Sami32 There seems to be a lot of confusion about the WM/ASF definitions. I found this definition, which might actually be a reasonable way to separate the two:

> The .wma file name extension is used for Windows Media Audio (WMA) files that contain audio compressed with one of the Windows Media Audio codecs.
> 
> The .wmv file name extension is used for Windows Media Video (WMV) files that contain both audio and video compressed with the Windows Media Audio and Video codecs.
> 
> The .asf file name extension is used for content that includes audio only or both audio and video. The content in an .asf file could be encoded by a variety of different codecs, including Windows Media Audio and Video codecs.

Sami32:
@Nadahar It what i tried to say to you about ASF.

BTW, nice CPU encoding tests. I'm happy to see that my 1 core CPU is more optimum than your for MPEG2 :rofl:

Nadahar:
@Sami32 
>  I'm happy to see that my 1 core CPU is more optimum than your for MPEG2 🤣

How do you figure that? You can't compare if the source file is different, and as far as I know you don't have my source file 😉

Nadahar:
@Sami32 If you want to compare, I've [uploaded the file I used](http://superbum.org/downloads/oa-1962-0502-htrafik-bussar.asf). I used no other FFmpeg options, I simply used:
```powershell
ffmpeg -threads <n> -i <inputfile> -threads <n> test.ts
```
I did *not* transcode to ```null```, even though that might have been more relevant as disk speed would have less impact. In any case, I never meant to compare this with other computers, so I didn't think of that as relevant. It was a lazy way of setting the output parameters simply using the output extension 😉

Sami32:
@nadahar Dreams are the only thing that no one can steal you :stuck_out_tongue:

Nadahar:
@Sami32 Now I'm very exited about the results for a really optimized CPU 👼

Sami32:
@Nadahar How much do you want to optimize mine ? :rofl: 

P.S. I confess that i don't know how to interpret most of the Emoji, outside of the very basic ones.

Nadahar:
@Sami32 I don't know either, most of the emojis are so badly rendered that I can't even hope to guess what they mean. I used ```:angel``` above, but I replaced it with ```:innocent``` which fit what I meant to express better. It's simply "old habbit", because on skype ```(angel)``` is similar.

Sami32:
@Nadahar Thanks for sharing.
I didn't even know `:rofi:` before seing you used it, and i didn't found easy to search the meaning in all the Github Emoji. As it's the second time that i see you use this one i was wondering about that one as well. So now i can go sleep smarter :wink: 
Angel i can understand, son of god, but innocent ? no kidding :laughing:

Nadahar:
@Sami32 You know that it is `:rofL`, not `:rofI` (Rolling On the Floor Laughing)? 😉 

You can't go to bed until you've given me some *optimized* results though.

Sami32:
@Nadahar I didn't, i told you that i mix l and 1, but it seem that i should add some more LOL.
Why they didn't choosen `:lol:` ?

About which results are you talking ? too late for me now.

Nadahar:
@Sami32 Results from you converting the file I've uploaded to MPEG2: https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1327#issuecomment-311841833

Nadahar:
@Sami32 I mean the results produced by you *converting* the uploaded file on your computer 😉 

I'm just kidding you though, you were the one that claimed:
>  I'm happy to see that my 1 core CPU is more optimum than your for MPEG2 🤣

Sami32:
@Nadahar As i said same same; did i mention that the decimal need to be moved on the left ? just a details though :rofl:

Note: Thanks for the sample, again one that make our media tool to bug :wink:

Nadahar:
@Sami32 
> As i said same same; did i mention that the decimal need to be moved on the left ?

Which decimal/number?

Sami32:
This one: `2.8x`

Nadahar:
@Sami32 As long as it's above 1, it should work. Let's not talk about how much above 😇

PR #1328, opened 2017-07-01
NPE in RendererConfiguration.calculatedSpeed()

Nadahar:
While looking at [this output](https://github.com/UniversalMediaServer/UniversalMediaServer/issues/1325#issuecomment-312385834) in #1325, it seems like the NPE is caused by ```getSpeedInMBitsStored()``` returning ```null```. I've tried to avoid that situation. This "fix" will lead to a maximum value of ```0``` being returned if no results are availabe from "speedstats" and no maximum bandwidth is configured on the renderer. That's not ideal, but I don't know what else to return. It's surely better than crashing with a NPE.

The related code is generally extremely fragile and unstructured. I see a lot of potentiel problems, both with thread-safety and missing handling of almost anything "not going according to plan". I won't get involved with another one of those major refactorings now, so this is just an attempt to patch *one of* all the problems with this code.

I've never looked at the "speedstats" code before, but now that I have I'd say it's in so bad shape that I think hiding it until it's fixed would probably be a better option that using it as it is. This isn't the first time I've seen it end up with some very questionable results. The "math" isn't really explained, but I can see from the log that the parsing fails regularly and that all the larger packet sizes are rejected by the underlying OS (macOS) in the log I looked at in #1325. Using only small packet sizes probably means that the results are very inaccurate.

In any case, it can't hurt to prevent a NPE, can it?

### DIFF
--- a/src/main/java/net/pms/io/BasicSystemUtils.java
+++ b/src/main/java/net/pms/io/BasicSystemUtils.java
@@ -223,6 +223,11 @@ public class BasicSystemUtils implements SystemUtils {
 		return timeString;
 	}
 
+	@Override
+	public int getPingPacketFragments(int packetSize) {
+		return ((packetSize + 8) / 1500) + 1;
+	}
+
 	/**
 	 * Return the proper tray icon for the operating system.
 	 *
--- a/src/main/java/net/pms/configuration/RendererConfiguration.java
+++ b/src/main/java/net/pms/configuration/RendererConfiguration.java
@@ -1,5 +1,6 @@
 package net.pms.configuration;
 
+import static org.apache.commons.lang3.StringUtils.isBlank;
 import com.sun.jna.Platform;
 import java.awt.Color;
 import java.awt.event.ActionEvent;
@@ -9,6 +10,7 @@ import java.io.IOException;
 import java.io.Reader;
 import java.net.InetAddress;
 import java.util.*;
+import java.util.Map.Entry;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 import java.util.regex.Matcher;
@@ -2538,25 +2540,26 @@ public class RendererConfiguration extends UPNPHelper.Renderer {
 
 	public String calculatedSpeed() throws InterruptedException, ExecutionException {
 		String max = getString(MAX_VIDEO_BITRATE, "");
-		for (InetAddress sa : addressAssociation.keySet()) {
-			if (addressAssociation.get(sa) == this) {
-				Future<Integer> speed = SpeedStats.getInstance().getSpeedInMBitsStored(sa, getRendererName());
-				if (max == null) {
-					return String.valueOf(speed.get());
-				}
-				try {
-					Integer i = Integer.parseInt(max);
-					if (speed.get() > i && i != 0) {
-						return max;
-					} else {
+		for (Entry<InetAddress, RendererConfiguration> entry : addressAssociation.entrySet()) {
+			if (entry.getValue() == this) {
+				Future<Integer> speed = SpeedStats.getInstance().getSpeedInMBitsStored(entry.getKey(), getRendererName());
+				if (speed != null) {
+					if (max == null) {
 						return String.valueOf(speed.get());
 					}
-				} catch (NumberFormatException e) {
-					return String.valueOf(speed.get());
+					try {
+						Integer i = Integer.parseInt(max);
+						if (speed.get() > i && i > 0) {
+							return max;
+						}
+						return String.valueOf(speed.get());
+					} catch (NumberFormatException e) {
+						return String.valueOf(speed.get());
+					}
 				}
 			}
 		}
-		return max;
+		return isBlank(max) ? "0" : max;
 	}
 
 	/**
--- a/src/main/java/net/pms/configuration/RendererConfiguration.java
+++ b/src/main/java/net/pms/configuration/RendererConfiguration.java
@@ -2542,7 +2542,7 @@ public class RendererConfiguration extends UPNPHelper.Renderer {
 		String max = getString(MAX_VIDEO_BITRATE, "");
 		for (Entry<InetAddress, RendererConfiguration> entry : addressAssociation.entrySet()) {
 			if (entry.getValue() == this) {
-				Future<Integer> speed = SpeedStats.getInstance().getSpeedInMBitsStored(entry.getKey(), getRendererName());
+				Future<Integer> speed = SpeedStats.getInstance().getSpeedInMBitsStored(entry.getKey());
 				if (speed != null) {
 					if (max == null) {
 						return String.valueOf(speed.get());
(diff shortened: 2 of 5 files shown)


SubJunk:
Reviewed the code but didn't test

Nadahar:
I haven't tested it either, I don't know how to reproduce the problem.

Nadahar:
I've done some very basic tests without getting any exception. The measurements seems to be way off on macOS though. 

While testing from my Windows 7 computer, I got reasonable results:
```powershell
Renderer TX-NR807 has an estimated network speed of 52.48892905803443 Mb/s
Renderer Kodi (Bozo) has an estimated network speed of 934.3658666666668 Mb/s
```
My network is gigabit, but I suspect that the Onkyo might have a 100 Mb/s NIC, which would make 52 a very reasonable value (about half). Kodi is running on the same computer, so it's probably limited only by the NIC or the OS itself.

On my macOS VM however, the results seems very questionable:
```powershell
Renderer TX-NR807 has an estimated network speed of 5.357146229676806 Mb/s
Renderer Kodi (Bozo) has an estimated network speed of 9.066508312054433 Mb/s
```

On a Kubuntu VM on the same server (i.e. the same physical NIC and network), the results look much more reasonable (although Kodi is gigabit, I suspect that there's some interference on that network cable from running along the 230V AC being only Cat 5):
```powershell
Renderer TX-NR807 has an estimated network speed of 45.32595304217296 Mb/s
Renderer Kodi (Bozo) has an estimated network speed of 284.16341832115165 Mb/s
```

I'd say that short of anyone fixing automatic maximum bandwidth on macOS, it should be disabled to that OS. No exceptons related to SpeedStat were logged on either test though.

Nadahar:
I've made some adjustments to try to make ```SpeedStats``` a bit more accurate for macOS without having to refactor the whole thing.

The results from my macOS VM with the new code is:
```powershell
Renderer TX-NR807 has an estimated network speed of 112.06554128037271 Mb/s
Renderer Kodi (Bozo) has an estimated network speed of 233.80491300062758 Mb/s
```
Even though I think the TX-NR807 speed is about the double of what I'd expect, it's only an assumption that the its network interface is 100 Mb/s. The results are fairly consistent, and the codi result looks reasonable, so I'm not quite sure why this happens.

Nadahar:
I've tested the latest commit on Windows and Linux, and the results are the same as they were. I've been trying to figure out why the Onkyo results are too high from macOS, and I think I have an idea. 

The reason ```SpeedStats``` comes up with crazy low values for maxOS as it is today, is because macOS limits ICMP packet size to ~8400 bytes. 

UMS use these packets sizes when "measuring":
```java
int[] sizes = {512, 1476, 9100, 32000, 64000};
```
As you can see, only the two smalles are allowed, and the others errors out with "message too large". The speed is calculated by dividing the packet size (* 2) with the time it takes to echo, and the average for 3 pings were used - I've updated this to 5 to get a more precise average. Small packets give the least reliable values, which is probably why the values came out completely wrong. One could argue that the small sizes shouldn't be used at all, as they are pretty much useless to measure speed. 

The way I "solved" the problem with the small maximum packet size it that if the package size is above 8000 bytes, I split them into multiple ping requests in rapid succession. That means that if a request for 5 packets of 32000 bytes are requested, on macOS 20 packets of 8000 bytes are sent instead (which isn't rejected by macOS and amounts to the same amount of data). The problem is that macOS waits between sending each packet, so there's a "small" gap there. I've set this delay to 100 ms, which is the lowest allowed value without admin privileges. The problem is that 100 ms is a long time in this context.

Network speed isn't the way we think of it, it's no like the packets travel at a certain speed that is limited by the bandwidth. They always travel at the speed of light, what is limiting the bandwidth is when the network interfaces can't cope or when the data is corrupted. If the data is corrupted (checksums are used to determine integrity), the receiving end asks the transmitting end to retransmitt the package. This can happen multiple times, and is why Wifi networks are typically so much slower in reality than in theory. If the network interfaces can't cope, packets are queued in a buffer and (for TCP) signals are sent back to the sender to take it easy as the buffer is filling up. For UDP there is no such control mechanism, so packets are simply dropped/discarded.

Whether it's due to queuing or retransmissions, the effect is more or less the same: The speed is high as long as the traffic is low. When transferring small amounts of data over a line, there's no difference in speed from a 10 Mb/s and a 1 Gb/s link. The difference is that the 1 Gb/s link can take much more before it starts queuing packets and throttle down TCP traffic. 

Because of the 100 ms delay between the packets, the queues/buffers has some time to clear themselves in between packets, and you don't get the same congestion as you would if the data was sent continuously. In short, there's no way to know how long it takes to send a large packet using multiple smaller packets as long as they aren't sent without any delay or in parallel. If we really wanted to take this all the way, we could spawn one thread per packet and issue them all "at once". That would probably give us a more accurate result when taking the average, but it would require a lot of refactoring.

My guess it that my Onkyo has a slow network interface which "benefits" especially from the pauses between the packages, which is why there's a particular deviation for its results.

I don't see any other way to do this except using multiple threads. There's something called TCP echo, but I don't know enough about it or if the typical renderer will reply to it. It could be a better solution since it's would support larger payload, and since the transfer of media use TCP it would also give a more accurate measurement of what we need to know. But this is a lot of refactoring and testing too, so I'll leave it alone for now.

What do you guys think? I think these values are much more useful, even if they are a bit too high in some circumstances on macOS. The current implementation is completely useless on macOS because the values are so low that playback is in reality impossible.

src/main/java/net/pms/io/BasicSystemUtils.java
```diff
@@ -223,6 +223,11 @@ public String parsePingLine(String line) {
 		return timeString;
 	}
 
+	@Override
+	public int getPingPacketFragments(int packetSize) {
+		return ((packetSize + 8) / 1500) + 1;
+	}
```
Sami32: Do the arbitrary MTU choice was done because it's cumbersome to acces the different OS registries, or because it'll not make much difference in the speed estimation anyway ?

Nadahar: Both I guess. It wasn't my choice, the code was already like this - but I can imagine that obtaining the MTU size for the network interface in question isn't trivial from Java. Also, you'd need to know the MTU size of the device on the other end and of any routers in between. It's not really viable to get it.

I don't think it make a whole lot difference to the speed estimate either. The MTU size determines how many fragments the packages that's larger than the MTU is divided into. The only "consequence" that has when measuring throughput is that there's a small overhead (20 bytes or so) per fragment because of fragment headers. Even if the MTU size is changed (which is rare AFAIK), chances are that it wouldn't be changed enough to make a real difference.

MTU size of 1500 is the maximum allowed by a standard [Ethernet frame](https://en.wikipedia.org/wiki/Ethernet_frame) as defined in [IEEE_802.3](https://en.wikipedia.org/wiki/IEEE_802.3). Some server/high-speed environmens use [jumbo frames](https://en.wikipedia.org/wiki/Jumbo_frame), but I really don't think it's relevant in this context. For jumbo frames to work, all equipment must support and be configured to use it. I doubt there's any renderer device out there where you can adjust the MTU size at all.

PR #1332, opened 2017-07-03
Update OSHI to 3.4.3

Nadahar:
This resolves the JNA and slf4j dependency version conflicts between OSHI and UMS.

@Sami32 It would be nice if you could test if this works on your computer without changing the JNA version.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -507,17 +507,7 @@
 
 				Check the net.java.dev.jna:jna-platform version before upgrading
 			-->
-			<version>3.4.2</version>
-			<exclusions>
-				<exclusion>
-					<groupId>org.slf4j</groupId>
-					<artifactId>slf4j-api</artifactId>
-				</exclusion>
-				<exclusion>
-					<groupId>net.java.dev.jna</groupId>
-					<artifactId>jna-platform</artifactId>
-				</exclusion>
-			</exclusions>
+			<version>3.4.3</version>
 		</dependency>
 	</dependencies>
 	<build>


Nadahar:
I've done basic tests (Maven build + running UMS) without any issues on Windows 7, macOS 10.12.5 and Kubuntu 16.04.1.

Sami32:
I tested it, and that still doesn't work on Windows XP SP3 32-bit.

So JNA 3.4.0. seem to be problematic in my case.

Nadahar:
@Sami32 
> So JNA 3.4.0. seem to be problematic in my case.

You mean 4.4.0, right?

How does it not work, what happens?

Sami32:
Yes, 4.4.0.
[Same same](https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1208#issuecomment-310133119) as already explained.

I can always make my build, and don't see much enhancements between JNA 4.3.0 and 4.4.0 anyway, so it's up to you.

Nadahar:
@Sami32 It's not about the enchancements between JNA 4.3.0 and 4.4.0. We have to "keep in sync" with other libraries or we'll get in a lot of trouble, and we also need to be able to use future JNA versions. Because of that we need to figure out what the problem is.

Are you saying that you get a JVM crash when running UMS or building UMS, or does the build process work just fine?

Nadahar:
@Sami32 The strange thing is that it works perfectly fine on my WinXP VM:
```powershell
18:36:17 INFO  System information:
  JVM: Java HotSpot(TM) Client VM 1.8.0_131 (32-bit) by Oracle Corporation
  OS: Microsoft Windows XP SP3 build 2600 32-bit
  CPU: Intel(R) Core(TM)2 Quad CPU    Q9650  @ 3.00GHz with 1 core
  Physical Memory: 3.0 GiB
  Free Memory: 2.6 GiB
  Maximum JVM Memory: 742.4 MiB
```

This time I've done it all the "old fashion way": 
* Built with ```mvn package -DforkCount=0```
* Ran the resulting ```UMS.exe trace```

Sami32:
With Java 8 version.

As i did have the same issue with Java 8 as well, though i don't have tested it again here, it have hight probability to be related to my old CPU, so i think that it could reasonnably be merged, IMO.

Nadahar:
@Sami32 While this PR doesn't make it any worse, I was expecting it to solve your issue. I'd really like to figure out why it doesn't work.

Sami32:
Thank you.
JNA 4.3.0 work fine for me, so i'm suffisantly lucky :wink:

PR #1336, opened 2017-07-11
AVI / MJPEG support enhancement

Sami32:
Following [this user report](http://www.universalmediaserver.com/forum/viewtopic.php?f=9&t=10097&p=31696#p31696) on the forum, that tried to play his family AVI videos, taken from cameras and phones years ago, without success.
It give me the idea that he should not be the only one, and that our AVI / MJPEG support was poor, so i did this PR.
```logtalk
DEBUG 2017-06-16 20:21:01.494 [New I/O worker #13] net.pms.dlna.DLNAResource Starting transcode/remux of 2004-jonathansroomjanuary1.avi with media info: container: avi, bitrate: 9217159, size: 122127360, video tracks: 1, video codec: und, duration: 00:01:46.00, width: 320, height: 240, frame rate: 5.000, thumb size: 26502, mime type: video/avi
```
```logtalk
DEBUG 2017-06-16 20:21:01.869 [ffmpeg64.exe-25-2] net.pms.io.OutputTextLogger     ISBJ            : JPGVideo
DEBUG 2017-06-16 20:21:01.869 [ffmpeg64.exe-25-2] net.pms.io.OutputTextLogger   Duration: 00:01:46.00, start: 0.000000, bitrate: 9217 kb/s
DEBUG 2017-06-16 20:21:01.869 [ffmpeg64.exe-25-2] net.pms.io.OutputTextLogger     Stream #0:0: Video: rawvideo, bgr24, 320x240, 9233 kb/s, 5 fps, 5 tbr, 5 tbn, 5 tbc
DEBUG 2017-06-16 20:21:01.869 [ffmpeg64.exe-25-2] net.pms.io.OutputTextLogger Codec AVOption b (set bitrate (in bits/s)) specified for output file #0 (\\.\pipe\ffmpegvideo_65_1497669661510) has not been used for any stream. The most likely reason is either wrong type (e.g. a video option with no video streams) or that it is a private option of some encoder which was not actually used for any stream.
```

Documentation of reference:
- https://en.wikipedia.org/wiki/Motion_JPEG_2000
- https://en.wikipedia.org/wiki/Motion_JPEG
- https://en.wikipedia.org/wiki/Indeo
- https://en.wikipedia.org/wiki/H.261
- [DIVX](https://support.divx.com/s/article/ka140000000bmIcAAI/Genuine-DivX-and-Diagnosing-Playback-Problems?language=en_US)
- [WMV](https://github.com/MediaArea/MediaInfoLib/blob/725929c736412576308cd024f7e585c88cfa2494/Source/Resource/Text/DataBase/CodecID_Video_Riff.csv#L650-L654), also from [Microsoft](https://msdn.microsoft.com/en-us/library/windows/desktop/bb970509(v=vs.85).aspx)
- [ISML](https://github.com/MediaArea/MediaInfoLib/blob/f9e3216447eba7c7dc435e602f1c5c88cae1680e/Source/Resource/Text/DataBase/CodecID_General_Mpeg4.csv#L26), also from [Microsoft](https://blogs.iis.net/samzhang/live-smooth-streaming-publishing-point-advanced-settings)

Can also be found in MOV container:
- [RGB](https://en.wikipedia.org/wiki/Uncompressed_video), [here](https://github.com/MediaArea/MediaInfoLib/blob/182baa2b699b0d24b3647094c8f07368a9369d86/Source/Resource/Text/DataBase/Codec.csv#L1-L8) and [there](https://github.com/MediaArea/MediaInfoLib/blob/31cbb6661f7121e8af7c10ac3cae4abece5a207d/Source/Resource/Text/DataBase/CodecID_Video_Matroska.csv#L1).
- [YUV](https://en.wikipedia.org/wiki/YUV), [here](https://github.com/MediaArea/MediaInfoLib/blob/4468063f90bdf485a2d43b9fe66878dc37b9a15a/Source/Resource/Text/DataBase/Format.csv#L61)

**Note**: Samples are availables on request.

### DIFF
--- a/src/main/external-resources/renderers/DefaultRenderer.conf
+++ b/src/main/external-resources/renderers/DefaultRenderer.conf
@@ -566,13 +566,13 @@ CharMap =
 #    3gp      3rd Generation Partnership Project            video/3gpp
 #    3g2a     3rd Generation Partnership Project 2 Audio    audio/3gpp2
 #    3g2      3rd Generation Partnership Project 2          video/3gpp2
-#    aac-lc   Advanced Audio Coding (Low Complexity)        audio/aac
+#    aac-lc   Advanced Audio Coding (Low Complexity)
 #    ac3      Audio Codec 3 (Dolby Digital)                 audio/vnd.dolby.dd-raw
 #    acelp    Algebraic Code Excited Linear Prediction                                     Audio codec
-#    adpcm    Adaptive differential pulse-code modulation   audio/x-adpcm
+#    adpcm    Adaptive differential pulse-code modulation
 #    adts     Audio Data Transport Stream                   audio/vnd.dlna.adts
-#    aiff     Audio Interchange File Format                 audio/L16
-#    alac     Apple Lossless Audio Codec                    audio/m4a
+#    aiff     Audio Interchange File Format                 audio/L16                      Also audio/aiff or audio/x-aiff
+#    alac     Apple Lossless Audio Codec                                                   Need audio/x-m4a or audio/m4a
 #    als      Audio Lossless Coding (MPEG-4 ALS)
 #    amr      Adaptive Multi-Rate audio codec               audio/amr
 #    ape      Monkey's Audio                                audio/x-ape
@@ -593,13 +593,17 @@ CharMap =
 #    flv      Flash Video                                   video/x-flv
 #    g729     G.729                                                                         Also known as Coding of speech-ACELP
 #    gif      Graphics Interchange Format                   image/gif
-#    h263     H.263 video compression standard              video/h263
-#    h264     H.264/MPEG-4 Part 10 (AVC)                    video/h264
-#    h265     High Efficiency Video Coding (HEVC)           video/h265
-#    he-aac   Advanced Audio Coding (High-Efficiency)       audio/aacp
-#    jpg      JPEG                                          image/jpeg
+#    h261     H.261 video compression standard
+#    h263     H.263 video compression standard
+#    h264     H.264/MPEG-4 Part 10 (AVC)
+#    h265     High Efficiency Video Coding (HEVC)
+#    he-aac   Advanced Audio Coding (High-Efficiency)
+#    indeo    Intel Indeo Video                                                             IV31, IV32, IV41 and IV50
+#    jpg      Joint Photographic Experts Group              image/jpeg
+#    jpeg     Motion JPEG A
+#    jpeg2000 Motion JPEG 2000
 #    lpcm     Linear Pulse-Code Modulation                  audio/L16
-#    mjpeg    Motion JPEG                                   video/x-motion-jpeg
+#    mjpeg    Motion JPEG B
 #    mka      Matroska Audio                                audio/x-matroska
 #    mkv      Matroska Video                                video/x-matroska
 #    mlp      Meridian Lossless Packing                     audio/vnd.dolby.mlp
@@ -619,35 +623,37 @@ CharMap =
 #    ogg      Ogg container                                 video/ogg                          Applies for Vorbis I files only, use with FLAC and Theora should use oga and ogv respectively
 #    ogv      Ogg container with video content              video/ogg
 #    ogx      Ogg container                                 application/ogg
-#    opus     Ogg Opus                                      audio/ogg
+#    opus     Ogg Opus                                                                         Need audio/ogg or audio/opus
 #    png      Portable Network Graphics                     image/png
 #    qdmc     QDesign Music Codec (RaveSound)
 #    ralf     RealAudio Lossless Format                     audio/vnd.rn-realaudio
+#    rgb      Uncompressed video (Red, Green and Blue)
 #    rm       RealMedia (RMVB)                              application/vnd.rn-realmedia-vbr
 #    shn      Shorten                                       audio/x-shorten
 #    sipro    Sipro Lab Telecom
-#    sls      Scalable to Lossless                          audio/mp4                          Also known as MPEG-4 SLS
-#    sor      Sorenson Spark                                video/x-sorenson-spark
+#    sls      Scalable to Lossless                                                             Also known as MPEG-4 SLS
+#    sor      Sorenson Spark
 #    tiff     Tagged Image File Format                      image/tiff
-#    theora   Ogg Theora                                    video/ogg
+#    theora   Ogg Theora
 #    tta      True Audio codec                              audio/x-tta
 #    truehd   Dolby TrueHD                                  audio/vnd.dolby.mlp
-#    vc1      Microsoft VC-1 (WMV 9)                        video/vc1                          Also known as WMV 9 and WMV3
-#    vorbis   Ogg Vorbis                                    audio/ogg
+#    vc1      Microsoft VC-1                                                                   Also known as WMV 9 and WMV3
+#    vorbis   Ogg Vorbis
 #    vp6      On2 TrueMotion VP6
 #    vp7      On2 TrueMotion VP7
-#    vp8      Google VP8                                    video/webm
-#    vp9      Google VP9                                    video/webm
+#    vp8      Google VP8
+#    vp9      Google VP9
 #    wavpack  WavPack                                       audio/wavpack
 #    wav      Waveform Audio File Format                    audio/x-wav
 #    weba     Webm Audio                                    audio/webm
 #    webm     WebM Video                                    video/webm
 #    wma      Windows Media Audio                           audio/x-ms-wma
-#    wma10    Windows Media Audio 10 Professional           audio/x-ms-wma
-#    wmalossless      ""          Lossless                  audio/x-ms-wma
-#    wmapro           ""          Pro                       audio/x-ms-wma
-#    wmavoice         ""          Voice                     audio/x-ms-wma
+#    wma10    Windows Media Audio 10 Professional
+#    wmalossless      ""          Lossless
+#    wmapro           ""          Pro
+#    wmavoice         ""          Voice
 #    wmv      Windows Media Video                           video/x-ms-wmv                     Matches WMV1, WMV2, WMV7 and WMV8 (see "vc1" above for more). Tag also used for asf files
+#    yuv      Uncompressed video                                                               Luminance (Y) and two chrominance (UV)
 #
 #    und      Undetermined, if the parser did not recognize one of above
 #
--- a/src/main/java/net/pms/configuration/FormatConfiguration.java
+++ b/src/main/java/net/pms/configuration/FormatConfiguration.java
@@ -72,14 +72,18 @@ public class FormatConfiguration {
 	public static final String FLV = "flv";
 	public static final String G729 = "g729";
 	public static final String GIF = "gif";
+	public static final String H261 = "h261";
 	public static final String H263 = "h263";
 	public static final String H264 = "h264";
 	public static final String H265 = "h265";
 	public static final String HE_AAC = "he-aac";
 	public static final String ICNS = "icns";
 	public static final String ICO = "ico";
+	public static final String INDEO = "indeo";
 	public static final String ISO = "iso";
 	public static final String JPG = "jpg";
+	public static final String JPEG = "jpeg";
+	public static final String JPEG2000 = "jpeg2000";
 	public static final String LPCM = "lpcm";
 	public static final String M4A = "m4a";
 	public static final String MKV = "mkv";
@@ -115,6 +119,8 @@ public class FormatConfiguration {
 	public static final String RAW = "raw";
 	public static final String REALAUDIO_14_4 = "ra14.4";
 	public static final String REALAUDIO_28_8 = "ra28.8";
+	/** Used as a "video codec" when sequences of raw, uncompressed RGB or RGBA is used as a video stream in AVI, MP4 or MOV files */
+	public static final String RGB = "rgb";
 	public static final String RM = "rm";
 	public static final String SHORTEN = "shn";
 	public static final String SIPRO = "sipro";
@@ -143,6 +149,8 @@ public class FormatConfiguration {
 	public static final String WMAPRO = "wmapro";
 	public static final String WMAVOICE = "wmavoice";
 	public static final String WMV = "wmv";
+	/** Used as a "video codec" when sequences of raw, uncompressed YUV is used as a video stream in AVI, MP4 or MOV files */
+	public static final String YUV = "yuv";
 	public static final String MIMETYPE_AUTO = "MIMETYPE_AUTO";
 	public static final String und = "und";
 
--- a/src/main/java/net/pms/dlna/LibMediaInfoParser.java
+++ b/src/main/java/net/pms/dlna/LibMediaInfoParser.java
@@ -470,15 +470,16 @@ public class LibMediaInfoParser {
 		} else if (value.equals("qt") || value.equals("quicktime")) {
 			format = FormatConfiguration.MOV;
 		} else if (
-			value.equals("isom") ||
-			value.startsWith("mp4") ||
-			value.equals("20") ||
-			value.equals("m4v") ||
-			value.startsWith("mpeg-4") ||
-			value.contains("m4a") ||
-			value.contains("xvid")
-		) {
-			format = FormatConfiguration.MP4;
+				value.contains("isom") ||
+				value.startsWith("mp4") ||
+				value.equals("20") ||
+				value.equals("isml") ||
+				value.startsWith("m4a") ||
+				value.startsWith("m4v") ||
+				value.equals("mpeg-4 visual") ||
+				value.equals("xvid")
+			) {
+				format = FormatConfiguration.MP4;
 		} else if (value.contains("mpeg-ps")) {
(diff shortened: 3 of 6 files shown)


Sami32:
@valib Thank you :smiley:

valib:
@UniversalMediaServer/developers it seems that Travis CI has a problem to build OS X part. All Java versions passed the test but for OS X it not even start.

Sami32:
@valib Yes, they have [a major issue](https://www.traviscistatus.com/incidents/v4s112tzj198), since more than 6 hours now.

Nadahar:
They will succeed, it only takes a while before they start as there's still quite a backlog of OS X builds at Travis CI.

src/main/external-resources/renderers/DefaultRenderer.conf
```diff
@@ -566,14 +566,14 @@ CharMap =
 #    3gp      3rd Generation Partnership Project            video/3gpp
 #    3g2a     3rd Generation Partnership Project 2 Audio    audio/3gpp2
 #    3g2      3rd Generation Partnership Project 2          video/3gpp2
-#    aac-lc   Advanced Audio Coding (Low Complexity)        audio/aac
+#    aac-lc   Advanced Audio Coding (Low Complexity)        audio/mp4
```
Nadahar: I don't understand how it's relevant to specify a mime type for a codec at all. Both the old and the new values are wrong, the truth is that there is no mime-type for ```aac-lc```.

src/main/java/net/pms/dlna/LibMediaInfoParser.java
```diff
@@ -490,12 +491,31 @@ private static void getFormat(StreamType streamType, DLNAMediaInfo media, DLNAMe
...
+				value.equals("wmv2")
+			) {
+				format = FormatConfiguration.WMV;
+		} else if (
+				value.contains("mjpg") ||
+				value.startsWith("mjpeg") ||
+				value.equals("mjpa") ||
+				value.equals("mjpb") ||
+				value.equals("jpeg") ||
+				value.equals("jpeg2000")
```
Nadahar: Is it really safe to parse ```jpeg``` and ```jpeg2000``` as ```MJPEG```? Both are actually image formats.

src/main/java/net/pms/dlna/LibMediaInfoParser.java
```diff
@@ -514,14 +534,30 @@ private static void getFormat(StreamType streamType, DLNAMediaInfo media, DLNAMe
...
+				value.startsWith("div") ||
+				value.equals("dx50") ||
+				value.equals("dvx1")
+			) {
+				format = FormatConfiguration.DIVX;
+		} else if (value.startsWith("indeo")) { // Intel Indeo Video: IV31, IV32, IV41 and IV50
+			format = FormatConfiguration.INDEO;
+		} else if (value.equals("yuv")) {
+			format = FormatConfiguration.YUV;
+		} else if (value.equals("rgb") || value.equals("rgba")) {
```
Nadahar: What exactly is identified here? ```yuv```, ```rgb``` and ````rgba``` sounds like image color models to me, not a codec or a format. How is this used?

Sami32: As you already know, i don't appreciate that you don't even bother looking at the links that i take the effort to post :disappointed: 
That's said, i should remember that you're near always doing that...
[VRMLuncompressed.avi.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/1143491/VRMLuncompressed.avi.txt)
[miss_am.avi.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/1143492/miss_am.avi.txt)

src/main/java/net/pms/dlna/LibMediaInfoParser.java
Sami32: With that it is, IMO, and its better than have "jpeg" displayed for video :P
```logtalk
if (media.getContainer() != null && !media.getContainer().equals(FormatConfiguration.JPG)) {
				format = FormatConfiguration.MJPEG;
			}
```
You can always believe your own eyes :tongue: 
[Moviefone_JPEG.mov.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/1143501/Moviefone_JPEG.mov.txt)

src/main/java/net/pms/dlna/LibMediaInfoParser.java
Nadahar: @Sami32 I read links if you explain what they're for and what to look for. I don't read through tons of text when I don't know what to look for.

Sami32: @Nadahar Its look blabla to me. Let me go cry i'll return after the ads :P

Nadahar: That said, I know what ```yuv```, ```rgb``` and ```rgba``` are without reading the links. What I'm asking for is the logical connection from a color space to a ```FormatConfiguration``` that can either be a file format or a codec.

src/main/java/net/pms/dlna/LibMediaInfoParser.java
Sami32: It didn't worked since long ago, for simple MJPEG.
https://github.com/MediaArea/MediaInfoLib/blob/ed4e03b2b86418195e78fa44fbee07ad10659db9/Changes.txt#L86-L88

src/main/external-resources/renderers/DefaultRenderer.conf
Sami32: No comment, as we'll never agree on our DLNA reading.

src/main/java/net/pms/dlna/LibMediaInfoParser.java
Sami32: If you don't bother checking at the MediaInfo links showing it as format, at least look at the sample i posted for you...

src/main/external-resources/renderers/DefaultRenderer.conf
Nadahar: @Sami32 This doesn't have to do with DLNA, but the fact that mime-types identify *formats*, not codecs. There is no "aac" format, thus there is no "aac mime type".

src/main/java/net/pms/dlna/LibMediaInfoParser.java
Nadahar: If see that MediaInfo use them as codecs (not formats) - which I can understand as they probably doesn't have a specific "codec name". They are however only relevant within the format that specifies that they can be used. It's the constraint that these MediaInfo values should only mean this in the specified context I'm requesting, to avoid parsing bugs and general confusion. A "codec" called ```YUV``` doesn't make any sense until you define that it's within the context of the AVI standard, which permits using a stream of raw, uncompressed images with certain color models.

Sami32: That just mean that you don't read MediaInfo lookup tables as i do: (link already posted above)
https://github.com/MediaArea/MediaInfoLib/blob/4468063f90bdf485a2d43b9fe66878dc37b9a15a/Source/Resource/Text/DataBase/Format.csv#L61

As you see, it can be inside a MOV or a MP4 containers.

src/main/external-resources/renderers/DefaultRenderer.conf
Sami32: If you thing that callin them "MPEG-4 Part 3" or "MPEG-4 AAC" will help the users...good luck.

src/main/java/net/pms/dlna/LibMediaInfoParser.java
Sami32: As a format or codec ID, my only purpose is to identify them as such, as i'm sure that they give issues to users as they're transcoded with they color matrix, color range and space that will not be supported by their renderers, at least for the users using MPEG2 video transcoding.

It seem that you're waiting for something, but i don't understand what...

Nadahar: @Sami32 I'm not waiting for anything in particular, I just don't know how to explain it so that you understand what I mean. One major thing that I see time and time again that we think completely different about, is MediaInfo. As I see it, MediaInfo is just one of several ways to parse media, and UMS must store information in a way that makes sense regardless of what method of parsing is used. You seem to be very focused on MediaInfo as the only source of information, and use decisions made in MediaInfo as "proof" that something is correct.

Sami32: Yes, i'm, as it's our main parsing tool, but i don't forget _FFmpeg_ though.

Did the latest modifications meet your requirements ?

Sami32: Of course not, but does it seem more acceptable to you ?

src/main/java/net/pms/dlna/LibMediaInfoParser.java
Nadahar: I can't really seem to explain this either, but the problem similar to the case with with ```rgb``` etc. This only makes sense within a particular context and is misleading otherwise, but the context isn't ensured. Your MediaInfo link states that the stream kind must be used to make sense of the value, but the check above doesn't actually check the stream kind.

src/main/java/net/pms/dlna/LibMediaInfoParser.java
Nadahar: I can't see any new commits with any change, so I assume that you rebased it. What did you change?

Nadahar: Ah, now I see the change. Yes, that helps. For the MJPEG case it would be perfect, here I think some further explanation is needed about these very unintuitive ```FormatConfiguration```s.

Sami32: You was always asking for amending, so now i do it more ;-)

Explanations are more your cup od tea. What i could say ?
For me its just YUV and RGB format ID that imply some special attention while been transcoded as their color range, space and matrix need to be converted.

Sami32: Restrict to MJPEG will not be correct, as i explained you, so i cannot do that.
Mov container is also popular for the cameras, i don't know about MP4, so MediaInfo already restrict it, and the fact of having used `equals` restrict again more possible issue.

src/main/java/net/pms/dlna/LibMediaInfoParser.java
Sami32: You make a point, and i've given you the stick :stuck_out_tongue:

Sami32: Done.

It look the code look more nice, but that will not change the parsing result though :P

src/main/java/net/pms/dlna/LibMediaInfoParser.java
```diff
@@ -490,12 +491,31 @@ private static void getFormat(StreamType streamType, DLNAMediaInfo media, DLNAMe
...
-		} else if (value.startsWith("h263") || value.startsWith("s263") || value.startsWith("u263")) {
-			format = FormatConfiguration.H263;
+		} else if (
+				value.startsWith("windows media") ||
+				value.equals("wmv1") ||
+				value.equals("wmv2")
+			) {
+				format = FormatConfiguration.WMV;
+		} else if (streamType == StreamType.Video &&
+				(
```
Nadahar: 👍

src/main/java/net/pms/configuration/FormatConfiguration.java
```diff
@@ -115,6 +119,7 @@
 	public static final String RAW = "raw";
 	public static final String REALAUDIO_14_4 = "ra14.4";
 	public static final String REALAUDIO_28_8 = "ra28.8";
+	public static final String RGB = "rgb";
```
Nadahar: Can't you just add a description above this constant that clarifies what it's used for? For example:
```java
/** Used as a "video codec" when sequences of raw, uncompressed rgb or rgba is used as a video stream in AVI files */
```

src/main/java/net/pms/configuration/FormatConfiguration.java
```diff
@@ -143,6 +148,7 @@
 	public static final String WMAPRO = "wmapro";
 	public static final String WMAVOICE = "wmavoice";
 	public static final String WMV = "wmv";
+	public static final String YUV = "yuv";
```
Nadahar: A similar description here..

src/main/external-resources/renderers/DefaultRenderer.conf
Nadahar: @Sami32 I think you miss my point completely. You have license plates on your car because the authorities require you to so that they can easily identify you, but you don't have license plates on your washing machine (I hope) because it isn't a car.

src/main/java/net/pms/dlna/LibMediaInfoParser.java
Nadahar: The restriction for MJPEG is perfect the way you've done it now. 

It's a good thing that you're amending, it just makes it hard to see the changes when you ask me if the changes made it better. I do the same myself sometimes, and it's useful to learn that when we're making changes that we're discussing, it's better to create new commits and then rebase them later. Reviewable is so much better here, it will let you review the changes regardless of if they are rebased, amended or committed separately.

src/main/java/net/pms/dlna/LibMediaInfoParser.java
Nadahar: I'm not so sure about that. As always I'm just as focused on the unintended consequences as the intended ones. I won't change the intended parsing, but it might avoid some unintended parsing errors 👍

src/main/external-resources/renderers/DefaultRenderer.conf
Sami32: I'll say that your point is...only your.
As i'm not the one that called them AAC, years ago, but perhaps your friend Apple ?
Whatever, most of the renderers identify them as AAC while displaying infos, so for the users it's more easy to understand, remember, IMO.

Anyway, if it was a suggestion for deleting the MIME type, i'll not follow you on that matter.

src/main/java/net/pms/configuration/FormatConfiguration.java
Sami32: Honestly, i don't share your love for verbose, but i'll anyway add it.

src/main/external-resources/renderers/DefaultRenderer.conf
Nadahar: @Sami32 Yes, the problem is that there's mime-types for codecs, which is wrong. Only the entries representing formats should have mime-types. Providing wrong information is just confusing to everyone. That's not something this PR has introduced though, so I don't mean that you have to change it, I just had to mention it since you updated one of the meaningless mime-types.

My Onkyo also "identify" both MP4 and ADTS audio files with an AAC stream inside as AAC, but that's not wrong. It's identifying the codec, not the file format.

To try to explain once again - what the above line implies is that a file containing ```aac-lc``` should have a mime-type of ```audio/mp4```. Can't you see that it is wrong? The file could be ADTS, MKV or any other container format that supports AAC. The mime type depends on the format, *not* the codec.

src/main/java/net/pms/dlna/LibMediaInfoParser.java
Sami32: "perfect" ? Are you going well ? sick or depressive ? :rofl: 
I'm kidding you because i never read you say that, and i still cannot imagine that you could ever say that. It's anyway good to take :smile: 

Amending on such specific little subject as RGB or YUV is fine, IMO, tough reviewing get uneasy if you don't tell the others that you changed your code.
So after amending, i can apologize as well :wink:

src/main/java/net/pms/configuration/FormatConfiguration.java
Nadahar: Don't you see how that will make the code much easier to maintain in the future? Now, if you have never seen this code before, or have forgotten why you did this, and is to refactor something that touches this - how are you to understand what ```RGB``` or ```YUV``` means as a ```FormatConfiguration```? It doesn't make sense since those are different color encodings and not any particular codec or format. With the description, you have a chance to understand what it means, it means a stream of multiple uncompressed images where the bits are coded as ```RGB``` or ```YUV```. That makes it much easier to make the correct decision when refactoring, and that is what I'm after. I look at it as preventing future bugs.

src/main/external-resources/renderers/DefaultRenderer.conf
Sami32: > the problem is that there's mime-types for codecs

Aaahhh I get it now :P I'll change it. If you explained like that before ;-)

Nadahar: @Sami32 I've tried, again and again..... as always it comes down to us having communication problems.

Sami32: Done.
I know that i leaved the one for ALAC, but as i remember that yourself seem to have some issue with it, at least we don't understand each other on that, so i thought that it could help users more than confuse them (I wished to add it to the MP4 lines but it is already full)

PR #1340, opened 2017-07-16
Use standard Java method instead of third party.

valib:


valib:
It is working only for Java 8 so it will be postponed until we stop support for Java 7.

Nadahar:
@valib I don't know if you or I am getting senile, but as far as I can remember you have done the same change before only to be reversed because it only works in Java 8. Either my mind is pulling a trick on my, or... 🙄 😇

valib:
@Nadahar I think it was me but to be senile is my right when I am almost 60 years old man. :smiley:

Issue #915, opened 2016-06-12
"Cache" browsing problem


Sami32:
I tried to browse the "cache" on my _Panasonic Viera_TV  for the first time and discover that **if more than 20 entries need to be browse by scrolling (including repertories) the displaying become empty and stuck to it** until i use _Return_ on my remote to go up in the tree of repertories.
If i choose to browse up, using up arrow, to go directly to the end of the directory, i get same result : empty displaying.
It give me the **bigger LOG file** ever, 31MB, in few time, so i don't know if it's normal when using "cache" ?
https://www.wetransfer.com/downloads/46f02d07969f40ecf738abb66ed8038c20160612125705/17985dbfcc0ad9eefc3e6b832eb94cdb20160612125705/d78d94
https://we.tl/ckQg9NI7Y2

I hope that i make it understandable...but feel free to ask if not.

**P.S.** I can't remember wich developper suggest to change the name of "cache" in the forum, but i second that :+1: 
Because i discovered that it is well structured, like a "Media Library" or "DataBase" what ever customers like more, and "cache" don't suggest it, do honor to this...Only my opinion.

SubJunk:
There's an issue for this with a lot of discussion at https://github.com/UniversalMediaServer/UniversalMediaServer/issues/653 so please use that one instead. Any help is very appreciated!
I agree that the whole cache browsing feature needs work, it could be much better

Sami32:
@SubJunk Thanks to point me there :-) i missed it...

Issue #653, opened 2015-08-23
New Media / Cache folders fail to load past 20 items


SubJunk:
On my Panasonic VT60 TV, when I go into the New Media or Cache folders they load perfectly until I scroll past the 20th item in the main folder, then the whole folder goes blank.
It doesn't happen on subfolders - they work perfectly, as do all other non-New Media folders.
So it seems that something about the New Media folder is stopping the renderer from receiving the next set of 20 items.

I also tested version 3.3.0 to see if it was a recent issue but it happened with that version too.

To produce these logs I started with a clean install of the latest code, set a folder to share and include in the new media folder, went straight to the New Media folder and scrolled down, which usually loads the next 20 items:

http://www.universalmediaserver.com/forum/viewtopic.php?f=14&t=1656&p=17787#p17787

@SharkHunter any ideas?

SharkHunter:
The media monitor is no special from a normal folder. Looks like there are new requests for anything but the first 20 items.
How does it look in the web (for example)?

SubJunk:
@SharkHunter it loads perfectly on the web interface

SubJunk:
@SharkHunter I also tested by enabling "include in the new media folder" on a folder full of videos, instead of full of folders like the the logs, and the same thing happened

SubJunk:
This is also now happening with all monitored folders in https://github.com/UniversalMediaServer/UniversalMediaServer/pull/673 and I'm still not sure why

Nadahar:
@SubJunk I pulled #673 and added the whole library to "New media" (which wasn't localized by the way). I couldn't reproduce the bug on either the PS3 or the Onkyo renderers that I have available. The PS3 doesn't fetch one page at a time, so that's no big surprise, but the Onkyo does - and was still fine. It's hard to figure out what's happening when you can't reproduce it. If I were you I'd install Wireshark (if you haven't already), it's really great for tracking down such problems. Start capturing before trying to browse, reproduce the problem and stop capturing. You can then filter and do all sort of this with the packages, but what I've found easiest is to find the package with the correct browse request, rightclick it and select ```Follow TCP Stream```. That will assemble all the packages into a long "text document" with all the requests and responses after each other in the correct order. It makes it extremely easy to see what is requested and what is delivered as a response, and see where the communication is going wrong.

[...]

PR #1343, opened 2017-07-18
Fix problem when the TotalMatches for the FileTranscodeVirtualFolder is changed during browsing

valib:
in the FileTranscodeVirtualFolder. The problem should be in the FileTranscodeVirtualFolder when all players, audio and subtitles combinations are made but I could't find it.

### DIFF
--- a/src/main/java/net/pms/network/Request.java
+++ b/src/main/java/net/pms/network/Request.java
@@ -43,6 +43,7 @@ import net.pms.dlna.DLNAMediaInfo;
 import net.pms.dlna.DLNAMediaSubtitle;
 import net.pms.dlna.DLNAResource;
 import net.pms.dlna.DLNAThumbnailInputStream;
+import net.pms.dlna.FileTranscodeVirtualFolder;
 import net.pms.dlna.MediaType;
 import net.pms.dlna.Range;
 import net.pms.dlna.RealFile;
@@ -757,8 +758,12 @@ public class Request extends HTTPResource {
 						if (xbox360 && containerID != null) {
 							uf.setFakeParentId(containerID);
 						}
-						if (uf.isCompatible(mediaRenderer) && (uf.getPlayer() == null || uf.getPlayer().isPlayerCompatible(mediaRenderer))) {
-							response.append(uf.getDidlString(mediaRenderer));
+						if (uf.isCompatible(mediaRenderer) && (uf.getPlayer() == null
+							|| uf.getPlayer().isPlayerCompatible(mediaRenderer))
+							// do not check compatibility of the media for items in the FileTranscodeVirtualFolder because we need
+							 // all possible combination not only those supported by renderer because the renderer setting could be wrong. 
+							|| files.get(0).getParent() instanceof FileTranscodeVirtualFolder) {
+								response.append(uf.getDidlString(mediaRenderer));
 						} else {
 							minus++;
 						}
--- a/src/main/java/net/pms/network/RequestV2.java
+++ b/src/main/java/net/pms/network/RequestV2.java
@@ -802,7 +802,8 @@ public class RequestV2 extends HTTPResource {
 
 						if (uf.isCompatible(mediaRenderer) && (uf.getPlayer() == null
 							|| uf.getPlayer().isPlayerCompatible(mediaRenderer))
-							 // do not check compatibility of the media when it is in the FileTranscodeVirtualFolder because the renderer setting could be wrong 
+							 // do not check compatibility of the media for items in the FileTranscodeVirtualFolder because we need
+							 // all possible combination not only those supported by renderer because the renderer setting could be wrong. 
 							|| files.get(0).getParent() instanceof FileTranscodeVirtualFolder) {
 								response.append(uf.getDidlString(mediaRenderer));
 						} else {
--- a/src/main/java/net/pms/network/RequestV2.java
+++ b/src/main/java/net/pms/network/RequestV2.java
@@ -802,7 +802,7 @@ public class RequestV2 extends HTTPResource {
 
 						if (uf.isCompatible(mediaRenderer) && (uf.getPlayer() == null || uf.getPlayer().isPlayerCompatible(mediaRenderer))) {
 							response.append(uf.getDidlString(mediaRenderer));
-						} else {
+						} else if (!(files.get(0).getParent() instanceof FileTranscodeVirtualFolder)) {
 							minus++;
 						}
 					}
--- a/src/main/java/net/pms/network/RequestV2.java
+++ b/src/main/java/net/pms/network/RequestV2.java
@@ -800,9 +800,12 @@ public class RequestV2 extends HTTPResource {
 							uf.setFakeParentId(containerID);
 						}
 
-						if (uf.isCompatible(mediaRenderer) && (uf.getPlayer() == null || uf.getPlayer().isPlayerCompatible(mediaRenderer))) {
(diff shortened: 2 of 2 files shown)


Nadahar:
@valib I've recently studied the related code a bit, so I might have an idea of how to solve it if you could only make me understand the problem. Are you saying that ```TotalMatches``` is wrong because the number of items in the ```Transcode``` folder change? ```TotalMatches``` are per folder, it doesn't include items in subfolders. Why/how does it change?

valib:
@Nadahar the problem is that my Panasonic doesn't like when the `TotalMatches` is changed. It seems to me that there is an not supported item in the children list which in this part of the code is removed (using`minus`) and change the `TotalMatches` value. This occured when renderer is asking for the next part of the browsing (the TV is asking for the next 20 items). For less then 20 items in the folder it works correct.

valib:
@Nadahar the problem is that my Panasonic doesn't like when the `TotalMatches` is changed. It seems to me that there is an not supported item in the children list which in this part of the code is removed (using `minus`) and change the `TotalMatches` value. This occured when renderer is asking for the next part of the browsing (the TV is asking for the next 20 items). For less then 20 items in the folder it works correct.

Nadahar:
@valib I still don't understand why it would change. What happens is that all the ```DLNAResources``` for that renderer is added to a ```List```, and the count of that list is used. Even if that list includes items that shouldn't be there, it shouldn't *change* because the exact same list is built when the next request comes. If your renderer use the "DLNATreeHack" this doesn't apply though, as that logic seems completely crazy to me.

I've done a fix to ```TotalMatches``` in ```ImagesFixes```, have you checked if that helps with your problem? What I fixed was the fact that if the renderer makes a request where ```startingIndex``` is bigger than the index of the last item (asks for startingIndex 20 if there are only 20 items), the ```TotalMatches``` returned would be 0 - which is wrong.

Sami32:
It seem related to that issue: https://github.com/UniversalMediaServer/UniversalMediaServer/issues/653
So that mean at least 2 years old bug.
https://github.com/UniversalMediaServer/UniversalMediaServer/issues/915

Nadahar:
@valib @Sami32 I still don't quite understand what happens, why the ```TotalMatches``` would change. Could one of you create a trace log (without this PR) when this happens so that I can try to understand what happens using [ImagesFixes](https://github.com/UniversalMediaServer/UniversalMediaServer/tree/ImagesFixes)? The reason is that there are currently extra logging for this in that branch, which could be useful for figuring out what happens.

valib:
@Nadahar I can explain that. It is not related to the issue mentioned by @Sami32 and maybe I explained it in the comment in my last commit. The problem is that in the `FileTranscodeVirtualFolder` we need all possible combination of the players and media audio and subtitles. This is the only chance to test also combinations whose are not allowed by renderer setting. In my case I enabled the `tsMuxeR` engine wich produces only the MPEG-TS format which is not supported by my Panasonic TV and that's why during the check this item was checked as not supported and the `minus` was increased and changed the `TotalMatches` value.

Nadahar:
@valib I still don't understand how it would change during browsing. I do see that they way this is done is logically wrong though, but I think there should be a proper fix. I'm afraid that this "hack" can lead to more problems than in solves.

Regarding your comment, do you mean this one?
```java
// do not check compatibility of the media for items in the FileTranscodeVirtualFolder because we need
// all possible combination not only those supported by renderer because the renderer setting could be wrong. 
```
If so, it makes me understand even less. Why should we include incompatible entries? Because the renderer configuration might be wrong? Isn't it better to fix the renderer configuration then, than to show all kind of invalid entries to all renderers?

valib:
@Nadahar I have found the problem. Everything depends on the order of the not compatible player in the transcoders list and the way we test the compatibility. When I tested it and for me not compatible player `tsMuxeR` was at the end of the list the browsing was working until the last items should be showed. Then the `TotalMatches` and  `NumberReturned` were decreased by `minus`. When I changed the order and the non compatible player was in the range of items requested by player then media folder is empty because the `TotalMatches` are bigger that requested number of items and `NumberReturned` is less than requested. The problem is that [here](https://github.com/UniversalMediaServer/UniversalMediaServer/blob/d129c7d80839f3399868d061a8855a51bf9982af/src/main/java/net/pms/network/RequestV2.java#L778) we start to check only files defined by range requested by renderer not all files in the folder.
On the other side the #TRANSCODE# folder from my point of view is mentioned to show all possible combinations of media, players and subtitles doesn't matter if those combination are relevant for renderer. There is a great chance that some renderer configuration is missing something and not offer the best setting. This is a chance how to test it than change configuration and see what happen.

Nadahar:
@valib I think I'm beginning to see the problem, I was just confused by the claim that it "changed during browsing". The whole "minus" thing is an ugly hack that shouldn't be there. If a renderer requests index 0-19 and then 2 is removed with "minus" so that 0-17 are returned (and ```TotalMatches``` is larger than 20) UMS has basically "broken the contract" with the renderer. Returning only 18 items would be ok if there only were 18 items, but as long as ```TotalMatches``` indicates that there are more, the renderer won't know what to do - how to get the two "missing" items. If the renderer is very clever, it will then ask for 18-37 the next time (although most probably doesn't do that). Even if the renderer is clever, it will still fail because UMS doesn't take the two "minuses" into account when calling ```getDLNAResources()``` with ```startingIndex``` - which will mean that the first two items in the reply will probably be the same as the last two in the previous reply (unless the last two were those that were removed). This will probably confuse the renderer even further. In short, UMS' approach is broken. It's not just for the transcode folder, this bug is universal. The problem is the "minus". 

To solve it the logic must be changed. Either the filtering of the resources must be done in ```getDLNAResources()``` (so that the correct number of items is always returned and no "minus" is needed), or the code calling ```getDLNAResources``` will have to become smarter and "fill up" ```files``` if there are more items available and we're returning fewer than the renderer asks. In both cases, the "minus" won't be needed anymore.

valib:
@UniversalMediaServer/developers it is fine that we included the checking of the OS X in the Travis but until they have problems to make it working it prevents our PR to be merged. Maybe we should remove it until they solve it.

Nadahar:
@valib I don't understand, it is working - they fixed that problem the other day. The "problem" you see is that it takes a bit longer before it starts. That's simply because there's often a longer queue on the OS X builds. This is probably because Travis has fewer OS X build containers than the demand, but in my experience it usually doesn't take more than 10 minutes extra. Are you really that much in a hurry to merge something?

valib:
@Nadahar the last commit in this PR I made was a day ago and until I restarted the test 20 minutes ago it was wrong. So it is not working properly.

valib:
@Nadahar agree that `minus` is ugly hack and the whole concept should be rewritten but this PR fix the current problem with `FileTranscodeVirtualFolder` and has no side effect. For the rest of real or virtual folders it will work as up to now. There are no other complains by users as I can see but the only one I made. We can focus on that problem later.

Nadahar:
@valib I'm sure that there's others that have problems as well, remember that a very small percentage of users actually report problems. That said, I'm not saying it's a big problem - but I don't think it would be that hard to solve "properly" either. I haven't had the time to look at it though.

valib:
@Nadahar what about to review this code to be merged when there are no side efects and leave the final solution to the future when you haven't had the time to look at it and I am not capable to do that?

Nadahar:
@valib I have a very hard time to believe that there's no side effects from doing this. We already know that there's one side effect - that renderers will be presented with incompatible formats.

PR #1350, opened 2017-07-30
Matroska MJPEG fix

Sami32:
Correct my last modification, as i had forgotten the Matroska specific case.


Sami32:
@valib Thanks.

PR #1356, opened 2017-08-05
Force interface on multicast socket

reorder:
On macOS (at least) multicast binds to 0.0.0.0 and first available interface despite passing specific address, this is causing ALIVE not being sent and, eventually, playback disconnect. Setting previously calculated interface on the socket fixes that.

### DIFF
--- a/src/main/java/net/pms/network/UPNPHelper.java
+++ b/src/main/java/net/pms/network/UPNPHelper.java
@@ -290,7 +290,23 @@ public class UPNPHelper extends UPNPControl {
 
 		InetSocketAddress localAddress = new InetSocketAddress(usableAddresses.get(0), 0);
 		MulticastSocket ssdpSocket = new MulticastSocket(localAddress);
-		ssdpSocket.setNetworkInterface(networkInterface);
+
+        try {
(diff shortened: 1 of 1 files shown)


SubJunk:
I remember reading something about it being good to use 0.0.0.0 on Windows so I'm not sure about this one. Maybe it could be wrapped in `if platform is macOS` to be safe?

reorder:
It is apparently a no-op if the interface was already selected correctly by MulticastSocket itself. Smoke testing on Windows does not reveal any problem too.

reorder:
That won't hurt, obviously.

In fact I struggle to imagine a scenario where UPNP needs to bind to different *interface* (not the address) than the one server binds to, but probably there is a reason.

OP-Klaus:
Ok great, if you make the change I'll merge :)

SubJunk:
Ok great, if you make the change I'll merge :)

SubJunk:
@reorder thanks!

SubJunk:
@valib ok, maybe we should revert?

valib:
@SubJunk I try to fix it in  #1620

PR #1370, opened 2017-08-23
Fix support for Windows XP

SubJunk:


Nadahar:
@SubJunk I'm not sure what you mean by "newer", you mean "different" right? The problem is that that to bundle an extra executable, the size of the download will grow. But, I don't see why we couldn't just replace the 32-bit FFmpeg with the one built for XP. It's not like that shouldn't work on newer OSes as far as I can understand.

Sami32:
@Nadahar I think he is really meaning newer, recent, as it the trunk build for WinXP that i linked to.
Well, that should b the more correct way to do this, but the version that i suspect him to use don't have SSE2 support, so by default for all 32-bit OS hum...
That's said, nobody complain about the other excecutable as Flac, MEncoder or tsMuxeR so...

@SubJunk I'm sure that you understood that it will still not work under Windows XP 32-bit, because of JNA build incompatibility.

Nadahar:
@Sami32 What I'm trying to say is that it's required to use a newer FFmpeg version for Windows XP - it's required to use one compiled without MSVC 2015 or newer and with the correct switches set.

SubJunk:
@Nadahar this doesn't add more executables to the package, it replaces them, so it will be the same size.

@Sami32 yes I think we will need more changes in this branch like a JNA rollback too. Feel free to make changes here too

Sami32:
@SubJunk That's up to you.
Or you can do as @Nadahar have done, or do regression to the precedent JNA version as i've done in my fork.

Thanks, but i began with your fork, after i tried @Nadahar and finally i ended working on mine :wink: 
Because i would like fix many things, also if it's not in a perfect way, at least i can make changes and  builds for some users, without bothering anyone.

mirh:
You can target XP even with MSVC 2017 actually. 
And as long as the software is not dumb at branching the various x86 extension sets in code, I think even compiling with AVX2 should do it.

444nonymous:
Are you going to release an exe to test?

SubJunk:
There was a lot of confusing information going around about this and I'm not sure I even know what is needed anymore. Is it just JNA and FFmpeg? I still don't understand why a newer FFmpeg build would fix support for the older OS but I can blindly put it out there to test

mirh:
That is a newer version is the reason for you to want to update. 
But the fix for older OS actually comes from using a separate build/branch in compiling ffmpeg, afaik.

Nadahar:
@SubJunk It's all explained above. It's not the FFmpeg *version* that determines XP compatibility, it's the compile options. You have to use a version that is compiled to work with Windows XP.

It's basically the same for JNA. JNA comes with a small compiled library bundled (jnidispatch.dll), and every version after 4.2.2 comes with an incompatible DLL for Windows XP because of the way it is compiled. I've solved that in DMS by compiling ```jnidispatch.dll``` myself (for JNA 4.4.0, haven't yet done it for JNA 4.5.0) and doing these changes: https://github.com/DigitalMediaServer/DigitalMediaServer/commit/4a3b3beb0c838e557f64b8e917e0ca26f8a0f96c

The bad thing about it is that a new DLL has to be compiled and embedded for every new JNA versions, but I've tried to point out the problem to JNA and they seems to think there's some other reason and we never really get anywhere. The version I compile myself seems to work perfectly well both on XP and later versions of Windows though.

SubJunk:
That sounds good about the DLL in DMS, any plans to bring it to UMS?

Nadahar:
@SubJunk The commit is there for anyone to use/cherry-pick: https://github.com/DigitalMediaServer/DigitalMediaServer/commit/4a3b3beb0c838e557f64b8e917e0ca26f8a0f96c. I asked you about it when I was doing it, but didn't get much of a reply.

The problem with this approach is that a "custom" DLL must be compiled for every new JNA version. Somebody has to do that, and I'm guessing that nobody else than me will bother to set up the environment to do it. That would make upgrading of JNA dependant on me in reality, which might not be a good idea. I haven't yet compiled one for JNA 4.5.0, but it's on my todo list - but so are so many other things...

SubJunk:
Ok cool, thanks. I don't anticipate a need for us to update JNA urgently so it should be ok to use your DLL

DeFlanko:
Isant Windows XP EOL with Microsoft?

mirh:
That has never been a bar for anything, by the same token you'd have to drop 7 too. 

On the other hand, I do suppose after 4 more years, the market really is changed. 
By all means, even after ffmpeg 4.0 officially dropped it, you can still actually get XP to [build and work](https://forum.doom9.org/showthread.php?t=181802) with the last version.. but you get to wonder how many people even stand to benefit and care.

SubJunk:
I'm not likely to work on this and I don't think any of the current developers will either, so will close it. I welcome any pull requests and further comments.

PR #1371, opened 2017-08-24
Cache dependencies during build

atamariya:


Nadahar:
Does this actually help? I was thinking about caching them as well, but then I read [this](https://docs.travis-ci.com/user/caching/#Things-not-to-cache). It states specifically that compiled binaries should *not* be cached.

I don't think changing the "binaries" folder path is that simple, have you verified that packages for all platforms will build correctly? I would think changes would be needed at least in the NSIS files, and maybe also for the building the TAR and the macOS bundle.

SubJunk:
Good to see you @atamariya :)

atamariya:
Thanks @SubJunk .

@Nadahar I fixed the Nsis script. The build time does indicate some improvement.
untitled

Nadahar:
@atamariya I've looked at this, and I don't quite understand how the cache works. It seems to me that it will make a separate cache for each PR and JVM, which means that it would only help for subsequent builds of the same commit..?
```powershell
attempting to download cache archive
fetching PR.1371/cache-linux-trusty-e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855--jdk-openjdk7.tgz
found cache

adding /home/travis/.m2 to cache
adding /home/travis/UniversalMediaServer/bin to cache
creating directory /home/travis/UniversalMediaServer/bin
```
The binaries doesn't seem to benefit from being cached, as they are placed in different folders than where the cache is extracted:
```powershell
main:
    [mkdir] Created dir: /home/travis/build/UniversalMediaServer/UniversalMediaServer/bin
      [get] Getting: http://universalmediaserver.com/svn/binaries/linux/ffmpeg?p=67
      [get] To: /home/travis/build/UniversalMediaServer/UniversalMediaServer/bin/ffmpeg
      [get] Getting: http://universalmediaserver.com/svn/binaries/linux/ffmpeg64?p=67
      [get] To: /home/travis/build/UniversalMediaServer/UniversalMediaServer/bin/ffmpeg64
      [get] Getting: http://universalmediaserver.com/svn/binaries/linux/tsMuxeR?p=67
      [get] To: /home/travis/build/UniversalMediaServer/UniversalMediaServer/bin/tsMuxeR
      [get] Getting: http://universalmediaserver.com/svn/binaries/linux/tsMuxeR-new?p=67
      [get] To: /home/travis/build/UniversalMediaServer/UniversalMediaServer/bin/tsMuxeR-new
      [get] Getting: http://universalmediaserver.com/svn/binaries/linux/tsMuxeR_licence.txt?p=67
      [get] To: /home/travis/build/UniversalMediaServer/UniversalMediaServer/bin/tsMuxeR_license.txt
[INFO] Executed tasks
```
The only difference I can see is that it seems to download fewer Maven artifacts (```.m2``` is also cached). Still, if it only helps when restarting a build, I'm not sure if there's any benefit. I think the build must be organized differently to benefit from the cache by having a "main build" that populates the cache which then starts the "sub builds" which can retrieve data from the cache. I'm not sure though.

SubJunk:
Is this redundant now since we added a similar thing recently, or is this still useful?

Nadahar:
I'm not sure what was recently changed, but this seems to be useful when rebuilding a branch that has been built before. Great care should be taken when verifying that the moving of the "bin" folder doesn't cause issues.

PR #1384, opened 2017-09-02
Revive parent folder thumbnails

Nadahar:
This is based on [this thread](http://www.universalmediaserver.com/forum/viewtopic.php?t=10592). This is not a bugfix, it's a question of how we want it to work.

There have been complaints in the past that folder thumbnails would override thumbnails for individual items. During #1061 and the introduction of the generic thumbnails, I removed this so that generic thumbnails would be shown instead. Folder thumbnails (```folder.jpg``` and ```folder.png```) is still used as the thumbnail for the folder itself, but not for the media items in the folder.

When I read the above thread, I was thinking that it might be better if a folder thumbnail would be shown if no thumbnail was found for the individual media item instead of a generic icon, if a folder thumbnail exists.

I'm not sure what is the "best" behaviour, ideally I think it should be possible to configure thumbnail priorities as I've mentioned before. This is a small tweak that makes folder thumbnails the "second last" priority only preferred over generic thumbnails, without taking on the larger project of making it configurable.

@UniversalMediaServer/developers Do you think this is an improvement, or is generic thumbnails a better solution?

### DIFF
--- a/src/main/java/net/pms/dlna/RealFile.java
+++ b/src/main/java/net/pms/dlna/RealFile.java
@@ -317,10 +317,32 @@ public class RealFile extends MapFile {
 				result = getMedia().getThumbnailInputStream();
 			}
 		} catch (IOException e) {
-			result = null;
-			LOGGER.debug("An error occurred while getting thumbnail for \"{}\", using generic thumbnail instead: {}", getName(), e.getMessage());
+			LOGGER.debug(
+				"An error occurred while getting thumbnail for \"{}\", using generic thumbnail instead: {}",
+				getName(),
+				e.getMessage()
+			);
 			LOGGER.trace("", e);
 		}
+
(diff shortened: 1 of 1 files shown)


Nadahar:
The point is that we can't have both. Either the folder thumbnail or generic thumbnails will be used when the individual media items doesn't have their own thumbnail and a folder thumbnail exists, so we need to decide how we want it to work before deciding if this should be merged.

Sami32:
@Nadahar As i said, i prefer has it is now, but its up to majority vote.

Nadahar:
I've discovered a problem with this PR, I'll have to get back to it.

Nadahar:
This PR was made in error, it already works as this. I don't know what the user in the originating thread means by "not working", but it works when I test it.

PR #1391, opened 2017-09-17
Fix replaceExtension for usage with directories on non-windows OS

sammyrc34:
On Linux or unix like OSs, previously replaceExtension appended a period before the separator, eg "Dir./folder.jpg". Fixed method for replacement in all cases, files and folders, and added associated unit tests.

### DIFF
--- a/src/main/java/net/pms/dlna/RealFile.java
+++ b/src/main/java/net/pms/dlna/RealFile.java
@@ -282,7 +282,7 @@ public class RealFile extends MapFile {
 
 		if (file.isDirectory()) {
 			cachedThumbnail = FileUtil.replaceExtension(file, "folder.jpg", true, false);
-			
+
 			if (cachedThumbnail == null) {
 				cachedThumbnail = FileUtil.replaceExtension(file, "folder.png", true, false);
 			}
--- a/src/main/java/net/pms/dlna/RealFile.java
+++ b/src/main/java/net/pms/dlna/RealFile.java
@@ -281,10 +281,10 @@ public class RealFile extends MapFile {
 		}
 
 		if (file.isDirectory()) {
-			cachedThumbnail = FileUtil.replaceExtension(file, "/folder.jpg", true, false);
-
+			cachedThumbnail = FileUtil.replaceExtension(file, "folder.jpg", true, false);
+			
 			if (cachedThumbnail == null) {
-				cachedThumbnail = FileUtil.replaceExtension(file, "/folder.png", true, false);
+				cachedThumbnail = FileUtil.replaceExtension(file, "folder.png", true, false);
 			}
 		}
 
--- a/src/main/java/net/pms/util/FileUtil.java
+++ b/src/main/java/net/pms/util/FileUtil.java
@@ -859,6 +859,7 @@ public class FileUtil {
 			file.getParentFile(),
 			file.getName(),
 			extension,
+			file.isDirectory(),
 			nullIfNonExisting,
 			adjustExtensionCase
 		);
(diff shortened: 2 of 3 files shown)


Nadahar:
@sammyrc34 Thanks, but it would be helpful if you explained what the problem is. Replacing the extension of a folder doesn't make sense to me, so what is it that's failing with the new code?

sammyrc34:
No directory thumbnails were working in Linux. The old code would return a File for "Dir./folder.jpg" instead of "Dir/folder.jpg" when searching for thumbnails (note the extra period). Also, the old code used only "/", which is not the separator for Windows.

Nadahar:
@sammyrc34 ```java.io``` and ```java.nio``` translates ```/``` to ```\``` on Windows automagically, so that's not a problem. What I don't understand is why ```Dir./folder.jpg``` is correct. I admit I feel a bit slow right now, but I don't get it.

sammyrc34:
Hehe, somehow I've made this confusing ;). 

`Dir./folder.jpg` is certainly not correct :). That extra period at the end of the directory is literal in Linux. This is what my change fixes. 

Eg old code did this:
`baseFileName + "." + extension`

My change is more elaborate, but essentially removes the extra period which previously created `Dir./folder.jpg`, therefore creating `Dir/folder.jpg`.

Otherwise I figured Windows was automatically swapping separators, as folder images were working in Windows.

Nadahar:
@sammyrc34 The reason I'm confused is probably because I recently refactored that code. Thus, when you were talking about the "old" code I assumed you meant the code before the refactoring. When I think about it it's obvious that you didn't know about that and that your "old" code was my "new" code 😉 

I think the problem should be solved at its root, and from what I can see the real problem is that there's a call which use ```replaceExtension``` in a hackish way to add ```folder.jpg``` to an existing path.

Nadahar:
@sammyrc34 I very recently refactored this, so I don't understand how I could have overlooked that. This should be done in an entirely different way, so I'll take a look at that.

sammyrc34:
I was tempted to create a new method for this, but wanted to keep a smaller footprint. Also, while Windows functions with "Dir./folder.jpg" somehow, it was still something that needed fixing when I looked at it. 

What do you think, a new method to just join path elements and test for existence?

Nadahar:
@sammyrc34 No, when I did the refactoring I also changed this to work in a more general way with any number of different extensions. I don't remember exactly what the methods are called at the moment, but I think they are found in ```MapFile```. I simply forgot to refactor that part.

sammyrc34:
I did check all calling code. Most entry points were passing in a file, looking for a different extension. 
Eg:
`file = FileUtil.replaceExtension(audioVideoFile, extension, false, true);`

The case in `RealFile` was the only example I could see for usage with directories. Even then 

The best change I can think of is to add `FileUtil.getFileInFolder( File folder, String file", boolean nullifnotexist, boolean case)`.

Nadahar:
@sammyrc34 There's no need for a method for that, ```new File(folder, filename)``` will take care of it. Existence is then checked with ```file.exists()```. What I want to do is to apply the same logic for folder thumbnails that's being used for other "image file thumbnails". I can look at it tomorrow, but if you're impatient this is the commit where I refactored this and where you can see the general logic: 143807c0e7dcca4f9648fa3f632aa46c56ab0029

sammyrc34:
Oh, I vaguely see your direction. Well so long as you're aware of the issue for Linux, and remodelling this is on your todo list, I might leave it to you.

As such, I'll close this request.

Nadahar:
@sammyrc34 I'll absolutely fix it, it's a regression from my very recent refactoring.

PR #1393, opened 2017-09-24
Ask to restart UMS if user changes folders config

SubJunk:


SubJunk:
@UniversalMediaServer/developers hey guys, this is working for me and it would be cool to get it tested and reviewed. The better change is to rewrite code to not need a full program restart in the first place, but this was a fast improvement. We also have other settings that need restarts to take effect so we can grow the list over time.

Nadahar:
@SubJunk I thought I fixed the change of shared folders a while ago (without restarting UMS, only by pressing the "restart server" button. Are you saying that this is still broken?

Nadahar:
@SubJunk We've already got a restart function that I've made sure works on non-Windows platforms as well, why create a new one?

https://github.com/UniversalMediaServer/UniversalMediaServer/blob/1a648a0ac25a9b7a41e40c146b9145973742df7a/src/main/java/net/pms/util/ProcessUtil.java#L250-L344

SubJunk:
@Nadahar I didn't know we already had that ability. Yes the folders don't refresh for me without a full program close

Nadahar:
@SubJunk I tested it and it works - although the restart method is quite limited. I would remove the restart methods you added and use the one that's already there - I've done a fair amount of testing and fixes on that.

That said, I think it's quite an annoyance to have the restart pop-up every time you've changed something in the folders. Chances are that people want to make more than one change, and I'm betting that many will think that they "have to" restart for that change to be "saved".

As you know I'm not that much of a fan of "quick fixes". I think it would be much better if whatever problem keeping this from being registered was found and corrected. Clicking the "restart" button restarts both the HTTP server, the web interface and reloads all renderers so there must be some stupid thing that keeps this from working.

If looking for the proper solution is too much, I'd rather modify the "restart button" so that it either restarted just the internals (like it does today) or restarted the whole application depending on what was changed. That way, users could make all their changes before applying them, and the whole application would be restarted when that was needed.

Sami32:
@Nadahar I think that you are not talking about the same thing.
It's working fine in the master, at least for me, but @SubJunk and @ExSport reports seem to be exclusively about the 7.0.0 branch.
So i suspect that 7.0.0 have broken it again. I could be wrong though.

Nadahar:
@Sami32 That would make sense, in which case the fix should be done in 7.0.0. I haven't tested it now, but I thought I had fixed this some time ago.

@SubJunk Have you tested this in 6.x.x, or only in 7.0.0?

ExSport:
@Nadahar
It was me who reported issues with shared files.
Yes I tested v7 branch only. From UMS start it launches media parsing of all files in the system. When you change it to the specific folder(s) only and click on restart icon, UMS still scans root volumes and fills DB with unneeded files as new settings is ignored until full UMS restart.

Nadahar:
@ExSport I though this was about the super-old bug that has been there where UMS doesn't recognize/show new folders added in the renderer, which I thought I had fixed.

The scanning is changed a lot in 7.0 so I don't really know how that works, but it should be easy to stop/restart the scanning if the shared folders are changed.

SubJunk:
Good point, I hadn't considered it would be specific to 7.0.0 because I remembered that super old bug too. I'll look into it.

SubJunk:
@Nadahar I just tested it on latest master and it didn't work for me, do you want logs?

Nadahar:
@SubJunk I retested as well, and I've having mixed results. Some times it works, some times not. I've had success with adding new shared folders (although not always), while removing seems worse.

It seems to me that the "age old" bug is fixed, but there are now new layers of bugs at work. I'm not exactly user of the details here, but there's both the global IDs that's still referencing removed resources and it's the root folder structures themselves. The root folders (one per renderer) probably needs to be rebuilt from scratch as well. I tried just clearing the global IDs, and that didn't work and gave additional issues. I think the reason is that the per-renderer root folder still references those IDs. I didn't go any further now, as I'm actually working on something else.

What seems to work perfectly well is connecting a new renderer after the "restart button" has been pushed. The new renderer will see the updated tree, which indicates to me that the problem is indeeded with the existing "root folders".

That said, if this is about stopping a running scan, a method for doing that must be made. Once the scan has fetched the "top level"/shared folders, it won't recheck them during the scan - so it must be stopped and restarted after the information has been refreshed.

src/main/java/net/pms/PMS.java
```diff
@@ -2015,7 +1992,7 @@ public void askToRestart() {
 			Messages.getString("Dialog.Question"),
 			JOptionPane.YES_NO_OPTION);
 		if (option == JOptionPane.YES_OPTION) {
-			PMS.get().restartApplication();
+			ProcessUtil.reboot();
```
Nadahar: @valib 👍

valib:
@UniversalMediaServer/developers I retested this PR and found funny thing. The restart used in this PR doesn't work. During initiation of the restarted UMS there is an error:

```
ERROR 2017-10-31 22:16:43.352 [main] A serious error occurred during Universal Media Server initialization: Address already in use: bind
TRACE 2017-10-31 22:16:43.368 [main] 
java.net.BindException: Address already in use: bind
	at sun.nio.ch.Net.bind0(Native Method)
	at sun.nio.ch.Net.bind(Unknown Source)
	at sun.nio.ch.Net.bind(Unknown Source)
	at sun.nio.ch.ServerSocketChannelImpl.bind(Unknown Source)
	at sun.nio.ch.ServerSocketAdaptor.bind(Unknown Source)
	at sun.net.httpserver.ServerImpl.<init>(Unknown Source)
	at sun.net.httpserver.HttpServerImpl.<init>(Unknown Source)
	at sun.net.httpserver.DefaultHttpServerProvider.createHttpServer(Unknown Source)
	at com.sun.net.httpserver.HttpServer.create(Unknown Source)
	at net.pms.remote.RemoteWeb.<init>(RemoteWeb.java:90)
	at net.pms.PMS.init(PMS.java:651)
	at net.pms.PMS.createInstance(PMS.java:1147)
	at net.pms.PMS.main(PMS.java:1307)
```
The same command `ProcessUtil.reboot()` used when the UMS is forced to reboot after GUI language is changed works well. The same error ocurs when I tested original @SubJunk implementation. Maybe it is related to the Java version. Can anybody retest it?

Nadahar:
@valib I think this is a bad idea to start with, but the reason for this is that the servers aren't stopped first. Servers need to be stopped and sockets release before restart. I don't remember how or if I solved that for the language restart.

valib:
@Nadahar I think that you don't solve that but it works for language restart. Maybe it depends on where the reboot command is initializated. I will try to stop the server before the reboot is called.

valib:
@Nadahar that is not the problem. I stopped  the server and nothing changed. The problem is that `System.exit(0)` should stop the current JVM but it doesn't and after reboot the are two JVM instances.

Nadahar:
@valib ```System.exit(0)``` doesn't immediately shut down the VM if threads don't respond to interrupt. That's why I've been complaining about the bad interruption handling for a long time. It's kind of random if UMS will shut down properly or not. 

The whole way the application restart is done is also questionable, because a new JVM is started before the previous has exited, and for it to work you're depending on the timing of the two and that the previous process stops before the new process tries to allocate the resources. For the same reason, you will see that renaming of the logfile usually fails when the user use "Restart in TRACE mode".

There's no easy fix for it though, because there are so many places in the code that has to be changed. Ideally, you wouldn't need to restart UMS at all - just close down everything and release all instances and run the initialization again. That would be the "proper" way to do it, but there's so many reasons why that isn't possible without months of refactoring.

To increase the chance of success you should shut down as many things as possible, but in the end it will never be reliable. This is the main reason why I think restarting UMS instead of actually solving the bugs is a very bad idea.

SubJunk:
The code I committed was working for me, I tested it a lot

valib:
@SubJunk I retested it now on the different PC with the same result. 
`JVM: Java HotSpot(TM) 64-Bit Server VM 1.8.0_144 (64-bit)
  OS: Microsoft Windows 10 build 16299 64-bit`

SubJunk:
@valib thanks for testing. Do you mean you tested the code I pushed (with my new restart function) or you tested it after https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1393/commits/8e21d3dd24b88b6dd211a9003a6249112b93c71d ?
The one that was working for me was my original code

valib:
@SubJunk I tested both. Your commit and my update with the same result. As @Nadahar wrote it seems that the place where the reboot is called is important. The @Nadahar implementation (which is comparable to your implementation) works when it is called from the GeneralTab and the LanguageSelection but not working for me when called from the main class PMS.java.

Nadahar:
It's not decided by where it's called from, but what threads and processes are currently running at the time ```System.exit()``` is called. The probem is very simple to understand, yet hard to fix.

Whenever ```System.exit()``` is called, registered shutdown hooks are started in parallel (libraries register shutdown hooks as well as UMS). It's a time of "complete chaos" because the order in which these are run are random, and some things might fail because something else has already happened/been disposed/destroyed. The JVM will wair for the shutdown hooks to complete, and will then (AFAIK) interrupts any remaining non-daemon threads. The JVM will then probably wait a given timeout and then kill the remaining threads and stop the JVM process. As we've seen before, things can keep this from ever happening, in which case the JVM never terminates (unless killed manually or the computer is restarted).

The "restart" behaviour in UMS (it's not implemented by me, I just fixed some bugs and crashes) is fundamentally flawed. Since it's impossible to do anything after ```System.halt()``` is called, the new JVM is started just before ```System.halt()``` is called. The problem is that this is also before the whole shutdown sequence of the old JVM, which can take anything from a few milliseconds to many seconds or more depending on the local environment and what is currently running in UMS. 

The whole approach depends on that the shutdown of the old JVM will happen fast enough so that the same resources (files and sockets) that are to be allocated by the new JVM has been deallocated before the new JVM tries to allocate them. This is completely random, and testing this on any given computer is pretty much useless because the time spend on both shutdown and startup depends on a lot of factors that UMS has no control or knowledge about.

I did check a bit around this a while ago, and there are several approaches used but all are really just hacks to make it somewhat more likely that it will succeed. Introducing some kind of delay in the startup of the new JVM is obviously beneficial, but as there's no way to know how long you should wait, it will just be a guess that will or won't work depending on the condition. If you added a 30 seconds delay, I'd say that it would probably work in 99% of cases, but it would also mean that restarting would take 30 seconds for everybody.

There are also wrappers that can take care of the restarting, which is a slightly better approach but it's challenging to get it to work on all platforms.

My conclusion was simply that there's no "clean" way to restart the JVM from within a Java program, at least not that I could find.

It would probably be possible to find a way if a lot of effort was put into this, but as I see it making sure that no restart is needed requires a lot less effort. I didn't pursue it any further for that reason, restarting is after all just a lazy way to avoid solving the real problems.

SubJunk:
Closing this since it seems it has failed testing

PR #1400, opened 2017-10-03
Fix charset handling during subtitle conversions

### LINKED ISSUE #1396: Subtitle encoding issue with Samsung E+ Series
Hi guys,
The srt subtitles show bizarre simbols on Samsung E+ Series TV
The only way to fix it is adding this line to the bottom of Samsung cfg profile for Samsung E+ Series TV

`RemoveTagsFromSRTSubtitles = false`

Please fix it for the next releases.
Is annoying add this line every time you update, and also are so many people who can't figurate how to do it.
Cheers.

Nadahar:
This is based on #1396. I found that when converting subtitles, the character set wasn't being kept and the output file was always UTF-8. This PR makes sure that the converted subtitle character set is the same as the original as long as it's properly detected, and it's supposed to fix #1396.

There might be other ways to get around the problem in #1396, for example by updating the subtitle charset information to indicate UTF-8. I think it's nice to have a way to output the same character set as the input in any case.

@valib This is completely untested and something I just threw together rather quickly. I hope you can take a look at it and do some tests, since you're the one that knows this best.

### DIFF
--- a/src/main/java/net/pms/util/FileUtil.java
+++ b/src/main/java/net/pms/util/FileUtil.java
@@ -1505,16 +1505,44 @@ public class FileUtil {
 		return filename;
 	}
 
+	/**
+	 * @deprecated Use {@link #createBufferedReaderDetectCharset(File, Charset)} instead.
+	 */
+	@Deprecated
 	public static BufferedReader bufferedReaderWithCorrectCharset(File file) throws IOException {
+		return createBufferedReaderDetectCharset(file, StandardCharsets.UTF_8).getBufferedReader();
+	}
+
+	/**
+	 * Attempts to detect the {@link Charset} used in the specified {@link File}
+	 * and creates a {@link BufferedReader} using that {@link Charset}. If the
+	 * {@link Charset} detection fails, the specified default {@link Charset}
+	 * will be used.
+	 *
+	 * @param file the {@link File} to use.
+	 * @param defaultCharset the fallback {@link Charset} it automatic detection
+	 *            fails. If {@code null}, the JVM default {@link Charset} will
+	 *            be used.
+	 * @return The resulting {@link BufferedReaderDetectCharsetResult}.
+	 * @throws IOException If an I/O error occurs during the operation.
+	 */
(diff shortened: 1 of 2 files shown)


Nadahar:
@valib Have you tested it?

valib:
@Nadahar yes I tested it and it works for me. I had the same idea to use for the output stream the detected charset but to use the `defaultCharset` instead of the `UTF-8` when the charset is not recognized is clever.

Nadahar:
@valib It's not as clever as it might seem, because that will still be UTF-8 for the time being. In the future, it might be something more useful. The reason for this is that UMS doesn't properly handle character sets a lot of places in the code. I'm sure you've seen that I've been correcting that when I come by it. 

A lot of operations will use the default character set if you don't explicitly set one. The default character set is set to that of the local OS by the JVM. I'm sure that has caused a lot of problems at one stage (either during PMS or UMS), so somebody "clever" has made sure that the default character set it forced to UTF-8 when UMS starts. As a result the default character set is *always* UTF-8 and we have lost the information about the OS' "native" character set.

I believe that removing the forcing of UTF-8 will create a lot of problems as there's still a lot of code that use the default character set. Until all that code is updated, we can't remove the forced UTF-8 setting and thus won't be able to know the OS character set.

valib:
@Nadahar on the other hand I think we should make `RemoveTagsFromSRTSubtitles = false` as a default option because there could be renderers whose are capable to handle HTML tags and can correctly show all that subs formatting. If the renderer is no capable to do that users have still an option to enable it.

Nadahar:
@valib I don't know/remember why we made it the default. It might be better if it was off by default, but I doubt very many renderers support it. The reason is that it's a complete lack of any standard for "SRT tags" and only a very few software players actually "support" this. I have a really hard time to believe that those making renderers at Sony, Panasonic, Samsung etc. somehow implements something that you have to search around the internet to even get a vague idea of how could potentially work.

valib:
@Nadahar I think that this PR should be merged. If it solve the problem reported in  #1396 my comment  about the default setting `RemoveTagsFromSRTSubtitles = true` is not relevant.

Nadahar:
I merged this after successful user test in #1396.

Issue #1401, opened 2017-10-05
Bad hack breaks contentFeatures when transcoding to non-NTSC


Nadahar:
Based on [this forum thread](http://www.universalmediaserver.com/forum/viewtopic.php?f=9&t=10749). Although I can't say for sure that this is the cause for the user's problem, I can say for sure that this is a terrible hack that will break a lot. It simply seems like somebody didn't finish their code and it has stayed this way.

When looking at one of the problematic files, this is the DLNA information given in the DIDL-Lite:
```powershell
DLNA.ORG_PN=MPEG_PS_PAL;DLNA.ORG_OP=01
```
Later when the renderer asks for ```contentFeatures``` for the same video, this is sent:
```powershell
DLNA.ORG_PN=MPEG_PS_NTSC;DLNA.ORG_OP=01;DLNA.ORG_CI=0;DLNA.ORG_FLAGS=01700000000000000000000000000000
```
According to DLNA these are required to be **identical**. I suspect that CI and FLAGS doesn't cause much havoc though, but the PN definitly will.

The reason is also easy to see:
https://github.com/UniversalMediaServer/UniversalMediaServer/blob/5d1a197f6e78de70aa271d0d4ace2ea80521ca88/src/main/java/net/pms/dlna/DLNAResource.java#L442-L447

```localizationValue == 1``` simply means NTSC, so NTSC will always be used for ```contentFeatures```. Although there's a comment there, that doesn't help. The code is still broken. I don't see any obvious way to know what value to use either, since UMS doesn't track what ```<res>``` element the renderer picks. From what I can see from the very messy code generating the DIDL, one version is offered for each "localization" if the renderer "require localization". That isn't the case for the renderer in question here, so only one is returned. How that ends up being PAL is beyond me, but it's probably due to some other hack. The code seems to me to simply offer NTSC to any renderer that doesn't "require localization". 

This will fail also if the renderer is set to "require localization" because the renderer will then be offered different versions of the file (or at least with different DLNA profiles while actual compliance to the particular profile is ignored). If the renderer doesn't pick the first (NTSC/NA version), and later requests ```contentFeatures``` the wrong result will be returned again - since it's hardcoded to 1.

[...]

PR #1402, opened 2017-10-05
Dirty fix for testing the issue #1401

valib:
The DIDL-Lite sends always the MPEG_PS_PALLocalizedValue but it should be determined if renderer is
capable to use it by using ProtocolInfo.

### DIFF
--- a/src/main/java/net/pms/dlna/DLNAResource.java
+++ b/src/main/java/net/pms/dlna/DLNAResource.java
@@ -34,7 +34,6 @@ import net.pms.configuration.FormatConfiguration;
 import net.pms.configuration.PmsConfiguration;
 import net.pms.configuration.RendererConfiguration;
 import net.pms.dlna.DLNAImageProfile.HypotheticalResult;
-import net.pms.dlna.protocolinfo.ProtocolInfo;
 import net.pms.dlna.virtual.TranscodeVirtualFolder;
 import net.pms.dlna.virtual.VirtualFolder;
 import net.pms.dlna.virtual.VirtualVideoAction;
@@ -444,7 +443,7 @@ public abstract class DLNAResource extends HTTPResource implements Cloneable, Ru
 		int localizationValue = 1;
 		String dlnaOrgPnFlags = null;
 		if (getType() == Format.VIDEO || getType() == Format.AUDIO) {
-			String orgPnValue = requestedURL.substring(requestedURL.lastIndexOf("_") + 1, requestedURL.lastIndexOf("."));
+			String orgPnValue = requestedURL.substring(requestedURL.lastIndexOf("/") + 1, requestedURL.indexOf("_"));
 			if (orgPnValue != null) {
 				dlnaOrgPnFlags = "DLNA_ORG_PN=" + orgPnValue;
 			}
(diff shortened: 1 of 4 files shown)


valib:
@Nadahar can you help me and give me an advise how to use the renderer ProtocolInfo to determine its capability to play the given media type?

Nadahar:
@valib Each ```Renderer``` (and ```RendererConfiguration```) have a ```deviceProtocolInfo``` field that contains the protocolInfo information for that renderer. It doesn't seem like I made a getter for it, so access is directly via the public field for now. It's easy to add a getter in ```UPnPControl.Renderer``` though, where the field is defined. I was working on an expansion of the protocolInfo stuff that SubJunk didn't have time for me to finish, so I just dropped it then and haven't looked at it since. I might had it planned there, but it's easy to add anyway. 

This isn't a big deal though, as the class held in ```deviceProtocolInfo``` is ```DeviceProtocolInfo```, which is thread-safe. The field is also ```final``` so it's not possible to "set" the field anyway.

That said, your "dirty fix" won't properly fix it. I bet there are renderers that support more than one of these "localized" profiles (I'm pretty sure I've seen that in many log files), we'd have to offer mutiple versions to the renderer (as is, in a way, done today via the renderer setting). The root problem is that when we get the request from the renderer, there's no way to know which of the links were chosen by the renderer (as they all share the same resource id). Using protocolInfo will thus only "solve" it for renderers that only supports one "version".

Nadahar:
@valib When it comes to the use of ```DeviceProtocolInfo``` itself, I don't remember now, but I think it's pretty well documented. It behaves much like a ```List```, but I see now that I didn't implement ```Iterable``` which means you can't use it in a simple ```for``` loop. When I look at it now it also seems that I forgot to include a ```get()``` method, so you'd have to use ```toArray(DeviceProtocolInfoSource<?> type)``` to iterate over the ```ProtocolInfo``` instances.

In the loop, you'd have to check for the profile you're interested in. Those listed are those reported by the renderer. Not all renderers report anything, so the list might be blank.

When you have a ```ProtocolInfo``` instance, getting the profile is easy: ```getDLNAProfileName()```. What you get then is a ```DLNAOrgProfileName``` instance, on which you can call ```getValue()``` to get the string content (everything after ```DLNA.ORG_PN=```).

valib:
@Nadahar agree but it means to add new DLNAResource for each combination of the DlnaOrgPnFlags to be able to be choosen the propper one by renderer. This could be wasting the resources but will be better than current implementaion.

Nadahar:
@valib No, that's not necessary. I've done a kind of a "hack" for this for images, which could be used for other things as well. I add the PN to the link for images, and then I parse it from the link when the request is received. That way I know what profile the renderer picked ;)

valib:
@Nadahar nice but can you add it also for video files?

Nadahar:
@valib It should be possible, take a look at the code used in DLNAResource for ```getURL()``` (AFAICR).

valib:
@UniversalMediaServer/developers can somebody test it?

Nadahar:
@valib I don't think I have any renderer that can test this. You'd need a renderer that use ```ContentFeatures``` and that requests non-US DLNA profiles. The easiest way to test it is probably to get users that has such renderers to do it.

I have a hard time reviewing it as well, as I don't quite understand why you do the changes that you do. You're mixing in changes that seems unrelated to this, and I don't understand what you use ```ProtocolInfo``` for in here.

valib:
@Nadahar the `ProtocolInfo` was in the first attempt when I tried to use it to recognise renderers capability but is not used anymore. I take your "hack" and simply add the PN to the link for video and audio. and check it when `ContentFeatures` are requested.

Nadahar:
@valib Why is that code still part of this PR then?

src/main/java/net/pms/dlna/DLNAResource.java
```diff
@@ -1928,10 +1936,11 @@ private String getDlnaOrgOpFlags(RendererConfiguration mediaRenderer) {
 	 * @return String representation of the DLNA.ORG_PN flags
 	 */
 	@SuppressWarnings("deprecation")
-	private String getDlnaOrgPnFlags(RendererConfiguration mediaRenderer, int localizationValue) {
+	private String getDlnaOrgPnFlags(RendererConfiguration mediaRenderer, int localization) {
```
SubJunk: Why rename this parameter here and then reassign it to the old name on line 1943?

Nadahar:
@valib My "problems" with this PR are:

* I don't see the connection between this PR and updating the dependencies. I'm not against updating dependencies, but as always I think there should be a reasonable effort done to check what has changed and estimate the probability that it will affect UMS. I think these changes should at least be in their own commit, and I'd like some words on what considerations have been done about potential issues as a consequence of updating these dependencies.
* I think it's unnecessary messy to leave the code you ended up not using in the PR. As I prefer to review on a commit basis, it is confusing.
* I don't quite understand why you didn't use the existing logic used for images or atleast a very similar logic. We want to make sure these codes are "compatible" and doesn't interfere with each other, so the "formatting rules" should be very strict.

SubJunk:
@valib can this be closed?

PR #1404, opened 2017-10-15
Fix DTS and AC-3 passthrough regression

Nadahar:
This fixes a regressions in c4de4b52b47cd05e5322a78bd9333e5d086ca7ab that was caused by me (while rebasing), although the commit was by @Sami32. I had changed the capitalization of some words in ```DLNAMediaAudio.toString()``` which broke the bad hack made in 5e5338c7ed4d6f6521b574754d88ca5c9f49b8ca. Thanks to @Sami32 for finding this sneaky bug 👍 

The hack made was very unsafe and was bound to break sooner or later. As such I didn't simply "fix" it by correcting the capitalization, as that would just break the next time someone did something with the log output.

@SubJunk Neither your commit message nor your code makes it quite clear to me what the purpose of 5e5338c7ed4d6f6521b574754d88ca5c9f49b8ca was. To correct it I've thus had to guess a bit. It seems to me that the intention was for the video to be transcoded not only if the selected audio track is DTS or AC-3, but if *any* of the audio tracks are. My fix has thus implemented that. If the intention was somewhat different, so please tell me if what I've done is correct or wrong.

I've also added DTSHD to this, although I don't know if DTSHD works with audio passthrough. I assume that some of you already know this, so please let me know if DTSHD should stay or be removed from this condition.

### DIFF
--- a/src/main/java/net/pms/dlna/DLNAResource.java
+++ b/src/main/java/net/pms/dlna/DLNAResource.java
@@ -929,17 +929,33 @@ public abstract class DLNAResource extends HTTPResource implements Cloneable, Ru
 			} else if (!format.isCompatible(media, renderer)) {
 				isIncompatible = true;
 				LOGGER.trace(prependTraceReason + "it is not supported by the renderer", getName());
-			} else if (
-				configurationSpecificToRenderer.isEncodedAudioPassthrough() &&
-				getMediaAudio() != null &&
-				(
-					"AC3".equals(getMediaAudio().getAudioCodec()) ||
-					"DTS".equals(getMediaAudio().getAudioCodec())
-				)
-			) {
-				isIncompatible = true;
-				LOGGER.trace(prependTraceReason + "the audio will use the encoded audio passthrough feature", getName());
-			} else if (format.isVideo() && parserV2 && renderer != null) {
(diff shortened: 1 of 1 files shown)


SubJunk:
Nice change, yes you interpreted that correctly

Nadahar:
I had to remove DTSHD as that didn't work. I'm going to merge this now.

ExSport:
As I posted somewhere in past, Shagrath said HD audio is not possible so in DTS-HD case only its dts core can be encapsulated for passthrough...

Nadahar:
@ExSport 👍 Yes, that's why I removed it. DTS-HD isn't supported by the S/PDIF protocol.

ExSport:
Don't know how dtshd is detected and then excluded for passthrough but what I remember is that all dtshd include dts core (standard?) so if such audios will not be excluded, then great👍

Nadahar:
@ExSport Yes, all DTS versions always have the "core" DTS stream which is in the original DTS format. I didn't want to dig into that now, only fix the regression. It should be created as a separate issue/enhancement. The core DTS would have to be extracted from the HD "stream" and S/PDIF encapsulated,

ExSport:
Great, thanks. So mentioned as a note for future enhancement:)

SubJunk:
Nice work

PR #1406, opened 2017-10-24
Fix exception in GetProtocolInfo

Nadahar:
It turns out that Cling throws and ```IllegalArgumentException``` if you ask for an "action" that doesn't exist in the output. Usually, and per UPnP specification AFAICR, ```Sink``` will be among the actions, but I've just studied a user log where that wasn't the case - resulting in an unhandled (runtime) ```Exception```.

I've rewritten the parsing slightly to avoid that situation.


Nadahar:
I have done basic testing, it should be safe.

Issue #1411, opened 2017-10-31
UMS scans the entire disk instead of chosen media folder


TacticalCode:
I downloaded the latest UMS yesterday to test streaming to a FireTV Stick. I'm running Linux (arch) and started UMS by typing java -jar ums.jar in a terminal. After setting some options with help of the first-start-wizard and GUI, I added a folder to share and noticed strange things coming up in the terminal I launched UMS in:

> INFO  00:01:05.869 [main] Universal Media Server is now available for renderers to find
> INFO  00:02:35.330 [AWT-EventQueue-0] Setting default renderer to "Chromecast"
> INFO  00:03:04.407 [AWT-EventQueue-0] Setting Audio thumbnails import: None
> WARN  00:04:55.787 [Library Scanner] Can't read directory: /boot/lost+found
> WARN  00:04:56.008 [Library Scanner] Can't read directory: /etc/audisp/plugins.d
> WARN  00:04:56.448 [Library Scanner] Can't read directory: /etc/cups/ssl
> WARN  00:04:56.811 [Library Scanner] Can't read directory: /etc/libvirt/secrets
> WARN  00:04:56.827 [Library Scanner] Can't read directory: /etc/luks-keys
> WARN  00:04:56.837 [Library Scanner] Can't read directory: /etc/lvm/cache
> WARN  00:04:56.889 [Library Scanner] Can't read directory: /etc/NetworkManager/system-connections
> WARN  00:04:56.895 [Library Scanner] Can't read directory: /etc/openvpn/client
> WARN  00:04:56.896 [Library Scanner] Can't read directory: /etc/openvpn/server
> WARN  00:04:56.902 [Library Scanner] Can't read directory: /etc/pacman.d/gnupg/openpgp-revocs.d
> WARN  00:04:56.902 [Library Scanner] Can't read directory: /etc/pacman.d/gnupg/private-keys-v1.d
> WARN  00:04:56.938 [Library Scanner] Can't read directory: /etc/polkit-1/rules.d
> WARN  00:04:56.963 [Library Scanner] Can't read directory: /etc/samba/private
> WARN  00:04:57.372 [Library Scanner] Can't read directory: /etc/sudoers.d
> WARN  00:04:57.449 [Library Scanner] Can't read directory: /etc/vmware/ssl

[...]

PR #1412, opened 2017-11-03
Update wizard

valib:
Add file chooser to the wizard to partialy solve the issue #1411. Set the wizard as mandatory for the clean install.

### DIFF
--- a/src/main/java/net/pms/PMS.java
+++ b/src/main/java/net/pms/PMS.java
@@ -568,6 +568,10 @@ public class PMS {
 			chooser.setMultiSelectionEnabled(false);
 			if (chooser.showOpenDialog(null) == JFileChooser.APPROVE_OPTION) {
 				configuration.setFolders(chooser.getSelectedFile().getAbsolutePath());
+			} else {
+				// If user cancel this option set the default directory which depends on the operating system.
+				// It is typically the "My Documents" folder on Windows, and the user's home directory on Unix.
+				configuration.setFolders(chooser.getCurrentDirectory().getAbsolutePath());
 			}
 
 			// The wizard finished, do not ask them again
--- a/src/main/java/net/pms/PMS.java
+++ b/src/main/java/net/pms/PMS.java
@@ -470,15 +470,6 @@ public class PMS {
 			if (splash != null) {
 				splash.setVisible(false);
 			}
-			// Inform user that the wizard will run
-			JOptionPane.showMessageDialog(
-				null,
-				Messages.getString("Wizard.1"),
-				Messages.getString("Dialog.Information"),
-				JOptionPane.INFORMATION_MESSAGE
-			);
-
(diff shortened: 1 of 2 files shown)


SubJunk:
I'm excited about this, will test soon!

SubJunk:
When it starts up we should remove the alert that says "Starts the UMS configuration wizard". It isn't an option so it should just start. I might make that change now

SubJunk:
@UniversalMediaServer/developers what do you think about this now? I'm happy to merge if you are

valib:
@SubJunk agreee with your update and I rebased all commits. If others agree and somebody review it we can merge it.

Nadahar:
Don't wait for me to approve it, I think it's a rather hackish way to do it. If you guys think it's good, go ahead.

If you wonder why I think it's hackish, I think that should be obvious, but in case it isn't: I don't think it really solves anything. It's probably just confusing to users to suddenly get an open file dialog presented and only being able to select one folder. Who shared just one folder anyway? What happens if the cancel or close the dialog? If they have to navigate to the navigation tab to add the rest of their folders, what's the point with the whole thing?

In my opinion it would be better to either isolate the part of the GUI that handles the shared folders to a separate JPanel subclass and then simply display this panel as a part of the wizard. Alternatively, as simpler solution would be to make UMS focus on the correct tab and section after starting if the wizard was running.

The bugs that prevents the configured folders form actually being used without a restart would have to be fixed either way, or it still won't make any sense to users.

valib:
@Nadahar mayby this partially satisfied your expectation.

SubJunk:
I agree with you @Nadahar that having the real folder chooser in the wizard would be better. I tried to do that a while ago and failed. I believe in iterative improvements so the question for me is "is this better than what we currently have?" rather than "is this perfect?". For me the answer is "yes, it is better than what we have, and it can and should be improved later".

SubJunk:
@valib thanks for the extra work on it. I noticed that the others have done some related work in DMS too https://github.com/DigitalMediaServer/DigitalMediaServer/pull/26

I'm not happy with the new language so will edit that before merging

Nadahar:
@SubJunk I don't know since when I became "others". I would have mentioned https://github.com/DigitalMediaServer/DigitalMediaServer/pull/26 if I thought it was relevant. It isn't as I see it, since it breaks plugins and reverses #140. The beauty of working with DMS is that I don't have to care about that.

Making this work without the reversal and breaking the plugins would've made it a much bigger task, which I don't think is worth it.

Nadahar:
@SubJunk @valib My original comment wasn't meant as a complaint, I was simply trying to say "Don't wait for my approval, do as you see fit". I generally try to stay away from things I don't agree with or see the point with if I consider them relatively harmless. As always, I explain myself a bit too much which makes it sound like critisism, which wasn't really my intention. I was just trying to explain why I'm not *for* this.

PR #1415, opened 2017-11-13
Fix codepage 932 Japanese

Nadahar:
This is just a "small" bug I found in a user log, I don't know if it actually have had any impact. It's still a bug though.

The issue is that when trying to find the OS' native charset for Windows, UMS calls the Windows API and gets the code page number in return. This is then prefixed with ```cp``` and used to find the corresponding Java ```Charset```.

For other codepages this works well, but for some reason Java doesn't have an alias for ```cp932```, making this fail for Japanese. I'd consider it a Java bug, but I found that prefixing it with ```MS``` instead of ```cp``` makes it work. 

I've thus made this PR, which simply tries both prefixes, and the error went away when tested by the affected user.

### DIFF
--- a/src/main/java/net/pms/PMS.java
+++ b/src/main/java/net/pms/PMS.java
@@ -26,7 +26,10 @@ import com.sun.net.httpserver.HttpServer;
 import java.awt.*;
 import java.io.*;
 import java.net.BindException;
+import java.nio.charset.Charset;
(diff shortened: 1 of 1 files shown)


Nadahar:
@valib Thanks. I thing the risk for complications is extremely low, so I'm merging this.

PR #1372, opened 2017-08-24
Fix bug for subs recognition

valib:
Rebased PR #1098 and fixed the bug reported in #1364.

@UniversalMediaServer/developers sorry guys that I did not continue on PR #1098. I tried to add some features but without success so I left it for some time and forgot on it 😞

**Note**: Should fix #1364, #1060, #1338 and #1434

PR #1419, opened 2017-11-25
Update ICU4J to version 60.1 and update subs recognition. 

valib:
It is made to avoid repeated subs parsing and code page detection.

### DIFF
--- a/src/main/java/net/pms/dlna/DLNAMediaInfo.java
+++ b/src/main/java/net/pms/dlna/DLNAMediaInfo.java
@@ -1117,7 +1117,7 @@ public class DLNAMediaInfo implements Cloneable {
 				}
 			}
 
-			postParse(type, inputFile);
+			postParse(type, inputFile, thumbOnly);
 			mediaparsed = true;
 		}
 	}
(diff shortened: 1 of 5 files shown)


src/main/java/net/pms/encoders/Player.java
```diff
@@ -409,8 +409,8 @@ public static void setSubtitleOutputParameters(String fileName, DLNAMediaInfo me
 		 */
 		if (!configuration.isDisableSubtitles() && params.sid == null && media != null) {
 			// Check for subtitles again
-			File video = new File(fileName);
-			FileUtil.isSubtitlesExists(video, media, false);
+//			File video = new File(fileName);
+//			FileUtil.isSubtitlesExists(video, media, false);
```
Nadahar: @valib I don't think leaving commented out code is a good practice unless there's a good reason, the code ends up being even more unreadable. If there's a good reason, it should be explained in a comment IMO. @SubJunk has said the same before.

Nadahar:
@valib I don't know what the code does/how it's used, so I'm unable to test it unless you describe how to use it and what has changed.

src/main/java/net/pms/encoders/Player.java
valib: @Nadahar I leave it here only for the case when testing will be wrong. It could be deleted when developers approve that thich change is working.

valib:
@Nadahar the test is simple. Just test if your setting for subs language meet your preferencies. Before you can see in the log that the charset recognition was done twice. After this change it is done only once.

SubJunk:
Ok so I spent some time figuring out the reproduction steps for this.

If a file has *external* subtitles, and its parent folder is being scanned, the charset recognition happens twice (verified with the log entry "Set detected charset".

In this branch is happened three times, so it seems to have got worse. Want logs?

valib:
@SubJunk thanks for testing. The problem is hidden when the database is enabled and thumbnails are taken from it. I overlooked it.

src/main/java/net/pms/dlna/DLNAMediaInfo.java
```diff
@@ -1649,8 +1670,8 @@ public void postParse(int type, InputFile f) {
 		}
 
 		// Check for external subs here
-		if (f.getFile() != null && type == Format.VIDEO && configuration.isAutoloadExternalSubtitles()) {
-			FileUtil.isSubtitlesExists(f.getFile(), this);
+		if (file.getFile() != null && type == Format.VIDEO && configuration.isAutoloadExternalSubtitles()  && !thumbOnly) {
+			FileUtil.isSubtitlesExists(file.getFile(), this);
 		}
```
Nadahar: This code is being reworked in #1372.

valib: @Nadahar OK I will wait until you finished the #1372 but it can give you some proposal.

SubJunk:
Cool I will wait for #1372 before testing again

SubJunk:
@valib so this was waiting for that subs branch to merge and it has now, is this still relevant?

valib:
This was reworked in  #1372 #1483 and #1615

PR #1420, opened 2017-11-26
Improved MusicBrainz match logic for cover art retrieval

atamariya:
I added some improvement to MusicBrainz match logic. This has helped me retrieve 1000s of cover art for my MP3 collection. Hope it helps others too.

If cover art is still missing , check if the MBID is referring to correct release on MB site.

```
select * from MUSIC_BRAINZ_RELEASES where mbid in (select mbid from COVER_ART_ARCHIVE where cover is null) order by album
```

If release is correct and thumbnail shows "Cover art from Amazon", make sure you update the cover art by uploading a "FRONT" thumbnail for the release.

### DIFF
--- a/src/main/java/net/pms/util/CoverArtArchiveUtil.java
+++ b/src/main/java/net/pms/util/CoverArtArchiveUtil.java
@@ -19,6 +19,8 @@
  */
 package net.pms.util;
 
+import static org.apache.commons.lang3.StringUtils.isBlank;
+import static org.apache.commons.lang3.StringUtils.isNotBlank;
 import fm.last.musicbrainz.coverart.CoverArt;
 import fm.last.musicbrainz.coverart.CoverArtException;
 import fm.last.musicbrainz.coverart.CoverArtImage;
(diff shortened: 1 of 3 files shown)


src/main/java/net/pms/util/CoverArtArchiveUtil.java
```diff
@@ -534,24 +544,26 @@ private String buildMBReleaseQuery(final CoverArtArchiveTagInfo tagInfo, final b
...
-				query.append(urlEncode(fuzzString(tagInfo.artist)));
-			} else {
-				query.append(urlEncode("\"" + StringUtil.luceneEscape(tagInfo.artist) + "\""));
-			}
-			added = true;
-		}
+		// Release (album) artist is usually the music director of the album. Track (Recording) artist is usually the singer.
+		// Searching release with artist here is likely to return no result
+//		if (StringUtil.hasValue(tagInfo.artistId)) {
+//			if (added) {
```
SubJunk: Can these lines just be removed instead of commented out? Maybe the note could remain, to remind people not to reimplement that logic

Nadahar:
@valib Argh, you did it again.... 😭 Why those pesky merge commits everywhere?

Nadahar:
I've made some suggested changes in https://github.com/UniversalMediaServer/UniversalMediaServer/tree/MusicBrainz.

atamariya:
This looks good.

Nadahar:
@atamariya I know I'm too detail focused, but I like to make sure that every commit compiles in case somebody wants to check out random commits laters to look for a regression or similar. Right now that's not the case here, because the order of the commits is wrong.

The branch I've rebased has the order correct, and I've also fixed a small spelling mistake I made. I don't know what your "remote name" for the UMS repo is, but pretending it's ```upstream``` all you have to do when being checked out to this branch is:
```powershell
git fetch upstream
git reset --hard upstream/MusicBrainz
```

atamariya:
@Nadahar done.

PR #1423, opened 2017-12-07
Fixed issue with the new MediaInfo versions

valib:
Problem reported in http://www.universalmediaserver.com/forum/viewtopic.php?f=5&t=11643&p=34540#p34540


SubJunk:
Is this meant to work with MediaInfo 17.12? If so I will update to that and merge this for our next release

valib:
@SubJunk yes it works also for the MediaInfo 17.12 but this is only the quick fix because I think that the concept of the `getFormat` class in the LibMediaInfoParser should be rewritten. The current problem is that the MediaInfo detects for some videos the audio codec `mp4a-40-2` which is taken in the LibMediaInfoParser as a container. Some time ago I divided format recognition to the container, video and audio parts but after some changes it is back like in the PMS era and the format is recognised doesn't matter if it is container, video  or audio codec. Maybe we can try to not check the audio codec https://github.com/UniversalMediaServer/UniversalMediaServer/blob/876b7cda6c8ccce067f16e9150345a3bd617a857/src/main/java/net/pms/dlna/LibMediaInfoParser.java#L198 when the format is already parsed. The same applies to the video codec.

src/main/java/net/pms/dlna/LibMediaInfoParser.java
```diff
@@ -471,7 +471,7 @@ private static void getFormat(StreamType streamType, DLNAMediaInfo media, DLNAMe
 			format = FormatConfiguration.MOV;
 		} else if (
 				value.contains("isom") ||
-				value.startsWith("mp4") ||
+				(streamType != StreamType.Audio && value.startsWith("mp4")) ||
```
Sami32: It will still leave you with some other issue, like MP4 SLS

Nadahar: @Sami32 I didn't quite understand, are you saying that this breaks SLS?

Sami32: Just that it will fail on some format and still have a bad, incorrect parsing.

SubJunk:
This looks good to me, but I'll test it some more to make sure

SubJunk:
@valib I have tested that this correctly picks up AAC-LC and HE-AAC audio streams within MP4 containers, but it doesn't seem to handle MP4 audio files properly, as @Sami32 said:

```
TRACE 2018-04-20 12:15:26.140 [Library Scanner] Matched format MPG to "sls2211_aot22_048_16.mp4"
TRACE 2018-04-20 12:15:26.160 [Library Scanner] Parsing results for file "sls2211_aot22_048_16.mp4": Container: MP4, Size: 2081596, Bitrate: 342981, Duration: 0:00:48.000, Audio Tracks: 2 [Audio Codec: und, Bitrate: 92297, Channels: 2, Sample Frequency: 48000 Hz], [Audio Codec: mp4, Bitrate: 244493, Channels: 2, Sample Frequency: 48000 Hz], Mime Type: video/mpeg
TRACE 2018-04-20 12:15:26.239 [Library Scanner] Adding new child "sls2211_aot22_048_16.mp4" with class "RealFile"
TRACE 2018-04-20 12:15:26.239 [Library Scanner] This video does not have subtitles
TRACE 2018-04-20 12:15:26.239 [Library Scanner] Getting player for resource "sls2211_aot22_048_16.mp4"
TRACE 2018-04-20 12:15:26.239 [Library Scanner] Returning compatible player "FFmpeg"
TRACE 2018-04-20 12:15:26.239 [Library Scanner] Matched support line f:.+
TRACE 2018-04-20 12:15:26.239 [Library Scanner] Matched support line f:.+
TRACE 2018-04-20 12:15:26.239 [Library Scanner] Final verdict: "sls2211_aot22_048_16.mp4" will be streamed
TRACE 2018-04-20 12:15:26.239 [Library Scanner] Matched support line f:.+
TRACE 2018-04-20 12:15:26.239 [Library Scanner] Matched support line f:.+
TRACE 2018-04-20 12:15:26.239 [Library Scanner] File "sls2211_aot22_048_16.mp4" will be sent with MIME type "MIMETYPE_AUTO"
```

SubJunk:
Ok I have tested this current code with a focus on 3 MP4 files:
MP4 with ER BSAC/SLS audio
MP4 with AAC-LC/SLS audio
MP4 with H.264 video and AAC-LC audio

All of them are correctly detected.

@UniversalMediaServer/developers does anyone have any other concerns about this or can it be merged?

Sami32:
@SubJunk Good idea to have added ER BSAC 👍

SubJunk:
@UniversalMediaServer/developers thanks for the comments guys, I have addressed them now. Does it look good to merge?

valib:
@SubJunk what about to use my recommendation to divide the recogniton to the container, video and audio? I know that some formats will be repeated in the container, video or audio parts but it will be clean what the UMS is doing. Also the format recognition from the "CodecID" is arguable. It can match to different formats because it is only the codec version which can supports different formats.

SubJunk:
@valib I think that's a very good recommendation and it will clean up that file a lot. I added it as a `todo` in this branch. I think we should do that in a separate branch so it is easier to compare though so maybe we merge this one and then do that?

valib:
@SubJunk I have no objections to do it in the way you proposed.

PR #1428, opened 2017-12-21
Update database to store mimetype.

valib:
This PR could a little bit speed up browsing when the database is used. The mime type is set during parsing and now stored in the database so there is not need to call the `media.postParse` when the database is enabled.


SubJunk:
This looks great, feel free to merge

SubJunk:
Even better ;)

valib:
@SubJunk @Nadahar it seems to me that the Travis OSx build is the most annoing feature we implemented. After several minutes this feature is not responding and the PR is not able to be merged.

Nadahar:
@valib As I've said before, I don't see why the big hurry to merge. They have a bit low capacity on macOS builds, but it's usually finished in an hour or two. 

You can see the backlog here: https://www.traviscistatus.com/ under "Backlog macOS Builds for Open Source projects".

SubJunk:
It's important to know whether a PR breaks tests before merging it but maybe we do need to reduce the number of builds. Currently it is running the tests on 5 different environments at a time, but probably we can reduce that to 2 or 3

Nadahar:
This doesn't seem to be very well thought through. I wasn't going to say anything, but I took at look at ```postParse()``` now and saw that this is also where the scan for subtitles is done. Removing this when fetching from the cache will mean that subtitles won't be refreshed once something is in the cache, so it won't even help to restart UMS after adding subtitles. To get subtitles to register, the database actually needs to be cleared.

I'm sure that will be a big hit. I also looked at the code in the method, and excluding the subtitle scan the code is very quick and I actually doubt this will save time at all. Have this actually been tested/timed, and what is the gain? I'm sure we're on the nanosecond level, but I'm not sure what is faster.

The reason I say anything despite that I wasn't going to is that this will make it even harder to finish #1372. I'm already pretty fed up with trying to resolve insanely messy code, so if you insist on breaking subtitles even more I'm not sure if I can get myself to finish it.

SubJunk:
Since this affects existing work I'm closing it, still thanks though, @valib !

Nadahar:
@valib @SubJunk Don't get me wrong, I don't think that the basic idea of this PR is a bad one. It just seems rushed, not properly thought through and tested to me when I can see several issues only be looking quickly at the code.

The MIME-type handling in UMS is basically a mess as I see it. There should be a "standard" MIME-type stored with the media that is then overridden for the renderer as needed. This "standard" MIME-type could and should be stored in the database. 

What made me make the comment above is simply that I'm so tired of new bugs being added quicker than they are solved. I don't understand the rush when adding stuff, why not try to think through the potential consequences of something and test it under different circumstances instead of just making "quick/dirty fixes" that turns out to cause more problems than they solve.

I think this PR could and should be improved to address the issues and make sure nothing is broken along the way.

src/main/java/net/pms/dlna/RealFile.java
```diff
@@ -180,7 +180,6 @@ public synchronized void resolve() {
 
 						if (medias.size() == 1) {
 							setMedia(medias.get(0));
-							getMedia().postParse(getType(), input);
```
SubJunk: @valib I've made this branch up to date with master now, but I'm not sure about this line since `postParse` does more than just set the MIME type. It can also do `secondaryFormatValid = false` and check for external subtitles. Are you sure we want to stop that happening here?
I don't know but I just wanted to ask to make sure it has been considered.

SubJunk:
@valib since the MIME type can change per renderer it might be good to store the renderer name in the database field too, e.g.
"video/mpeg;Panasonic ST60,video/mkv;Samsung 5500" then our code can see if we have cached the mime type for that file/device combination yet

SubJunk:
Closing this as it needs to take into account the renderer, and it's stale

PR #1429, opened 2017-12-26
Added Bengali translation (empty)

SubJunk:
We had a request to add this on Crowdin so I've done that


Nadahar:
I've added a commit that adds the missing step and modified the existing commit accordingly.

SubJunk:
Thanks for finishing this up @Nadahar

PR #1432, opened 2017-12-27
Update web interface to play OGA audio plus update the image formats supported by it.

valib:
Tested on Firefox and Chrome. This audio format is not supported by MSIE and EDGE.

### DIFF
--- a/src/main/java/net/pms/remote/RemoteUtil.java
+++ b/src/main/java/net/pms/remote/RemoteUtil.java
@@ -179,7 +179,9 @@ public class RemoteUtil {
 				mime.equals(HTTPResource.AUDIO_OGA_TYPEMIME) ||
 				mime.equals(HTTPResource.AUDIO_MP3_TYPEMIME) ||
 				mime.equals(HTTPResource.PNG_TYPEMIME) ||
-				mime.equals(HTTPResource.JPEG_TYPEMIME)
+				mime.equals(HTTPResource.JPEG_TYPEMIME) ||
+				mime.equals(HTTPResource.GIF_TYPEMIME) ||
+				mime.equals(HTTPResource.TIFF_TYPEMIME)
 			)
 		) {
 			return true;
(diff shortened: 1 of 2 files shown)


SubJunk:
@valib the formatting change looks good, reducing the duplication of MIME type definitions.
We already detect which browser is connected, maybe this change could be conditional upon the browser being Firefox or Chrome, to prevent passing a broken video to other browsers?

valib:
@SubJunk this could be done only when we introduce the browser's.conf and rewrite the `WebRendrer` where is set configuration for all browsers in the https://github.com/UniversalMediaServer/UniversalMediaServer/blob/11acac7c3bd521219f41d97a60ef50c0f0884722/src/main/java/net/pms/configuration/WebRender.java#L115
because the configuration setting is done during initialization and there is not possible to distinguish the current browser and set it properly.

src/main/java/net/pms/remote/RemoteUtil.java
```diff
@@ -179,12 +173,15 @@ public static boolean directmime(String mime) {
...
-				mime.equals(MIME_JPG)
+				mime.equals(HTTPResource.MP4_TYPEMIME) ||
+				mime.equals(HTTPResource.WEBM_TYPEMIME) ||
+				mime.equals(HTTPResource.OGG_TYPEMIME) ||
+				mime.equals(HTTPResource.AUDIO_OGA_TYPEMIME) ||
+				mime.equals(HTTPResource.AUDIO_MP3_TYPEMIME) ||
+				mime.equals(HTTPResource.PNG_TYPEMIME) ||
+				mime.equals(HTTPResource.JPEG_TYPEMIME) ||
+				mime.equals(HTTPResource.GIF_TYPEMIME) ||
+				mime.equals(HTTPResource.TIFF_TYPEMIME)
```
Sami32: Except for Safari users, all the others should get TIFF issue.

valib:
@SubJunk I tested it on Firefox and Chrome and it mostly works. The TIFF format is strange and there is not a web browser which supports all of combinations of this this format. I think that this the best way how to offer users to show the TIFF pictures.

Sami32:
https://github.com/openseadragon/openseadragon/issues/1380
https://en.wikipedia.org/wiki/Comparison_of_web_browsers#Image_format_support
https://bugzilla.mozilla.org/show_bug.cgi?id=160261

Nadahar:
TIFF are already handled by the automatic image conversion so I can view them without this PR. I don't think saving the small conversion time is worth losing support for many of them.

Issue #1414, opened 2017-11-08
Fix Java version check on startup


TQwan:
I recently upgraded to Java 9 (SDK and SE) and uninstalled java 8 se and java 8 sdk, because they contain vulnerabilities.

Anyway when I now start UMS 6.7.4 it says Java is not installed. I press Ok.
Then it tries to install Java. I close this window by clicking on the X (the esc also closes the window but nothing heapons)

Anyway UMS seems to start after closing the install screen (I guess it never checked if Java was really properly installed :D )

My TV is still able to stream from my PC with UMS 6.7.4 running Java SDK 9 (and SE 9).

Anyway please update the Java version check to support version 9, since this works

The problem is in this file
https://github.com/UniversalMediaServer/UniversalMediaServer/blob/86502e6a72005cf18055f8eb118f85d2d833a004/src/main/external-resources/nsis/pms.nsi

I think line 193 is the problem:
; Check if JRE7 is installed
${VersionCompare} ${JRE7_VERSION} $R1 $R2

This just checks if JRE 7 is installed, while it should check if 7, 8 and 9 are installed.
Note: Line 149 "MessageBox MB_ICONINFORMATION "${PRODUCT_NAME} uses Java Runtime Environment ${JRE7_VERSION}+"
This line specifically says JAVA 7 and higher are supported.

SubJunk:
I have reproduced this

SubJunk:
This is fixed in 6.8.0 and 7.0.0-rc2

PR #1433, opened 2017-12-29
Fixed detection of JRE 9 (#1414)

SubJunk:
Also includes minor formatting and comments to make this horrible, confusing file slightly easier to read

### DIFF
--- a/src/main/external-resources/nsis/pms.nsi
+++ b/src/main/external-resources/nsis/pms.nsi
@@ -25,7 +25,7 @@ VIProductVersion "${PROJECT_VERSION_SHORT}.0"
 !define REG_KEY_SOFTWARE "SOFTWARE\${PROJECT_NAME}"
 
 ; Definitions for Java
-!define JRE7_VERSION "7.0"
+!define MIN_JRE_VERSION "7.0"
 !define JRE_URL "http://downloads.sourceforge.net/project/unimediaserver/Dependencies/win32/jre-x86.exe"
 !define JRE64_URL "http://downloads.sourceforge.net/project/unimediaserver/Dependencies/win32/jre-x64.exe"
 
@@ -117,8 +117,22 @@ Function GetJRE
 		Call CheckJREVersion
 		IfErrors CheckRegistry3 JreFound
 
-	; 4) Check for registry
+	; 4) Check the registry for JRE (location used by JRE 9)
 	CheckRegistry3:
+		ClearErrors
+		${If} ${RunningX64}
+			SetRegView 64
+		${EndIf}
+		ReadRegStr $R1 HKLM "SOFTWARE\JavaSoft\JRE" "CurrentVersion"
+		ReadRegStr $R0 HKLM "SOFTWARE\JavaSoft\JRE\$R1" "JavaHome"
+		StrCpy $R0 "$R0\bin\${JAVAEXE}"
+		IfErrors CheckRegistry4
+		IfFileExists $R0 0 CheckRegistry4
+		Call CheckJREVersion
+		IfErrors CheckRegistry4 JreFound
+
+	; 5) Check for registry
+	CheckRegistry4:
 		ClearErrors
 		${If} ${RunningX64}
 			SetRegView 32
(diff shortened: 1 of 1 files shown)


SubJunk:
@UniversalMediaServer/developers I have tested this with JRE 8 and JRE 9 on Windows 7, and it works with both.
With JRE 9 the heap size check seems to break (so it sets 768 instead of 1280) but that is a different issue.

src/main/external-resources/nsis/pms.nsi
```diff
@@ -117,16 +117,14 @@ Function GetJRE
 		Call CheckJREVersion
```
Nadahar: @SubJunk I'm really confused by the whole ```GetJRE``` function. First of all, the syntax isn't entirely clear to me and pushing and popping (I assume) to/from the stack is confusing me. Since you have C experience I guess some of this makes more sense to you.

Assuming that the above haven't thrown me completely off here, it seems to me that there are some problems here: 
* I can't see that steps 1, 2 or 4 is performed. Step 1. would be futile, but shouldn't it then be removed from the description? The other steps would be useful in case Java isn't installed with Oracle's installer.
* Some registry locations seems to be checked multiple times, while the use of ```SetRegView``` 32/64 and ```Wow6432Node``` really confuse me. As far as I understand the ```Wow6432Node``` is the "32 bit view" on a 64 bit OS, and shouldn't need to be access directly. Setting the "reg view" to 32 should give the same result without explicitly specifying ```Wow6432Node```. This seems to be a combination of both strategies, with redundant checks. Your Java 9 check will, as far as I can tell, fail to detect a 32 bit Java 9 installed on a 64 bit OS. What I think needs to be checked is:
  * On a 64 bit OS both ```HKLM "SOFTWARE\JavaSoft\Java Runtime Environment"``` and ```HKLM "SOFTWARE\JavaSoft\JRE"``` in both 32 and 64 "registry view" - a total of 4 checks.
  * On a 32 bit OS both ```HKLM "SOFTWARE\JavaSoft\Java Runtime Environment"``` and ```HKLM "SOFTWARE\JavaSoft\JRE"```. Since "registry view" isn't a concept for 32 bit OS'es, that means a total of 2 checks.

Nadahar:
@SubJunk I've been trying to understand why Java 9 would "break" the heap size, but I can't. The heap size is set by examining the amount of RAM and isn't related to the JVM. The command line parameter seems to be the same for Java 9 (I haven't found any information that it has changed, but haven't found explicit documentation for Java 9 either). 

There's not more to it than that, the information is stored in the registry during installation and then used by the launcher. My only guess as to why this would fail is that the calls to 32/64 registry views mess up so that the value is read and written to different parts of the registry. If the launcher doesn't find the information, it will default to 768.

SubJunk:
@Nadahar yes I totally agree, the logic seems messed up and I spent a whole afternoon trying to navigate my way through it.
Some brave person may want to refactor it to remove unused code and give variables good names and add more comments.

I think it would be good to get this change merged as it is if you agree, since right now we have an urgent problem that people on Windows with JRE9 can't use UMS unless they are clever, so I suggest we merge this and I do a release of it

Nadahar:
@SubJunk Not cleaning it up "now" is one thing, but as far as I can understand that code won't be able to detect a 32 bit Java installation on a 64 bit Windows. I think that should be adressed at the very least. 

From what I can understand from the Java download page, it will offer the 32 bit Java version to anyone using a 32 bit browser and the 64 bit to 64 bit browsers. Until lately, 32-bit has been the standard version for most browsers and you'd have to explicitly pick the 64 bit version, but it seems that time time has finally come for the browsers to switch to 64 bit. Because of this, I bet we will see a lot of 64 bit OS'es with either Java version for a long time to come. Oracle has been following the browser bitness because of the Java plugin, but it is becoming less and less relevant and is rumored to be dropped in Java 10. In any case, we can't expect the Java bitness to be consistent on 64 bit Windows.

Without understanding everything of this, it seems to me that it would just take a copy-haste of yet another ```CheckRegistryX``` and check for ```SOFTWARE\JavaSoft\JRE``` also with a 32 bit "view".

SubJunk:
Good thinking, I will test that and update the branch if it doesn't work.
I'm on a 64-bit OS and using default Chrome and it only offers me 64-bit Java versions but I'll find a 32-bit link

SubJunk:
@Nadahar it seems that Java 9 is only officially released in 64-bit, according to their docs https://docs.oracle.com/javase/9/migrate/toc.htm#JSMIG-GUID-59701F80-5BB1-416D-835D-C39A9112FC1E

SubJunk:
Also I can't reproduce the Heapmem issue anymore so it may have been an anomaly.

So AFAICT there are no problems with this PR

Nadahar:
@SubJunk The missing 32-bit Java 9 is some bullshit Oracle is pulling. It seems they won't explain why, and it's not because it doesn't exists as they had it up and removed it: https://stackoverflow.com/questions/46356345/can-java-9-run-on-a-32-bit-os

They probably was paid off by Microsoft, which is still trying very hard to kill XP, or they found it a threat to their non-free "embedded" Java. In any case, it says it is possible to build the 32 bit version, so somebody probably will do that on a regular basis if Oracle will keep up their bullshit.

As a rush-fix I guess you're right: most users will have 64 bit Java 9 or will have to stick to Java 8.

This whole logic fundamentally wrong though, as it only checks for Oracle's JREs, and only the one that is marked as ```CurrentVersion``` in the registry - which is the last one installed as far as I know. There are many situations under which a suitable JVM can be available but this logic will try to download and install its own version.

SubJunk:
@Nadahar yeah it does seem strange that they are discontinuing 32-bit builds. This project is years away from having version 9 be the minimum anyway so it's not something that changes anything we do.

Yeah there are problems with the installer for sure. We get lucky that Windows users tend to like to do things the normal/default way :) I think this file has barely changed since PMS days, even since before I used PMS.

Anyway what needs to happen for this PR to be merged?

Nadahar:
@SubJunk You can simply merge it if you want to. I would have to test it on different OS'es to be able to approve it, but if you're confident that it won't break anything you can merge without approval.

Nadahar:
Tested on WinXP 32 without any problems.

SubJunk:
@Nadahar thanks for the review and testing!

PR #1449, opened 2018-02-09
Fix a couple of NPE's in DLNAResource.getDlnaOrgPnFlags()

### LINKED ISSUE #1447: NullPointerException when trying to navigate zip file on Android
Hi.  I am using BubbleUPnP on Android 7.0 on a Huawei MediaPad M3.  However, I tried other DLNA clients and they show the same error.  The text looks like it got an HTTP 500 Internal Server Error, but it got cut off.  (See attached image.)  The debug logs show a NullPointerException.

For some reason, I've been able to reproduce this with a zipfile full of jpg images, but if they are png images, it works fine.  Also, this works fine from my PS3.  I've tried on both 6.8.0 and 7.0.0 rc2.  I'm attaching debug packs from both versions, and a small zip file of jpgs that I used to reproduce the error.

Thanks.

[ums_dbg_6_8_0.zip](https://github.com/UniversalMediaServer/UniversalMediaServer/files/1705558/ums_dbg_6_8_0.zip)
[ums_dbg_7_0_0_rc2.zip](https://github.com/UniversalMediaServer/UniversalMediaServer/files/1705559/ums_dbg_7_0_0_rc2.zip)
upnp-error
[rends4.zip](https://github.com/UniversalMediaServer/UniversalMediaServer/files/1705560/rends4.zip)

Nadahar:
This should partially fix #1447. It only fixes the NPE, not the underlying problem with parsing media in archives.

### DIFF
--- a/src/main/java/net/pms/dlna/DLNAResource.java
+++ b/src/main/java/net/pms/dlna/DLNAResource.java
@@ -2354,10 +2354,10 @@ public abstract class DLNAResource extends HTTPResource implements Cloneable, Ru
 							}
 						}
 					}
-				} else if (mime.equals("video/vnd.dlna.mpeg-tts")) {
+				} else if (media != null && mime.equals("video/vnd.dlna.mpeg-tts")) {
 					// patters - on Sony BDP m2ts clips aren't listed without this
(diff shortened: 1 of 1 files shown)


Nadahar:
@valib TY 👍

PR #1450, opened 2018-02-10
Try to implement subtitles to TSMuxeR

valib:
@UniversalMediaServer/developers can anybody test it? TSMuxeR supports SUBRIP an PGS subtitles to be muxed to the output stream. I implemented the SUBRIP part but I am not able to test it because my TV supports only the MPEG-PS format.

### DIFF
--- a/src/main/java/net/pms/encoders/FFMpegVideo.java
+++ b/src/main/java/net/pms/encoders/FFMpegVideo.java
@@ -897,7 +897,7 @@ public class FFMpegVideo extends Player {
 					!params.sid.getType().equals(SubtitleType.PGS)
 			) {
 				deferToTsmuxer = false;
-				LOGGER.trace(prependTraceReason + "we need to burn subtitles because only SUBRIP and PGS are supported.");
+				LOGGER.trace(prependTraceReason + "there are external subtitles but only SUBRIP and PGS types are supported.");
 			}
 			if (deferToTsmuxer == true && avisynth()) {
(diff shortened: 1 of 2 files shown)


Vlck:
Hello, I want to test this implementation. Just tell me what do You need.

Vlck:
Unfortunately my renderer don't support external subtitles :(
I tried and one by one with theese settings:

OfferSubtitlesByProtocolInfo = true
OfferSubtitlesAsSource = true
SubtitleHttpHeader = CaptionInfo.sec

I am not sure, but i think that tsmuxer can handle PGS (BD) internal subtitles on some movies over DLNA.  I will make more tests, and share.

valib:
@Vlck sorry I made a mistake. It is not needed to have external subs supported because this code is muxing external subs to the MPEG-TS stream. Should be but is not. I found the way how to test it with Kodi and it is not working. Seems that TSMuxeR is not able to mux video and audio streams together with the subs in the file. It is producing the MPEG-TS stream and there is also muxed the subs stream but empty.

Vlck:
Ahoj valib, 
my poor knowledge tell me that tsmuxer make two separated streams - audio and video, but for the subtitles, I am not familiar. I want to help, but for now am still novice... I was read a lot, and I will be ready soon to give feedback! 
I respect your work and I am touched!
Best Regards!

valib:
It seems not working when tested with Kodi. This is producing video without the subtitles even if they are muxed into stream.

PR #1457, opened 2018-03-05
Update Maven site and other Maven plugins to the latest versions.

valib:


SubJunk:
Any reason for it other than be on the latest version?

valib:
@SubJunk no other reason than have the latest versions

SubJunk:
Cool. I'll test it on macOS before merging

SubJunk:
Unless you have already?

Nadahar:
Surefire ```2.20.1``` is the "bad" version that we've already had to reverse once. In addition, the assembly and compiler plugin are "scary" ones that needs thorough testing. The huge jump in version number for the enforcer plugin also makes me a bit worried.

I think the changelog of these 3 plugins need to be studied at the very least. Making sure that it builds on one specific computer doesn't really tell that much.

SubJunk:
@UniversalMediaServer/developers surefire 2.22.0 is released now, does it fix the issue that was preventing update?

SubJunk:
@valib I have synchronized this with master and I'm testing now

PR #1459, opened 2018-03-08
Fix the max bitrate when video is converted to the MPEG-2

valib:
This PR should solve the problem introduced in the http://www.universalmediaserver.com/forum/viewtopic.php?f=9&t=12285

### DIFF
--- a/src/main/java/net/pms/encoders/FFMpegVideo.java
+++ b/src/main/java/net/pms/encoders/FFMpegVideo.java
@@ -477,7 +477,7 @@ public class FFMpegVideo extends Player {
 			if (params.mediaRenderer.isHalveBitrate()) {
 				defaultMaxBitrates[0] /= 2;
 				LOGGER.trace("Halving the video bitrate limit to {} kb/s", defaultMaxBitrates[0]);
-			}			
+			}
 
 			int bufSize = 1835;
 			boolean bitrateLevel41Limited = false;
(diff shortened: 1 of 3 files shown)


src/main/java/net/pms/encoders/FFMpegVideo.java
```diff
@@ -468,10 +468,16 @@ public FFMpegVideo(PmsConfiguration configuration) {
...
+			// when video is converted to MPEG-2 to avoid the FFMpeg to warn the "packet too large, ignoring buffer limits to mux it"
+			if (dlna.getDefaultRenderer().isTranscodeToMPEG2() && defaultMaxBitrates[0] > 10080) {
+				defaultMaxBitrates[0] = 10080;
+			}			
+
 			if (params.mediaRenderer.isHalveBitrate()) {
 				defaultMaxBitrates[0] /= 2;
 				LOGGER.trace("Halving the video bitrate limit to {} kb/s", defaultMaxBitrates[0]);
-			}
+			}
```
Nadahar: @valib You should remove the whitespace added.

Nadahar:
@valib I've never understood the idea behind the "halfbitrate" setting, so I can't really determine the consequence of modifying this logic. Limiting the bitrate to the spec maximum seems reasonable though. 

Have you made a test-build for the user so you can confirm if it really solves the problem?

SubJunk:
@valib it would be good to check if this fixes the problem for the user. 1080p content quality will really suffer if it can't go above 10Mbit/s with MPEG-2. We have seen that message in the MEncoder and FFmpeg logs (it comes from the libav libraries that both those projects use) for years, since the PMS days, and AFAICT it doesn't cause any actual problems.

valib:
@SubJunk you are right. The quality is horrible.

PR #1469, opened 2018-04-03
Faster hashing v2

SubJunk:


SubJunk:
I propose to merge this branch to `master`, which is equivalent to using method 1 from the BenchMark class. Then when #1372 is merged it should make the minor switch to using method 2, since that branch already includes all of the refactoring to use `Path` instead of `File`. I attempted to merge that stuff into this branch but it's hard to do because of the extent of the refactoring, which is fine. Method 1 is still a good improvement.

@Nadahar @valib if you agree please approve this PR, and if you want to merge it make sure to use the merge with squash button to hide my sloppy commit history ;)

Nadahar:
I was planning to cherrypick it into #1372. Isn't that a better route, so that we can test with this as well?

SubJunk:
Yes that would be good

PR #1471, opened 2018-04-05
Delete Roku-TV8.conf during UMS unistalling

valib:
We forgot to add the new .conf to the setup.nsi to be deleted during the UMS uninstalling.


valib:
@Nadahar can you merge it?

PR #1474, opened 2018-04-07
Fix bug with JPEG_LRG profile check

Nadahar:
I've made a little typo in the ```JPEG_LRG``` evaluation so that support for ```PNG_LRG``` is checked. That means that renderers that doesn't support ```PNG_LRG``` won't be offered ```JPEG_LRG```.

This originates from [this thread](http://www.universalmediaserver.com/forum/viewtopic.php?f=9&t=11617&start=20). Testing has been done by both my and the affected user, and the error is quite obvious. Thus, I don't think any further testing is needed.


Nadahar:
@valib Thanks

PR #1481, opened 2018-04-09
Test prettifier

Nadahar:
This is 6.8.0 (d4e30cd0a93fe457a8ecaef6069c87522657c26b) pluss the test added in #1372 that "works in master". It's very strange to me that if also seems to fail on this "master code". 

@SubJunk Do you have an explanation how this can also fail here, since it, in your view as a *user*, is a bug in #1372?

### DIFF
--- a/src/test/java/net/pms/util/FileUtilTest.java
+++ b/src/test/java/net/pms/util/FileUtilTest.java
@@ -122,6 +122,7 @@ public class FileUtilTest {
 
 		// Video of a movie
 		assertThat(FileUtil.getFileNamePrettified("Universal.Media.Server.2015.720p.mkv", null)).isEqualTo("Universal Media Server (2015)");
+		assertThat(FileUtil.getFileNamePrettified("Universal_Media_Server_(2015)_[720p,BluRay,flac,x264]_-_FANSUBBERS.mkv", null)).isEqualTo("Universal Media Server (2015)");
 
 		// Video of a special edition of a movie
 		assertThat(FileUtil.getFileNamePrettified("Universal.Media.Server.Special.Edition.2015.720p.mkv", null)).isEqualTo("Universal Media Server (2015) (Special Edition)");


SubJunk:
This isn't `master`. See https://github.com/UniversalMediaServer/UniversalMediaServer/commit/d9d1ee008625d6ee53689a4e30a1631e5c9bad07#diff-ffb022ec9a3337993c5917459df75a74R125 and its corresponding passing build https://travis-ci.org/UniversalMediaServer/UniversalMediaServer/builds/363968449

Nadahar:
@SubJunk This is master the last time #1372 was synced with it. If this bug is fixed since then in master, it will be resolved when #1372 is updated to use the new master. The bug still has nothing to do with #1372. What exactly is your point? I only ever said that it's not related to #1372, which is still the case.

SubJunk:
It's bad enough that I already have to restore and backup different versions of the UMS profiles because that branch is so far behind master, now I have to go back and figure out which bugs may have been inadvertently fixed as master and 7.0.0 went along for the last 1 1/2 years? No I just won't do more testing until your branch is synchronized with master. It can't be merged before that happens and I would always want to do that anyway to guard against unanticipated merge conflicts so it doesn't affect anything.

Nadahar:
@SubJunk I see that you stick with your line of manipulating the truth. This PR was up to date with master until you merged #942 on January 23, and you know that very well. How long you've been working on #942 is hardly relevant.

Since the merge, master has serious issues (that I've repeatedly pointed out before the merge) as I see it. The fact that these still haven't been addressed as of April 9 isn't my responsibility or fault. You can argue that it's  "too long" all you want, it's not up to me.

The fact that you do as you do now, tries to blame somebody else for your own bugs, is exactly why I don't want to synchronize it until the problems in master is addressed. I know from experience that it's highly likely that you will try to blame something I did on a problem that is really a consequence of #942 as soon as the code is mixed. I'm done with that.

SubJunk:
You aren't getting blamed for the bug that was fixed potentially over a year ago, that's paranoia. You are getting blamed for taking months to get your branch up to date with master but whatever, if you don't want your branch merged with master I can just cherrypick the parts I want into there to move things along.

SubJunk:
The bug is in your branch because it has since been fixed in master. The bug wasn't caused by you but it exists in your branch. I don't know how to be more clear about that.

PR #1483, opened 2018-04-10
Rebased fix subs recognition PR  #1372

valib:
@UniversalMediaServer/developers the rebasing of  #1372 to master was horrible so please review this PR and test it.


SubJunk:
Will test over the next couple of days, I'm busy hacking away at the memory use bug

valib:
Rebased to updated master.

SubJunk:
The memory use bug seems to be fixed now (well it's not fixed but it's back to how it has been since v5) so I'll test this now

SubJunk:
Getting a lot of these errors on the startup scan:

```
org.h2.jdbc.JdbcSQLException: Column "EXTERNALFILE" not found [42122-196]
	at org.h2.message.DbException.getJdbcSQLException(DbException.java:345)
	at org.h2.message.DbException.get(DbException.java:179)
	at org.h2.message.DbException.get(DbException.java:155)
	at org.h2.jdbc.JdbcResultSet.getColumnIndex(JdbcResultSet.java:3154)
	at org.h2.jdbc.JdbcResultSet.get(JdbcResultSet.java:3246)
	at org.h2.jdbc.JdbcResultSet.getString(JdbcResultSet.java:314)
	at net.pms.dlna.DLNAMediaDatabase.getData(DLNAMediaDatabase.java:626)
	at net.pms.dlna.RealFile.resolve(RealFile.java:173)
	at net.pms.dlna.DLNAResource.syncResolve(DLNAResource.java:1608)
	at net.pms.dlna.DLNAResource.run(DLNAResource.java:1405)
	at net.pms.dlna.RealFile.isValid(RealFile.java:66)
	at net.pms.dlna.DLNAResource.addChild(DLNAResource.java:652)
	at net.pms.dlna.MapFile.manageFile(MapFile.java:273)
	at net.pms.dlna.MapFile.analyzeChildren(MapFile.java:329)
	at net.pms.dlna.MapFile.analyzeChildren(MapFile.java:309)
	at net.pms.dlna.RootFolder.scan(RootFolder.java:285)
	at net.pms.dlna.RootFolder.scan(RootFolder.java:244)
	at net.pms.dlna.DLNAMediaDatabase.run(DLNAMediaDatabase.java:1413)
	at java.lang.Thread.run(Unknown Source)
```

valib:
@SubJunk have you deleted the database before test or use clean install of the UMS?

SubJunk:
I didn't do anything to the database and I was running `7.0.2-SNAPSHOT` before installing

valib:
I will test it.

valib:
@SubJunk it seems like wrong rebasing. I reproduced this issue and I will try to fix it. Thanks.

valib:
@SubJunk fixed in the last commit

SubJunk:
Thanks, testing it now

SubJunk:
@valib the new folders in the Media Library (like TV Shows, Movies, etc.) are returning blank pages in this branch

valib:
@SubJunk you introduced in the 7.0.0 version  the `OpenSubtitle.backgroundLookupAndAdd` which also populates the Media Library but in the meantime @Nadahar refactored the OpenSubtitle and introduced `registerExternalSubtitles` which also seems like a very good improvement but this doesn't populate the Media Libraly. I tried to combine those different approaches and I fail. @SubJunk @Nadahar guys please get rid of the animosity and cooperate to implement this PR which is definitely worth to be implemented.

Nadahar:
It seems you guys don't quite get my position. This isn't about "animosity" on my part, but about wasting my time. I want to spend my time on coding, not intrigues and unneccesary bugs and problems.

This is the very reason I started DMS, to be able to actually fix the code instead of having to deal with endless arguments about whose code I change and why it should be changed. Pushing bugs and unfinished commits to master and rushing to get things released instead of properly fix all known issues falls into the same category of "waste of time" for me. It *always* ends up costing more time in total when you have to deal with the results of ignoring a problem in the first place, having to explain to users and go through log files to try to pin down the issues.

Since we have a very different philosophy, I think it's better than I work "my way" in DMS and you use whatever you want of that code. It's that simple.

When it comes to this PR being finished, it's not in my eyes. The reason is very simple, there are still *known issues*. That's all it takes for me, if there are known issues, it's not finished.

The issues I know about here is only about "live subtitles". There are multiple problems with the "quality" of the results, the similarity algorihm used and how we use it, the fact that duplicate subtitles are listed and more. The reason there are duplicate subtitles is that multiple queries are done to OpenSubtitles (depending on the results), and some of the subtitles returned by the different queries often point to the same subtitle file, although the exact information returned isn't exactly the same.

Apart from the OpenSubtitles part, I don't know of any other issues, but I can't think of a reason why anyone would want to release something with known issues and deal with user frustration and reports as a consequence.

When it comes to merging of OpenSubtitle.java I don't quite see the problem. Even though ```backgroundLookupAndAdd()``` is in the same class, the code isn't really related AFAICT. The background lookup tries to find the IMDB ID and some other information about the video itself, while the rest of the class deals with subtitles.

src/main/java/net/pms/dlna/DLNAResource.java
```diff
@@ -1649,12 +1649,18 @@ protected String getDisplayNameEngine(PmsConfiguration configuration) {
...
 	 */
-	@SuppressWarnings("unused")
 	protected String getDisplayNameBase(PmsConfiguration configuration) {
+		String nameToShow;
+		if (isNotBlank(displayNameOverride)) {
+			nameToShow = displayNameOverride;
+		} else {
+			nameToShow = getName();
+		}
+
```
Nadahar: @valib What does this fix?

valib:
@Nadahar thanks. This is a nice example how I stuck somewhere between Visual Basic and object oriented programing. My solution is working but is not so elegant as yours. This is what you should expect from the 60 years old man. 😃

SubJunk:
I'm working on this now

SubJunk:
Haven't tested it much but I pushed some work on it. I'll test more tomorrow

Nadahar:
@valib The bug is in master too, it's the hammering of OpenSubtitles I've tried to raise attention to multiple times before. It's in 6.x as well, and it probably has been there ever since ```InfoDb``` was conceived.

Because of abuse by applications like UMS, OpenSubtitles have had to [crack down on this](https://forum.opensubtitles.org/viewtopic.php?f=8&t=16072). I must say that making a lookup for every file that's parsed strikes me as very close to insane.

The reason you haven't seen this before is that the old code in ```OpenSubtitle``` simply hid exceptions with the standard ```catch (Exception e) {}``` used by this author.

The only thing that is new now is that the exceptions are logged. The cause isn't a "bug", but a flawed design.

SubJunk:
Yes as we have said many times before it would be good to have a feature to cue the OpenSubs lookups so we can limit the rate and also retry failed ones periodically. It is a waste of bandwidth to keep performing lookups that will fail so it would be good to pause them for a time when they start to fail. 

At least since we don't rely on the OpenSubs lookups for anything really important, this doesn't cause real problems. Probably the most noticeable thing is that TV episodes might not have their episode titles appended if they aren't in the filename. But it still isn't very kind to OpenSubtitles for us to do this to them so I fully support a cue.

SubJunk:
@valib the database seems to be populated correctly now and I have checked that marking folders as played/unplayed works. Is there anything else this is waiting for me to do?

SubJunk:
Possible regression in the GUI, I thought the text here was shortened so it fits?

image

valib:
@SubJunk I think that now it is working and only the OpenSubs lookups should be solved.

SubJunk:
@valib I noticed another bug - if I go into the #--TRANSCODE--# folders there is now only one entry in there

What do you mean the OpenSubs lookups should be solved? If you mean making it smarter with retries and further rate limiting, I think that can happen in another branch since it isn't related to this one - it is also a problem in `master`

SubJunk:
I just got home from work and saw this and fixed it. Somehow Apache encountered an error that made it stop, I've never seen that before. Thanks for the heads up

SubJunk:
I will try to work on this this weekend, I don't want it to fall too far behind

valib:
@SubJunk I retested this PR and it seems that I made somewhere a mistake because folders in the TRANSCODE folder are wrongly populated. There is only one item in folders when there should be multiple items for different audio and subtitles.

SubJunk:
@valib cool, I left a comment about that after I last tested it too https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1483#issuecomment-386471770 I guess you missed it

SubJunk:
I have fixed that bug now

SubJunk:
And now it is up to date with `master`

SubJunk:
@valib @Nadahar do you guys think we need a full database re-initialization after this update? I realised after I did the updating code that maybe this is one time we just need to redo it because the new `NOT NULL` fields need to be populated

Nadahar:
@SubJunk It's too long since I did this for me to remember why I used ```NOT NULL``` for that field. When I look at the code now it seems like a mistake, although I cannot be sure I'm not missing something now.

Generally, ```NOT NULL``` fields should also have a ```DEFAULT``` value. From what I can see, this field is supposed to be set to ```""``` for internal subtitles, which means that is used as the "null value". If the table definition included the default, this would work. Altering a table by adding a new column that is required to be non-null will probably lead to an error, since the existing rows will be expanded and the lack of a default means they will be null...

The reason I don't think simply adding ```DEFAULT ""``` is the "solution" is that I can't see why ```""``` is better than null. I hate to disagree with myself, which makes it obvious that I either did a mistake then or I am doing it now :wink: In either case, there's a mistake on my part :innocent: 

The only thing I do remember is that it was supposed to be guaranteed that *if* the subtitles are external, ```EXTERNALFILE``` must be populated. The logic built around this assumes that this is true and bugs will result in the rule is broken. Therefore, the update code actually needs to delete all existing subtitles entries that is external. But, since it seems that there is no field for this (I think ```EXTERNALFILE``` is also used to determine if the subtitles is indeed external), I'd say that all subtitles entries would have to be deleted. There's another reason for this too, as this PR redefines how the ```ID``` field is used. I guess we could use the old definition for ```ID``` (where ```ID``` >= 100 used to mean "external") to only delete the external tracks, but this should be tested to make sure it behaves like expected.

To be clear, I mean something like this before the ```ALTER``` statement:

```sql
DELETE FROM TABLE SUBTRACS WHERE ID >= 100
```

This code is only written here on GitHub and is completely untested, but I think the final result should be something like this (given that the ```NOT NULL``` requirement is dropped and that works fine):

```java
try (Statement statement = conn.createStatement()) {
	statement.execute("DELETE FROM SUBTRACKS WHERE ID >= 100")
	StringBuilder sb = new StringBuilder();
	sb.append("ALTER TABLE SUBTRACKS ADD EXTERNALFILE VARCHAR2(").append(SIZE_EXTERNALFILE).append(")");
	statement.execute(sb.toString());
	sb.setLength(0);
	sb.append("ALTER TABLE SUBTRACKS ADD CHARSET VARCHAR2(").append(SIZE_MAX).append(")");
	statement.execute(sb.toString());
	statement.execute("ALTER TABLE SUBTRACKS DROP constraint PKSUB");
	statement.execute("ALTER TABLE SUBTRACKS ADD constraint PKSUB primary key (FILEID, ID, EXTERNALFILE)");
}
version = 15;
```

Nadahar:
I think remember why it is ```NOT NULL``` :bulb: : I don't think you can have null in the primary key. That would mean that the solution should be to use ```""``` as the default, so something like this:

```java
try (Statement statement = conn.createStatement()) {
	statement.execute("DELETE FROM SUBTRACKS WHERE ID >= 100")
	StringBuilder sb = new StringBuilder();
	sb.append("ALTER TABLE SUBTRACKS ADD EXTERNALFILE VARCHAR2(").append(SIZE_EXTERNALFILE).append(")").append(" DEFAULT '' NOT NULL");
	statement.execute(sb.toString());
	sb.setLength(0);
	sb.append("ALTER TABLE SUBTRACKS ADD CHARSET VARCHAR2(").append(SIZE_MAX).append(")");
	statement.execute(sb.toString());
	statement.execute("ALTER TABLE SUBTRACKS DROP constraint PKSUB");
	statement.execute("ALTER TABLE SUBTRACKS ADD constraint PKSUB primary key (FILEID, ID, EXTERNALFILE)");
}
version = 15;
```

The default would also have to be set in the ```CREATE TABLE``` code of course.

valib:
@Nadahar non of you suggestion works. The H2 SQL grammar doesn't implement the "DEFAULT" or "NOT NULL" for "ALTER TABLE ADD" command. I tried to use your code but it doesn't work. Especially `statement.execute("DELETE FROM SUBTRACKS WHERE ID >= 100")` is nonsens because the ID is not implemented to distinguish the internal subs to have ID lower that 100 and for external subs higher that 100. Maybe I made a mistake during rebasing but the ID is bad. When I checked the H2 database by the "H2 console" the ID is not unique.

SubJunk:
Ok cool, thanks for those notes. I'll try to make this work

Nadahar:
@valib 
> The H2 SQL grammar doesn't implement the "DEFAULT" or "NOT NULL" for "ALTER TABLE ADD" command. I tried to use your code but it doesn't work. 

As I wrote, this was only pseudocode written on GitHub, so you can't expect the syntax to be 100%. That said, according to the H2 documentation, [ALTER TABLE ADD](http://www.h2database.com/html/grammar.html#alter_table_add) does reference [columnDefinition](http://www.h2database.com/html/grammar.html#column_definition) which does support this construct. It's probably just a syntax error, maybe I used the wrong type of quotes for the empty string for example.

> Especially statement.execute("DELETE FROM SUBTRACKS WHERE ID >= 100") is nonsens because the ID is not implemented to distinguish the internal subs to have ID lower that 100 and for external subs higher that 100.

This is for upgrading from previous versions of the database and won't work on a database that is already upgraded. The ID *was* used like that in previous versions, but isn't anymore. Those that previously had an ID >= 100 no longer has any ID and has ```EXTERNALFILE``` set instead. Thus, ID is *no longer* unique per subtitles track, but it *was* in previous versions.

valib:
@Nadahar 

> As I wrote, this was only pseudocode written on GitHub, so you can't expect the syntax to be 100%. That said, according to the H2 documentation, ALTER TABLE ADD does reference columnDefinition which does support this construct. It's probably just a syntax error, maybe I used the wrong type of quotes for the empty string for example.

I tried all that but even if the columnDefinition should support `NOT NULL` it doesn't  work for ALTER TABLE ADD.

src/main/java/net/pms/dlna/DLNAMediaDatabase.java
```diff
@@ -268,20 +268,17 @@ public synchronized void init(boolean force) {
...
-								sb.append(", LANG     VARCHAR2(").append(SIZE_LANG).append(')');
-								sb.append(", TITLE    VARCHAR2(").append(SIZE_MAX).append(')');
-								sb.append(", TYPE     INT");
-								sb.append(", EXTERNALFILE VARCHAR2(").append(SIZE_EXTERNALFILE).append(") NOT NULL");
-								sb.append(", CHARSET VARCHAR2(").append(SIZE_MAX).append(")");
-								sb.append(", CONSTRAINT PKSUB PRIMARY KEY (FILEID, ID, EXTERNALFILE))");
+								sb.append("ALTER TABLE SUBTRACKS ADD EXTERNALFILE VARCHAR2(").append(SIZE_EXTERNALFILE).append(") DEFAULT ''");
 								statement.execute(sb.toString());
+
+								statement.execute("ALTER TABLE SUBTRACKS ALTER COLUMN EXTERNALFILE SET NOT NULL");
```
Nadahar: :+1:

SubJunk:
Seems to be updating properly after my latest commit. I don't understand the deletion of subtracks with greater than ID 99 but I haven't really worked on this branch so I will trust those who have.

Before we merge I would like to do another code review. Once that is done are you guys happy for it to be merged? @valib @Nadahar

Nadahar:
@SubJunk Before this PR external subtitles tracks had an ID >= 100 while internal started from 0 or 1 (don't remember). 

After this PR, external subtitles doesn't have an ID (probably -1) but instead keeps the path to the external subtitles file in `EXTERNALFILE`. This information is necessary for the "new" logic to work, which means that any external subtitle tracks stored in the database are useless as they are missing crucial information. 

By deleting these during upgrade UMS will have to re-parse and re-insert the external subtitles tracks (which should happen while browsing). If we left them be, they would cause errors because they would be interpreted as internal subtitles with an ID that is nowhere to be found in the source media.

This is because the "external is >= 100" rule has been removed since it caused bugs. Some internal subtitles actually have ID's above 100 in some container type. Also, since having an ID for external subtitles isn't useful (since it's all made up anyway and can't be used for anything), the new indication that a track is external is defined as ```EXTERNALFILE``` being non-blank.

Nadahar:
@valib I saw that @SubJunk did it in two steps, and I don't really know what the problem is, but I found this supposedly valid H2 SQL:
```sql
ALTER TABLE PUBLIC.EMPLOYEE ADD NEW_COL CHAR(25) DEFAULT '11' NOT NULL
```
It seems to be exactly as I imagines it should be.

valib:
@Nadahar I retested it and I don't understand what I did before when it doesn't worked for me. Maybe the  `DEFAULT ''` before the `NOT NULL` is the clue. Without the `DEFAULT ''` it fails. Yes it is working but during testing I found other problem. If database is used and when from the table SUBTRACKS are deleted external subtitles tracks with ID >= 100 they are never added again. The same problem is when external subs are deleted or added. The UMS is not checking if something changed with subtitles and uses only what is stored in the database. Only in the TRANSCODE folder are all relevant subtitles listed because this folder is populated in different way.

Nadahar:
@valib If you try to constrict it to NOT NULL without setting a default, it will fail. That is because the database have to add a value for the new column to all existing rows. When no default is set, this value is NULL - which is forbidden. By adding a new column to an existing table while requiring that the column is non-null and has no default, you're asking the database to do something impossible. If you do this if the table is empty (or when you create the table), it should work because there are no existing rows that will get this dilemma. You will however be unable to add rows that doesn't have a value for the NOT NULL column.

External subtitles that is discovered should be added to the database. This is one of the things I explicitly fixed. This has either stopped working again as a result of a bug, or it is caused by the rebasing. You should check if this is working in the original PR first.

valib:
@Nadahar I tested the original PR to be sure where the problem is. This is here https://github.com/UniversalMediaServer/UniversalMediaServer/blob/37989b3f77983190b31c41c685f2fbba8d79845f/src/main/java/net/pms/dlna/RealFile.java#L177 and the line before.

SubJunk:
@Nadahar since it is in the original PR do you have any interest in fixing it or should I do that? I want to keep this moving along and hopefully merge soon

Nadahar:
If any of you find a way to fix it, please go ahead. I am planning to look at it, but not until I've finished some other stuff.

When I think about it I'm not sure how best to solve it (when these should be added to the database), or if it's useful to have external subtitles in the database at all. Internal subtitles follows the video file, so there's no reason to reparse these over and over again, but you never really know if new external subtitles have shown up or some have been removed since you last checked, so my question is: When it the cached information about external subtitles actually useful?

This PR makes sure that a scan for external subtitles is always done unless it has been cached in memory AFAICR, so I'm not sure the information from the database is actually used. These things have faded a bit in my memory now, so this all has to be verified - but I'm not sure the "fix" is to push updates to the database. If so, there should also be a way to remove cached external subtitles that's no longer there.

Nadahar:
@valib I haven't looked in-depth on what you're doing, but it looks to me like you're screwing up the logic.

valib:
@Nadahar the logic for subtitles stored in the database is wrong. When they are changed there is not the way how to rescan them to update the current situation. So I use your code the `registerExternalSubtitles(false);` which populates the subtitles cache and MediaInfo for that media.

Nadahar:
@valib I understand that part, it's which way it should be solved I'm talking about. As I mentioned I'm not even sure if external subtitles are useful to keep in the database, but I since realize that they might be needed by the media library.

To avoid rescanning and reparsing a lot, I'd think that the correct thing to do if they are useful in the database, is to add them when they are actually parsed. That means that whenever `registerExternalSubtitles()` finds new subtitles they should probably be inserted into the database right after parsing.

To remove subtitles that have been deleted from the database, some additional logic would be needed (not much though). It could be done in the same method by simply marking any existing external subtitles tracks that aren't found during refresh for deletion.

SubJunk:
@valib will this be worked on more or should we give up on it? to be honest it makes the list look ugly anyway since almost all popular release files have subtitles so I wouldn't be sad to see that part go

SubJunk:
I just reread what you said @valib and realise that I replied wrong. When I said ugly, I meant the text part, like `Filename (internal subtitles: english)`. I think the flags in the thumnails look really good!

valib:
@SubJunk I am on holiday abroad so to react is a little bit complicated. It could be merged because the main part of this PR is working. What is missing is the OpenSubtitles part Nadahar finished in the DMS. For me to merge those changes is a nightmare. If you are willing to do that please feel free.

Nadahar:
I won't normally interfere here anymore, but I just have to add:

@SubJunk Whether audio and subtitles information should be added and how verbose it should be is configurable, so complaining about that it is "ugly" just shows how little attention you've paid to this PR. I don't remember what the default it, but I'm pretty sure it's not the most verbose, and it can easily be changed to what you think is most "beautiful".

@valib I'm not sure which part of my XXE implementation you're thinking of, but my XXE fix for Cling doesn't work. That is, I don't think it's anything wrong with the implementation itself, but it seems that UPnP relies on DTD being enabled. I haven't dived into the details of the XML communication to know exactly what is needed, but simply turning it off breaks UPnP communication. I think this must be solved in Cling, and it seems like it is abandoned. So, I don't know at the moment how to handle XXE. I'd wish Cling itself would be flagged as vulnerable, because then something might happen. Numerous other software also use Cling and is equally vulnerable as a result, so if this was made public knowledge something might happen.

SubJunk:
@valib ok cool I will attempt to finish it while you're away, I hope you have a great holiday! I am going to merge most of the stuff from DMS I think, but it will take a little while because we have peer reviews and testing. This one is a high priority for me.

@Nadahar actually it was made customisable after I tested and suggested it, and fixed bugs in it along the way too, but sure keep painting me as the bad guy in all situations for some reason I will never understand

SubJunk:
@valib ok cool I will attempt to finish it while you're away, I hope you have a great holiday! I am going to merge most of the stuff from DMS into UMS I think, there is some good stuff there. This one is a high priority for me.

Nadahar:
>@Nadahar actually it was made customisable after I tested and suggested it, and fixed bugs in it along the way too, but sure keep painting me as the bad guy in all situations for some reason I will never understand

I've never tried to paint you as the "bad guy" in any situations. As I've said before, I strongly dislike dishonesty, and you keep trying to "twist" things to your advantage. For me, that is dishonesty, and I respond accordingly.

I never said that you didn't cause it to be configurable, I have no memory of how it came to be configurable. What I said was:
> complaining about that it is "ugly" just shows how little attention you've paid to this PR

You're in your full right to think this PR shouldn't be merged, either because you think it reduces the quality or the functionality, because you don't like me and my involvement in it, or because you had a vision telling you it will lead to the end of the world. That's fine for me, but don't come up with stupid excuses that simply doesn't hold water.

You should know that @valib also has worked on this, so you can't demean this to get to me without also being contemptuous to @valib. I understand if you think that sucks, but I've done plenty of other things that you should be able to do without stepping on somebody else.

SubJunk:
@Nadahar I'm not your enemy, I have no interest in being that. I give you compliments all the time and I just want us to be peaceful and respectful with each other. I removed my comment about the text being ugly and I removed the text that you quoted in an attempt to be more peaceful.

SubJunk:
Also I have worked on this branch too and its predecessor, and will continue to work on it to try to get it merged, so hopefully you can see that I'm not an outsider to it.

Nadahar:
I'm all for that too and I try not to interfere, but in this case I felt that @valib was being stepped on so I chose to say something.

SubJunk:
What a hero

Nadahar:
> What a hero

Believe what you want, it is the truth. You keep revealing your true colors though.

SubJunk:
Valib and I have been working together since before your time here and will continue after you are gone, and he can stick up for himself. I really doubt he sees the fact that I commented negatively about a small piece of UI in a branch as "stepping on", but if he does I hope he will let me know.

In the meantime I will continue to contribute to getting this branch finished and merged so our users can benefit from the many good changes in here.

SubJunk:
I had a look through the DMS changelog and didn't see anything about fixing OpenSubtitles so I'm not sure what to do, maybe I missed it in the commit messages. I'll have another look tomorrow

SubJunk:
@valib I just merged master into here, I got a few conflicts which you can see in the merge commit, it would be good if you can look at those 3 files to make sure I resolved the conflicts correctly if you have time

SubJunk:
There seems to be some bug in this branch that is causing a lot of memory use and eventually crashing the program, related to OpenSubtitles. I'm not getting the crash on `master`. The log fills with:

```
TRACE 2018-09-28 18:40:57.246 [OpenSubtitles background worker 9] Error in OpenSubtitles parsing:
java.io.IOException: Insufficient system resources exist to complete the requested service
	at sun.nio.ch.FileChannelImpl.map0(Native Method)
	at sun.nio.ch.FileChannelImpl.map(Unknown Source)
	at net.pms.util.OpenSubtitle.computeHash(OpenSubtitle.java:175)
	at net.pms.util.OpenSubtitle.getHash(OpenSubtitle.java:151)
	at net.pms.util.OpenSubtitle.getInfo(OpenSubtitle.java:1788)
	at net.pms.util.OpenSubtitle.getInfo(OpenSubtitle.java:1783)
	at net.pms.util.OpenSubtitle$2.run(OpenSubtitle.java:4815)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
	at java.lang.Thread.run(Unknown Source)
```

Sami32:
@Valib Nice work, there are many good things here. As i guess that you and @subjunk have already tested it because you are using subtitles a lot it will be good if all your work get rewarded IMO.
It should work better than the docker anyway ;)

Sami32:
@valib Nice work, there are many good things here. As i guess that you and @SubJunk have already tested it because you are using subtitles a lot it will be good if all your work get rewarded IMO.

Sami32:
Hum, it seem some minor conflicts appeared since the last time.

OP-Klaus:
@Sami32 thanks for looking over this. However I'm not happy about the memory use bug which I have reproduced in this branch but not in `master`. Also I noticed some very slow responses to browse requests but I will benchmark to see if that is related to this branch.

I agree there is a lot of good work done in this branch by many developers so I am interested to get this into a mergeable state!

SubJunk:
@Sami32 thanks for looking over this. However I'm not happy about the memory use bug which I have reproduced in this branch but not in `master`. Also I noticed some very slow responses to browse requests but I will benchmark to see if that is related to this branch.

I agree there is a lot of good work done in this branch by many developers so I am interested to get this into a mergeable state!

OP-Klaus:
@Sami32 the memory usage bug prevents this from being merged as it leads to the whole program crashing

SubJunk:
@Sami32 the memory usage bug prevents this from being merged as it leads to the whole program crashing

OP-Klaus:
Based on the logs after the crash I am guessing there is a resource leak introduced in this branch related to OpenSubtitles, but that is just a guess and I will look into it more.

SubJunk:
Hopefully I synchronised this correctly with `master` now

src/main/java/net/pms/dlna/RootFolder.java
```diff
@@ -94,29 +95,24 @@ public boolean isValid() {
...
-	public void discoverChildren(boolean isAddGlobally) {
 		if (isDiscovered()) {
 			return;
 		}
 
 		if (configuration.isShowMediaLibraryFolder()) {
 			DLNAResource libraryRes = PMS.get().getLibrary();
 			if (libraryRes != null) {
-				addChild(libraryRes, true, isAddGlobally);
+				addChild(libraryRes);
```
SubJunk: @valib why remove the `isAddGlobally`? It is what prevents the startup scan from taking up memory that is never released

valib: @SubJunk because your merging stopped UMS working not only on my TV but also on the web interface. The virtual folders were missing.

SubJunk: Ah ok, I will need to re-add those and resolve it a different way, because this is an important thing that stops the program crashing on startup. I'm working on this now :)

SubJunk: Which virtual folders were missing? Are they working now after my latest commit?

src/main/java/net/pms/dlna/RootFolder.java
```diff
@@ -112,7 +117,7 @@ public void discoverChildren() {
 				PMS.getConfiguration().getDataFile("UMS.last"),
 				PMS.getConfiguration().getInt("last_play_limit", 250),
 				Playlist.PERMANENT|Playlist.AUTOSAVE);
-			addChild(recentlyPlayed);
+			addChild(last, true, isAddGlobally);
```
valib: @SubJunk why is not here `recentlyPlayed`

src/main/java/net/pms/dlna/RootFolder.java
```diff
@@ -95,7 +95,12 @@ public boolean isValid() {
 		return true;
 	}
 
+ 	@Override
 	public void discoverChildren() {
+		discoverChildren(true);
```
valib: @SubJunk this was not in your merging and that is why all virtual folders were missing.

src/main/java/net/pms/dlna/RootFolder.java
SubJunk: I don't remember getting a conflict at that line *shrugs*

SubJunk:
@valib I can't reproduce the memory leak bug on the latest version so maybe it was a problem with a merge that was fixed later. I had reproduced it 3 times in a row on the older version but now I can't.

I have one question about a change in behavior that causes more transcoding than `master` - it seems that if a video has a subtitles format that my device does not support, that video is transcoded, even if those subtitles do not match my language preferences.

An example in my library is a video that has the language set to English for the audio and subtitles, but the subtitles are PGS which my TV does not support. On `master` the file is sent without transcoding, but in this branch it is transcoded. Which one is the intended behavior? Is it a bug or a feature?

Sorry if it has been discussed before, I can't remember all of the discussion on this feature :)

If this is intended behavior I will approve and merge this PR. If it is a bug I will try to fix it.

SubJunk:
> this is a question. If the video have external subtitles they have a priority against the internal one. If external subtitles language is not recognised than video is transcoded with this subtitles assuming that user add it intentionally and match his requirements. The same logic applies for internal subtitles. If they are not supported by renderer they are transcoded. So is up to users to add external subtitles to override not wanted internal subtitles or convert the video without subtitles. Do you have any other idea how to solve it?

@valib yeah it is hard to know what is the most intuitive way to handle all of these cases

In the case I noticed in my library, the subtitles were internal and the languages were correctly flagged, so it was quite a good quality file. I think in that particular case, where the subtitles do not match our language preference, we should not transcode it. I think if a device doesn't support a certain stream within a container it supports, it will just ignore it, but maybe that isn't always the case.

Maybe we should have a renderer config option like `AllowUnsupportedInternalSubtitles` to indicate that the renderer ignores the unsupported subtitles instead of not playing the file.

This level of complexity makes me really want automated testing for this too, since that is a great opportunity to document the behaviour we want at the same time as ensuring it works. At my day job we do TDD and it is really effective at protecting against regressions especially when we hire a new person and they aren't aware of all the little parts of logic, it acts like a learning tool for them

> the main problem is that there are a lot of videos with internal subtitles whose do not have set the subtitles language. How do we resolve that they are not valid?

It would be cool to use automatic language detection on the files and store the results in the database for speed. I haven't looked much at our automatic language detection so I don't know what it already does, but that would be cool

> Maybe we can recommend to add dummy external subtitles with the preferred language like video_name.eng.srt to override internal subtitles.

That doesn't seem very user-friendly to me. Ideally it would be as automatic as possible

valib:
> It would be cool to use automatic language detection on the files and store the results in the database for speed. I haven't looked much at our automatic language detection so I don't know what it already does, but that would be cool

The language detection is not 100% successful and there are a lot of false recognition. To do that we  must extract internal subtitles from video and test it which takes some time and will slow browsing.  What is more important the language detection for picture based subtitles is impossible.

SubJunk:
@valib is it possible to know how confident the automatic detection is, so that we only use the automatic detection if it reports 99%+ confidence for example?

The detection could be part of either the current file scan or even a secondary scan, to prevent it from blocking the browsing, in a similar way to how we do OpenSubtitles lookups - we create a cue of requests for all videos with audio or subtitles with undefined languages, and then those get processed in the background. If a browse request happens before the lookup for a file has happened then we use the current logic.

It is the exact same type of progressive enhancement that we do with the filename prettifying where we initially use the data from the filename and then enhance it with OpenSubs data in the future.

SubJunk:
@valib anyway, maybe we should merge this and then deal with that question separately?

SubJunk:
Getting an error when I try to open the web interface on this branch:

```
20:42:53 INFO  Exception in thread "AWT-EventQueue-0" java.lang.NullPointerException
20:42:53 INFO  	at net.pms.io.BasicSystemUtils$3.actionPerformed(BasicSystemUtils.java:156)
20:42:53 INFO  	at java.awt.MenuItem.processActionEvent(Unknown Source)
20:42:53 INFO  	at java.awt.MenuItem.processEvent(Unknown Source)
20:42:53 INFO  	at java.awt.MenuComponent.dispatchEventImpl(Unknown Source)
20:42:53 INFO  	at java.awt.MenuComponent.dispatchEvent(Unknown Source)
20:42:53 INFO  	at java.awt.EventQueue.dispatchEventImpl(Unknown Source)
20:42:53 INFO  	at java.awt.EventQueue.access$500(Unknown Source)
20:42:53 INFO  	at java.awt.EventQueue$3.run(Unknown Source)
20:42:53 INFO  	at java.awt.EventQueue$3.run(Unknown Source)
20:42:53 INFO  	at java.security.AccessController.doPrivileged(Native Method)
20:42:53 INFO  	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(Unknown Source)
20:42:53 INFO  	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(Unknown Source)
20:42:53 INFO  	at java.awt.EventQueue$4.run(Unknown Source)
20:42:53 INFO  	at java.awt.EventQueue$4.run(Unknown Source)
20:42:53 INFO  	at java.security.AccessController.doPrivileged(Native Method)
20:42:53 INFO  	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(Unknown Source)
20:42:53 INFO  	at java.awt.EventQueue.dispatchEvent(Unknown Source)
20:42:53 INFO  	at java.awt.EventDispatchThread.pumpOneEventForFilters(Unknown Source)
20:42:53 INFO  	at java.awt.EventDispatchThread.pumpEventsForFilter(Unknown Source)
20:42:53 INFO  	at java.awt.EventDispatchThread.pumpEventsForHierarchy(Unknown Source)
20:42:53 INFO  	at java.awt.EventDispatchThread.pumpEvents(Unknown Source)
20:42:53 INFO  	at java.awt.EventDispatchThread.pumpEvents(Unknown Source)
20:42:53 INFO  	at java.awt.EventDispatchThread.run(Unknown Source)
``

**Comment by SubJunk** [comment]
@valib why did you add the `not ready` label? Are you planning on implementing that language detection, or something else?

**Comment by SubJunk** [comment]
@valib do you mean this NPE?

```
java.lang.NullPointerException: null
	at net.pms.dlna.DLNAMediaDatabase.insertOrUpdateSubtitleTracks(DLNAMediaDatabase.java:664)
	at net.pms.dlna.DLNAMediaDatabase.insertOrUpdateData(DLNAMediaDatabase.java:1003)
	at net.pms.dlna.DLNAResource.registerExternalSubtitles(DLNAResource.java:3949)
	at net.pms.dlna.DLNAResource.syncResolve(DLNAResource.java:1563)
	at net.pms.dlna.DLNAResource.run(DLNAResource.java:1358)
	at net.pms.dlna.RealFile.isValid(RealFile.java:65)
	at net.pms.dlna.DLNAResource.addChild(DLNAResource.java:627)
	at net.pms.dlna.DLNAResource.addChild(DLNAResource.java:589)
	at net.pms.dlna.SubSelFile.discoverChildren(SubSelFile.java:66)
	at net.pms.dlna.DLNAResource.discoverChildren(DLNAResource.java:1449)
	at net.pms.dlna.DLNAResource.discoverWithRenderer(DLNAResource.java:1312)
	at net.pms.dlna.DLNAResource.getDLNAResources(DLNAResource.java:1234)
	at net.pms.remote.RemoteBrowseHandler.mkBrowsePage(RemoteBrowseHandler.java:43)
	at net.pms.remote.RemoteBrowseHandler.handle(RemoteBrowseHandler.java:220)
	at com.sun.net.httpserver.Filter$Chain.doFilter(Filter.java:79)
	at sun.net.httpserver.AuthFilter.doFilter(AuthFilter.java:83)
	at com.sun.net.httpserver.Filter$Chain.doFilter(Filter.java:82)
	at sun.net.httpserver.ServerImpl$Exchange$LinkHandler.handle(ServerImpl.java:675)
	at com.sun.net.httpserver.Filter$Chain.doFilter(Filter.java:79)
	at sun.net.httpserver.ServerImpl$Exchange.run(ServerImpl.java:647)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
```

I can look into it when I have time

**Comment by SubJunk** [comment]
@valib that bug is fixed now, I tested it and can see the live subtitles results

**Comment by valib** [comment]
@SubJunk can you please fix conflicts. I try out it but finished in the middle and not undestand how to continue.

**Comment by SubJunk** [comment]
@valib yes I'll try

File: src/main/java/net/pms/dlna/DLNAMediaDatabase.java [hunk 6]
```diff
@@ -574,39 +569,49 @@ public synchronized boolean isOpenSubtitlesMetadataExists(String name, long modi
...
+
+					elements.close();
+					subs.setLong(1, id);
+					elements = subs.executeQuery();
+					while (elements.next()) {
+						String fileName = elements.getString("EXTERNALFILE");
+						File externalFile = isNotBlank(fileName) ? new File(fileName) : null;
+						if (externalFile != null && !externalFile.exists()) {
+							LOGGER.trace("Deleting cached external subtitles from database because the file \"{}\" doesn't exist", externalFile.getPath());
+							deleteRowsInTable("SUBTRACKS", "EXTERNALFILE", externalFile.getPath(), false);
```
SubJunk: @valib it occurred to me that this might be better suited to go into our database cleanup step instead of happening here, since then it is asynchronous to the request to `getData` and therefore will be more performant. Do you agree?

valib: @SubJunk the database cleanup is called when the scannig of shared folders finished but what about the case when user is browsing folders and subs were deleted in the mean time and scanning is not finished?

SubJunk: @valib that makes sense, thanks for explaining

valib:
.@SubJunk I think that we should return back before you merged the version 7.0. Until that the original PR #1372 was mergeable with master without any problems. Because I and seems that also you are not capable to follow Nadahars inputs in this PR it could be better to first merge the original PR and than your 7.0 because you know what you introduced there and maybe it will be easier for you to fix conficts. And than add all following changes as cherry picks.

SubJunk:
> .@SubJunk I think that we should return back before you merged the version 7.0. Until that the original PR #1372 was mergeable with master without any problems. Because I and seems that also you are not capable to follow Nadahars inputs in this PR it could be better to first merge the original PR and than your 7.0 because you know what you introduced there and maybe it will be easier for you to fix conficts. And than add all following changes as cherry picks.
> Time is running and it will be harder and harder to merge this PR.
> 
> EDIT: your approach could also works. The main problems occured between versions 7.5 and 7.6.

Yes it is pretty hard for developers to continue from where another developer left off, especially since you and I have a very different coding style to Nadahar too (that's not a bad thing about anyone, programming is a very unique and individual thing).

However I think we are doing really well, we have worked hard on this and it felt stable before these recent conflicts, and I agree with you that we should hurry up. We also have users who are running this build exclusively and they have reported good success with it so I feel it already has good manual testing behind it.

If you have any ideas about the specific commits that have caused our user problems please let me know, or revert them yourself. I will try to find time soon to go through the problems users have reported, but help is always appreciated because as always I am short on time (as I'm sure you are too :))

valib:
@SubJunk I tried to cherry pick some commits and stopped on commit 30d7c085257b1641ddbb5ee18ef9f9d88ca3f6cd which is above my skills.

SubJunk:
@valib that's probably a good approach to take to syncing this, going one commit at a time. I guess we should question whether to keep or revert that commit first, and if we are keeping it I can try to sync it. It is the first part in a series of a few commits, and they are all quite heavy.

SubJunk:
@valib thanks for the further work, it's getting there!

SubJunk:
@valib are you working on that TODO? I am about to start looking at it if not

SubJunk:
@valib I tried to reproduce the problem and it seemed to be working for me, maybe you forgot to reset your database before testing it or maybe I misunderstand the problem?

valib:
@SubJunk my mistate. The TODO commit I will revert.

valib:
@SubJunk now after testing it should be merged.

SubJunk:
@valib I will do thorough testing on it tomorrow

SubJunk:
@valib UMS ran out of memory on its initial scan for me, but I might have fixed it. It takes a long time to test it so I am still testing and will post whether it is fixed, then I'll do testing of the functionality in this branch

SubJunk:
Yeah that is working now so I will continue to test `b46011b`

SubJunk:
I'm getting failed playback on subtitled videos on the web interface, will investigate more

SubJunk:
@valib since this branch is now included in the `8.0.0` branch, do you want to work on it in there instead of here and close this PR? It was part of the `8.0.0-a1` release

valib:
@SubJunk it is up to you if this PR will be merged to the master or` 8.0.0-a1`. I continued with this PR to be  compatible with master.

valib:
The last commit works only with the MediaInfo version 18.12. Previous versions have problem that the `Format/String` sometimes returns `null`.

SubJunk:
@valib I propose merging this branch into the `8.0.0` and continuing work on it there since I did a nice squash :p I can also update MediaInfo in our SVN repo. Also do you want access to our SVN repo or happy for me to continue managing it?

If you are happy for work on this to continue in that branch please let me know and I can merge it in

valib:
@SubJunk no problem. If you merge this branch it could be closed.

SubJunk:
Closing this since work is continuing in `8.0.0` https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1615

PR #1486, opened 2018-04-14
Don't use global ID repository during cache scan

SubJunk:
This resolves the initial memory use problem. It's a simple boolean that disables adding to global repository during a scan


SubJunk:
I'd like to release this very soon in 7.0.1 since we have a lot of users experiencing this problem. I have tested it and it works for me, the memory manages itself after scanning over 10k files. If there are objections we can revert it. I usually want to wait for reviews but this is urgent.

SubJunk:
So unless anyone sees a big problem with it I'll merge in a few minutes and try to do a release today @UniversalMediaServer/developers

PR #1490, opened 2018-04-28
Fix potential errors of audio stream ffmpegvideo

onon765trb:
This PR fixes potential errors. Not used options can be a cause for errors. Extra options can result in decrease in quality or behavior users unexpect. Needs to consider ffmpeg often changes the behavior of options.


src/main/java/net/pms/encoders/FFMpegVideo.java
```diff
@@ -356,13 +356,17 @@ public FFMpegVideo(PmsConfiguration configuration) {
...
 				} else if (type() == Format.AUDIO) {
 					// Skip
-				} else if (renderer.isTranscodeToAAC()) {
-					transcodeOptions.add("-c:a");
-					transcodeOptions.add("aac");
 				} else {
-					if (!customFFmpegOptions.contains("-c:a ")) {
-						transcodeOptions.add("-c:a");
-						transcodeOptions.add("ac3");
+					if (!customFFmpegOptions.contains("-(c:a|codc:a|acodec)")) {
```
SubJunk: Should `codc:a` be `codec:a`?

Nadahar: I don't think ```contains()``` support regex..

valib: The ```contains()``` definitelly not support regex. @onon765trb have you ever tested your PRs?

src/main/java/net/pms/encoders/FFMpegVideo.java
```diff
@@ -1007,12 +1011,12 @@ public synchronized ProcessWrapper launchTranscode(
 					channels = configuration.getAudioChannelCount();
 				}
 
-				if (!customFFmpegOptions.contains("-ac ") && channels > 0) {
+				if (!customFFmpegOptions.contains("-ac ") && channels > 0 && params.aid.getAudioProperties().getNumberOfChannels() > configuration.getAudioChannelCount()) {
```
SubJunk: Would it be better to use the `channels` variable instead of `configuration.getAudioChannelCount()` on this line?

src/main/java/net/pms/encoders/FFMpegVideo.java
```diff
@@ -356,13 +356,17 @@ public FFMpegVideo(PmsConfiguration configuration) {
...
 				} else if (type() == Format.AUDIO) {
 					// Skip
-				} else if (renderer.isTranscodeToAAC()) {
-					transcodeOptions.add("-c:a");
-					transcodeOptions.add("aac");
 				} else {
-					if (!customFFmpegOptions.contains("-c:a ")) {
-						transcodeOptions.add("-c:a");
-						transcodeOptions.add("ac3");
+					if (!customFFmpegOptions.matches(".* -(c:a|codec:a|acodec) .*")) {
```
SubJunk: Is it intentional to have space characters before the `.*` here? It might limit the matching e.g. if the command is the last custom command in the `customFFmpegOptions` string

PR #1495, opened 2018-05-04
Maintain aspect ratio in web interface thumbnails

SubJunk:


SubJunk:
@UniversalMediaServer/developers this one-liner makes images maintain their aspect ratio. I've tested it in fixed and dynamic modes, and it works with the different background settings too.

PR #1497, opened 2018-05-05
WIP Java 10 crash fix

### LINKED ISSUE #1496: Unable to start 7.0.1 with Java 10.0.1 on macOS High Sierra 10.13.4
Installed Java using: `brew cask install java`

`java -version` returns

```
java version "10.0.1" 2018-04-17
Java(TM) SE Runtime Environment 18.3 (build 10.0.1+10)
Java HotSpot(TM) 64-Bit Server VM 18.3 (build 10.0.1+10, mixed mode)
```

When trying to run Universal Media Server `7.0.1` a dialog is displayed: "Unable to load Java Runtime Environment". 

When trying to execute the script: `Universal Media Server.app\Contents\MacOS\UMS` the following error is displayed:

```
2018-05-05 00:52:45.441 UMS[1018:9999] No matching JRE or JDK found.
2018-05-05 00:52:45.445 UMS[1018:9999] Error launching JVM Runtime ((null)) Relative Path: '/Applications/Universal Media Server.app/Contents/PlugIns' (dylib: (null))
  error: Unable to load Java Runtime Environment.
```

screen shot 2018-05-05 at 01 00 24

Is this version of the JRE not detected / supported?

When running directly, a class not found exception is displayed:

```
java -jar /Applications/Universal\ Media\ Server.app/Contents/Java/ums-7.0.1-jar-with-dependencies.jar
Exception in thread "main" java.lang.NoClassDefFoundError: com/sun/java/swing/plaf/windows/WindowsLookAndFeel
	at java.base/java.lang.ClassLoader.defineClass1(Native Method)
	at java.base/java.lang.ClassLoader.defineClass(ClassLoader.java:1009)
	at java.base/java.security.SecureClassLoader.defineClass(SecureClassLoader.java:174)
	at java.base/jdk.internal.loader.BuiltinClassLoader.defineClass(BuiltinClassLoader.java:801)
	at java.base/jdk.internal.loader.BuiltinClassLoader.findClassOnClassPathOrNull(BuiltinClassLoader.java:699)
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClassOrNull(BuiltinClassLoader.java:622)
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:580)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:190)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:499)
	at net.pms.PMS.main(PMS.java:1145)
Caused by: java.lang.ClassNotFoundException: com.sun.java.swing.plaf.windows.WindowsLookAndFeel
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:582)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:190)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:499)
	... 10 more
```

Found this, might be useful:

https://github.com/jidesoft/jide-oss/commit/bce89cec677169503b820cb67ca5709f7c72e6ec

SubJunk:
#1496


src/main/java/net/pms/newgui/LooksFrame.java
```diff
@@ -181,23 +205,43 @@ public static void initializeLookAndFeel() {
...
 
 			lookAndFeelInitialized = true;
 		}
 	}
 
+	/**
+	 * Safely checks whether a particular look and feel class is set.
+	 *
+	 * @param lnf
+	 * @param lookAndFeelClassPath
```
Nadahar: No big deal, but the parameters should have a short description as well.

valib:
There is a difference how looks the new L&F implementation 
new laf
and the current look
ums old
The font is smaller and could be udjusted.

Nadahar:
@valib The problem is probably (as always) the font scaling because you use a different DPI. It's not simply a matter of adjusting the font, that will break if for everybody with "normal" DPI.

valib:
@Nadahar maybe I use different DPI but the setting is the same for the current L&F and the new implementation which uses definitelly smaller or different font and it has nothing with the scalling.

Nadahar:
@valib It has everything to do with font scaling. Since you use a different DPI, you have set windows to scale fonts to 125% if I remember correctly. If you set Windows to not scale fonts, I'm pretty sure the "old" and the "new" implementation would look identical.

As I understand it, the problem is that the "new" solution doesn't handle font scaling, while the "old" do. That's why I say that it isn't simply a matter of scaling up the font, that will only work for you. A solution must be found so that font scaling works in the new solution, and it will work both for you and everybody else.

valib:
@Nadahar I set windows from the font setting 125% (recommended by Windows) to 100% and the result is the same. Current look 
old look
nad new look
new look
so it has nothing to do with the font scalling. Do you compare the current look to the new setting?

Nadahar:
@valib That's strange, maybe you're right. I'll have to test it, but I can't do that right now.

ExSport:
@valib Did you logged out?

SubJunk:
Thanks for the testing and review guys, I will work on this more soon

SubJunk:
@valib it's because on this line https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1497/files#diff-75f202387ab8fdc0d0ebb9d63e299cf7R102 I changed from jgoodies to plaf. Maybe there is a parameter we can pass in, or maybe the plaf one just doesn't have font scaling support :(

Nadahar:
@SubJunk The old version of jgoodies that we use is open source. You should probably look at what it does when creating a new ```WindowsLookAndFeel```. I suspect that it still use Sun's implementation in the bottom, so it's probably just a "wrapper" providing additional functionality like font scaling and other things.

valib:
I think that I found the problem. In the `com.sun.java.swing.plaf.windows.WindowsLookAndFeel` there is a `private boolean useSystemFontSizeSettings`  which initial value is not set which means it is false. There is not a method how to set it true so the font scalling is not working.

SubJunk:
Maybe we can reproduce the things that flag does, at a later time. It looks like it does this:

```
            MenuFont = new WindowsFontSizeProperty("win.menu.font.height", Font.DIALOG, Font.PLAIN, 12);
            FixedControlFont = new WindowsFontSizeProperty("win.ansiFixed.font.height", Font.MONOSPACED,
                       Font.PLAIN, 12);
            ControlFont = new WindowsFontSizeProperty("win.defaultGUI.font.height", Font.DIALOG, Font.PLAIN, 12);
            MessageFont = new WindowsFontSizeProperty("win.messagebox.font.height", Font.DIALOG, Font.PLAIN, 12);
            WindowFont = new WindowsFontSizeProperty("win.frame.captionFont.height", Font.DIALOG, Font.BOLD, 12);
            ToolTipFont = new WindowsFontSizeProperty("win.tooltip.font.height", Font.SANS_SERIF, Font.PLAIN, 12);
            IconFont    = new WindowsFontSizeProperty("win.icon.font.height", Font.DIALOG, Font.PLAIN, 12);
```

SubJunk:
@valib @Nadahar I just pushed a temporary commit to show my attempt at doing this, since I have to stop working for the day. The program fails to start though, maybe one of you can see why that is if you're interested

Nadahar:
It's hard for me to see how setting the font size to a fixed size of 12 can make font scaling work. Such values should never be hardcoded in my experience. What am I missing?

src/main/java/net/pms/newgui/LooksFrame.java
```diff
@@ -320,6 +322,76 @@ public LooksFrame(AutoUpdater autoUpdater, PmsConfiguration configuration) {
 			UIManager.put("Tree.font", sf);
 			UIManager.put("Spinner.font", sf);
 		}
+					// *** Shared Fonts
+					Integer twelve = new Integer(12);
```
valib: `new Integer()` is deprecated in Java 9+. It is recommended to use static `Integer.valueOf()` instead which internally returns `new Integer(Value)`.

valib:
@Nadahar don't ask me but it works. 😲

SubJunk:
Our current `master` works for me on Windows 10 and Java 10, and I need to upgrade my macOS version soon to test on macOS, since I don't think I can get Java 10 on Sierra. So I'll test soon. Thanks for the work to clean this up @valib !

SubJunk:
Before this piece of work is finished we need to make it play nicely with the previous settings on line 291 that seem to be there for languages with special characters, since as it is now it will overwrite them

Nadahar:
@SubJunk GitHub has now updated their review requirements to make more sense. If you go to Settings -> Branches and click ```Edit``` for ```master``` you can now enable "Dismiss stale pull request approvals when new commits are pushed". This makes approvals a whole lot more meaningful, and less risky to give, as further changes to the PR will invalidate the approval and require the reviewer to approve the changes since the last review.

It's the crucial functionality that has been missing with GitHub reviews since they arrived as I see it, and my main (although not the only) reason the prefer Reviewable.

I absolutely thing you should enable this, I will certainly be less hesitant to approve PRs with this.

SubJunk:
Great change, I've enabled that

SubJunk:
This might be related to at least some of the problems in https://www.universalmediaserver.com/forum/viewtopic.php?f=10&t=12557

SubJunk:
We have confirmation from a user that this fixes the problem https://www.universalmediaserver.com/forum/viewtopic.php?p=37789&sid=c6b6f42c4b965bccb58ca8a457e1a4ea#p37789

SubJunk:
@valib sorry I should have noted that this still needs a little bit of work before merge as noted here https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1497#issuecomment-390900339

I will prioritize it since it seems to be causing problems for a lot of users

SubJunk:
@valib I have confirmed that English and Japanese languages works with these changes using Java 8 and 10. If you are happy to approve it again we can merge

SubJunk:
There are some differences between this font and the other one unfortunately, it was a bit nicer before. Also the new one is a bit wider so things don't fit as nicely. Attached are before and after shots:

before
after

One thing we could do is make it so the new font changes are only applied on Java 10+, so that our Java 8 and standalone users can still have the nicer font while we come up with a better solution

SubJunk:
This is about as critical as a bug can get (whole program doesn't start) and the fix has been released in another program for a long time, and this PR has previously passed review in May, and I have tested it manually with different languages and different operating systems, so I think we can merge it.

PR #1498, opened 2018-05-06
Improved support for Panasonic DX series TVs

SubJunk:
Almost identical to the CX but according to the manual they dropped DTS support in MKVs


SubJunk:
My friend tested it and it's working so I'll merge

PR #1502, opened 2018-05-29
Preserve the "sample" flag while prettifying

SubJunk:
This fixes a bug where:

```
Movie.2011.1080p.BluRay.DTS.x264-Releasers.mkv
Movie.2011.1080p.BluRay.DTS.x264-Releasers.Sample.mkv
```

Both show up in the server as "Movie (2011)" when prettified

### DIFF
--- a/src/main/java/net/pms/dlna/DLNAMediaDatabase.java
+++ b/src/main/java/net/pms/dlna/DLNAMediaDatabase.java
@@ -251,9 +251,10 @@ public class DLNAMediaDatabase implements Runnable {
 							version = 13;
 							break;
 						case 13:
-							// From version 12 to 13, we added 7 indexes
 							try (Statement statement = conn.createStatement()) {
-								statement.execute("ALTER TABLE FILES ADD EXTRAINFORMATION");
+								StringBuilder sb = new StringBuilder();
+								sb.append("ALTER TABLE FILES ADD EXTRAINFORMATION VARCHAR2(").append(SIZE_MAX).append(')');
+								statement.execute(sb.toString());
 							}
 							version = 14;
 							break;
--- a/src/main/java/net/pms/dlna/DLNAMediaDatabase.java
+++ b/src/main/java/net/pms/dlna/DLNAMediaDatabase.java
@@ -66,7 +66,7 @@ public class DLNAMediaDatabase implements Runnable {
 	 * The database version should be incremented when we change anything to
 	 * do with the database since the last released version.
 	 */
-	private final int latestVersion = 13;
+	private final int latestVersion = 14;
 
 	// Database column sizes
 	private final int SIZE_CODECV = 32;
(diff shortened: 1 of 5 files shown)


SubJunk:
@UniversalMediaServer/developers I have tested this and it seems to work well for now. It can maybe be extended more for different conventions instead of using `endsWith` eventually, but I think we can merge this if someone reviews it since it already fixes the bug for my media

src/main/java/net/pms/util/FileUtil.java
```diff
@@ -546,6 +547,10 @@ public static String getFileNamePrettified(String f, File file, DLNAMediaInfo me
 
 		formattedName = basicPrettify(filename);
 
+		if (formattedName.toLowerCase().endsWith("sample")) {
```
Nadahar: You should always use ```Locale.ROOT``` when using locale sensitive functions like `` toLowerCase()``` unless you actually wants it to do it in a locale-sensitive way. When you don't specify it, it will use the OS default locale, which means that the results are unpredictable.

SubJunk: @Nadahar I have added a change to use English "locale" because it is only doing lowercase to compare with an English string `sample`. If that is a good fix can you approve and merge, or if you think `Locale.ROOT` is better please explain or just make the change

Nadahar: @SubJunk ```Locale.English``` is perfectly fine AFAIK, the reason I use ```Locale.ROOT``` is because I understand it as that the "root" locale simply means "no special handling". I understand it as if the "root" locale is the "default"/basis/fallback for all the others, and that the other locales then basically just implements the deviations. When it comes to case conversion I'm pretty sure there's no "special handling" for English, so it will probably end up running the exact same code. I simply explicitly use ```Locale.ROOT``` to signal that I don't want any localized handling.

AFAICR Android doesn't have ```Locale.ROOT``` and recommends using ```Locale.English``` instead, so it's really just semantics.

When it comes to approving it, I guess I could since it's a small change and you probably knows this code well, but I haven't really studied what it does and would prefer if @valib could just update his review.

I understand the idea behind preserving the "sample" so you can tell the difference from "the real thing", but I'm not quite sure of how this relates to the "edition".

SubJunk: Cool, thanks for the reply. Yeah it is a bit of a hack to put it onto the end of the edition instead of making another parameter like `isSample` to `return`

SubJunk:
@Nadahar I made it less hacky now
@valib if you're still happy with it can you merge?

SubJunk:
@valib thanks!

src/main/java/net/pms/dlna/DLNAMediaDatabase.java
```diff
@@ -78,23 +78,19 @@
...
-	private final int SIZE_ARTIST = 255;
-	private final int SIZE_SONGNAME = 255;
 	private final int SIZE_GENRE = 64;
-	private final int SIZE_TVEPISODENAME = 255;
 	private final int SIZE_YEAR = 4;
 	private final int SIZE_TVSEASON = 4;
 	private final int SIZE_TVEPISODENUMBER = 8;
 
+	// Generic constant for the maximum string size: 255 chars
+	private final int SIZE_MAX = 255;
```
Nadahar: :+1:

src/main/java/net/pms/dlna/DLNAMediaDatabase.java
```diff
@@ -254,6 +250,14 @@ public synchronized void init(boolean force) {
...
 							version = 13;
 							break;
+						case 13:
+							try (Statement statement = conn.createStatement()) {
+								StringBuilder sb = new StringBuilder();
+								sb.append("ALTER TABLE FILES ADD EXTRAINFORMATION VARCHAR2(").append(SIZE_MAX).append(')');
+								statement.execute(sb.toString());
+							}
+							version = 14;
+							break;
```
Nadahar: Nice :+1:

PR #1503, opened 2018-06-02
Fix MIME type on Resume files

SubJunk:


SubJunk:
@valib does your Panasonic TV play "Resume" files currently? I had to update the MIME type in my renderer config to make it work so I wonder if the same change needs to be done to other Panasonic files

src/main/java/net/pms/dlna/DLNAResource.java
```diff
@@ -772,6 +772,8 @@ public void addChild(DLNAResource child, boolean isNew, boolean isAddGlobally) {
 
 					if (resumeRes != null && resumeRes.media != null) {
 						resumeRes.media.setThumbready(false);
+						resumeRes.media.setMimeType(HTTPResource.VIDEO_TRANSCODE);
+						resumeRes.media.setThumbready(false);
```
Nadahar: @SubJunk Why set "thumbready" twice?

SubJunk: Good question, not sure how that happened. Will fix now

SubJunk:
I resolved the issue that caused it to fail peer review so I'm merging

PR #1504, opened 2018-06-03
Fixed maximum memory setting in JRE9+

SubJunk:
Currently if a user has JRE9+ installed, and not JRE8- installed, NSIS will always default their maximum memory to 768MB. This restores our previous functionality of defaulting to 1280MB if they have higher than 4000GB of RAM.
Also some minor formatting to make the code more readable

### DIFF
--- a/src/main/external-resources/nsis/setup.nsi
+++ b/src/main/external-resources/nsis/setup.nsi
@@ -118,8 +118,11 @@ Function AdvancedSettings
 		SetRegView 64
 	${EndIf}
 	ReadRegStr $0 HKLM "SOFTWARE\JavaSoft\Java Runtime Environment" "CurrentVersion"
-	IfErrors SetMinMem
 	IfErrors 0 CheckMemAmnt
+
+	; This is where the registry entries go in JRE9+
+	ReadRegStr $0 HKLM "SOFTWARE\JavaSoft\JRE" "CurrentVersion"
+	IfErrors SetLowMemoryLimit CheckMemAmnt
 
 	; Get the amount of RAM on the computer
 	CheckMemAmnt:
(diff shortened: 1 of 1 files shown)


SubJunk:
I've tested it and it works for me, so it would be good if someone could review and merge

Nadahar:
@SubJunk My mind might be playing tricks on me now, but didn't you fix this quite some time ago? I'm pretty sure I remember some PR dealing with the changed registry path..

SubJunk:
Yeah it is the same fix but in a different file with a different purpose, this is the existing one https://github.com/UniversalMediaServer/UniversalMediaServer/blob/master/src/main/external-resources/nsis/pms.nsi#L120

I actually question whether we even need to check for Java here. Why does the fact that we found Java affect how much RAM we can use? It seems like faulty logic that can just be removed

Nadahar:
Much of the logic in the installer is indeed very flawed, but I wouldn't just remove it without replacing it with something that works.

What happens in the existing "logic" is that the heap size is set during installation and stored in the registry (by setup.nsi). When UMS.exe is started (pms.nsi), this value is then retrieved and applied to the command line that calls Java with ums.jar.

SubJunk:
I meant remove the part that checks whether Java exists before it sets the variable that populates the default for that field,, since the rest of the logic is independent from that part of it. It seems like pointless complexity when we go on to unconditionally set the HeapMem at https://github.com/UniversalMediaServer/UniversalMediaServer/blob/ec991d5f620a512918d3ae05229965d4705890e5/src/main/external-resources/nsis/setup.nsi#L170

Unless I am misunderstanding it which is possible, the logic in English is:

```
If we found Java:
  If RAM on the computer is greater than 4GB
    Offer the default value of 1280
  Else
    Offer the default value of 768
Else
  Offer the default value of 768

Then after the user saves their advanced options, save HeapMem in the registry (in a generic location, unaffected by previous things)
```

What I propose is cleaner:

```
If RAM on the computer is greater than 4GB
  Offer the default value of 1280
Else
  Offer the default value of 768

Then after the user saves their advanced options, save HeapMem in the registry (in a generic location, unaffected by previous things)
```

Then `pms.nsi` has its own logic to check and make sure the computer has Java before UMS starts

PR #1505, opened 2018-06-03
Fix PR  #1502

valib:
When tested PR #1502 I forgot to test clean install or delete the database to test (re)instalation.


SubJunk:
Thanks for fixing my mistake!

PR #1506, opened 2018-06-04
Rename translation keys to force Crowdin to allow new translations

SubJunk:
These are the strings that were changed in 7.0.0 but the changes aren't working properly in Crowdin so this will force it to work by just treating them as new strings instead


SubJunk:
I agree about the release timing. I doubt I will get to release today anyway since I want to include some other changes in the next release, unless you have time to review https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1504 and https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1503 today. No rush

SubJunk:
Ok I'm merging this now then I'll push it to Crowdin, we can do a release next week or the week after

PR #1509, opened 2018-06-09
Implement ParseLogger made by @Nadahar

valib:
@SubJunk I see this enhancement very usefull because we need to see the MediaInfo result for troubleshooting.  Now we have to ask users to provide the MediaInfo result to understand what is wrong.


SubJunk:
The code looks good to me but I haven't tested it. Have you tested it?

Nadahar:
Unless some errors have been introduced when cherry.picking, the code is relatively well tested.

PR #1510, opened 2018-06-09
Let FFMpeg to choose the propper video stream.

valib:
Thanks @Nadahar for the comment. I think that I found better way how to choose the video stream.


SubJunk:
I think that is intentional - the comment has been added to specify that FFmpeg will choose the video track with the highest resolution

PR #1514, opened 2018-06-16
Improvements to Windows auto updater:

SubJunk:
- Stop disabling close button
 - Stop disabling resizing
 - Set better minimum size
 - Save updated file to profile directory to get around most common permission problem
 - Update English translations
 - Fixed newline character not working
 - Updated javadocs
 - Attempt to run previously-downloaded file instead of saying "Must download before run"
 - Allow the whole process to happen without UMS being elevated
 - Added logging

### DIFF
--- a/src/main/java/net/pms/configuration/Build.java
+++ b/src/main/java/net/pms/configuration/Build.java
@@ -44,19 +44,19 @@ public class Build {
 
 	/**
 	 * the name of the subdirectory under which PMS config files are stored for this build.
-	 * the default value is "PMS" e.g.
+	 * the default value is "UMS" e.g.
 	 *
 	 *     Windows:
 	 *
-	 *         %ALLUSERSPROFILE%\PMS
+	 *         %ALLUSERSPROFILE%\UMS
 	 *
 	 *     Mac OS X:
 	 *
-	 *         /home/<username>/Library/Application Support/PMS
+	 *         /home/<username>/Library/Application Support/UMS
 	 *
 	 *     Linux &c.
 	 *
-	 *         /home/<username>/.config/PMS
+	 *         /home/<username>/.config/UMS
 	 *
 	 * a custom build can change this to avoid interfering with the config files of other
 	 * builds e.g.:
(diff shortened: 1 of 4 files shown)


SubJunk:
I have tested this a lot

SubJunk:
@UniversalMediaServer/developers after this is merged it might be a good time to consider enabling automatic update checking by default on Windows. At the moment it is an arduous process that requires restarting UMS to gain elevation but now it is much more user-friendly - the installer asks for permissions like normal but UMS itself never needs them anymore, unless the user has a custom setup with a profile folder that has no write permissions ofc

Nadahar:
@SubJunk If the user has a configuration with a read-only profile folder a lot of stuff will break like the database and all the other stuff that is stored temporary or permanently in the profile folder. I don't think that will be a problem, but what it the "normal" thing to do these days is to store the installer in the system provided temp folder. After installation, successful or not, the installer is simply deleted. The users will always have write permissions there, or everything will break. I think this is a nasty habit since it doesn't allow you to keep the installer file for later use and you will have to waste bandwidth by re-downloading if you ever need to run the installer again, but it's still very common. A better way in my opinion would be to let the user configure where to store them and if they should be deleted after use or not.

SubJunk:
@Nadahar are you saying this PR needs to be changed to use the temp folder before approval? As it is now, it already falls back to the temp folder if it can't write to the profile directory, and it will never save more than one installer file (it overwrites the old one if one exists) so I think that seems like a good system. Since we already have the ability to use a custom profile folder, by extension we now offer the ability to choose where the file is stored.

Nadahar:
@SubJunk No, I was only sharing my thoughts on the subject. To approve it I'd have to review the code. I took a quick look at it, but I don't really know the old autoupdate code well enough to properly see what this does.

SubJunk:
The autoupdate code is pretty bad to be honest. Someone had some good ideas for it I think, but executed them poorly, so its state management is full of bugs. This does improve it a lot though. Thanks for clarifying your thoughts

PR #1515, opened 2018-06-19
Remove Oracle JDK 7 test from Travis CI build

Nadahar:
It seems that Travis CI is finally decommissioning their Ubuntu Precise containers, which were the last to support Oracle JDK 7. This removes the tests so builds don't fail as a consequence.


SubJunk:
Thanks!

PR #1517, opened 2018-06-28
Use new FFmpeg on Windows 7+

SubJunk:
It's that time again, to see if FFmpeg will break things

My main reason for wanting the newer version is font/subtitles handling seems to have got faster in the update - with our 2016 version it is timing out for some files but the new one works better


SubJunk:
@valib any opinion on this?

SubJunk:
This isn't so simple because FFmpeg support for encoding to AAC has changed.

PR #1518, opened 2018-06-30
Run UMS with an internal, non-installed JRE

SubJunk:
This is working as it is on Windows (and AFAIK the same method will work on Linux and macOS). It adds an extra 40MB to the installer, which I think is probably ok given the 200+MB updates I get for every phone app these days.
Also we will probably need to start offering separate x64 downloads since I don't want to only use x86 JRE, and it would be silly to include both in the one installer.

Other than that it seems to offer us everything we were getting before with our NSIS runner, and we get to still use our NSIS installer - we are just replacing UMS.exe

I have confirmed that it uses the internal version of JRE, and for people who really want their own one, they can run the `jar` directly as before.

Help appreciated if anyone is interested in jumping in, especially with testing

Edit: Also we are including every file in the JRE when we probably don't need to, so a future task could be to trim those back, though not super important


SubJunk:
I have fixed the infrequent launching bug, it works every time for me now.

@UniversalMediaServer/developers can anyone please test this?

SubJunk:
@UniversalMediaServer/developers I think I will release this along with our next release, as a "standalone" version. So there would be like:

UMS-7.2.0.exe
UMS-7.2.0-standalone-x86.exe
UMS-7.2.0-standalone-x64.exe

Any objections?

SubJunk:
@UniversalMediaServer/developers I have worked more on this now by creating two new Maven profiles - `windows-standalone-x86` and `windows-standalone-x64`. This should also make it easier to review the code because the changes to `pom.xml` are separate from our existing profiles.

I have tested it and it is working well for me, and I think it is a good step forward for us. I would have preferred to get a multiple-platform solution and that is still my goal in the future, but this is working now for Windows, so I think it is ready for review. I'll release the first standalone versions as betas and ask users to provide feedback on them.

Anyone interested in reviewing?

SubJunk:
Our first release of this is released and a post explaining it is at http://www.universalmediaserver.com/forum/viewtopic.php?f=8&t=12676

One thing I realised after releasing it is that if the automatic updater is used, it will download the non-standalone build, so that should be fixed soon too.

SubJunk:
Ok the automatic updater should work for this branch now, but it is manually changed for this branch which means we can't merge this for now. I will add a label for that.

@UniversalMediaServer/developers I tried to make it happen dynamically by changing the project name based on profile but I couldn't figure out how to do it, any other ideas?

SubJunk:
@valib thanks for approving it, it seems to be working well but it can't be merged because of the automatic updating logic. maybe you can think of a way to make that work so that standalone builds update to standalone builds, and normal builds update to normal builds?

SubJunk:
This has been superseded

Issue #1527, opened 2018-07-14
Database thumbnail size


SubJunk:
I did some testing today on 3,500 files and found that if I remove the code that write thumbnails to the database, it's only 19MB, but with the thumbnails it is 900+MB.

Maybe we should look at some better compression for them and/or store them as files instead of the whole blob in the database - the database can store the path to the image file.

Another optimization would be to have a thumbnails table and check for exact matches, because at least with my music collection if I have an album of 12 songs it means there are 12 identical album covers that all get stored individually in our db, so people with a similar media setup to that could find the size reduced a lot.

Any thoughts @UniversalMediaServer/developers ?

Nadahar:
There are several things we can do here, reducing the thumbnail size is the most obvious. Some form of deduplication could probably achieve a lot too. I don't think there are much we can do to "improve" the compression except that we could certainly lower the "quality" setting for JPEGs. The blobs are already compressed as they are either JPEGs or PNGs,

When it comes to the audio files, I think we can do a huge improvement. Those that are retrieved from Cover Art Archive are already stored in `TableCoverArtArchive`, but per MusicBrainzID. That means they should already be deduplicated. If we made sure to just reference these instead of storing them with the media file, we should be able to make a little revolution with the database size.

SubJunk:
I tried to see what the quality setting was and it seemed like it is baseline huffman, does that algorithm allow any variation like to set it to 60% or 80% quality, as programs like Photoshop allow?

Good idea about the cover art archive. Some combination of all of these things should give us a big improvement.

A more dramatic improvement could just be to disable thumbnail caching by default. I noticed the server still worked nicely with that code removed, with the thumbnails just showing up when they were ready.

SubJunk:
The same database is 444MB at `0.6f` quality so that seems like an easy win. I'm working on the thumbnails table now, and then hopefully we can merge those two changes for a new release

Nadahar:
@SubJunk I've already got some work going on the thumbnails database solution. Your modification to the quality setting is to simple though. The code you changed is generic and used for many thing, you can't simply reduce it there. Many images are processes multiple times, and when using 60% for *each iteration* quality will suffer a lot. Not only thumbnails use that code either. You must make sure that the 60% quality is only applied to the last step before storage in the database to achieve your goal. I'm not sure I agree that 60% is acceptable quality though, but it might be good enough for thumbnails.

valib:
[...]

PR #1528, opened 2018-07-15
Database optimization

SubJunk:


src/main/java/net/pms/image/ImagesUtil.java
```diff
@@ -1479,7 +1479,7 @@ protected static Image transcodeImage(
 		Thumbnails.of(bufferedImage)
 			.scale(1.0d)
 			.outputFormat(outputFormat.toString())
-			.outputQuality(1.0f)
+			.outputQuality(0.6f)
```
Nadahar: As I already said in #1527 this is effectively a bug.

SubJunk: Is this better? https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1528/commits/c95f6b3bcc65dee219f526426fbb2657ccc96a84

src/main/java/net/pms/dlna/DLNAMediaDatabase.java
```diff
@@ -1252,6 +1256,62 @@ public synchronized void cleanup() {
...
+			 * Removes entries that are not referenced by any rows in the FILES table.
+			 */
+			ps = conn.prepareStatement(
+				"DELETE FROM SUBTRACKS " +
+				"WHERE NOT EXISTS (" +
+					"SELECT ID FROM FILES " +
+					"WHERE FILES.ID = SUBTRACKS.FILEID" +
+				");"
+			);
+			rs = ps.executeQuery();
```
atamariya: How about adding DELETE CASCADE constraint instead to child tables?

SubJunk: That's a good idea, thanks. That should work for all of them except the THUMBNAILS table

SubJunk: That's done now, not fully tested but is that what you meant?

atamariya: Yes, the constraints look fine.

valib:
@SubJunk there is a problem with the musicbrainz.org restriction to make only limited number of requests in the limited time othervise the server returns status code 503. The same apply for the @Nadahar implementation  https://github.com/DigitalMediaServer/DigitalMediaServer/commits/DeduplicateAudioThumbnails. This problem was reported by @Nadahar some time ago. You can reproduce this bug using the web interface. When this happen and exception is made and for the thumbnail is not used the general thumbnail but the last thumbnail made for different media.

Nadahar:
@valib 
> When this happen and exception is made and for the thumbnail is not used the general thumbnail but the last thumbnail made for different media.

If that's the case, it's a bug. I don't quite understand how this can be the case though, it seems very strange if it should use a thumbnail for a different media. Are you sure that's not just a bug with the web interface?

valib:
@Nadahar the web interface is making max 100 requests in paralel which set in the PMSConfiguration  `WEB_MAX_THREADS = 100` but is not the case of the standard renderers whose are obviously making requests only for number which is shown at the screen.

Nadahar:
@valib I wasn't thinking of the `503` error, but that the thumbnail for one media is shown for another. You can get the `503` error with normal renderers too, if you just browse a folder with many audio tracks in the, or browse multiple folders quickly. I just can't understand how a thumbnail for one media can be displayed for another, I would think that it would either show no thumbnail at all or a generic thumbnail.

Nadahar:
@valib Forget it, I misunderstood your original text. I thought you were talking about my implementation, but I see from the log file that you weren't. I haven't looked at the implementation details for the other, so I don't know if that might happen or not.

valib:
@Nadahar I am refering also to your DMS DeduplicateAudioThumbnails branch which has the same result..

Nadahar:
@valib I've been unable to confirm the bug you describe. I did however find another "problem" when testing this that causes a "similar" issue. I did get a lot of thumbnails that didn't belong to the correct files for those that couldn't be looked up. When looking into it, it turned out that it was because it falls back to "folder thumbnails" when no thumbnail can be found for that particular media. There is quite a lot of "mess" in the folder I used for testing, so it turns out it misidentified some images in the folder for being "folder thumbnails". My solution to that problem is: https://github.com/DigitalMediaServer/DigitalMediaServer/commit/16c3ccc5e8bf2bbd55e93161942b34290c039e81

In addition I had to delete a couple of files that actually had matching "folder thumbnail" names. After doing this, generic thumbnails are shown for those that can't be fetched.

Can you confirm if this fixes your issue too? As this isn't really related to neither my or @SubJunk's implementation, it's likely that this will cause the same issue when you test both branches.

valib:
@Nadahar I was wrong to inform that the last thumbnail is repeated. I had also the mess in my testing folder and there were the `folder.jpg` which was shown when the cover was not retrieved from the database because of the status code 503 returned. This is the main problem with the web interface and maybe with other renderers too when we not keep the MusicBrainz regulation `For "anonymous" user-agents we allow through (on average) 50 requests per second, and decline (http 503) the rest.`

Nadahar:
@valib I'm aware of the query rate limitation on musicbrainz.org. It wasn't there when I made this. To be able to control the query rate, a different solution would have to be implemented, as now the query is made when the thumbnail is requested. The results are cached in the database though, so it should ideally just query once per media file. But, one query can be up to 4 actual requests depending on if it finds something, so it's a bit hard to follow the restrictions they have. If you read below the text you quoted, there's also a limit per IP address - which is one request per second. That is very little if I read this correctly.

Nadahar:
@SubJunk I think it makes sense to look at how you're going to store the data to decide when to use joins. Since both the information from `FILES` and the thumbnail goes into `DLNAMediaInfo` I think it's correct to do a join there, especially since it doesn't seem that you will have any issues as long as you do a left outer join. Joining the other tables where there can be multiple entries per FILES entry is more of a challenge, and since they are used to create instances of different types, I can't see any point in using join there. First H2 would have to join all that information, and then our code would have to split it again, solving repeated information for the `FILES` table as you would have to do an outer join.

SubJunk:
I made the change from your review @valib

SubJunk:
@valib yes I think it would be a good idea to contact them about raising the limit for our user-agent

valib:
@SubJunk there is a bug at the line https://github.com/UniversalMediaServer/UniversalMediaServer/blob/04aebdb197d5cb53dbc79387c353bb37bba20aec/src/main/java/net/pms/dlna/DLNAMediaDatabase.java#L536

SubJunk:
Good catch, that's fixed now. Thanks

PR #1529, opened 2018-07-16
Fixed wrong logging

valib:


Nadahar:
@valib The previous code looks correct to me, and it looks like this PR will make it wrong. Have you tested with an image where you know the versions and verified that it is wrong?

valib:
@Nadahar it is possible that merging to the master was wrong but if the code is testing if `jfifVersion == null` and than is logged that `"JPEG DLNA compliance failed with too low Exif version \"%s\"",	exifVersion` and the same applies for `exifVersion == null` than it should be fixed.

Nadahar:
@valib I think you misread the code, that's why I asked if you've actually tested it. You must see those `if`'s in relation to the code above, some conditions have already been excluded, which is why `jfifVersion == null ` must mean that the Exif version is too low and vice versa.

valib:
@Nadahar thanks. I read the code more carefully and I see what you mean.

PR #1531, opened 2018-07-18
Implement the XXE restrictions for the UMS Xml code.

valib:
Based on recommendations from the https://www.owasp.org/index.php/XML_External_Entity_(XXE)_Prevention_Cheat_Sheet.

### DIFF
--- a/src/main/java/net/pms/network/UPNPControl.java
+++ b/src/main/java/net/pms/network/UPNPControl.java
@@ -9,7 +9,6 @@ import java.nio.charset.StandardCharsets;
 import java.util.*;
 import java.util.Map.Entry;
 import javax.xml.parsers.DocumentBuilder;
-import javax.xml.parsers.DocumentBuilderFactory;
 import net.pms.PMS;
 import static net.pms.network.UPNPHelper.sleep;
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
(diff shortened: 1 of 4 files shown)


Nadahar:
@valib I edited the text so we don't publish it.

Nadahar:
Great that you've looked into it :+1:

valib:
@UniversalMediaServer/developers can you test it? I don't have the Linux based PC in my network to do it.

valib:
@Nadahar I checked the `DocumentBuilderFactory` and the `TransformerFactory` in the UMS code and those are only occurences when they are implemented. The Cling is a different  issue but we can point that this is its problem only by testing which I am not able.

Nadahar:
@valib Ok. I've made another XML parser implementation (StAX I think) in #1372 at least. I think it's quite likely that Cling will have the same issue, since this is enabled by default in all the "normal" Java libraries.

valib:
@Nadahar done in https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1483/commits/add051ab7b386069aa912e147271879b56f60c8a

valib:
I have checked the Cling code and as expected the XXE vulnerability is not solved. So we need testing.

Nadahar:
@valib I've tried to set up a test, but haven't been able so far. The script throws stack exceptions that doesn't make much sense to me, and I really don't feel like start debugging Python.

If it's not solved in Cling, there's no need for testing - it must be fixed. We can either override the necessary classes or make a PR at https://github.com/4thline/cling.

Nadahar:
I found the bug in python script, so I managed to run the test. I can confirm that it's still exploitable.

SubJunk:
@valib is this still a good change even if it doesn't fix the problem? I am happy to merge if so

valib:
@SubJunk it is already implemented in master. Merged from #1615

PR #1535, opened 2018-07-28
Fixes an issue with the database upgrade script in 7.2.0

SubJunk:


SubJunk:
I will hopefully be releasing this as 7.2.1 today

valib:
@SubJunk what about to make it in  the different way #1536

PR #1537, opened 2018-08-01
Configurable media scanning

valib:
This allow users to choose if they let UMS to scan shared folders even if it slow down the UMS response.


.travis.yml
```diff
@@ -14,6 +14,7 @@ matrix:
     - os: linux
       jdk: openjdk8
     - os: osx
+      osx_image: xcode9.3
```
Nadahar: @valib It turns out 9.3 wasn't the best choice since it will soon be removed, so the previous 8.x default is probably a better choice.

valib: @Nadahar I rely on your macOS knowledge so what you reccomend in this case I will follow :-)

Nadahar: @valib Yes, @Sami32 made me aware that 9.3 wasn't the smartest choice after I did it, which is why I mentioned it. It doesn't relate to macOS per se, it's just that Travis CI has tied the Java version to the xcode version for some reason. 9.4, which is the new default, has Java 10, while the earlier images has Java 8.

valib: @Nadahar we should decide to use either xcode9.2 or xcode8.3 because we will not pass the Travis test until we fully implement Java 10 e.g. new LookAndFeel for Windows and other.

Nadahar: @valib Yes, that was my point to. 8.3 might be the best choice, it all depends on which version will stay available the longest at Travis CI. All that matters to us is the Java version, as we don't use xcode itself.

SubJunk:
@valib this is a good idea. Do you mind if I tweak the language a bit? Also we should maybe set the priority of scanning to the lowest one instead of just one below normal, maybe that will help the responsiveness

valib:
@SubJunk feel free to reword it. My english is not so good :-)
I am not sure about the lowest priority of scanning it will have any influence. As @Nadahar pointed out when I set the priority below normal

> While this is a good thing to do (I think it was one of the things I mentioned a long time ago in the 7.0.0 thread that has been ignored), don't expect too much of an effect. The scanning and UMS in general, maybe except transcoding, is I/O bound. Thus, I don't expect this to make much of a difference, the harddrive will be the bottleneck and the harddrive being busy is what will make the system less responsive.

But we can try.

Nadahar:
There's another thing that reduce the impact of setting thread priority too. Very few software actually use it (look in task manager and look at the priorities), so almost everything is running on "normal". That means that as soon as you use a priority lower than "normal", you're usually already "last in the queue". Any further reduction will only make any difference if there are other threads that use "lower".

SubJunk:
I pushed my tweaks and in my opinion this is good to merge

SubJunk:
I have added the new string to Crowdin in advance to give translators as much time as possible

.travis.yml
```diff
@@ -14,7 +14,6 @@ matrix:
     - os: linux
       jdk: openjdk8
     - os: osx
-      osx_image: xcode9.3
```
Nadahar: @SubJunk Why would you do that? Now it will fail.

SubJunk:
I'll put it back

valib:
@SubJunk agree that it could be merged but do you merge this PR even if it did not past the Travis CI?

SubJunk:
No I don't think we should merge things that don't pass on Travis. Cool I'm merging it now

PR #1538, opened 2018-08-10
Fixed marking files fully played

SubJunk:


Nadahar:
I don't really follow what you're doing with the database, so I can't review this. It will fail at Travis until the xcode version is changed to something that works though. But, 3 commits for one tiny change? I know we have different views on this, but I think this is pure madness.

IMO the changes you did to xcode in #1537 should be reversed and it should be merged. This should then first be squashed and then rebased on master. Then things will work without being completely mad.

SubJunk:
@Nadahar let me know if you want me to explain what's happening with the database

PR #1540, opened 2018-08-11
Fix potential resource leak

valib:


SubJunk:
@valib this could be good, but what made you do it? Was it an auto IDE thing or you noticed the leak or what?

SubJunk:
It seems Travis might be failing due to some temporary problem with `com.google`:

`package com.google.gson does not exist`

SubJunk:
I wonder what is causing the Travis failures. Maybe the caching?

valib:
@SubJunk actually it is an Eclipse warning that the `input` stream is not closed. After your questions I checked the code and I think that this implementation is wrong. I tried to test it but I could't find the way how to do it. The condition https://github.com/UniversalMediaServer/UniversalMediaServer/blob/2532c798cc32d0d20dbc04041b53bca62a1c1e5c/src/main/java/net/pms/io/BufferedOutputFileImpl.java#L310 is normally skipping the folowing code.

SubJunk:
I guess we could make some test code that loops through that function a lot of times to see if we run out of memory?

Also not related to this part of the code but related to resource leaks, we have a user on the forum (boss) who runs out of memory seemingly just from his TV sitting there not playing anything too, so I think it's good to look into resource leaks. He ended up setting a cron job to restart every 2 days, really sucks for him and I spent a lot of time trying to figure out what happens.
A current theory of mine is that maybe our renderer identification code isn't freeing resources for certain renderers when they send us their info so it piles up. Not sure but that one is annoying for the user and me!

Nadahar:
Resource leaks aren't just about memory. Depending on what the "resource" is, it can represent an open file (lock in the filesystem) or a network socket or TCP connection etc. There is almost always a limit to how many can be held open at the same time of these things, but the exact value varies by OS and system configuration.

I've seen another report on the forum about files being "impossible to delete" while UMS is running, which is evidence that a file lock hasn't been released.

Eclipse's "resource leak detection" isn't foolproof, some times it complains even though I can't see that the claim is valid. That said, I take any "complaint" as something that should be checked very closely. It's not enough to make sure that it's closed if "everything goes as planned", it much always be closed under any circumstances. I don't think "load testing" is a good way to determine such a problem, what can present no apparent issue on one system might fail on another.

In my opinion the only option is to have a water-tight logic. Sadly, that is often not the case in our code.

If you for example have created and instance that represents some resource, and then creates a new instance and assigns this to the same variable, the reference to the "old" resource is lost and it becomes impossible to close. It doesn't help if the code the finally closes the last resource it held, the "lost" on the way will stay open until the JVM is terminated.

Nadahar:
Regarding running UMS continuously, I've given that up a long time ago. I did mention this repeatedly in the beginning, but since I got more familiar with the code I realized that there are so many potential reasons for this that I consider it a task that it "impossible" to take on. My "solution" has been to try to clean up one thing at a time, so that hopefully some day it could be manageable to clean up the rest. I never got much interest from the rest of you when trying to clean things up though, so I stopped writing about things I find a long time ago.

SubJunk:
How can we make progress on this issue? Or is it even a bug or just a false positive from the IDE?

valib:
@SubJunk I think that my presumption was wrong and this code could make an error. Because I don't know how to test it am going to close this PR.

Issue #1542, opened 2018-08-12
Folders not able to be deleted with UMS running


SubJunk:
We have some reports of this, e.g. http://www.universalmediaserver.com/forum/viewtopic.php?f=9&t=12664
I haven't confirmed it myself so this is more of a placeholder until it can be confirmed. If it is a bug it would make sense because in v7 we added the functionality of watching folders for changes to files, so that could block the folder unintentionally.

PR #1543, opened 2018-08-14
Fix not being able to delete folders containing folders that are within watched folders (#1542)

SubJunk:


valib:
Tested and it is true that subfolders with media inside are not able to delete. This fix works but not fully. The subfolder is possible to delete also with media files inside but those media files are not deleted from the database.

SubJunk:
@Nadahar afaik it is a Windows-only bug

@valib thanks for testing, yes it seems like we should be more smart about how to handle deletion of folders. That isn't really part of this bug (it is probably the same in master right?) but it might make sense to fix them at the same time

Also it seems I screwed up the referential integrity stuff again so I am about ready to give up and just go back to our old sloppy way of using the cleanup step after scanning :p it is elegant to do it with foreign key constraint cascades but I don't think I'm good enough at using them at this stage

SubJunk:
I just pushed a try at support for the database call working when a folder is deleted, please test

SubJunk:
Tested and seems to be working for me

src/main/java/net/pms/dlna/RootFolder.java
```diff
@@ -1514,6 +1514,9 @@ public void notify(String filename, String event, FileWatcher.Watch watch, boole
 							} else {
 								LOGGER.trace("Folder {} is empty", filename);
 							}
+						} else if ("ENTRY_DELETE".equals(event)) {
```
valib: @SubJunk this doesn't work for me. It was also my idea to fix it in that way but there is a condition https://github.com/UniversalMediaServer/UniversalMediaServer/blob/5f8da71ce15a223fb90fb145166baf197ddce88d/src/main/java/net/pms/dlna/RootFolder.java#L1501 which is always false for folders.

SubJunk: Oh that's weird... I tested on Windows 7 so that must be why we got different results. Which version of Java?

SubJunk: We can do a hack around it anyway by detecting whether the string is a folder, and even if that detection gets it wrong (like if it is a file with no extension) it won't cause any big problems, I think it would just delete the file instead, and if it doesn't the file will be deleted during the cleanup step of the next scan

valib: @SubJunk I don't know if it works for WIN7 but it is logical that the code at line 301 doesn't work for deleted folders because the `WatchEvent` is raised after the OS deletes the folder and in that time no longer exists on the drive. The `Files.isDirectory()` is checking if the folder exists and if not returns `false` like it is not folder.

SubJunk:
@valib thanks for the commit. Does it work now for you? Is there more waiting to do on here?

PR #1545, opened 2018-08-20
Rollback referential constraints and do cleanup instead

SubJunk:


SubJunk:
@valib I think if we can get this one progressed, we have a good next version to release. Is there anything I can do to help this move along? I'm not in a rush, just asking :)

SubJunk:
@UniversalMediaServer/developers ok so now this database upgrade is covered by tests which gives me more confidence in it. We verify that no errors occurred, and that there are no referential constraints remaining after the upgrade. Any feedback or approval?

Nadahar:
@SubJunk I generally stay away from all bugfixes for the "7.0.0 code" as I didn't think it was finished when it was merged. As I pointed out then, there were several issues that was pointed out and ignored in #942. At that point I disengaged from everything related to this code, and that is still the case.

Nobody can guarantee that a code is bug free when released, but that's a long way from knowingly releasing "alpha quality" code. I think it's less work and results in better code quality when everything is handled when things are made, instead of spending the next year trying to patch it. In addition, the users don't have to suffer. As a result of not being heard on this principle, I don't feel any responsibility for correcting the situation.

SubJunk:
This pull request has nothing to do with 7.0.0 code, it fixes a bug that has existed for 20 days, since https://github.com/UniversalMediaServer/UniversalMediaServer/commit/aca7d5f006dea0cadecb9251f500bf2ae0c723b9

SubJunk:
This fixes a critical bug in our last release (the bug is that for some users - depending on the prior state of their database - the fully played status doesn't save to the database) and I have written unit tests to test the upgrade code and some related functionality, and the upgrade code verifies itself on the user's computer too.

I'm merging it now since it seems like otherwise it will just sit here.

PR #1546, opened 2018-08-20
Multi audio fix

SubJunk:
Possibly fixes http://www.universalmediaserver.com/forum/viewtopic.php?f=9&t=12694&p=37043#p37043

If user tests and gets back to us we can do a better fix because I know there was a reason for doing this https://github.com/UniversalMediaServer/UniversalMediaServer/commit/73be0ba86ee7f3c8f671fd3487e03c125cac7ca7#diff-880a7e67943a96acd14c61af9f392546 in the first place @valib

### DIFF
--- a/src/main/java/net/pms/encoders/FFMpegVideo.java
+++ b/src/main/java/net/pms/encoders/FFMpegVideo.java
@@ -954,6 +954,10 @@ public class FFMpegVideo extends Player {
 		// Map the proper audio stream when there are multiple audio streams.
 		// For video the FFMpeg automatically chooses the stream with the highest resolution.
 		if (media.getAudioTracksList().size() > 1) {
+			// Set the video stream
+			cmdList.add("-map");
+			cmdList.add("0:v:0"); //TODO find a way how to automatically select propper stream when media includes multiple video streams
+
 			cmdList.add("-map");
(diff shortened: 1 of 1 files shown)


src/main/java/net/pms/encoders/FFMpegVideo.java
```diff
@@ -954,6 +954,10 @@ public synchronized ProcessWrapper launchTranscode(
 		// Map the proper audio stream when there are multiple audio streams.
 		// For video the FFMpeg automatically chooses the stream with the highest resolution.
 		if (media.getAudioTracksList().size() > 1) {
+			// Set the video stream
+			cmdList.add("-map");
+			cmdList.add("0:v"); //TODO find a way how to automatically select propper stream when media includes multiple video streams
```
valib: @SubJunk maybe    `0:V`    would be better because in the FFMpeg Documentation is written that ’v’ matches all video streams, ’V’ only matches video streams which are not attached pictures, video thumbnails or cover arts.

SubJunk:
Ok I checked the old docs and they still say it so I think we can make that uppercase change https://web.archive.org/web/20160609011350/https://ffmpeg.org/ffmpeg.html#Stream-specifiers-1

SubJunk:
@valib that's done now

PR #1547, opened 2018-08-31
Support for a movie without audio

onon765trb:
Support for a movie without audio, based on Issue #1175 and closed Pull requests #1272.
This is because some renderers do not support for a movie without audio and remain DLNA compatibility.


src/main/external-resources/UMS.conf
```diff
@@ -866,6 +866,13 @@ ffmpeg_fontconfig =
 # Default: true
 ffmpeg_mencoder_problematic_subtitles =
 
+# Output silent audio
+# --------------------------------------------
+# Output silent audio when a movie do not have audio in ffmpegvideo. Some of renderers 
+# do not support a movie without audio. This can be perfomed by either UMS.conf or renderer.conf.
+# Default: false
+ffmpeg_output_silent_audio=
```
SubJunk: Really minor formatting thing - please add a space with the =, so `ffmpeg_output_silent_audio =`

src/main/external-resources/renderers/DefaultRenderer.conf
```diff
@@ -286,6 +286,10 @@ WrapEncodedAudioIntoPCM =
 # Default: true
 MuxLPCMToMpeg = 
 
+# Output silent audio when a movie do not have audio in ffmpegvideo. Some of renderers do not support a movie without audio. This can be perfomed by either UMS.conf or renderer.conf.
+# Default: false
+OutputSilentAudio=
```
SubJunk: again space, and please add some line breaks so it matches the style of the rest of the file

src/main/java/net/pms/configuration/PmsConfiguration.java
```diff
@@ -151,6 +151,7 @@
 	protected static final String KEY_FFMPEG_MENCODER_PROBLEMATIC_SUBTITLES = "ffmpeg_mencoder_problematic_subtitles";
 	protected static final String KEY_FFMPEG_MULTITHREADING = "ffmpeg_multithreading";
 	protected static final String KEY_FFMPEG_MUX_TSMUXER_COMPATIBLE = "ffmpeg_mux_tsmuxer_compatible";
+	protected static final String KEY_FFMPEG_OUTPUT_SILENT_AUDIO= "ffmpeg_output_silent_audio";
```
SubJunk: again space

src/main/java/net/pms/configuration/RendererConfiguration.java
```diff
@@ -166,6 +166,7 @@
 	protected static final String OFFER_SUBTITLES_BY_PROTOCOL_INFO = "OfferSubtitlesByProtocolInfo";
 	protected static final String OFFER_SUBTITLES_AS_SOURCE = "OfferSubtitlesAsSource";
 	protected static final String OUTPUT_3D_FORMAT = "Output3DFormat";
+	protected static final String OUTPUT_SILENT_AUDIO= "OutputSilentAudio";
```
SubJunk: again space

src/main/resources/i18n/messages.properties
```diff
@@ -98,6 +98,9 @@ FFmpeg.0=<html>Font setting is applied to embedded and external text-based subti
 FFmpeg.1=Defer to MEncoder when transcoding problematic subtitles
 FFmpeg.2=<html>MEncoder is more stable than FFmpeg at transcoding some kinds of subtitles, so this lets us choose which is the most stable option for each video.</html>
 FFmpeg.3=Use font settings (size, color ...)
+FFmpeg.4=Output silent audio
+FFmpeg.5=<html> Output silent audio when a movie do not have audio in ffmpegvideo.<br>Some of renderers do not support a movie without audio.<br>This can be perfomed by either UMS.conf or renderer.conf, also.
+</html>
```
SubJunk: This should probably be on the previous line, and do you mind if I do a commit to alter the English in here?

onon765trb: No problem. I will correct only spaces.

onon765trb: About " Are there any renderers in particular that this can be applied to straight away?"
Yes.

SubJunk: @onon765trb great, let's add this for at least one renderer in this Pull Request, because it would be good for the feature to have at least one use

onon765trb: @ SubJunk Does it mean that the changes of PmsConfiguration.java and GUI had better been deleted? I don't mind even if so. In so case, which will I or you do that?

onon765trb:
@SubJunk Does it mean that the changes of PmsConfiguration.java and GUI had better been deleted? I don't mind even if so. In so case, which will I or you do that?

src/main/resources/i18n/messages.properties
onon765trb: @SubJunk Nice! Do you commit it? I don't mind it. Do you think I had better add the feature to my other PRs, #1548 and #1549.

SubJunk: @onon765trb I don't know which renderers do not support movies with no audio track, so I can't do it

onon765trb:
@SubJunk I found one point that modification is need. It is a consideration of external Picture Subs. I will do it, soon before a merge.

onon765trb:
I resolved the conflicts with the master.

SubJunk:
This is still waiting for a renderer to use it - I don't think we should add a new config option that is unused

SubJunk:
In the text in this Pull Request it says:

> This is because some renderers do not support for a movie without audio and remain DLNA compatibility.

So I think we should add at least one of these in here

onon765trb:
It seems that I need more concretely.
Which document or  behavior what you say is?
In the present, this PR outputs silent audio when a movie does not audio.
This can be excused by either UMS.conf or renderer.conf (`(configuration.isFFmpegOutputSilentAudio() || params.mediaRenderer.isOutputSilentAudio())`). In other word, at least one。

SubJunk:
What is one renderer that should use this feature?

onon765trb:
Can you please explain concrete behavior which you image?

SubJunk:
The line `OutputSilentAudio = true` added to one renderer `.conf` file

onon765trb:
I close this PR, once.

esabol:
Please reopen!

SubJunk:
I agree, I tried to help as much as I could on this so it is a shame to see that work be wasted.
I just don't want to add unused code, so the ways to get this merged are:
1) Make a renderer use the setting (this is what I've been saying here many times), or
2) Remove the renderer setting so it is only a user setting (this would make it less useful though, so I really prefer option 1)

PR #1548, opened 2018-08-31
Add soxr in ffmpegvideo

onon765trb:
Stretching, squeezing, filling and trimming are performed by swr. [ffmpeg doccument about async of swr](https://ffmpeg.org/ffmpeg-resampler.html)

On the other hand, high quality is gotten by soxr of ffmpeg from [Has someone tested the cheby option of soxr lib ?](http://forum.minimserver.com/showthread.php?tid=4181&pid=27185). High end audio users are satisfied by the results.


SubJunk:
This could be good, can you do a before and after comparison so we can see?

onon765trb:
Sorry, I don't have it because I  always use soxr since facing bad samplings in a few years ago

SubJunk:
@onon765trb ok no problem, I was just interested to hear the difference. Do you think this is ready to merge? The code looks good to me

PR #1549, opened 2018-08-31
Add libfdk and audiotoolbox options in aac enoding

onon765trb:
There are discussion about the quality of ffmpeg aac encoder, as shown by [Topic: Personal Listening Test of AAC encoders available from FFmpeg](https://hydrogenaud.io/index.php/topic,111085.0.html) and  [Topic: FFMPEG AAC Native VBR Quality questions](https://hydrogenaud.io/index.php/topic,111996.0.html). This means that high end audio users aren't satisfied by the native aac encoder of ffmpeg.

AVISynth and VapourSynth users wants to use Nero AAC or [qaac](https://sites.google.com/site/qaacpage/). The latter is Apple' engine. Also, AviUtl adopts qaac as a default aac encoder. And the developper of FFmpeg, Carl Eugen, give credit Apple' engine audiotoolbox, as shown by https://ffmpeg.org/pipermail/ffmpeg-devel/2016-March/190487.html.

libfdk can output high quality aac as shown by [Topic: Personal Listening Test of AAC encoders available from FFmpeg](https://hydrogenaud.io/index.php/topic,111085.0.html). And [libfdk is included in Android](https://en.wikipedia.org/wiki/Fraunhofer_FDK_AAC).

libfdk and audiotoolbox can be used by ffmpeg, easily. As a result. I pullreuested this.


src/main/external-resources/UMS.conf
```diff
@@ -866,6 +866,20 @@ ffmpeg_fontconfig =
 # Default: true
 ffmpeg_mencoder_problematic_subtitles =
 
+# Use Libfdk for aac encoding
+# --------------------------------------------
+# Whether FFmpeg should a high quality aac encoder, libfdk. Note: need a build by yourself, due to Licence.
+# to be transcoded.
+# Both engines can handle subtitles, but MEncoder is more stable right now.
+# Default: false
+fffmpeg_libfdk_encoder =
```
SubJunk: Extra f

src/main/external-resources/UMS.conf
```diff
@@ -866,6 +866,20 @@ ffmpeg_fontconfig =
 # Default: true
 ffmpeg_mencoder_problematic_subtitles =
 
+# Use Libfdk for aac encoding
+# --------------------------------------------
+# Whether FFmpeg should a high quality aac encoder, libfdk. Note: need a build by yourself, due to Licence.
+# to be transcoded.
+# Both engines can handle subtitles, but MEncoder is more stable right now.
```
SubJunk: Why is there a note about MEncoder on a setting for FFmpeg? I don't understand what it means

Nadahar:
I thought the reason the Fraunhofer AAC library isn't used in FFmpeg is that the license conflicts with the h264 library, which is GPL. That means that you can only compile FFmpeg with one of them, not both, and normally that would make h264 the priority.

onon765trb:
@Nadahar You are right. FFmpeg including libfdk can't distributed. On the hand, there are users who can build by themselves. I consider them. For users to misunderstanding, I emphasis the licece by ".

SubJunk:
@onon765trb in this case, I think we should make this a config-only addition, instead of adding it to the GUI. We do that with highly customized options like this

onon765trb:
@SubJunk I have understood it.

onon765trb:
I will resolve the conflicts with the master after merging PR #1547 because another conflicts appears.

onon765trb:
I close this PR, once.

PR #1552, opened 2018-09-01
Wake On Lan to Power on The TV

artafakhari:
you only need to send a WOL command to the MAC Address of the TVs saved in a list.


SubJunk:
@artafakhari same thing has gone wrong with this PR as the other one

SubJunk:
@artafakhari I'm closing this since it should be an Issue instead of a Pull Request

Issue #1488, opened 2018-04-16
Retain filename with web interface downloads


SubJunk:
Suggestion from http://www.universalmediaserver.com/forum/viewtopic.php?p=30363&sid=241ed45080481f922ad482c23927db02#p30363

PR #1555, opened 2018-09-01
MINOR adds filename to download from webui

js-kyle:
Hi there,

I've picked up enhancement issue #1488 as my first contribution here. Let me know your feedback!


SubJunk:
This is a really elegant solution, I have never seen the `download` attribute before

Issue #1530, opened 2018-07-17
GPU acceleration on integrated Intel graphics (VAAPI)


goncalossilva:
Hi, and thanks for the awesome software!

I'm using ffmpeg in UMS (which I confirmed to be compiled with VAAPI support) and UMS.conf contains the following: `gpu_acceleration=true`.

To the best of my understanding, this should trigger an ffmpeg call similar to the following the following: `ffmpeg -hwaccel vaapi -hwaccel_device /dev/dri/renderD128 -i (...)`, provided that the device is available (it is). Inspecting `top`, these aren't included.

The end result is that I'm trying to do hardware accelerated transcoding on the integrated intel graphics, which in theory should work, but in the end the GPU has close to 0% utilization while the CPU is pegged at 100%.

At this point I'm questioning if this is supposed to work out of the box, or if I am expected to edit `CustomFFmpegOptions` to add the above lines. Thank you.

mirh:
hwaccel vaapi is for decoding. 

Encoding has [different switches](https://trac.ffmpeg.org/wiki/Hardware/VAAPI#Encoding).

goncalossilva:
I see, thanks for the explanation! I saw your PR in #1560, looking forward to it. :smile:

SubJunk:
This is merged now :)

PR #1560, opened 2018-09-06
Add FFmpeg GPU decoding

onon765trb:
Issues #1530 and the discussion of #1175
UMS can support for FFmpeg GPU decoding by This PR.
This significantly decreases the burden of PC.


SubJunk:
@onon765trb Also I tested it and it didn't work for me by default (I didn't set anything), I expected it to add `-hwaccel auto` to the params but it didn't. I hope I didn't break it! Can you have a look?

mirh:
What is decoding needed in ums?

onon765trb:
@SubJunk Your change is interesting. Do you check GPU acceleration? Your code work in my test.

I have a few on my mind by looking your code. How do you think the change from all variables name GPU Acccelation to GPU Decoding Accelation for the future.

This is because GPU acceleration has decoding and encoding. If using your code taking lists and `ffmpeg  -hide_banner -encoders | grep h264`, can set GPU acceleration encoding easily. Of course, the tuning of encoding set is need and it may take a long time.

At least, I can make the change from GPU Acccelation to GPU Decoding Accelation on my code, easily.

Nadahar:
> What is decoding needed in UMS?

When you *transcode*; that is, convert from one format to another, the existing format must be *decoded* and the new format must be *encoded*.

In most cases, encoding is significantly more resource intensive than decoding. This is by design, as the goal when codecs are designed is usually to make the decoding cost as low as possible, at the expense of the encoding cost. Thus, hardware accelerated encoding is where you're likely to see the biggest "gains", but hardware accelerated decoding also helps by freeing the CPU capacity that would otherwise have been used on decoding.

SubJunk:
@onon765trb that sounds like a good change

onon765trb:
I changed the names. But I did not touch variables reflects on other than ffmpeg.

PR #1567, opened 2018-09-14
DMS engine refactoring

SubJunk:


SubJunk:
Need to fix the engines not showing up, and revert the translation changes

SubJunk:
Next I will remove the connecting cable animation because I think our less experienced users may take it literally. It's a shame because it is a slick animation, but I don't want lots of support threads in the forum asking which cable we are telling them they need

SubJunk:
@valib please review this. It adds some hover and click transitions to icons and cleans up some of the code. I haven't modified it much from DMS and it was merged over there 2 years ago. I have also tested it and it seems to be working well for me.

valib:
@SubJunk sorry but there are still animated button from the DMS like https://github.com/UniversalMediaServer/UniversalMediaServer/blob/79b25a3d5968308f7f84db1714e1fbaac387593f/src/main/java/net/pms/newgui/NavigationShareTab.java#L97 which should be rewritten.

SubJunk:
@valib cool I'll make a new branch to clean it up

PR #1570, opened 2018-09-15
Fix various macOS bugs

SubJunk:


SubJunk:
@valib please review this when you get a chance, we have quite a few users experiencing the file chooser bug

SubJunk:
Fix confirmed on the forum https://www.universalmediaserver.com/forum/viewtopic.php?p=37227

onon765trb:
@SubJunk I did some tests. Minimizing is hidden. Good!

But specifying share folders is not good on wizard. To specify share folders, a few times selecting folders is forced.
Further, home folder `/Users/x` is selected automatically. This is not good on security or the burden of PC.

If sellecting automatically, I think that `/Users/x/Movies`, `/Users/x/Music` or `/Users/x/Pictures` had better been done.

SubJunk:
@onon765trb thanks for your review of this. I'm not sure what you mean by:

> But specifying share folders is not good on wizard. To specify share folders, a few times selecting folders is forced.

Can you please explain?

I agree that we should have better defaults too. Let's make another issue/PR for that

onon765trb:
@SubJunk 

> I made an issue for that at #1572

Good issue!

> Can you please explain?

Select button does not became valid after selecting a folder at the first time.

SubJunk:
Hmm it works for me. Which version of Java?

Issue #1561, opened 2018-09-08
Question: about audio remux


onon765trb:
I am working for audio remux in ffmpeg video to PR. And I have questions.
ll. 346-348 in FFMpegVideo.java
```
boolean isSubtitlesAndTimeseek = !isDisableSubtitles(params) && params.timeseek > 0;

if (configuration.isAudioRemuxAC3() && params.aid != null && params.aid.isAC3() && !avisynth() && renderer.isTranscodeToAC3() && !isSubtitlesAndTimeseek) {
```
Why is `!isDisableSubtitles(params)` needed?
Why is `params.timeseek > 0` needed? I think it is not needed if using a ffmpeg option, `-seek_timestamp 1` and `-muxdelay 0`.

ll. 814-822 in FFMpegVideo.java.
```	
if (
	configuration.isAudioRemuxAC3() &&
	params.aid != null &&
	params.aid.isAC3() &&
	!avisynth() &&
	renderer.isTranscodeToAC3() &&
	!isXboxOneWebVideo &&
	params.aid.getAudioProperties().getNumberOfChannels() <= configuration.getAudioChannelCount()
)
```
Why is `!isXboxOneWebVideo` needed? This is not used in ll. 346-348 in FFMpegVideo.java.

Thank you in advance.

OP-Klaus:
@onon765trb for your first question, we found that in FFmpeg, seeking does not work with the audio `-c:a copy` param. Maybe they have since changed it so it works which would be cool

Second question, it is because we transcode to WMV/WMA for Xbox, so it doesn't support the common audio formats like AC-3 and AAC

SubJunk:
@onon765trb for your first question, we found that in FFmpeg, seeking does not work with the audio `-c:a copy` param. Maybe they have since changed it so it works which would be cool

Second question, it is because we transcode to WMV/WMA for Xbox, so it doesn't support the common audio formats like AC-3 and AAC

onon765trb:
Thank for explanation.
About `!isDisableSubtitles(params)`
Similar reasons?

PR #1571, opened 2018-09-18
Add remux mode of some audio codecs and using LPCM in ffmpegvideo

onon765trb:
Add remux mode of some audio codecs and using LPCM in ffmpegvideo
This PR is related to Issue #1561.

I cannot confirm the problem of sync in seeking or subtitles by a ffmpeg option `-seek_timestamp`.
Its decisionn considers avisynth, XboxOneWebVideo, TsMuxeR for dts, seeking, subtitles, the number of audio channel, the sample rate of audio. Namely, can disable remux mode in seeking or subtitles, etc.
Disable remux by default.


SubJunk:
Looks like there is some syntax error causing it to fail automatic tests

onon765trb:
@SubJunk Thank you for  your pointing out. I will refix.

onon765trb:
@SubJunk Do you think that `lpcmUse` is need in line 552 of FFMpegVideo.java.
`lpcmUse` is a paramater to output LPCM.

SubJunk:
@onon765trb this line here? https://github.com/UniversalMediaServer/UniversalMediaServer/blob/f09760994e0179643a98e64d2b19268b2667be30/src/main/java/net/pms/encoders/FFMpegVideo.java#L552
I don't see how it is related to LPCM

onon765trb:
@SubJunk Oh-. Line 522

SubJunk:
@onon765trb yes that looks like a good change to me. I will review the rest of the changes soon

src/main/resources/i18n/messages.properties
```diff
@@ -500,7 +500,13 @@ TrTab2.26=Keep AC-3 tracks
 TrTab2.27=Use LPCM for audio
 TrTab2.28=Keep DTS tracks
 TrTab2.29=AC-3 re-encoding audio bitrate: (in Kbps) (e.g. 576, 640)
+TrTab2.30=Keep LPCM tracks
```
SubJunk: It would be better if new strings used words instead of numbers like `TrTab2.KeepLPCMTracks`

src/main/external-resources/UMS.conf
```diff
@@ -647,28 +647,55 @@ disable_transcoding =
...
 # Default: true
 audio_remux_ac3 =
 
+# Keep AC-3 tracks, ignoring profiles
+# ----------------
+# Remux AC-3 instead of transcoding it in the FFmpeg video engine
+# even when using a profile not set ac3 in renderer.conf.
+# For this option to work, have to set ForceMuxAC3ToMpeg  = true in renderer.conf.
+# Default: false
+audio_force_remux_ac3 =
```
SubJunk: Why is this setting needed? Isn't it better to make it automatic based on what the renderer supports?

src/main/resources/i18n/messages.properties
onon765trb: I do not mind such a change. Do you do so?

src/main/external-resources/UMS.conf
onon765trb: The behavior you refer is executed by `audio_remux_ac3 = true` in UMS.conf.
Its code is on `ac3Remux = params.aid.isAC3() && configuration.isAudioRemuxAC3() && params.mediaRenderer.isTranscodeToAC3();` in l. 1512 of FFMpegVideo.java.

The behavior of `audio_force_remux_ac3 = true` is to remux ac3 (`audio_remux_ac3 = true` is needed.) even when `TranscodeVideo =  MPEGTS-H264-AAC`. At least, it works well in my tests.

onon765trb: I exclude Profile to wmv,  to be on the safe side,

src/main/resources/i18n/messages.properties
SubJunk: I can do it

src/main/external-resources/UMS.conf
onon765trb: >  Some renderers will allow it while others won't

Yes. You are right. As a result, Both `ac3Remux = params.aid.isAC3()` and `audio_remux_ac3 = true` needed at the same time. This follows the behavior of remux DTS.

SubJunk: Why should a user need to enable it? If they have a renderer that supports it, then it should be used, because it is better to mux than transcode.
I am thinking of the ease for users. When possible, good things should "just work"

onon765trb:
Probably, your premise is that the transstreaming of mp4 or mkv,etc. is always possible by `Supported =`
As shown in the discussion of Issue #1202, most of japanese renderers do not support transstreaming of mp4 or mkv,etc. The most support only ts file or ts and mpegps.
Namely, most of japanese users are forced to reencode video and audio though @Sami32 made many efforts.
This PR is a part to solve it.

I want to confirm some.
In general, "mux" means reencoding and passthrough.
And "transcode" does reencoding.

`audio_force_remux_ac3 = true` realizes passthrough even  when setting `TranscodeVideo = MPEGTS-H264-AAC`. Is this "mux" that you refer?
As a result, change to`TranscodeVideo = MPEGTS-H264-AC3` when enjoying lossless ac3.
Reencoding decreases quality, even PCM, and increases the burdens of PCs and networks.
Most of japanese users are forced these.
This option can avoid them.
And this option can be disabled and is not forced, like remux DTS.

src/main/java/net/pms/newgui/TranscodingTab.java
```diff
@@ -675,16 +681,54 @@ public void itemStateChanged(ItemEvent e) {
...
 			public void itemStateChanged(ItemEvent e) {
 				configuration.setAudioUsePCM(e.getStateChange() == ItemEvent.SELECTED);
+				AudioRemuxLPCM.setEnabled((e.getStateChange() != ItemEvent.SELECTED));
 			}
 		});
-		builder.add(GuiUtil.getPreferredSizeComponent(forcePCM), FormLayoutUtil.flip(cc.xy(1, 4), colSpec, orientation));
+		builder.add(GuiUtil.getPreferredSizeComponent(forcePCM), FormLayoutUtil.flip(cc.xy(1, 8), colSpec, orientation));
+
+		AudioRemuxLPCM = new JCheckBox(Messages.getString("TrTab2.30"), configuration.isAudioRemuxLPCM());
+		AudioRemuxLPCM.setToolTipText(Messages.getString("TrTab2.84") + (Platform.isWindows() ? " " + Messages.getString("TrTab2.21") : "") + "</html>");
```
valib: Using `TrTab2.84` is wrong. Default value for `audio_remux_lpcm =` is `false` but the `TrTab2.84` says `Default: Enabled`. This must be corrected also for other sethings.

SubJunk:
@onon765trb I agree we should mux when possible, we just need to make our settings clear and easy to use. I don't want people to be confused about this. I will have a look at the settings again soon to see if I can think of a way to make them easier to understand, after the comment by @valib has been addressed

onon765trb:
I close this PR, once.

onon765trb:
I close this PR, once.

Issue #1572, opened 2018-09-18
Improve the default shared folders


SubJunk:
We should probably share the media folders by default.
On macOS they are users/x/Movies, users/x/Music, and users/x/Pictures.
Similar on Windows, and will have to research Linux distros

PR #1574, opened 2018-09-19
Use media folders by default on macOS and Windows

SubJunk:
Implements https://github.com/UniversalMediaServer/UniversalMediaServer/issues/1572

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,3 +1,11 @@
+7.4.0 - 2018-09-??
+
+	General:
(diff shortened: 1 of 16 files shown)


SubJunk:
There is a bigger implementation at https://github.com/DigitalMediaServer/DigitalMediaServer/commit/08eb4df8f593e488dd43bde4d77105e27aa44da2 we could use instead

SubJunk:
I have tested this on Windows, macOS and Ubuntu and it works so I'm merging

PR #1579, opened 2018-09-23
Cleanup our implementation of the DMS engine refactoring

SubJunk:


SubJunk:
@valib that should be fixed now

SubJunk:
I threw in a fix for another thing I noticed so the UI works with the startup scan

PR #1580, opened 2018-09-23
Bush Freeview renderer

atamariya:
Adding new renderer as per request here http://www.universalmediaserver.com/forum/viewtopic.php?f=5&p=37268#p37268


SubJunk:
Thanks for this @atamariya 
I might cut the empty lines and comments so it matches the format of our other files before merge

SubJunk:
Also it will need icons. I can do it this week

abarton1962:
Hi guys,
Thanks for doing this.
At the moment, I can add the config file into the UMS.app package until it is added into the renderers directory as par t of the build. Sorry if the file is a bit "verbose", It was my first time doing this and so just keep all the options (whether they were needed are not) in the file.

SubJunk:
@abarton1962 I have done some cleanup now, please let me know if you want more lines added like for images and ts before it merges

SubJunk:
@abarton1962 will you have time to review these changes at some point or should I just merge?

PR #1581, opened 2018-09-23
Docker image

atamariya:
Scripts to create a docker image for ums. Mount following volumes and ports to run the image:

    Media folder
    VOLUME /media

    Profile folder containing UMS.conf
    VOLUME /profile

EXPOSE 9001 5002 1044

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,18 @@
+FROM alpine
+
+MAINTAINER Anand Tamariya <atamariya@gmail.com>
+
+RUN apk add --no-cache ffmpeg mediainfo openjdk7-jre 
+
+WORKDIR /usr/src/ums
+
+VOLUME /media
+
+VOLUME /profile
+
+EXPOSE 9001 5002 1044
+
+CMD java -DUMS_PROFILE=/profile -Dfile.encoding=UTF-8 -Djava.net.preferIPv4Stack=true -Djna.nosys=true -cp ums.jar net.pms.PMS 
+
+COPY . /usr/src/ums
+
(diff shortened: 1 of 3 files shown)


docker/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+
+WORKDIR /usr/src/ums
+
+VOLUME /media
+
+VOLUME /profile
+
+EXPOSE 9001 5002 1044
+
+CMD java -DUMS_PROFILE=/profile -Dfile.encoding=UTF-8 -Djava.net.preferIPv4Stack=true -Djna.nosys=true -cp ums.jar net.pms.PMS
```
SubJunk: Does it need memory instructions like 1280mb?

atamariya: Added JVM_OPTS environment variable for runtime memory setting. However, memory allotted to container will also play an important role. https://github.com/akashche/docs/blob/4325be5ccb52721bfa26521beaee40e607c7b7af/openjdk/content.md

SubJunk: Ok great, thanks. That looks like a fine default. We have historically used 768mb for 32-bit and 1280mb for 64-bit systems. I don't know enough about docker to properly review this PR but I can do some reading about it, or do you think it is just safe to merge now?

atamariya: It's safe to merge.

atamariya:
@SubJunk You should publish a docker image at https://hub.docker.com/ and also add an announcement in forum for better visibility. 
```
cd docker
ant
docker tag ums $DOCKER_ID/ums
docker push $DOCKER_ID/ums
```

Users can pull the image using:
```
docker pull $DOCKER_ID/ums
docker run -it ums
```



```

SubJunk:
Ok, will do

atamariya:
@SubJunk created PR #1599

SubJunk:
@atamariya ok I have verified that the binaries are working now, they were in the wrong place in the image. I've also added instructions to the UMS release script I use so I can't forget to update it now.

PR #1585, opened 2018-09-27
Dms minor stuff

SubJunk:


SubJunk:
@valib  these seem harmless enough so can probably merge?

PR #1586, opened 2018-09-27
Disabled plugins

SubJunk:
DMS 60b77e2a57c480f41275a0f66cca64b46dd53244


SubJunk:
@valib what do you think? I don't think any of our plugins work and we are unlikely to have more made

Nadahar:
@SubJunk Would you care to explain why I'm banned from the forum?

SubJunk:
I can look into it, I didn't ban you

SubJunk:
I can see you're in the group "Registered users" and I tested your permissions, they seemed to work for me https://www.universalmediaserver.com/forum/memberlist.php?mode=viewprofile&u=4025

Nadahar:
@SubJunk Something have magically happened then, because now I can log in again. Thanks for checking.

SubJunk:
These are the only two entries in the logs for your user btw:

image

Might have something to do with the password change

SubJunk:
Yes very strange, maybe when I simulated your permissions on my account it fixed whatever was wrong. Maybe something to do with your role recently changing. I just hope it isn't happening to other users too without us knowing.

SubJunk:
@valib can you please approve or disapprove this? If it is disapproved I will close it. If it is approved it will make getting in sync with DMS easier as our code will conflict less

valib:
@SubJunk seems to me that keep very old not working plugins is useless when original developers doesn't  want to take care about them.

PR #1588, opened 2018-09-29
Dms minor fixes 2

SubJunk:


SubJunk:
There are a bunch of instances of PMS and UMS changing to DMS in here but I don't really care about code branding, do you agree @valib ?

valib:
@SubJunk I have no problem with using code from the DMS without changing descriptions.

SubJunk:
@valib cool thanks, that will make it easier for me. Do you think this needs more testing or can it be merged? Seems minor enough to me

SubJunk:
Also any opinion on removing the plugins code?

valib:
@SubJunk this can be merged. About plugins I don't know. I used only Jumpy but with new Kodi it stopped working. Are there any plugins still working?

SubJunk:
From memory, it was only the MovieInfo plugin that was working, and we have had a few reports on the forum of it not working anymore, so I think it is probably time to retire that part of the code since it will just cause users to waste time and have a bad experience

PR #1591, opened 2018-09-30
Fixed web interface menu display on Safari

### LINKED ISSUE #1351: Web interface doesn't work in Safari on macOS
Reported [here](http://www.universalmediaserver.com/forum/viewtopic.php?t=10377) and confirmed by me. It's a bit of a problem with the new "web interface" button that opens in the default browser, which is Safari by default on macOS. 

This might or might not be related to #977. In either case debugging the web interface really isn't for me.

SubJunk:
Also minor front page fix

Fixes #1351


SubJunk:
Thanks for the testing guys

Issue #1592, opened 2018-09-30
7.4.0 on macOS: "videotoolbox" listed twice in the ffmpeg GPU decoding acceleration method menu


esabol:
MacOS X High Sierra 10.13.6 (17G65)
ffmpeg version N-90990-g652b857b94-tessus

"videotoolbox" is listed twice in the ffmpeg GPU decoding acceleration method menu and in the UMS.conf option `fmpeg_available_fgpu_decoding_acceleration_methods`. This seems odd. The menu options listed are "auto", "videotoolbox", and "videotoolbox".

Also, it seems like a typo that the UMS.conf option is named `fmpeg_available_fgpu_decoding_acceleration_methods`. Shouldn't it be `ffmpeg_available_fgpu_decoding_acceleration_methods`?

I guess you can blame ffmpeg for the duplication since the output lists it twice (see below), but perhaps UMS should ignore duplicates when parsing the output.
```
% "/Users/xxxxx/Downloads/Universal Media Server.app/Contents/Resources/osx/ffmpeg" -hwaccels
ffmpeg version N-90990-g652b857b94-tessus Copyright (c) 2000-2018 the FFmpeg developers
  built with Apple LLVM version 9.1.0 (clang-902.0.39.1)
  configuration: --cc=/usr/bin/clang --prefix=/opt/ffmpeg --extra-version=tessus --enable-avisynth --enable-fontconfig --enable-gpl --enable-libaom --enable-libass --enable-libbluray --enable-libfreetype --enable-libgsm --enable-libmodplug --enable-libmp3lame --enable-libopencore-amrnb --enable-libopencore-amrwb --enable-libopus --enable-libsnappy --enable-libsoxr --enable-libspeex --enable-libtheora --enable-libvidstab --enable-libvo-amrwbenc --enable-libvorbis --enable-libvpx --enable-libwavpack --enable-libx264 --enable-libx265 --enable-libxavs --enable-libxvid --enable-libzimg --enable-libzmq --enable-libzvbi --enable-version3 --pkg-config-flags=--static --disable-ffplay
  libavutil      56. 18.100 / 56. 18.100
  libavcodec     58. 19.101 / 58. 19.101
  libavformat    58. 13.101 / 58. 13.101
  libavdevice    58.  4.100 / 58.  4.100
  libavfilter     7. 21.100 /  7. 21.100
  libswscale      5.  2.100 /  5.  2.100
[...]

PR #1594, opened 2018-09-30
Minor fix to avoid duplicating GPU acceleration methods and fix typo.

valib:
Reported in #1592


esabol:
Wow, that was fast, @valib ! Thanks!

PR #1595, opened 2018-09-30
Update the FFmpeg setting GUI

valib:


SubJunk:
I have tested this and it looks good

PR #1599, opened 2018-10-03
Docker image using "with-dependencies.jar"

atamariya:


SubJunk:
Looks good to me, is it ready to merge?

SubJunk:
I pushed it to docker as `universalmediaserver/ums`, it doesn't have any of the binaries though like FFmpeg, how do we get that in there?

atamariya:
@SubJunk all the binaries (linux) required can be copied to the docker folder before building.

SubJunk:
That's done now, and pushed

PR #1600, opened 2018-10-03
Dms sync

SubJunk:
Another one of these. This removes the tooltip links, how do you feel about that @valib since they were yours I think?


valib:
@SubJunk no problem with that.

PR #1606, opened 2018-10-05
Added more web samples and removed broken ones

SubJunk:


SubJunk:
This needs to be refactored into the new `8.0.0` way, then it will merge into that branch

PR #1621, opened 2018-10-16
Enhanced PS4 support

fgimenezm:
Adding supported image formats and SRT subtitles.
TIP: Convert subs to UTF8 if you have accented characters.


SubJunk:
@fgimenezm thanks!

PR #1623, opened 2018-10-17
Dms sync3

SubJunk:


SubJunk:
@valib I have done general testing of this and can't see any obvious bugs. The window position is remembered correctly. If you are happy with it I think we can merge

mzji:
The whitelist function is removed by this PR but document (including Github Wiki or comments in the `UMS.conf` file) related with this function has no change... Could any project maintainer explain that why this function is removed and do we have any plan to restore or reimplement this function? If no such plan exists, which function should be used as alternative to hide folders from some specific renderers?

SubJunk:
@mzji yes we should restore that functionality

mzji:
@SubJunk Thanks for your rapid reply! That's a good message for me.

PR #1625, opened 2018-10-18
Removed dependency on plexus StringUtils

SubJunk:
This fixes CVE-2018-12418


SubJunk:
@valib this one seems like it can probably go in without causing problems, what do you think?

Issue #1281, opened 2017-05-19
Build profile that doesn't download binaries


Nadahar:
Travis builds spend a lot of time and bandwidth on downloading binaries like FFmpeg, DCraw etc for every build. This also uses a lot of unnecessary bandwidth from your file host @SubJunk. I can't see that this is necessary to run the tests, and I think we should make a build profile that doesn't download these.

SubJunk:
Sure, that sounds good, they're not needed for tests currently

atamariya:
This #1371 should help.

Nadahar:
Probably won't happen.

PR #1626, opened 2018-10-19
Skip downloading binaries on tests

SubJunk:
Fixed https://github.com/UniversalMediaServer/UniversalMediaServer/issues/1281


SubJunk:
This seems to be working and shaves some time off our test suite as well as making it more resilient

PR #1630, opened 2018-10-20
Better structured filename re-writing testcase

stasinos:
This patch factors the expected behaviour of the filename prettifier out
of the testcase code and puts it into a JSON resource. The testcase
code goes over the test definitions in the JSON and executes them.


pom.xml
```diff
@@ -579,11 +579,12 @@
 				<!-- Enable the use of Maven properties -->
 				<filtering>true</filtering>
 			</testResource>
-			<!--*.srt files to test charset detection-->
+			<!--*.srt files to test charset detection and *josn to test filename prettification-->
```
SubJunk:
```suggestion
			<!--*.srt files to test charset detection and *json to test filename prettification-->
```

src/test/java/net/pms/util/FileUtilTest.java
```diff
@@ -38,6 +42,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+
```
SubJunk:
```suggestion
```

src/test/java/net/pms/util/FileUtilTest.java
```diff
@@ -91,52 +96,31 @@ public void testGetFileNameWithoutExtension() throws Exception {
...
-		// Video of a special edition of a movie
-		assertThat(FileUtil.getFileNamePrettified("Universal.Media.Server.Special.Edition.2015.720p.mkv")).isEqualTo("Universal Media Server (2015) (Special Edition)");
-		assertThat(FileUtil.getFileNamePrettified("Universal.Media.Server.2015.Special.Edition.720p.mkv")).isEqualTo("Universal Media Server (2015) (Special Edition)");
-
-		// Video of a sample of a movie
-		assertThat(FileUtil.getFileNamePrettified("Universal.Media.Server.2015.720p.Sample.mkv")).isEqualTo("Universal Media Server (2015) (Sample)");
+		JsonParser parser = new JsonParser();
+		
+		try {
+	        JsonElement tree = parser.parse(
```
SubJunk: Please indent inside the `try`

src/test/resources/net/pms/util/prettified_filenames_metadata.json
```diff
@@ -0,0 +1,223 @@
+[
+	{
+		"comment": "Video of a TV episode",
+		"filename": "Universal.Media.Server.S01E02.720p.mkv",
+		"prettified": "Universal Media Server - 102",
+		"metadata": {
+			"type": "tv-series-episodes",
+			"series": "Universal Media Server",
+			"season": 1,
+			"episodes": [2]
```
SubJunk: These metadata fields don't seem to be used, are these for a future improvement?

stasinos: Yes, as I noticed that at some other place in the code metadata is guessed from filename to double-check the downloaded info.

stasinos:
Please see also the catch clause in testGetFileNameWithRewriting()
Is there a logging convention for internal errors that should never happen in a well-installed UMS, such the JSON file missing? Should UMS just throw an exception or should the error be logged and "swallowed"?

src/test/resources/net/pms/util/prettified_filenames_metadata.json
SubJunk: @stasinos it would be cool to grow this unit test into more like an integration test so it can use this metadata to verify the database data was written correctly - not in this PR but at some point in the future. As our automated testing matures I'd like to see more of that across the board.

SubJunk:
@stasinos about the logging convention, I think we almost always want to swallow and log errors, as long as we keep the stacktrace. If you're unsure we can decide case-by-case, or if you think it should be another way that can be discussed too.

We like to discuss things so we arrive at the best approach instead of being too prescriptive with it.

I'm going to merge this branch now. Thanks for the extra test cases too :)

Nadahar:
If the exception is in a test, I'm not sure the usual "log and handle" approach is ideal. If the test file is missing, the test should either skip or fail IMO. Skipping might be the most correct option.

Skipping a test can be done like this: https://github.com/UniversalMediaServer/UniversalMediaServer/blob/d39d7325120bb40e27015c89ae2566013617dbcd/src/test/java/net/pms/util/jna/macos/corefoundation/CoreFoundationTest.java#L52-L60

PR #1632, opened 2018-10-21
Tests for metadata extraction from filename

stasinos:
Adds a new test function that checks the metadata
extracted from the filename.
Amends one of the testcases so that they correctly
reflect the current working of the extraction code.

### DIFF
--- a/src/test/java/net/pms/util/FileUtilTest.java
+++ b/src/test/java/net/pms/util/FileUtilTest.java
@@ -120,6 +120,141 @@ public class FileUtilTest {
 	}		
 
 
+	/**
+	 * Note: The method this is testing handles numerous inputs, so this test
+	 * could get very large. It should get much larger than it is now.
+	 *
+	 * @throws java.lang.Exception
+	 */
+	@Test
+	public void testGetFileNameMetadata() throws Exception {
+		JsonParser parser = new JsonParser();
+		
+		try {
+			JsonElement tree = parser.parse(
+					new java.io.FileReader(
+							FileUtils.toFile(
+									CLASS.getResource("prettified_filenames_metadata.json")
+									)
+							)
+					);
+
+	        JsonArray tests = tree.getAsJsonArray();
+	        for( JsonElement test: tests ) {
+	        	JsonObject o = test.getAsJsonObject();
+	        	String original = o.get("filename").getAsString();
+	        	JsonObject metadata = o.get("metadata").getAsJsonObject();
+
+	        	String[] extracted_metadata = FileUtil.getFileNameMetadata(original);
+	        	assert extracted_metadata.length == 6;
+	        	String movieOrShowName = extracted_metadata[0];
+	        	int year = -1;
+	        	try {
+	        		if(extracted_metadata[1]!=null) {
+	        			year = Integer.parseInt(extracted_metadata[1]);
+	        		}
+	        	}
+	        	catch(NumberFormatException ex) {
+        			throw(new AssertionError(ex)); 
+        		}
+	        	String extraInformation = extracted_metadata[2];
+	        	int tvSeason = -1;
+	        	try {
+	        		if(extracted_metadata[3]!=null) {
+	        			tvSeason = Integer.parseInt(extracted_metadata[3]);
+	        		}
+	        	}
+	        	catch(NumberFormatException ex) {
+        			throw(new AssertionError(ex)); 
+        		}
+	        	String tvEpisodeNumber = extracted_metadata[4];
+	        	String tvEpisodeName = extracted_metadata[5];
+
+        		JsonElement elem; 
+        		
+        		elem = metadata.get("extra");
+        		if( elem != null ) {
+        			for( JsonElement extra: elem.getAsJsonArray() ) {
+        				assertThat( extraInformation.indexOf(extra.getAsString()) > -1 );
+        			}
+        		}
+	        	if( "tv-series-episode".equals(metadata.get("type").getAsString()) ) {
+	        		// A single episode, might have episode title and date
+	        		elem = metadata.get("series");
+	        		if( elem != null ) {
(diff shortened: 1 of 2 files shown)


src/test/resources/net/pms/util/prettified_filenames_metadata.json
```diff
@@ -178,6 +193,7 @@
 		}
 	},
 	{
+		"id": 16,
```
SubJunk: Why do these need IDs when the filename is already unique?

SubJunk:
Cool!

src/test/resources/net/pms/util/prettified_filenames_metadata.json
SubJunk: That's a good thought to anticipate that. I'm not sure we would need a manual unique key in that case either. If we need an integer we can get the items' position in the JSON automatically, or just concatenate two property values to make a unique one (like `context + filename`).
I am just not a fan of manually incrementing numbers :)

src/test/resources/net/pms/util/prettified_filenames_metadata.json
```diff
@@ -55,15 +55,14 @@
 		}
 	},
 	{
-		"comment": "Video of an extended TV episode",
+		"comment": "Video of a TV episode, noting that EXTENDED is the group and not the edition",
```
SubJunk: Is this comment correct? I think the group is usually like `Universal.Media.Server.S01E02.720p-EXTENDED.mkv`

SubJunk:
@stasinos thanks, it looks good, I just added one question before merge

src/test/resources/net/pms/util/prettified_filenames_metadata.json
stasinos: That was my original thought, but then the extraction code is not working properly because EXTENDED is not extracted, it matches COMMON_FILE_ENDS_CASE_SENSITIVE and is removed in line 570.

This is what I mean with "Amends one of the testcases so that they correctly
reflect the current working of the extraction code" in the commit message.

SubJunk: Yeah I think you are right that it is a bug in the code in that case. I have identified other bugs while writing tests before too, it is one of the many great things about writing tests :)
Maybe we should remove that assertion from the file, and create a new issue for the bug?

stasinos: How about marking some of the items in the JSON as "todo" or s'thing similar? The test code will not assert those, but log success or failure. It will be useful while toying with the regexps to be sure that nothing that used to work broke, while also getting reported what didn't work and works now.

stasinos: I commited so you can see where this is going. It seems to be working, but it's not thoroughly tested yet.

stasinos:
Some corrections in both the testing code and the actual test cases. It looks solid now.

stasinos:
About the EXTENDED group vs. edition confusion, are we sure that that testcase comes from really seen filenames? Where an Extended edition was in all caps, and so it matched the group name?

SubJunk:
@stasinos I can't remember, I can search my media library for that soon to check if I have an example that matches it

src/test/java/net/pms/util/FileUtilTest.java
```diff
@@ -120,6 +120,238 @@ public void testGetFileNameWithRewriting() throws Exception {
...
+        					if(todo) { logger.warn( "testGetFileNameMetadata/season would fail for TODO test " + original ); }
+        					else { throw(err); }
+        				}
+	        		}
+	        		elem = metadata.get("episode");
+	        		if( elem != null ) {
+	    	        	try {
+	        				assertThat(tvEpisodeNumber).isEqualTo(String.format("%02d",elem.getAsInt()));
+	    	        	}
+	    	        	catch(NumberFormatException ex) {
```
SubJunk: These catches can be chained to make the code shorter, e.g.

```
catch(NumberFormatException | NullPointerException ex) {
```

src/test/java/net/pms/util/FileUtilTest.java
```diff
@@ -120,6 +120,238 @@ public void testGetFileNameWithRewriting() throws Exception {
...
+	        		}
+	        	}
+	        	else if( "movie".equals(metadata.get("type").getAsString()) ) {
+	        		logger.debug( "Doing movie " + original );
+	        		elem = metadata.get("title");
+	        		if( elem != null ) {
+	        			try  {
+	        				assertThat(movieOrShowName).isEqualTo(elem.getAsString());
+	        			}
+        				catch(NullPointerException ex) {
```
SubJunk: There is quite a lot in this file that doesn't match our codestyle but I can fix that after we merge

stasinos: Is the codestyle explained somewhere? I can try to fix if you have a pointer.

src/test/java/net/pms/util/FileUtilTest.java
stasinos: Nifty, didn't know this idiom.

src/test/java/net/pms/util/FileUtilTest.java
SubJunk: It's not something that we enforce because I really want to encourage contributions, I don't want to turn people away if they write a bit differently :) But if you are interested we basically follow the way Google recommends (or rather, they have started recommending it the way we already were doing it :)) https://google.github.io/styleguide/javaguide.html#s4.1.2-blocks-k-r-style

SubJunk:
By the way if you use NetBeans you can just select the contents of the file and go ctrl+shift+f (keys depending on operating system) and it will format the file in a similar way to what that website says :)

stasinos:
I think I caught all catches and elses that break line. Couldn't spot any other place where the Google style book is not followed.

PR #1634, opened 2018-10-21
Another DMS sync

SubJunk:


SubJunk:
@valib I had a chance to go through these changes and they look minor enough to me, it can probably be merged if you agree

PR #1636, opened 2018-10-22
Overhauled transcoding engine registration and GUI display

SubJunk:
This depends upon `DMS-Sync4` and again had a lot of conflicts so needs review and testing. I wanted to split it out to avoid `DMS-Sync4` getting too big


SubJunk:
Closed this because a lot of the conflicts it had are already resolved in https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1644 so let's just do that one instead

PR #1638, opened 2018-10-22
Replace all inefficient use of keySet iterator by entrySet iterator

valib:


SubJunk:
@valib is the difference in performance noticeable? Because readability is also a factor to consider :)

valib:
@SubJunk I am not sure if it is noticeable but the `keySet()` should be used only when to check if the key in the Map exists and the value is not important. The `entrySet()` should be used when also value of the Map is needed. See for example https://stackoverflow.com/questions/3870064/performance-considerations-for-keyset-and-entryset-of-map

SubJunk:
@valib the only thing is that the new code seems less readable to me, but performance is nice too so if you think this is worth merging for the performance we can do that. Do you think it is worth the trade off?

SubJunk:
@valib that looks better, good idea

SubJunk:
Looks good, does it need testing or have you tested it?

valib:
@SubJunk of course I tested it without any problems but who knows.

SubJunk:
@valib cool well we can merge if you want?

src/main/java/net/pms/configuration/RendererConfiguration.java
```diff
@@ -561,11 +561,12 @@ public boolean associateIP(InetAddress sa) {
...
 
 	public static void calculateAllSpeeds() {
-		for (InetAddress sa : addressAssociation.keySet()) {
+		for (Entry<InetAddress, RendererConfiguration> key : addressAssociation.entrySet()) {
+			InetAddress sa = key.getKey();
 			if (sa.isLoopbackAddress() || sa.isAnyLocalAddress()) {
 				continue;
 			}
-			RendererConfiguration r = addressAssociation.get(sa);
+			RendererConfiguration r = key.getValue();
```
Nadahar: I'd say it's very confusing that the entry is called "key"

SubJunk: That's a good point, it would be better to use more customized terms

valib: Thanks for comments. Done

SubJunk:
We should either close this or merge it after fixing conflicts. If you want to fix the conflicts @valib I'll merge

valib:
@SubJunk now it could be merged.

PR #1657, opened 2018-11-06
Handling the "Album Artist" field present in the audio files. 

maciekberry:
Album Artist is a field used in the music files to show the "main" artist of an album. It is quite useful for albums like "Apocalyptica makes an album themselves but for 2-3 songs they invite some guests" (in that case, the Album Artist is "Apocalyptica" while individual songs will have different "Artist" field).

There is an excellent guide (not mine :) )about the purpose of this field: https://www.blisshq.com/music-library-management-blog/2010/10/12/how-to-use-album_artist/

I have a fairly large music library and lack of this feature is making it very painful to use (the same album is present several times, each time with one song). So please find attached my first try at UMS (and at pull requests in general, so do not shoot if something is wrong).

Features implemented:
- new database field
- correctly reading it from mediainfo and saving in the database. Please note that if a song does not have an Album Artist, an SQL null and not empty string is used (see below).
- changing all SQLs and displaying the new field in the DLNA. Because I force to use null and not an empty string, I used COALESCE function and speed up things a bit.


src/main/java/net/pms/dlna/DLNAMediaAudio.java
```diff
@@ -607,6 +612,9 @@ public String toString() {
 		if (isNotBlank(getAlbum())) {
 			result.append(", Album: ").append(getAlbum());
 		}
+		if (isNotBlank(getAlbumArtist())) {
+			result.append(", Album Album: ").append(getAlbumArtist());
```
SubJunk: Just a typo here
```suggestion
			result.append(", Album Artist: ").append(getAlbumArtist());
```

src/main/java/net/pms/dlna/LibMediaInfoParser.java
```diff
@@ -231,6 +231,11 @@ public synchronized static void parse(DLNAMediaInfo media, InputFile inputFile,
 
 						currentAudioTrack.setAlbum(MI.Get(general, 0, "Album"));
 						currentAudioTrack.setArtist(MI.Get(general, 0, "Performer"));
+						
+						String albumperformer = MI.Get(general, 0, "Album/Performer");
+						currentAudioTrack.setAlbumArtist(albumperformer);
+						//LOGGER.trace("Album " + MI.Get(general, 0, "Album") + " - artist " + albumperformer );
```
SubJunk:
```suggestion
```

src/main/java/net/pms/dlna/DLNAMediaAudio.java
maciekberry: Of course, no problem with that!

src/main/java/net/pms/dlna/LibMediaInfoParser.java
maciekberry: Of course, no problem with that!

src/main/java/net/pms/dlna/DLNAMediaAudio.java
SubJunk: I think you can choose to apply my suggestions from within the browser, since I don't have permission to do it :)

SubJunk:
@maciekberry This looks good to me and I am interested in merging this before our next official release, to save our users having to reinitialize their database twice.

@valib what do you think about this change? Are you ok if I approve and merge?

SubJunk:
@maciekberry congratulations on your first merged code to UMS! You will see credit in our changelog in the next release too

Issue #1659, opened 2018-11-08
Can't add to playlist via web interface


SubJunk:
@acanive the recent work you did has caused the "Add to playlist" button on the right of media titles to have a flickering effect as it switches between two links, can you please look into it?

acanive:
solved in https://github.com/acanive/UniversalMediaServer/commit/887023a487bf41f62749a53e4964b498fb46334c

SubJunk:
@acanive thanks for the fast work, will test soon

PR #1660, opened 2018-11-08
Fix to some web interface navigation issues

acanive:
Fix to some web interface navigation issues.
Changed close button in virtual keyboard and added navigation to it.
Fix to Can't add to playlist via web interface #1659


SubJunk:
@valib ok well maybe we should revert it since that is a serious problem... it is possible that I made a mistake when I cherrypicked the commit

valib:
@SubJunk it is related only to the web interface. On my TV it works without any problems.

Issue #1639, opened 2018-10-23
Differentiate between requests from TV and app on TV


SubJunk:
It appears from this users' logs that even though the play request came from the Kodi app on the TV, we still treated it as if it was the TV's native DLNA handling, which led to a supported file format mixup https://www.universalmediaserver.com/forum/viewtopic.php?f=9&t=12914

DiksonH:
FYI, If kodi is active while I turn off the Sony TV (standby mode), the renderer becomes KD-43X7500 (using xbmc.conf or chromecast ultra.conf).

Nadahar:
This is nothing new, I pointed this out back in 2015. It's no "simple bug", it has to do with how UMS identifies renderers. If an identification has been made, that is cached and liked to the IP address. That means that all devices with multiple renderers on the same IP address will have an unstable and unreliable detection. You can experience the same problem simply by running multiple software renderers on a computer.

SubJunk:
> The reason is that while a renderer identifies itself properly during discovery/initial handshake, it does not identify itself when it later makes browse or play requests.

In this case the renderer does seem to identify itself when it makes the play request so I think we can do something about it

PR #1662, opened 2018-11-10
Enable Samsung built-in resume functionality

### LINKED ISSUE #665: Support for Samsung TV C5000's built-in Resume function
The TV's built-in media player has the function of resuming a previously stopped video using the remote's "D" button. I know that UMS already has a similar functionality but it transcodes the video on the fly and there are some limitations to it - such as you cannot FF or rewind and it's not as precise as the built-in option.
It seems like the same issue is already discussed here: http://forum.serviio.org/viewtopic.php?p=6504&sid=308fb26b61282ae7fa65fcd191ad4e7a#p6504
I can supply you with a wireshark dump if necessary. Thank you.

drakulis:
Solution taken from: http://www.serviio.org/forum/viewtopic.php?f=3&t=272&start=0#p6504

Closes #665

### DIFF
--- a/src/main/java/net/pms/database/TableFilesStatus.java
+++ b/src/main/java/net/pms/database/TableFilesStatus.java
@@ -61,7 +61,7 @@ public final class TableFilesStatus extends Tables {
 	 * definition. Table upgrade SQL must also be added to
 	 * {@link #upgradeTable()}
 	 */
-	private static final int TABLE_VERSION = 8;
+	private static final int TABLE_VERSION = 9;
 
 	// No instantiation
 	private TableFilesStatus() {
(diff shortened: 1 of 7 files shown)


SubJunk:
@drakulis this is a very exciting addition, thanks! I will review it tonight

src/main/java/net/pms/network/RequestV2.java
```diff
@@ -895,6 +911,8 @@ public ChannelFuture answer(
 				response.append(CRLF);
 				response.append(HTTPXMLHelper.SOAP_ENCODING_FOOTER);
 				response.append(CRLF);
+			} else {
+				LOGGER.info("Unsupported action received: " + content);
```
SubJunk:
```suggestion
				LOGGER.debug("Unsupported action received: " + content);
```

src/main/java/net/pms/network/RequestV2.java
```diff
@@ -31,12 +33,18 @@
 import java.util.Map.Entry;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+
```
SubJunk:
```suggestion
```

src/main/java/net/pms/database/TableFilesStatus.java
```diff
@@ -237,6 +237,76 @@ public static Boolean isFullyPlayed(final String fullPathToFile) {
...
+				try (ResultSet resultSet = statement.executeQuery(query)) {
+					if (resultSet.next()) {
+						result = resultSet.getInt("BOOKMARK");
+					}
+				}
+			} finally {
+				TABLE_LOCK.readLock().unlock();
+			}
+		} catch (SQLException e) {
+			LOGGER.error("Database error while looking up file status in " + TABLE_NAME + " for \"{}\": {}", fullPathToFile, e.getMessage());
```
SubJunk:
```suggestion
			LOGGER.error("Database error while looking up file bookmark in " + TABLE_NAME + " for \"{}\": {}", fullPathToFile, e.getMessage());
```

SubJunk:
@drakulis also just a note, I don't have a Samsung device to test on, so I am trusting you that this works :)

src/main/java/net/pms/database/TableFilesStatus.java
```diff
@@ -371,6 +441,12 @@ private static void upgradeTable(final Connection connection, final int currentV
 
 						version = 8;
 						break;
+					case 8:
+						try (Statement statement = connection.createStatement()) {
+							statement.execute("ALTER TABLE " + TABLE_NAME + " ADD BOOKMARK INTEGER DEFAULT 0");
```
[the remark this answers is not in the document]
drakulis: Ups, overlooked. Done.

src/main/java/net/pms/database/TableFilesStatus.java
drakulis: done

src/main/java/net/pms/network/RequestV2.java
drakulis: removed empty line.

src/main/java/net/pms/network/RequestV2.java
drakulis: done

drakulis:
@SubJunk I've just pushed changes regarding your comments.

drakulis:
I've implemented it using my Samsung TV (D-serie). Here is some comparison how it was before and after my change
ums-samsung-resume

SubJunk:
@drakulis thanks for your submission! We will have a lot of happy users I'm sure.
I wonder if there is a related optimisation we can do where we hide the "Resume" files that UMS creates if the device is Samsung, since they are redundant now. Do you think that would be good?

drakulis:
Well... I would leave this option. Its an option so could be disabled if user doesn't want it.
This option may be usable eg. on Samsung Android devices like tablets where some third party app is used. UMS will recognize such device as a Samsung device (which in fact it is) but DLNA app may not support Samsung bookmarking. What I wanted to say is already described in following issue #1639

SubJunk:
Good point. Another optimization of this work would be to make this a renderer setting instead of using `isSamsung()`

Issue #1664, opened 2018-11-11
Refactor "RequestV2" class


drakulis:
**Current state**
`RequestV2` class is insanely long and its main method (`answer`) has ~1000 lines of code with massive amount of `if / else` statements. It is highly unreadable and hard to debug such method

**Todo**
- extract `if / else` blocks into separate methods
- create common method for creation of response payload to avoid adding `XML_HEADER`, `SOAP_ENCODING_HEADER` and `SOAP_ENCODING_FOOTER` over and over again

**Goal**
Improve readability and debugging of the class.

@SubJunk I can take care of it. Are you interested with such contribution? If not just close this issue.

valib:
@drakulis it seems promising. Just do it. 👍

valib:
@SubJunk in the end I think that is time to decide if the UMS will use the old implementation or the Netty which is set as default and not confuse users with those two options.

SubJunk:
@drakulis yes this is a good idea

@valib yes let's at least remove the old one from the GUI, if not remove it entirely

drakulis:
OK, so refactoring such big class will be a challenge and it will be very easy to broke something during this process. Additionally reviewing the changes in PR would be a nightmare.

So I will do it in few steps and create PR when I finish each of them. Currently I see following:
- extract common `createResponse` method to avoid adding XML_HEADER, SOAP_ENCODING_HEADER and SOAP_ENCODING_FOOTER over and over again
- extract `searchHandler ` method
- extract other easy to extract handlers like `imageResourceHandler`, `mediaReceiverRegistrarHandler`, `getSystemUpdateIdHandler` and others
- extract `fileHandler` (movie, music etc). This one will be tricky.
- ...
- extract http method handlers - GET / POST / ...

SubJunk:
Yes I really like having small pull requests for reviews, it keeps it moving along much faster and safer

SubJunk:
@drakulis are you just taking a break or is this abandoned? I'm happy to leave it open if you're taking a break, no rush :)

drakulis:
@SubJunk just a break. I have tough project right now and I have to focus on it completely.

SubJunk:
Cool, no problem. See you when you get back :)

PR #1667, opened 2018-11-12
refactor(requestv2): extract "createResponse" method

drakulis:
First change regarding #1664

- [ ] extract `searchHandler` method
- [ ] extract other easy to extract handlers like `imageResourceHandler`, `mediaReceiverRegistrarHandler`, `getSystemUpdateIdHandler` and others
- [ ] ...
- [ ] extract http method handlers - GET / POST / ... or other groups of requests (eg. `ContentDirectoryHandler`, `ConnectionManagerHandler` etc)

### DIFF
--- a/src/main/java/net/pms/network/RequestV2.java
+++ b/src/main/java/net/pms/network/RequestV2.java
@@ -674,11 +674,9 @@ public class RequestV2 extends HTTPResource {
 
 			StringBuilder payload = new StringBuilder();
 			if (soapaction != null && soapaction.contains("IsAuthorized")) {
-				payload.append(HTTPXMLHelper.XBOX_360_2);
-				payload.append(CRLF);
+				payload.append(HTTPXMLHelper.XBOX_360_2).append(CRLF);
 			} else if (soapaction != null && soapaction.contains("IsValidated")) {
-				payload.append(HTTPXMLHelper.XBOX_360_1);
-				payload.append(CRLF);
+				payload.append(HTTPXMLHelper.XBOX_360_1).append(CRLF);
 			}
 
 			response.append(createResponse(payload.toString()));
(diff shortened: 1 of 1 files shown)


src/main/java/net/pms/network/RequestV2.java
```diff
@@ -671,79 +671,43 @@ public ChannelFuture answer(
...
-				response.append(CRLF);
-				response.append("<Id>").append(DLNAResource.getSystemUpdateId()).append("</Id>");
-				response.append(CRLF);
-				response.append(HTTPXMLHelper.GETSYSTEMUPDATEID_FOOTER);
-				response.append(CRLF);
-				response.append(HTTPXMLHelper.SOAP_ENCODING_FOOTER);
-				response.append(CRLF);
+				StringBuilder payload = new StringBuilder();
+				payload.append(HTTPXMLHelper.GETSYSTEMUPDATEID_HEADER);
+				payload.append(CRLF);
```
SubJunk: @drakulis It could be more readable to chain the `CRLF`s too since they take up a lot of space, e.g. `payload.append(HTTPXMLHelper.GETSYSTEMUPDATEID_HEADER).append(CRLF);` what do you think?
This branch is already an improvement anyway

drakulis: Good idea. Done.

If you ask me I would completely remove those `CRLF`. Sure, they may provide more readable string during debug but overall they blur the code. When I finish refactoring this class I will try to implement request / response objects as a Java object to get rid of playing around with concatenation of the strings. I will use JAXB to serialize / deserialize payloads like I already done in SamsungBookmark.

@SubJunk If you are fine with current changes please merge this PR. I will take care of next step from the list then.

src/main/java/net/pms/network/RequestV2.java
```diff
@@ -671,79 +671,38 @@ public ChannelFuture answer(
...
-				response.append("<Id>").append(DLNAResource.getSystemUpdateId()).append("</Id>");
-				response.append(CRLF);
-				response.append(HTTPXMLHelper.GETSYSTEMUPDATEID_FOOTER);
-				response.append(CRLF);
-				response.append(HTTPXMLHelper.SOAP_ENCODING_FOOTER);
-				response.append(CRLF);
+				StringBuilder payload = new StringBuilder();
+				payload.append(HTTPXMLHelper.GETSYSTEMUPDATEID_HEADER).append(CRLF);
+				payload.append("<Id>").append(DLNAResource.getSystemUpdateId()).append("</Id>").append(CRLF);
+				payload.append(HTTPXMLHelper.GETSYSTEMUPDATEID_FOOTER).append(CRLF);
```
SubJunk: @drakulis I think this would create a double `CRLF` when it has gone through `createResponse`, is that ok?

src/main/java/net/pms/network/RequestV2.java
SubJunk: There may be some requirement in DLNA with line-breaks? I would have to look it up. Either way it would be good to keep this a like-for-like refactoring, and make a separate issue for removing line-breaks afterwards, if we are focusing on keeping the PRs/branches separated for fast review. I'm happy to merge this if my last comment can be addressed (the double CRLF)

drakulis: DLAN protocol is based on SOAP. So payload is a regular XML. Line breaks and other white characters have no impact on it. If you take a look on payloads submitted from the devices to UMS you will notice that in mosts cases payload is a one-line string.

src/main/java/net/pms/network/RequestV2.java
drakulis: @SubJunk Yes it may happen. Thats why I don't like those `CRLF`. But it no harm to protocol or application. Nomatter what I removed this CRLF

src/main/java/net/pms/network/RequestV2.java
Nadahar: @SubJunk It's like @drakulis says, the CRLF's have no consequence and have either been added for human readability/debugging or lack of knowledge.

SubJunk: Cool it's good to be sure of that :)

PR #1670, opened 2018-11-14
 refactor(requestv2): extract "searchHandler"

drakulis:
Second change regarding #1664.

The handler itself is still huge and need further refactor but it is out of scope of this task. It will be done when we start splitting `RequestV2` into smaller classes

- [ ] extract other easy to extract handlers like `imageResourceHandler`, `mediaReceiverRegistrarHandler`, `getSystemUpdateIdHandler` and others
- [ ] ...


src/main/java/net/pms/network/RequestV2.java
```diff
@@ -703,173 +701,11 @@ public ChannelFuture answer(
...
-				if (soapaction.contains("ContentDirectory:1#Search")) {
-					response.append(HTTPXMLHelper.SEARCHRESPONSE_FOOTER);
-				} else {
-					response.append(HTTPXMLHelper.BROWSERESPONSE_FOOTER);
-				}
-				response.append(CRLF);
-				response.append(HTTPXMLHelper.SOAP_ENCODING_FOOTER);
-				response.append(CRLF);
-			} else {
+            } else if (soapaction != null && soapaction.contains("ContentDirectory:1#Browse")) {
```
SubJunk: It seems that all of the lines that have been changed (in all of the files) have had their tab-indentation replaced by space-indentation, probably just an automatic IDE thing, please put it back how it was :)

drakulis: Done.

BTW. Current code formatter seems to be outdated. When I've executed it, code has been changed drastically

SubJunk: @drakulis yeah we should look into that.. I didn't have anything to do with that formatter, I have mine set locally in NetBeans config

drakulis: I'm using IntelliJ. I just googled a bit about that and it looks that if we want to share formatter across Eclipse / Netbeans / IntelliJ we have to share it in eclipse formatter format. NetBeans and IntelliJ have plugins which provides importing of eclipse formatter

drakulis: @SubJunk just committed rest of files with proper indentation

Nadahar: @SubJunk Please go ahead and "look at it" and see if you can make it adhere to the "UMS standard" (whatever that is). You make it sound like you weren't involved, but that's simply not true if you look at #606. The fact that you didn't bother to test it doesn't mean that "it has nothing to do with you".

I haven't been able to make the formatter do everything "correctly" so I don't use it blindly. I actually just use it as a starting point if the code formatting is very bad, and then manually go through it and correct the rest. I doubt that it's possible to make it do everything as you'd want it, and even if it did, it would still drastically change pretty much any class you'd run it on simply because the code isn't coherent style-wise.

drakulis: @Nadahar yeah, I also noted it. We should set up common formatter (or use existing one) and apply it on all classes. Otherwise it will be really hard to contribute well-formatted code. During this refactor I'm not using formatter at all (i really miss it) because it change the code to much.

Nadahar: @drakulis I simply found the rules to be too limited, especially when it comes to splitting lines. It's been a couple of years now, so I don't remember all the details. I won't claim that it's impossible to improve the current file, but I remember giving up the idea that it would do everything correctly.

SubJunk: @drakulis most IDEs support applying formatting rules only to the lines you touch, so there is no need to worry about conflicting configs. Years ago I committed a big formatting change to make it more consistent but yes maybe we are due for another one after we come up with good rules :)

The automatic rules we make should match the style guide we use as defined at https://github.com/UniversalMediaServer/UniversalMediaServer/blob/master/STYLEGUIDE.md#code-style

SubJunk: Also @drakulis I'll look at this very soon and probably merge :)

SubJunk: I've created an issue for us to continue discussion and maybe work on this https://github.com/UniversalMediaServer/UniversalMediaServer/issues/1671

PR #1673, opened 2018-11-15
Make global ID repository memory-aware

SubJunk:
I have just tested this by disabling the changes in https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1486 and scanning my whole computer, and UMS didn't run out of memory, so this works.

It's not an ideal fix - but it is much better than crashing and it will combine nicely with other fixes to come

### DIFF
--- a/src/main/java/net/pms/dlna/GlobalIdRepo.java
+++ b/src/main/java/net/pms/dlna/GlobalIdRepo.java
@@ -2,6 +2,8 @@ package net.pms.dlna;
 
 import java.util.ArrayList;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
+import net.pms.PMS;
+import net.pms.util.UMSUtils;
 import org.apache.commons.lang.StringUtils;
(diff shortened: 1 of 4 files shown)


src/main/java/net/pms/dlna/GlobalIdRepo.java
```diff
@@ -19,9 +19,6 @@
 	// The number of IDs to remove at a time when memory is getting tight
 	private final int DEBOUNCE_AMOUNT = 10;
```
valib: @SubJunk it could be set higher to have more memory space.

valib:
@SubJunk sorry I read it wrong.

SubJunk:
@valib thanks for the approval, I have to do some more testing to make sure that it doesn't break things after removing them, then I'll merge if it works well

SubJunk:
I think this has been superceded by https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1695

PR #1675, opened 2018-11-15
Update buttons images to have diferences between the normal state

valib:
and mouse over and pressed.


SubJunk:
@valib the buttons are already different but maybe the difference is too subtle. The idea is that the colors get more saturated on hover and press. Maybe I can try to do the same thing but increase the saturation so it is more noticeable, instead of having the green and pink colours?

valib:
@SubJunk the differences are not visible so I only changed the RGB red and green value to make it more visible. Feel free to make it in your way.

SubJunk:
@valib ok I will do a commit soon with a similar change and please give your opinion on whether it is visible enough

valib:
@SubJunk OK.

SubJunk:
@valib how about this? https://github.com/UniversalMediaServer/UniversalMediaServer/compare/BiggerDifferencesBetweenIconStates?expand=1&short_path=dabd79e#diff-dabd79e7a4f33cba40357a3b2326cc00

valib:
@SubJunk it looks good. You should continue with the rest of icons and I will close this PR

valib:
Closed. Finished in #1763

PR #1683, opened 2018-11-19
Trace the GlobalIdRepo add

valib:
This is only for testing purpose because there is a bug when all renderers are adding their own GlobalIdResource IDs for the same media. The DLNAResource is added for renderer `null` which is OK but after that it is added again for particular renderer during browsing which is a bug.


SubJunk:
@valib this looks good, would it be helpful if I pushed the branch I've been using to play around with too? It is mostly just logging but you might find it useful

valib:
@SubJunk feel free to add whatever you want.

SubJunk:
Don't think we need this anymore

PR #1689, opened 2018-11-23
refactor(requestv2): extract request handler methods 

drakulis:
regarding #1664. Second approach for extracting handler methods. The changes in this PR contains point 2 and 3 from the list below.
The only change (apart of those already known) is located in `pom.xml` where I've added new dependencies.

I've tested it on all supported Java versions: 7 up to 11.

- [ ] `inputStream` as a local variable instead of class property. (required for further refactor)
- [ ] create separate methods for `streamResponse`, `xmlResponse`, `headerResponse` 
- [ ] ...


SubJunk:
@drakulis thanks for this, I will test it on my environment where the previous one failed just to be double sure before merging :)

SubJunk:
Works for me where it failed before, nice work

PR #1691, opened 2018-11-25
 refactor(requestv2): inputStream as local variable

drakulis:
regarding #1664. 

Main goal of this change is to convert `inputStream` class property to local variable. There is no reason to keep it as a class property and further refactor would be more complex to achieve.

Additionally I've made some other small code polishing.

**Refactor progress**
- [x] `inputStream` as a local variable instead of class property.
- [ ] create separate methods for `streamResponse`, `xmlResponse`, `headerResponse` 
- [ ] ...


valib:
👍

src/main/java/net/pms/network/RequestV2.java
```diff
@@ -799,51 +796,55 @@ private InputStream imageHandler(HttpResponse output) {
...
-					"</service>" + CRLF);
-			} else {
-				s = s.replace("Universal Media Server", configuration.getServerDisplayName());
-				if (mediaRenderer.isSamsung()) {
-					// register UMS as a AllShare service and enable built-in resume functionality (bookmark) on Samsung devices
-					s = s.replace("<serialNumber/>", "<serialNumber/>" + CRLF
-							+ "<sec:ProductCap>smi,DCM10,getMediaInfo.sec,getCaptionInfo.sec</sec:ProductCap>" + CRLF
-							+ "<sec:X_ProductCap>smi,DCM10,getMediaInfo.sec,getCaptionInfo.sec</sec:X_ProductCap>");
-				}
+	private String preparePmsSpec(String pmsXml) {
```
SubJunk: Not a big deal but it might make more sense for the new instances of `pms` to be `ums`

SubJunk:
@drakulis this looks good, have you also tested it manually? If so we can merge

src/main/java/net/pms/network/RequestV2.java
drakulis: @SubJunk yeah, I was not sure about that. I see that everywhere `pms` is used so I wanted to be consequent. But I agree that we should start use `ums` prefix where it is needed.

Currently I'm on business trip. Will be back home on Thursaday. Then I will change it.

drakulis:
@SubJunk yes, I'm moving all my changes on my server automatically.

src/main/java/net/pms/network/RequestV2.java
drakulis: @SubJunk done

PR #1693, opened 2018-11-26
Fixed a reversion in Sony Bravia TV support since 6.7.3

SubJunk:


SubJunk:
@valib this is the fix for the thing that paradox is talking about on the forum

valib:
@SubJunk this overrides the MediaInfo and should be fixed like I do in https://www.universalmediaserver.com/forum/viewtopic.php?f=9&t=12970

SubJunk:
@valib cool, I agree that is a better fix, let's see if he can confirm that it works :)

Issue #1680, opened 2018-11-18
GlobalIdRepo overuse


SubJunk:
Based on https://github.com/UniversalMediaServer/UniversalMediaServer/issues/1652#issuecomment-439658453

SubJunk:
@theafien 240k instances of GlobalIdRepo is indeed too many so I am interested in seeing how that happened. I haven't had much to do with that area of the code before, the global ID feature was implemented by @SharkHunter and was a bugfix for how some renderers expected things to be, I think in particular Xboxes.

I am trying to learn more about that area now and this will issue will help me learn more about it :)

@SharkHunter any idea why 600 videos would result in 240k instances of this class?

SubJunk:
@skeptical also did a lot of work on it, would be cool to see him around the place again :)

SharkHunter:
First up the globalIdRepo is to be able to use the remote control feature of DLNA. SInce each object will have a unique ID one render can tell another render to play video X. Now if you look more closely there is not 240K instances of globalIdRepo (there is always one) there is 240K of globalIdRepo$ID and that is @skepticals fault :). The reason for 240K is obviously that there are 240K DLNAResources created. THe question is why there are 240k DLNAResources created. I would say all other servers give each media a unique id but then agian they don't do discover on the fly.

theafien:
@SharkHunter this resources include videos, srt, audio? my library have 1000 files including image, video, srt, txt, url shortchuts no more than that

SubJunk:
@theafien it includes around 6 entries per video file, which are the virtual files within the `#-TRANSCODE-#` folders. There is certainly no reason for there to be 240k entries in there.

One of our users has a TV that keeps spamming browse requests when it is sitting there and might be creating a new set of global IDs sometimes - so that can eventually lead to 240k+. That is my suspicion about a lot of these memory problems anyway. I am approaching solutions from multiple angles:

1) Reduce the memory used by each resource, mostly by optimizing the compression of thumbnails since they are the thing that takes memory
2) Deduplicate the thumbnails in memory - we already dedupe them in the database but that needs to be abstracted to happen in memory too, and it has already been implemented in a fork so we will grab that soon
3) If the memory does eventually get full we will remove the oldest entries, I have a PR for this already but it is incomplete

When those 3 things are done I think we will be much closer to stability over time

valib:
[...]

PR #1615, opened 2018-10-10
8.0.0

SubJunk:
This major release of UMS will focus on making our handling of web feeds and streams easier to use by making it manageable via GUI, and fixing bugs with it including the implementation of more types of feeds and streams.

In general I will try not to commit directly to this branch, instead I will merge related branches into here, except for quick fixes noticed after those branches are merged.

PR #1695, opened 2018-11-27
GlobalIdRepo: use weak references

skeptical:
Not an actual PR, for reference in #1680.

### DIFF
--- a/src/main/java/net/pms/dlna/DLNAResource.java
+++ b/src/main/java/net/pms/dlna/DLNAResource.java
@@ -1103,7 +1103,7 @@ public abstract class DLNAResource extends HTTPResource implements Cloneable, Ru
 			if (child != found) {
 				// Replace
 				child.parent = this;
-				child.setIndexId(GlobalIdRepo.parseIndex(found.getInternalId()));
+				PMS.getGlobalRepo().replace(found, child);
 				children.set(children.indexOf(found), child);
 			}
 			// Renew
--- a/src/main/java/net/pms/dlna/GlobalIdRepo.java
+++ b/src/main/java/net/pms/dlna/GlobalIdRepo.java
@@ -1,5 +1,7 @@
 package net.pms.dlna;
 
+import java.lang.ref.WeakReference;
+import java.lang.ref.ReferenceQueue;
 import java.util.ArrayList;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 import org.apache.commons.lang.StringUtils;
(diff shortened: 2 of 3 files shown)


valib:
@skeptical there is a bug see the log
[debug.log](https://github.com/UniversalMediaServer/UniversalMediaServer/files/2622003/debug.log)

skeptical:
How are you @valib?  As mentioned in the other thread this is untested and for information purposes only and comes without any guarantees :). I googled the error and have added an untested fix for your sake, but it isn't my intention to support this extensively.  The bug has to do with concurrency which is outside the main point being illustrated here, I just attempted to transcribe my old pre-concurrency commits to the current code.  @Nadahar is the experienced locksmith here and would probably have more useful suggestions as to getting that aspect of it right.

SubJunk:
By the way for testing this it will be useful to temporarily revert https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1486 since running a scan is the fastest way to reproduce the bug.

Also another note is that generally audio files will produce it faster if they have cover art, in my experience.

Nadahar:
> Also another note is that generally audio files will produce it faster if they have cover art, in my experience.

I don't think audio files makes the problem any bigger, they will simply use more memory and thus deplete the available memory faster. The reason for this is that I chose to store audio thumbnails in higher resolution than other thumbnails, since they are also used as "cover art" and are often shown as a background or a larger image while a music track is playing. Thus I think they are limited to 640 x 480 instead of 320 x 240 for other thumbnails. The thumbnails aren't really a part of the problem though, they only make the symptoms easier to see.

skeptical:
@SubJunk :+1: glad one of us is actually thinking here :P.

SubJunk:
@skeptical I don't get accused of that too often, it's a nice change! :)

SubJunk:
Tonight I had what might be a side-effect of this PR - I had some folders stop working. The experience on the TV was that I was browsing folders and some of them didn't have our custom folder icon anymore, and nothing happened when I went into them, but the folders with the icon still worked so I guess the parent hadn't been removed but those specific children had. I am guessing that's a predictable side-effect of this approach.

I wonder if it would be better to not include the thumbnails in what we cache in the global ID repo, since then it could store many times more IDs without needing to discard potentially needed ones?

skeptical:
@SubJunk maybe I'm too far out of it at this point :), but you'll need to translate because it's not computing for me:

> some of them didn't have our custom folder icon anymore

Is this something to do with renderers that support custom folder icons?

> I wonder if it would be better to not include the thumbnails in what we cache in the global ID repo, since then it could store many times more IDs without needing to discard potentially needed ones?

What do you mean by "caching thumbnails" and "discarding needed IDs" in the global id repo? Are you running this in conjunction with #1673?

SubJunk:
My PC is back up and working, luckily the damage was limited to my power supply so replacing that fixed it. I will continue to test this branch. Unfortunately I am back to the first day again so will need to keep it running for a few more days to try to reproduce the bug I was seeing last time, and then I will look into it more if it happens again.

We also have word from the user who is helping with testing, he has been running it for 5 days when it would usually crash in 3 days, so it seems promising.

SubJunk:
Well I have been running this for 7 days with no problems and no noticeable increase in memory usage, and the user who is helping me test has also had no problems for 6 days and then another chunk of 5 days, after being interrupted by a power outage.
Is anyone else testing this currently? Should I put together a public beta release for it?

SubJunk:
@valib do you think we should just merge this instead of doing a beta release?

I have been running it continuously all this time and still haven't noticed any problems, not sure why I got issues the first time I ran it but I haven't reproduced it so that might have been something else.

SubJunk:
OK well we have a few big branches in progress (this one, subtitles, and 8.0.0) so I'm gonna make a frankenstein branch of those 3 and release it as `8.0.0 alpha 1` in the next 24 hours. If anyone has objections let me know :)

SubJunk:
8.0.0-a1 is released with this in it, let's see what people say

SubJunk:
Closing this since it is in https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1615

PR #1697, opened 2018-11-29
Support for Samsung Q9 series

valib:


SubJunk:
Ideally there would be a `@2x` version of the image and they would have the 10px side paddings but those things are optional

PR #1702, opened 2018-12-02
Use H.264 in Chrome web interface

SubJunk:
I have tested this on Chrome and Firefox on macOS and it correctly automatically differentiates the browsers - it uses libvorbis on Firefox still, but uses the Chrome-specific args for Chrome.

I think this can be merged as it is and we can continue to make similar changes for other browsers to make sure they all work.

### DIFF
--- a/src/main/external-resources/UMS.conf
+++ b/src/main/external-resources/UMS.conf
@@ -1381,11 +1381,6 @@ web_flash =
 # Default: false
 web_subtitles_transcoded =
 
-# web videos MIME types spoofing
-# ------------------------------
-# When Chrome browser is used, send videos with a spoofed WEBM MIME type.
-# Default: false
-web_chrome_mkv_as_webm_spoof =
 # When Firefox is used on Linux, send videos with a spoofed MP4 MIME type.
 # Default: false
 web_firefox_linux_mp4 =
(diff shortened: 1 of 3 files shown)


SubJunk:
@valib I've also made it use x264 on Firefox by default, and stopped it from downscaling resolution unless it's using libvorbis. Should be good to merge and makes the web interface videos much faster and nicer on Chrome and Firefox

valib:
@SubJunk it works but I see one difference on Chrome. It stops time seeking.

SubJunk:
@valib thanks for testing, I will try some different options to see if that can work on Chrome

SubJunk:
@valib I am trying to reproduce the ability to time seek on `master` but I can't do it when it is transcoding. How are you doing it?

valib:
@SubJunk in fact the time seek worked mostly only back in time and partially forward depends on how fast was transcoding but now video player shows only "Live" without any chance to seek. The question is if it is possible to set the web player to show the video duration not the time available to the end of the actually send partially transcoded video. This is working on my Pansonic TV and also on my new Samsung TV so the time seeking is available.

SubJunk:
@valib AFAIK the "live" message in the browser is happening on `master` too, so does this have anything to do with these code changes? Or are we just discussing another problem that is in the same area?

This is a screenshot from master showing "LIVE":

image

valib:
@SubJunk as I wrote before this is not real seeking support. For .OGG format was possible to return back to the beginning or somwere between from the current playing time. Also was possible to jump forward but only between current time and the end of the last transcoded chunk not the duration time.

SubJunk:
@valib I see, well it looks like there are a lot of video.js plugins that we can possibly use to implement a kind of chapter support, but maybe that could be part of a different branch. Do you think this is enough to stop this branch from being merged?

PR #1704, opened 2018-12-02
Fix xvid recognition

valib:
The XVID video codec is not properly recognized. This change I tested on my new Samsung TV and also on my old Panasonic TV which is capable to support only MPEG and DIVX format.


valib:
Wrong commit

PR #1705, opened 2018-12-02
Fix XVID recognition

valib:
The XVID video codec is not properly recognized. This change I tested on my new Samsung TV and also on my old Panasonic TV which is capable to support only MPEG and DIVX format.

### DIFF
--- a/src/main/java/net/pms/dlna/LibMediaInfoParser.java
+++ b/src/main/java/net/pms/dlna/LibMediaInfoParser.java
@@ -477,12 +477,6 @@ public class LibMediaInfoParser {
 		media.getSubtitleTracksList().add(currentSubTrack);
 	}
 
-	@Deprecated
-	// FIXME this is obsolete (replaced by the private method below) and isn't called from anywhere outside this class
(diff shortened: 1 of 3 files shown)


SubJunk:
This looks good. It would probably also be easy to add a new section `testGetFormat()` into `LibMediaInfoParserTest.java` for this. I think soon we will be making that file more comprehensive by hooking it up to MediaInfo but for now this would give us some coverage at least. If you want me to do it I can, let me know

src/main/java/net/pms/dlna/LibMediaInfoParser.java
```diff
@@ -480,22 +480,20 @@ public static void addSub(DLNAMediaSubtitle currentSubTrack, DLNAMediaInfo media
 	@Deprecated
 	// FIXME this is obsolete (replaced by the private method below) and isn't called from anywhere outside this class
 	public static void getFormat(MediaInfo.StreamType streamType, DLNAMediaInfo media, DLNAMediaAudio audio, String value) {
```
drakulis: personally I would rename also this method. Or even completely remove it if it is `deprecated` and unused as described in comment

valib: @drakulis no problem.

PR #1707, opened 2018-12-05
Update the .conf

valib:


SubJunk:
@valib it could be good to also create a generic base config for `Samsung Q*` TVs so that more people get these fixes. If your TV works better without chunkedtransfer it is likely that the `Q8 Series` also works better without it, for example. We can use `LoadingPriority` to make sure people get the most specific config that we have

PR #1700, opened 2018-12-01
refactor(requestv2): extract dlna stream handler

drakulis:
regarding #1664. 

This one was a real nightmare. The code was really messed up and fragile. The process flow were implemented very twisted way and related parts of functionality were spread out across whole `answer` method. I had to gather them together and put into one method. I hope I didn't break anything during this refactor. Please do the code review and test this branch on your devices. I will test this changes for few days to be sure it still works as it should.

Additionally I've made some other small code polishing - removed unused methods and moved variable declarations to the scope where they belongs.

The size of `answer` method has been already reduced from ~1000 to >100 lines of code, and the code has been organized into single responsibility methods. It is still far from perfect but it is enough to extract separate classes for eg `ContentDirectoryHandler`, `ConnectionManagerHandler` etc so we could reduce all those `if` / `else` statements and size of `RequestV2` class itself

## Testers are very welcome!

**Refactor progress**
- [x] extract common `createResponse` method to avoid adding `XML_HEADER`, `SOAP_ENCODING_HEADER` and `SOAP_ENCODING_FOOTER` over and over again
- [x] extract `searchHandler` method
- [x] extract other easy to extract handlers like `imageHandler`, `mediaReceiverRegistrarHandler`, `getSystemUpdateIdHandler` and others
- [x] `inputStream` as a local variable instead of class property.
- [x] create separate methods for `streamResponse`, `xmlResponse`, `headerResponse` 
- [x] extract fileHandler (movie, music etc). This one will be tricky.
- [ ] rename `argument` to `uri` (as in fact it is) and make some improvements in handlers related with it
- [ ] extract service handlers (`ContentDirectoryHandler`, `ConnectionManagerHandler` etc)
- [ ] drop support for first version of request (`Request` class)

PR #1709, opened 2018-12-07
#1664 image fix

drakulis:
This is fix for image load [issue](https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1700#issuecomment-445046186) found by @valib after recent refactor regarding #1664


SubJunk:
@drakulis thanks for this, will merge when @valib confirms :)

valib:
@Nadahar here is used `ByteArrayInputStream` which extends `InputStream` and it uses `byte[]` as a source of the stream. So implemented method `available()` is accurate.

Nadahar:
That said, as long as the source is a `DLNAThumbnailInputStream`, you can use `getImageInfo().getSize()`.

drakulis:
@SubJunk we have green light from @valib

Issue #1674, opened 2018-11-15
Incorrect Genre for some music files


maciekberry:
Hello,

I have run into the same issue than described here:
<https://www.universalmediaserver.com/forum/viewtopic.php?t=9032>
=> Genre are shown as "Genre_013" instead of "Pop" , "Genre_017" instead of "Rock"...

This only happens for MP3 files, but it's still PITA.

After investigating, I have realized that UMS sets the parameter "Language" to "raw" when creating and instance of the mediainfo library (file LibMediaInfoParser, line 45). Setting this parameter to "en" fixes the issue and (after 10 seconds of testing :) ) does not seem to break anything else

We have therefore two choices it seem: 
1. Setting Language to "en"
2. Introducing a translator for Genres based on the CSV files distributed with MediaInfo and replacing the correct Genre_XXX by the readable version locally

What do you think? If setting language to "en" is out of question, I can try to provide a pull request with the second option of course.

SubJunk:
How about a combination of 1 and 2, where we set MediaInfo language to the same language that UMS is configured for? We would just have to make sure that the rest of our code isn't relying on that raw language, but I think we use the keys instead of the values so it should be ok.

SubJunk:
Thanks for coming back, nice investigation :) Then maybe we should use `en` language for MediaInfo, and then add the genres to our translation files so they can be translated on `Crowdin`, what do you think about that approach?

maciekberry:
Fortunately, regardless of the language, the mediainfo library always accepts the "raw" key name. So even if the langauge is set to "en", the key name is "Format_Version".

Please see the attached screenshot (it's a breakpoint while reading a file and as you can see, the language parameter is set to "en").

As I said: I will accept any decision from you. I can also make an enum with the genres.

![1543275281173](https://user-images.githubusercontent.com/44756987/49048822-8a49ed80-f1dc-11e8-86f3-1f582ba7b1d9.jpg)

valib:
@maciekberry nice catch. I didn't check what language is using the MediaInfo as default. If it is correct than your suggestion is acceptable.

maciekberry:
This can be closed then.

maciekberry:
ARGH, you were 2 seconds quicker than me!

HFBoroni:
[...]

PR #1711, opened 2018-12-09
Solving issue #1674

maciekberry:
Solving issue #1674 (correct language passed to MediaInfoLib and avoiding incorrect genre labels) as per discussion in the ticket

### DIFF
--- a/src/main/java/net/pms/dlna/LibMediaInfoParser.java
+++ b/src/main/java/net/pms/dlna/LibMediaInfoParser.java
@@ -42,7 +42,7 @@ public class LibMediaInfoParser {
 		if (MI.isValid()) {
 			MI.Option("Internet", "No"); // avoid MediaInfoLib to try to connect to an Internet server for availability of newer software, anonymous statistics and retrieving information about a file
 			MI.Option("Complete", "1");
-			MI.Option("Language", "en");
+			MI.Option("Language", "raw");
 			MI.Option("File_TestContinuousFileNames", "0");
 			LOGGER.debug("Option 'File_TestContinuousFileNames' is set to: " + MI.Option("File_TestContinuousFileNames_Get"));
 			MI.Option("ParseSpeed", "0");
--- a/src/main/java/net/pms/dlna/LibMediaInfoParser.java
+++ b/src/main/java/net/pms/dlna/LibMediaInfoParser.java
@@ -42,7 +42,7 @@ public class LibMediaInfoParser {
 		if (MI.isValid()) {
 			MI.Option("Internet", "No"); // avoid MediaInfoLib to try to connect to an Internet server for availability of newer software, anonymous statistics and retrieving information about a file
 			MI.Option("Complete", "1");
-			MI.Option("Language", "raw");
+			MI.Option("Language", "en");
 			MI.Option("File_TestContinuousFileNames", "0");
 			LOGGER.debug("Option 'File_TestContinuousFileNames' is set to: " + MI.Option("File_TestContinuousFileNames_Get"));
 			MI.Option("ParseSpeed", "0");


maciekberry:
OK @SubJunk , I give up, I do not know how I can make this pull request that will only contain my change (src/main/java/net/pms/dlna/LibMediaInfoParser.java).

Every time I try to create a pull request, github automatically takes all the commits made to the branch (so mainly my synchronisation commits). I have read everything I could find on Github help but I do not know whether it is normal or not.

Perhaps (I am speculating here), those "synchronisation merges" will be ignored and only the change I want will be merged upstream?

Or - more probably - I am a github noob and somebody needs to help me (unless you do not want to see any more commits from me :) )

Can you help?

Nadahar:
@maciekberry I don't think this is a GitHub issue, but a Git issue. The problem is that you weren't "in sync" with UMS master when you made your commit.

On the GitHub side, it's not too late to "revive" this PR. By fixing the branch `maciekberry:master` this PR will be updated accordingly. But, this will fail if the PR is closed, so make sure to open the PR before you make any changes to the branch.

A PR will contain all commits that is different between UMS' master branch and then PR branch. Thus, the commits you see is those that differ from UMS' master branch. When making a PR, you're supposed to make sure that only the changes you intend the be merged are in the PR branch.

First off, you should never use your "master" branch for PRs, even though that's what most of us does when we make our first PR :wink: Instead, you should make one branch per PR you create, so that you can make changes to them individually. Since you haven't done this, the remains of #1657 is causing some trouble.

What happened is that @SubJunk "squashed" some commits when he merged #1657. "Squashing" means to combine several commits into one. That means that the 4 commits [in #1657](https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1657/commits) (451abc9caa737494a4e75286477f8530b59d4063, b7b7177b5567b065e23c6af0c691cde4c6c68822, 3420de1666c8f3a434b728a06fdcd589a942c512 and 0b909aad4cf18a7197e4d36eb8cb88cc735aabe3) were combined to a "new" commit ee550c6ef81336b5f538949939d76f0a88e3fafc. Although the code is identical, the "Git tree" is not. Since then a lot of other commits have been added to UMS master, and you have added some more "sync commits" before you made 5db7739a8aa21eef651f25bc0cec1dd7c5fe3756, which is what I suppose this PR was supposed to be.

There are multiple ways to solve this, the first thing I would do is to create a new branch in your repo called for example `backup`. It will be a copy of your `master` as it is now, which means that you can always "start again" if something you attempts fails.

There are basically two ways you can then "clean up" your `master branch` that I can think off. You could either `reset` your `master` to be equal to UMS's `master` and then `cherry-pick` the last commit from your `backup` branch. The other way is to `rebase` your `master` branch on UMS' `master` branch.

They would both produce the same result, but I think the rebasing is easiest to do, so I'll suggest that you use that. I don't know how you use Git, but I'll give you the command line commands which should be available to you regardless.

First, make sure you're checked out to your local `master` branch:
```powershell
git checkout master
```

Then fetch UMS' repo (to make sure you're using the latest version of it) and then do the rebase:
```powershell
git fetch upstream
git rebase upstream/master
```

That should be all to fix your local branch, but to push it to GitHub you need to use "force push" since you have to overwrite the commits that are already there:
```powershell
git push --force-with-lease origin master
```

Assuming that you reopened this PR before making the push, everything should be in perfect order when this is done. If some of the code changed by this PR has changed in UMS' `master` in the meantime, you might get a `conflict` (it means that Git can't figure out how to merge the code). If that's the case, it gets a bit more complicated as you have to manually merge the part of the code that Git can't figure out. I don't think that will happen in this case though.

maciekberry:
Thanks @Nadahar ! Your answer not only makes perfect sense but is also very helpful :)

I went for the rebase. For some obscure reasons, I did got 2 conflicts (the same changes that Subjunk suggested to my original PR), but a bit of git mergetool seems to have fixed that. 

Let me know if all is ok now (I have three commits in this PR, but in fact it is the same one with a revert in the middle)

valib:
@maciekberry @Nadahar it can be simply squashed and merged here in GitHub.

maciekberry:
Yes please. For now, I have broken my repo by messing up the commits while rebasing as per Nadahar advice :(
So I would appreciate if you could squash it while merging on "your" side, please.

Nadahar:
@valib I'm aware of that, but I was thinking that it was best to learn how to deal with such issues.

@maciekberry How have you broken your repo? I have done a lot of strange things with Git, but I haven't managed to "break" a repo yet. In which way is it "broken"?

valib:
@Nadahar I understand your effort but even if  @maciekberry  commited a lot of partial commits we can still squash them into one commit here by one click.

maciekberry:
OK, it is not broken of course. 
git rebase --abort fixed it all.

However, the valib solution seem to be a bit more pragmatic than forcing you to spend more time teaching me all the git arcanes.

I promise, next time I will learn this.

Nadahar:
@valib I know that, this was about how to "clean up" when things have gone wrong. As always I take the "educational route" :wink:

valib:
@Nadahar that is why I like you 👍

SubJunk:
@maciekberry Git can be tricky to learn at first, great to see all the help being given in here :)

PR #1713, opened 2018-12-09
refactor(RequestV2): small code improvements

drakulis:
Regarding #1664.

- rename `argument` to `uri`
- use `HttpMethod` instead of strings for http method checks
- remove `http10` flag while it is not used in `RequestV2` at all
- make `nettyRequest` local variable instead of class property

**Refactor progress**
- [x] `inputStream` as a local variable instead of class property.
- [x] create separate methods for `streamResponse`, `xmlResponse`, `headerResponse` 
- [x] extract fileHandler (movie, music etc). This one will be tricky.
- [x] rename `argument` to `uri` (as in fact it is) and other small improvements  and code cleanup


SubJunk:
@drakulis seems good to me, is this ready to merge or does it need more manual testing?

drakulis:
@SubJunk Those are just cosmetic changes. I don't expect any issues with them. I can be merged.

Issue #1706, opened 2018-12-03
Web interface broken hover effect on Firefox


SubJunk:
Since the redesign of the web interface there is a weird effect on Firefox that can be seen below. Reproduce by hovering over thumbnails:

![image](https://user-images.githubusercontent.com/584228/49358615-240f1e80-f738-11e8-884d-8cd1b5daae8f.png)

Any interest in fixing this @acanive ?

acanive:
sure, let me check and report back

acanive:
@SubJunk there is also a bug while trying to browse using remote in header area, content is not highlighted so you never know where is the focus. it was a line of code in css that was removed when the code was merged. should I repair this too or leave it as it is?

acanive:
@SubJunk job done, after merge, I will upload a device configuration file.
**note**: if you see double F thats due to my keyboard, it is learning to write on his own ;)

SubJunk:
@acanive if you mean the white highlight around the logo in the header, I removed that because it was permanently enabled instead of on hover. Thanks for the fixes :)

acanive:
@SubJunk ah ok, yes, that was the behaviour, I will rewrite the code to fix that.

PR #1717, opened 2018-12-12
 Fixed issue #1706, Web interface broken hover effect on Firefox.

acanive:
Fixed broken hover effect on Firefox.
Fixed navigation issue in header area.
Removed unused layout in virtual keyboard code.


SubJunk:
@acanive thanks!

PR #1737, opened 2018-12-27
Avoid downloading a JRE if user has already installed **ONLY** JDK

siboXD:
I've noticed that running UMS in an environment where you've installed **only** the JDK but not a standalone JRE, there were always a prompt that notified the user to install a "newer" version of Java.

I digged through the code and found that, checks were made only for presence of a _Java Runtime Environment_ and not for a _Java Development Kit_.

I managed to add those checks... because a JDK has a JRE inside and so it should't have any problems running UMS.


SubJunk:
@siboXD thanks!

PR #1740, opened 2019-01-02
Add examples for all binary tools paths to UMS.conf

Amadeus-:
Given the name of the configuration variables, as well as the default values provided, it is easy to assume that the settings for things like "mplayer_path", etc. should be the path only, not the path + the executable name.

I confirmed that the rest of the source is expecting the custom settings to include the binary file name, so I added examples to the UMS.conf file so as to make sure it was obvious that the setting should include the binary file name.

There were other "Examples" in UMS.conf, and they were listed before the "Default" values for the setting, so I duplicated this practice.

Please feel free to edit as desired.   Thanks!


SubJunk:
@Amadeus- thanks!

PR #1742, opened 2019-01-03
Fix bug when ffmpeg defers to mencoder

valib:


SubJunk:
@valib this looks like the right thing to do, instead of creating a new instance of MEncoderVideo, but can you tell me what the bug is so I can put it in the changelog?

valib:
@SubJunk because you implemented the `PlayerFactory` from DMS then initialization of the MEncoder by using `new MEncoderVideo()` doesn't work because the `cmdList.add(getExecutable())` in the `MEncoder.launchTranscode` returns null and the video is transcoded by FFMpeg not switched to MEncoder.

SubJunk:
It already works for me but ok let's merge

PR #1751, opened 2019-01-16
Update to Roku 6, 7, 8, 9

palopezv:
This update is based on a Roku Express+, model 3910MX with original firmware 8.x.x updated to version 9.0.0 build 4083-51 and the sdk documentation available at https://sdkdocs.roku.com/display/sdkdoc/Audio+and+Video+Support and https://sdkdocs.roku.com/display/sdkdoc/Closed+Caption+Support as of early January 2019.

I have tested these settings for the last two months.


palopezv:
Hmmm... I let "SeekByTime = true" fall through the cracks. It can be safely dropped if this is merged as it doesn't work with Roku Media Player. My bad.

SubJunk:
Well the tricky thing is that this breaks Roku TV detection because we are being too simplistic. We need to detect by the extra numbers in brackets too like e.g. https://forums.roku.com/viewtopic.php?t=89214

SubJunk:
There is a real problem that this is trying to solve, but this isn't the correct solution, as commented above. Please feel free to make another branch that improves detection

palopezv:
Agreed, I'm sorry to have left this slip though the cracks, but I can say I've been giving a lot of thought to the problem.

PR #1756, opened 2019-01-28
Update AVC profile in the MediaInfo parsing.

valib:
Do not use getters for variables defined in the same class.

EDIT: @SubJunk we should check all the code to avoid use getters and setters to be used for variables defined in the classes whose defined them and methods in those classes calls them. It can a little bit speed up the UMS because using getters and setters takes some extra time than using the variables themselfs.

### DIFF
--- a/src/main/java/net/pms/dlna/LibMediaInfoParser.java
+++ b/src/main/java/net/pms/dlna/LibMediaInfoParser.java
@@ -191,6 +191,7 @@ public class LibMediaInfoParser {
 						value = MI.Get(video, i, "Format_Profile");
 						if (!value.isEmpty() && media.getCodecV() != null && media.getCodecV().equals(FormatConfiguration.H264)) {
 							media.setAvcLevel(getAvcLevel(value));
+							media.setH264Profile(getAvcProfile(value));
 						}
 
 						if (parseLogger != null) {


Nadahar:
@valib You're right that it has a small time penalty, but it is so small that this only really "counts" in loops where the same call is done thousands of times.

If you start changing this without checking the logic, you're in for some nasty surprises. It breaks inheritance because when using the local variable instead of the getter, getters that are overridden by subclasses won't be called and thus the logic will break.

valib:
@Nadahar I am aware of that but if getters are overriden by subclasses than the basic code should be still working and classes which overide it should apply a new code or I am wrong?

Nadahar:
@valib I'm not quite sure what you mean. I'll try to make a tiny example to show you what I mean.

```java
public class Car {
  private Color color = Color.GREEN;

  public Color getColor() {
    return color;
  }

  public String toString() {
    return "Car (" + getColor() + ")";
  }
}
```

If I make a subclass like this:
```java
public class Ambulance extends Car {

  @Override
  public Color getColor() {
    return Color.RED;
  }
}
```

..calling `toString()` will still print the correct color. If you change the code in `toString()` to use the local variable `color` instead, the wrong color will be printed. I realize that the above "pattern" wouldn't be very smart, but it was the only simple example I could think of at the moment. The principle applies though: Some times subclasses want to modify what is returned even when they can't or won't change the local variable.

valib:
@Nadahar ahh I see. That example I overlooked. I anticipate that subclass could override setters to change the value but your nice example shows that I was wrong.
Thanks for the lesson. 👍 
I will change it back.

Nadahar:
@valib I'm not saying it's always wrong to change this, just that you must consider the use pattern and the general logic in each case. In some cases there won't be any reason to use the getter internally, in which case using the variable is the "correct thing to do" in my opinion. Other times it won't. I just don't think it's possible to make simply rules saying "this is always right" or "this is always wrong". If you want to stick to a simple rule without having to study the overall logic, you should stick to using the getters also internally.

valib:
@Nadahar from this perspective should we avoid using local variables in methods and prefere using getters and setters? Not strictly of course.

PR #1757, opened 2019-02-02
Use move instead of renameTo

SubJunk:
The other way doesn't give any useful errors to log, just a boolean. This is also more readable since we aren't renaming, we are moving.


SubJunk:
I have now tested this an it gives a nice message in the logs like "there is not enough free space on the disk". Merging now

PR #1768, opened 2019-03-03
Updated h2database to 1.4.198

SubJunk:
They report reduced memory usage, fixed memory leaks, better automatic selection of indexes, and more


SubJunk:
I have been running this on the 8.0.0 branch for a few days and it seems stable to me

PR #1783, opened 2019-04-03
- restarting system idle timer in doAllowSleep

scriptorron:
Hi all,

First at all: UMS is the best media server I could find. I tried many free and commercial products and only UMS was satisfying me. Thank you all for providing this software!!!!

The correct handling of the Windows power management is essential for me. 

My setup is:
- Media server: PC with Windows 10 (32bit)
  - runs MediaPortal for watching and recording TV
  - runs UMS for music streaming
  - Windows power management is configured to hibernate after 20 minutes inactivity
  - has an USB connected IR remote control which allows to control the MediaPortal
- UPnP/DLNA control point: Android tablet with BubbleUPnP
- UPnP/DLNA rederer: same Android tablet with BubbleUPnP; alternatively a Samsung blue-ray player that allows streaming

My typical use case for music listening is:
1. wake up the media server PC from hibernate with the remote control
2. start BubbleUPnP on tablet and listen to music
3. stop BubbleUPnP when I do not want to listen to music anymore
4. 20 minutes after stopping the music the media server must automatically go to hibernate (I am too lazy to switch off the PC!)

UMS has already implemented the functionality for preventing hibernate during playback. Unfortunately it is not working on my computer! After analyzing the debug log file ([debug-ErrorAt10h16m.log](https://github.com/UniversalMediaServer/UniversalMediaServer/files/3037669/debug-ErrorAt10h16m.log)) I found 2 issues:

1. When playing a 6 minutes song from a playlist the StopPlaying event comes 5 seconds after the StartPlaying event. The StartPlaying for the next song comes 6 minutes later:
`INFO  2019-03-17 09:54:25.334 [StartPlaying Event] Started playing M83 - Oblivion (Original Motion Picture Soundtrack) - 00 04 - M83 - Tech 49.mp3 on your BubbleUPnP`
...
`INFO  2019-03-17 09:54:30.729 [StopPlaying Event] Stopped playing M83 - Oblivion (Original Motion Picture Soundtrack) - 00 04 - M83 - Tech 49.mp3 on your BubbleUPnP`
...
`INFO  2019-03-17 10:00:29.325 [StartPlaying Event] Started playing M83 - Oblivion (Original Motion Picture Soundtrack) - 00 05 - M83 - The Library.mp3 on your BubbleUPnP (STV100-4)`
It seams that the BubbleUPnP renderer loads the stream within 5 seconds in a cache. The problem is that after the StopPlaying UMS is not aware that the renderer is still in playback. If BubbleUPnP caches more than 20 minutes stream the server PC will go to hibernate while the renderer is still playing. I am not sure if other renderer also have such a large cache.

2. About 20 minutes after my last mouse action the PC went to hibernate! This happened 2 minutes after starting a playback:
`INFO  2019-03-17 10:14:25.242 [StartPlaying Event] Started playing M83 - Oblivion (Original Motion Picture Soundtrack) - 00 10 - M83 - Unidentified Object.mp3 on your BubbleUPnP (STV100-4)
`
...
`INFO  2019-03-17 10:14:36.385 [StopPlaying Event] Stopped playing M83 - Oblivion (Original Motion Picture Soundtrack) - 00 10 - M83 - Unidentified Object.mp3 on your BubbleUPnP (STV100-4)`
...
`TRACE 2019-03-17 10:16:47.766 [UPNP-AliveMessageSender] Setting SSDP network interface: name:eth5 (NVIDIA nForce-Netzwerkcontroller)`
This was the last log message before hibernate!
...
`DEBUG 2019-03-17 10:26:02.622 [UPNP-AliveMessageSender] Sending ALIVE...`
This is the first log message after I woke up the PC again!

I do not see a solution for issue 1. It is likely that other renderer do not have such a large cache. Furthermore I do not have so many songs longer than 20 minutes. I can easily live with issue 1.

More critical is issue 2. It seems to me that the StartPlaying and StopPlaying events do not reset the system idle timer. In the code both events execute the `doPreventSleep` and `doAllowSleep` methods of the `SleepManager`. There is done:
1. `KERNEL32.SetThreadExecutionState(Kernel32.ES_SYSTEM_REQUIRED | Kernel32.ES_CONTINUOUS);` at StartPlaying and
2. `Kernel32.INSTANCE.SetThreadExecutionState(Kernel32.ES_CONTINUOUS);` at StopPlaying.

The Windows documentation is not clear for that. A possible interpretation that fits to my observation is that sleep is prevented for the time between `ES_SYSTEM_REQUIRED|ES_CONTINUOUS`  and `ES_CONTINUOUS` but the system idle timer does not get restarted. In worst case the PC hibernates immediately after `ES_CONTINUOUS`!

My proposed fix is to add send a `ES_SYSTEM_REQUIRED` in the `doAllowSleep` method. This restarts the system sleep timer and solved issue 2. I have this modification running since weeks and have no problems anymore.

### DIFF
--- a/src/main/java/net/pms/network/Request.java
+++ b/src/main/java/net/pms/network/Request.java
@@ -53,6 +53,7 @@ import net.pms.formats.v2.SubtitleType;
 import net.pms.image.ImagesUtil;
 import net.pms.io.OutputParams;
 import net.pms.io.ProcessWrapper;
+import net.pms.service.Services;
 import net.pms.util.FullyPlayed;
 import net.pms.util.StringUtil;
 import net.pms.util.SubtitleUtils;
@@ -365,6 +366,7 @@ public class Request extends HTTPResource {
 					appendToHeader(responseHeader, "Connection: keep-alive");
 				} else if (dlna.getMedia() != null && dlna.getMedia().getMediaType() == MediaType.IMAGE && dlna.isCodeValid(dlna)) {
 					// This is a request for an image
+					Services.sleepManager().postponeSleep();
 					DLNAImageProfile imageProfile = ImagesUtil.parseImageRequest(fileName, null);
 					if (imageProfile == null) {
 						// Parsing failed for some reason, we'll have to pick a profile
--- a/src/main/java/net/pms/network/RequestV2.java
+++ b/src/main/java/net/pms/network/RequestV2.java
@@ -53,6 +53,7 @@ import net.pms.image.ImageInfo;
 import net.pms.image.ImagesUtil;
 import net.pms.io.OutputParams;
 import net.pms.io.ProcessWrapper;
+import net.pms.service.Services;
 import net.pms.network.message.BrowseRequest;
 import net.pms.network.message.BrowseSearchRequest;
 import net.pms.network.message.SamsungBookmark;
@@ -329,6 +330,7 @@ public class RequestV2 extends HTTPResource {
 				iStream = dlnaThumbnailHandler(output, dlna, fileName);
 			} else if (dlna.getMedia() != null && dlna.getMedia().getMediaType() == MediaType.IMAGE) {
 				// This is a request for an image
+				Services.sleepManager().postponeSleep();
 				iStream = dlnaImageHandler(output, dlna, fileName);
 			} else if (dlna.getMedia() != null && fileName.contains("subtitle0000")) {
 				// This is a request for a subtitles file
--- a/src/main/java/net/pms/service/SleepManager.java
+++ b/src/main/java/net/pms/service/SleepManager.java
@@ -108,6 +108,12 @@ public class SleepManager {
 		playingCount--;
 		if (isPreventSleepSupported() && playingCount == 0 && sleepPrevented && mode == PreventSleepMode.PLAYBACK) {
 			allowSleep();
+			/* 
+			 * Between ES_SYSTEM_REQUIRED|ES_CONTINUOUS and ES_CONTINUOUS the Windows does not go to sleep.
+			 * But both do not restart the system idle timer. In worst case Windows goes
+			 * to sleep short after ES_CONTINUOUS. To avoid this the system idle timer gets restarted here.
+			 */
+			postponeSleep();
 		}
 	}
 
@@ -445,12 +451,6 @@ public class SleepManager {
 		protected synchronized void doAllowSleep() {
 			LOGGER.trace("Calling SetThreadExecutionState ES_CONTINUOUS to allow Windows to go to sleep");
 			Kernel32.INSTANCE.SetThreadExecutionState(Kernel32.ES_CONTINUOUS);
-			/* 
-			 * Between ES_SYSTEM_REQUIRED|ES_CONTINUOUS and ES_CONTINUOUS the Windows does not go to sleep.
-			 * But both do not restart the system idle timer. In worst case Windows goes
-			 * to sleep short after ES_CONTINUOUS. To avoid this the system idle timer gets restarted here.
-			 */
-			Kernel32.INSTANCE.SetThreadExecutionState(Kernel32.ES_SYSTEM_REQUIRED);
 
 			sleepPrevented = false;
 		}
--- a/src/main/java/net/pms/service/SleepManager.java
+++ b/src/main/java/net/pms/service/SleepManager.java
@@ -445,6 +445,12 @@ public class SleepManager {
 		protected synchronized void doAllowSleep() {
 			LOGGER.trace("Calling SetThreadExecutionState ES_CONTINUOUS to allow Windows to go to sleep");
 			Kernel32.INSTANCE.SetThreadExecutionState(Kernel32.ES_CONTINUOUS);
+			/* 
+			 * Between ES_SYSTEM_REQUIRED|ES_CONTINUOUS and ES_CONTINUOUS the Windows does not go to sleep.
+			 * But both do not restart the system idle timer. In worst case Windows goes
+			 * to sleep short after ES_CONTINUOUS. To avoid this the system idle timer gets restarted here.
+			 */
+			Kernel32.INSTANCE.SetThreadExecutionState(Kernel32.ES_SYSTEM_REQUIRED);
 
 			sleepPrevented = false;
 		}


Nadahar:
Hi, and thank you for your contribution. Although I'm not longer with this project, I'm the one that wrote this code, so I guess I should answer this.

I never use hibernation/sleep myself, so my testing was limited to "faked" scenarios. Since the code was released, there have been little feedback so I've assumed that it works correctly.

You're correct that there's very little we can do about 1). When it comes to caching the behavior varies a lot, most renderers don't cache that much in my experience, while some might cache large parts of or a whole video file. The only way I see to counter this is to increase the sleep timeout in the OS.

Regarding 2) I'm a little surprised that my testing didn't cover this. It's been a while now, but I seem to remember vaguely that I checked whether or not the timer was being reset when the sleep prevention ended. I seem to remember having an issue with this in macOS. In any case, I can't see that the code actually resets the timer at playback end. Maybe I forgot to actually implement it :wink: 

There is a method for this though, which is used when displaying images (since they have no "start" and "stop" playing events, the timer is simply reset each time a new image is requested). By using this method, this should be fixed on other platforms than Windows as well.

My suggestion is though to instead change `stopPlaying()` to:
```java
  public synchronized void stopPlaying() {
    if (playingCount == 0) {
      LOGGER.error("Sleepmanager cannot decrease playing reference count as it's already zero");
      return;
    }
    playingCount--;
    if (isPreventSleepSupported() && playingCount == 0 && sleepPrevented && mode == PreventSleepMode.PLAYBACK) {
      allowSleep();
      postponeSleep();
    }
  }
```

The only change is that `postponeSleep()` is added after `allowSleep()`. Would you be willing to test if this solves the problem as well?

scriptorron:
@Nadahar: 
Thank you for your answer. I will test your proposed change but I am not sure if I get the time today. You will get the results as soon as possible.

Nadahar:
@scriptorron Thanks - there's no hurry - you're the first to complain about this in the 2 years since I wrote the code :wink:

scriptorron:
@Nadahar: 
Yesterday night I tested your proposal and could successfully listen to music. But I plan to do more tests on weekend:
* watching movies
* watching pictures
* using my TV as renderer
* using my blue-ray player as renderer
* checking if the media server goes to sleep after stopping playback

All this will take a lot of time.

A challenge is to make sure that no other program (for instance Windows Update) is preventing sleep. It is a nightmare to debug Windows power management! With `powercfg.exe /requests` (run as administrator!) you get a list of processes that prevent sleep. But it does not tell you the remaining time until sleep. Furthermore `powercfg.exe` reports a "single shot" only and you can not run it on command line without restarting the system idle timer. 
To get more information I made a Python script ([ShowSystemIdleTime.zip](https://github.com/UniversalMediaServer/UniversalMediaServer/files/3042905/ShowSystemIdleTime.zip)) that runs in a loop and tells me the remaining time and if other processes are preventing sleep. Sorry that I did not made it in Java - I have no clue of Java and my Python skills are better.
By the way I made some interesting observations with this script. Windows reports the remaining time until sleep only with 30s resolution and at the beginning of an idle period it takes 60s before the time is decremented by 30s. If you disable the standby mode in the Windows power options the system idle timer does not get stopped but it is set to 4294967295 seconds (135 years) and still counts down. It does not matter how you configure your system - latest after 135 years your windows will go to sleep. All this makes me believe that the Windows power management is just patchwork.

I will report my test results next week.

Nadahar:
@scriptorron Wow - you sound almost like me :laughing: - very thorough :+1: The reason I asked you to test was simply because I have some "aversion" to testing this since the last time I did it. I know very well how painful it is to test, I did tests on Windows XP, Windows 7, Windows 10, macOS 10.8.5 and macOS 10.12 when I made this. I hoped that these tests were enough to extrapolate the results to other versions of the OSes, although you can never know for sure..

I "cheated" in that I put the sleep timeout to 1 minute to avoid making it take forever. I still don't know how I missed this one, as I remember I was onto the problem with the state of the timer when sleep prevention was cancelled. I don't know what I concluded or if I forgot to implement something though, two years is too long for my memory :innocent: 

Python, Perl etc is fine for scripting IMO, Java isn't really well suited for that anyway (as it must be compiled and then run through a JVM). 

I'm happy for all the tests you want to do, I spent enough time on testing this that I feel that I "never want to do it again", but what I meant to ask you was simply to confirm that doing the fix the way I suggested had the same results as your fix :wink: It seems to me like you have confirmed this already, so thank you.

SubJunk:
@scriptorron thanks for this contribution and your kind words about UMS! :)
I'll wait to see the results of your testing of the different fix

Nadahar:
@scriptorron In [my commit](https://github.com/DigitalMediaServer/DigitalMediaServer/commit/9373d32501760946f02b7f0ef691651076c28afd) I also did another little fix that I found while looking at it. You may want to included that as well.

Nadahar:
I definitely created a bug instead of fixing one. The whole goal of the `Math.max()` is to prevent the value from ever being negative (if for example the computer clock is adjusted by NTP, manually etc). I somehow managed to "trick myself" when reading the code to interpret it as if it would always be 0 and thus never actually wait - but it was just self delusion :blush: 

So, forget all about my previous comment :innocent:

scriptorron:
I did some tests with Nadahar's proposal today. The short summary is:
For MP3 and video content it works as intended! But streaming images does not restart the system idle timer.

Long description:
I used my script to see which processes prevent sleep and when the idle timer gets restarted. Before I started streaming I waited until the PC went to sleep - just to be sure that the weapons are armed. I observed some Windows processes that set power requests (probably with  `ES_SYSTEM_REQUIRED|ES_CONTINUOUS`) but only when I shifted the mouse or pressed a key the idle timer got restarted. I also have seen a long lasting windows process `defragsvc` that set a power request and was running longer than the idle timeout: the system idle timer counted until its timeout but the system did not go to sleep (because the process requested power!). About 1 minute after `defragsvc` removed its power request the PC went to sleep. All this makes me sure that we do right when we restart the idle timer after `ES_CONTINUOUS`.

After that I used different controller (BubbleUPnP, VLC for Android and my TV) to stream content to different renderer (BubbleUPnP, VLC on Android, TV and my BlueRay player). Results are:

- Browsing the library does not restart the idle timer. FMO this is acceptable.
- Streaming images does not restart the idle timer. I run a slide show of photographs and the server went to sleep after the configured 20 minutes. Probably this is something that needs to be fixed. Nadahar already mentioned a method that is used for streaming images and that the `postponeSleep()` needs to be inserted there.
- When streaming MP3 I can see the power request (from `ES_SYSTEM_REQUIRED|ES_CONTINUOUS` to `ES_CONTINUOUS`) during the data transmission and followed by the restart of the system idle timer. The Android renderer have a large cache. For instance a 4:15 song is transmitted in about 8 seconds. While the renderer is playing the data from the cache the idle time is already running. Other renderer (my TV and my BlueRay player) have smaller caches: they cached only the last minute of the song. FMO it is acceptable that the user must configure the system idle timer of the UMS server to a longer time than the cached content in the renderer.
- Streaming video is very similar to streaming MP3. I see the power request followed by the idle timer restart. Only the size of the cached data is different: VLC for Android and my TV have cached the last 80 seconds of the video (XviD in SD quality). 
- I jumped around from one MP3 (and video) to the other while playing to simulate the "irresolute user". The power requests and timer restarts were as expected. When I stopped playback the power requests disappeared, the idle timer got a last restart and after time out the UMS server went to sleep. Perfect!

scriptorron:
@Nadahar I did my tests without your last code change. In fact I read your last 2 comments right now after writing my prosaic message. :-)

Nadahar:
@scriptorron Thanks, that sounds great except for your finding with images. This puzzles me as the `postponeSleep()` method used in this fix already existed exactly because of the images.

That either means that it's not called for images, or that the same command doesn't reset the timer in that case.

scriptorron:
@Nadahar What a pity - I forgot to set the log level to TRACE for my tests. Then the `doResetSleepTimer()` method would leave a message in the log. 

Give me some minutes to install UMS on my test-PC again and run some experiments.

Nadahar:
@scriptorron Mystery solved about the images. `postponeSleep()` is never called, since `StartStopListenerDelegate` isn't used for images in `Request` and `RequestV2`.

When I created `SleepManager` I didn't focus on the calls being made from unrelated code, I simply replaced another implementation for handling the sleep management itself. As you can see from #1226 I never did look into whether the calls were made from the correct places (the task is still not "completed"), but testing feedback indicated that things were working correctly despite this. Apparently this wasn't entirely true :wink:

Nadahar:
@scriptorron You don't have to test that it's not called, I have searched the references in the code and it isn't called :wink:

scriptorron:
@Nadahar Thank you. Tell me if you want me to test something.

Nadahar:
@scriptorron I did a quick fix for the issue [here](https://github.com/DigitalMediaServer/DigitalMediaServer/commit/5075a7408dec4a20e0e534cd8cf52db001a48814). I think that should be all.

I haven't followed the UMS code for a long time now, so I don't know if `Services` will do the same there as in DMS, but assuming it does, the same change should work in UMS too.

scriptorron:
@Nadahar What is the right "git"-way to merge your changes into my working copy? I need to learn git!!!

Nadahar:
For information, I considered adding it to other requests too (like thumbnails). But, when thinking about it, I remembered that some renderers "live their own life" and does all kind of browsing/refreshing on their own, without any user action. To avoid that some of these "creative" renderers prevent sleep completely, I found it safer to only make the call when an image is requested.

scriptorron:
@Nadahar I fully agree! The restart of the idle timer should only be done when the user interacts or when content is played for the user.

Nadahar:
@scriptorron As always with Git, there are many, many ways to do it. The easiest might be to cherry-pick my commit and then squash it into yours. In this case it's not so easy though, as you'd have to add the DMS repo as an additional remote to your local repo to be able to cherry pick from it. 

In this case, I think copy/paste is the easiest way to go :wink:

scriptorron:
@Nadahar Thanks. Give me some minutes to copy/paste, compile and test.

scriptorron:
@Nadahar ReqestV2.java looks completely different in my working copy. I have something like this and will copy the `postponeSleep` in the line after the "request for an image" comment:

`		InputStream iStream = null;
		if (dlna != null) {
			// DLNAresource was found.

			if (!dlna.isCodeValid(dlna)) {
				LOGGER.error("Invalid code for: " + dlna.getName());
				throw new HttpException(HttpResponseStatus.BAD_REQUEST);
			}

			if (fileName.startsWith("thumbnail0000")) {
				iStream = dlnaThumbnailHandler(output, dlna, fileName);
			} else if (dlna.getMedia() != null && dlna.getMedia().getMediaType() == MediaType.IMAGE) {
				// This is a request for an image
				iStream = dlnaImageHandler(output, dlna, fileName);
			} else if (dlna.getMedia() != null && fileName.contains("subtitle0000")) {
				// This is a request for a subtitles file
				iStream = dlnaSubtitlesHandler(output, dlna);
			} else {
				// This is a request for a regular file.
				iStream = dlnaMediaHandler(output, startStopListenerDelegate, dlna, id);
			}
		}
`

Nadahar:
@scriptorron That should probably work just fine. It looks like the code has just be reorganized without changing the logic itself.

scriptorron:
@Nadahar It works! I see the idle timer restarting when viewing images. You did a great job! 
I will commit and push the last version of my code.

scriptorron:
I have not checked again all combinations of controller/renderer I have. The changes are certainly independent on this. I can also not imagine that the MP3 and video streaming got damaged with the last changes.

Nadahar:
@scriptorron Yes, these changes are so simply that the risk of regressions to other things should be minimal.

scriptorron:
@Nadahr I think you solved the "Prevent Sleep" issue for all kinds of content. At least for me this is really an important feature. Thank you!!! It was a pleasure for me to meet you on the keyboard ;-)

It would be nice when the code finds its way in the master branch. If you don't mind I would like to switch off the computer and spend the rest of the day with my family. Tell me if you want me to do more tests in the next days.

Nadahar:
@scriptorron Please enjoy your family time :smile: As far as I'm concerned this has been dealt with, it's already in master in DMS. It's up to the UMS developers if they want to merge this PR, but I would suggest that they did.

Thank you for doing all the hard work :wink:

SubJunk:
Great work you guys, thanks a lot. It will be included in the 8.0.2 release which will be out either this week or next

PR #1784, opened 2019-04-05
H2database 199

SubJunk:


SubJunk:
@valib I identified that the problem with h2database 197+ was a mistake in our syntax and it is working now. I also searched the code for other occurrences of that bug but this was the only one

Issue #1788, opened 2019-04-08
virtual_folders can't handle multiple parameters


begleysm:
I am admittedly very new to UMS but I believe I've found a bug.

I am attempting to implement two virtual folders.  The UMS.conf file says that the format should be:
`VFolderName1|Folder1Path,Folder2Path;VFolderName2|Folder1Path`

The line in my UMS.conf is:
`virtual_folders = TV\\ Shows|/mnt/media/Video/TV_Shows;Web\\ Shows|/mnt/media/Video/Web_Shows`

After saving UMS.conf and restarting UMS the websever is unavailable.  Checking debug.log shows the following:
```
DEBUG 2019-04-08 21:14:36.570 [main] Reading virtual_folders: "TV\ Shows|/mnt/media/Video/TV_Shows;Web\ Shows|/mnt/media/Video/Web_Shows" (default: "")
WARN  2019-04-08 21:14:36.620 [main] Folder not found: File "/mnt/media/Video/TV_Shows;Web\ Shows" not found
ERROR 2019-04-08 21:14:36.620 [main] A serious error occurred during Universal Media Server initialization: com.google.gson.stream.MalformedJsonException: Use JsonReader.setLenient(true) to accept malformed JSON at line 1 column 75 path $
DEBUG 2019-04-08 21:14:36.622 [main]
com.google.gson.JsonSyntaxException: com.google.gson.stream.MalformedJsonException: Use JsonReader.setLenient(true) to accept malformed JSON at line 1 column 75 path $
        at com.google.gson.Gson.assertFullConsumption(Gson.java:864)
        at com.google.gson.Gson.fromJson(Gson.java:854)
        at com.google.gson.Gson.fromJson(Gson.java:802)
        at net.pms.configuration.MapFileConfiguration.parseVirtualFolders(MapFileConfiguration.java:118)
        at net.pms.dlna.RootFolder.getVirtualFolders(RootFolder.java:477)
        at net.pms.dlna.RootFolder.discoverChildren(RootFolder.java:195)
        at net.pms.dlna.RootFolder.discoverChildren(RootFolder.java:117)
        at net.pms.configuration.RendererConfiguration.getRootFolder(RendererConfiguration.java:507)
        at net.pms.PMS.getRootFolder(PMS.java:159)
        at net.pms.PMS.init(PMS.java:659)
        at net.pms.PMS.createInstance(PMS.java:900)
        at net.pms.PMS.main(PMS.java:1070)
[...]

PR #1789, opened 2019-04-11
Fix conversion of VirtualFolders string syntax into JSON syntax

valib:
fix issue #1788

### DIFF
--- a/src/main/java/net/pms/configuration/MapFileConfiguration.java
+++ b/src/main/java/net/pms/configuration/MapFileConfiguration.java
@@ -11,6 +11,7 @@ import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.lang.reflect.Type;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
 import net.pms.PMS;
@@ -66,7 +67,7 @@ public class MapFileConfiguration {
 			conf = null;
 
 			try {
-				conf = FileUtils.readFileToString(file);
+				conf = FileUtils.readFileToString(file, StandardCharsets.US_ASCII);
 			} catch (FileNotFoundException ex) {
 				LOGGER.warn("Can't read file: {}", ex.getMessage());
 				return null;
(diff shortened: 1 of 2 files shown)


src/main/java/net/pms/configuration/MapFileConfiguration.java
```diff
@@ -66,60 +67,65 @@ public MapFileConfiguration() {
 			conf = null;
 
 			try {
-				conf = FileUtils.readFileToString(file);
-			} catch (FileNotFoundException ex) {
-				LOGGER.warn("Can't read file: {}", ex.getMessage());
-				return null;
+				conf = FileUtils.readFileToString(file, StandardCharsets.US_ASCII);
```
SubJunk: @valib is there a reason why you used US_ASCII instead of UTF8?

valib: @SubJunk I tried to use something general. The question is what is the general charset for all .confs? If I change it in my WIN10 the charset used by Notepad or other text editor is Windows CP1250.

SubJunk: I think UTF-8 is a de facto standard that supports a lot of different characters but maybe my information is out of date. It tends to be the encoding type that databases are initialised to unless there is a good reason otherwise

Nadahar: UTF-8 is the best choice when you choose a text file encoding for something new. However, it doesn't help to read something as UTF-8 if the source isn't encoded as UTF-8. The first 127 codes of UTF-8 are the same as US-ASCII, but that's also true for many others, like for example ISO 8859-1.

The Java "properties" files are defined as using ISO 8859-1, which is why unicode characters must be specified using `\uxxxx`. Since the `.conf` files used by UMS are properties files, they all have to be encoded as ISO 8859-1. It might be confusing for a user to be expected to use ISO 8859-1 in most configuration files, but UTF-8 in the virtual folders configuration file. At the same time, if this is a JSON file, the `\uxxxx` notation probably isn't supported so that would make it impossible to use unicode characters in virtual folders.

PR #1791, opened 2019-04-18
Enhance install instructions

rubo77:
This will use openjdk-8 instead of 7 and the formatting is enhanced. The file could also be renamed to `.md` to make it display nicely here on GitHub.

maybe we should also add an instruction how to create a systemd-file?

### DIFF
--- a/INSTALL.txt
+++ b/INSTALL.txt
@@ -3,28 +3,40 @@ Installers for Windows and Mac OS X and tarballs for Linux can be found at:
     http://sourceforge.net/projects/unimediaserver/files/Official%20Releases/
 
 To install UMS from the tarball on Linux, open a terminal and enter the
-following commands (replace <version> with the version you're installing):
-Note: These instructions have been tested on Ubuntu 12.04, but something similar
+following commands (set VERSION to the version you're installing):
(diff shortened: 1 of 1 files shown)


INSTALL.txt
```diff
@@ -3,28 +3,40 @@ Installers for Windows and Mac OS X and tarballs for Linux can be found at:
...
-    wget http://sourceforge.net/projects/unimediaserver/files/Official%20Releases/Windows/UMS-<version>.tgz/download
+        sudo apt-get install mediainfo dcraw vlc-nox mplayer mencoder
+      
+  you can also optionally install dcraw and VLC: `sudo apt-get install dcraw vlc`
 
-3) Extract the tarball into a ums-<version> directory:
+2. Download the tarball e.g.:
 
-    tar xzvf ums-<version>.tgz
+    get the direct link from http://sourceforge.net/projects/unimediaserver/files/Official%20Releases/
```
SubJunk: While we are updating this link anyway let's change it to https://www.fosshub.com/Universal-Media-Server.html

rubo77:
maybe you can just take my suggestion and add it in a new commit, then you can close this PR unmerged. I think it is easyer, cause you know some more info, that is valuable to add there anyway.

PR #1793, opened 2019-05-06
Fixes issues with starting UMS as a service.

pponce:
Also allows UMS to remove the service once created via the UMS UI.  The remove service button I believe is expecting the name of the service to be Universal Media Server


SubJunk:
Nice catch, thanks @pponce !

PR #1794, opened 2019-05-06
Implement latest apache http engine

valib:


valib:
@Nadahar thanks for your warning. I reverted the commit.

SubJunk:
@valib can you please summarise the changes you made to the automatic updater? Are there also bugs fixed with it?

valib:
@SubJunk this is only implementation of the org.apache.httpcomponents instead of commons-httpclient 3.1 which is not maintained since 2007. This latest org.apache.httpcomponents allows me a little bit change the logic because download is fully controlled by CloseableHttpAsyncClient and using ZeroCopyConsumer returns directly the downloaded file without an intermediate in-memory buffer. I only had to make a copy of the ZeroCopyConsumer to implement callback to show the progress of download.

SubJunk:
Cool it seems like a good change especially since I am wanting to make the automatic updater work in other operating systems. Thanks!

PR #1803, opened 2019-05-21
Single page AngularJS application & JAX-RS

mpiggott:
Uhh Hi,

While poking around I fell down a hole and ended up converting the HTTP layer to JAX-RS (jetty + resteasy) and the UI to an AngluarJS single page application.

I attempted to keep the changes isolated to the web UI code, and I did it incremntally (e.g. first 3 commits are an attempt to convert the endpoints w/o ripping out templates).

I assume there are issues I didn't notice, that I'm not setup to notice, but the ones I know about:
* Polling can NPE, its not clear to me why not to create a default rendering on poll?
* layout (grid / dynamic) is temporary 
* distinct bump port is unavailable 

Is this something worth pursing further?


valib:
@mpiggott it looks promissing but I see one problem. When the user click to browse an folder which has  more items than a lot of time nothing happened and the web page looks like finished but empty without any message "Please wait" or showing e.g. runnig circle. In the left collumn there is only "Back" folder (button) and the main page is empty. Current UMS behaviour shows almost immediately the tree in the left page and items in the main page which are than populated with thumbnails. Here the code wait until everything is done and after that it is shown.

mpiggott:
> @mpiggott it looks promissing but I see one problem. When the user click to browse an folder which has more items than a lot of time nothing happened and the web page looks like finished but empty without any message "Please wait" or showing e.g. runnig circle. In the left collumn there is only "Back" folder (button) and the main page is empty. Current UMS behaviour shows almost immediately the tree in the left page and items in the main page which are than populated with thumbnails. Here the code wait until everything is done and after that it is shown.

Hmm, maybe I overlooked something but I believe the current behaviour is that the server constructs the entire HTML for a given browse node in memory then returns it. Changing the endpoints I tried to limit my changes to simply ripping the templates and instead creating objects, assuming I didn't muck something up I would assume that the overall page load would take similar or perhaps slightly less time (as the template isn't rendered on the server).  Though at the moment it would definitely be disconcerting for a user as the page appears loaded and there is no indication of activity, adding an indicator should be relatively simple.

SubJunk:
Hi @mpiggott and welcome to UMS! :)

I tried to test this but the branch causes a failure in the `makensis` part of the Maven build:

```
[INFO] [MAKENSIS] File: "\src\main\external-resources\web" -> no files found.
[INFO] [MAKENSIS] Usage: File [/nonfatal] [/a] ([/r] [/x filespec [...]] filespec [...] |
[INFO] [MAKENSIS]    /oname=outfile one_file_only)
[INFO] [MAKENSIS] Error in script "/src/main/external-resources/nsis/setup.nsi" on line 206 -- aborting creation process
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  50.068 s
[INFO] Finished at: 2019-05-31T11:20:01+12:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.codehaus.mojo:nsis-maven-plugin:1.0-SNAPSHOT:compile (build-setup-windows) on project ums: Execution of makensis compiler failed. See output above for details.
```

so I guess that path will need to be updated

Nadahar:
@SubJunk You don't need to make a Maven build to test it. Just run it from your IDE, or if it doesn't support that, just build the JAR (simply Ctrl + C when Maven starts with NSIS, the JAR is already in the `target` folder by then). and run that (`java -jar <filename>`).

I gave this a quick test earlier, and to me it looked pretty much as before (I assume that was the purpose). What I'm wondering is what's the gain by replacing "one way" with "another way" that produce the same result.

Since I'm a complete noob when it comes to web-technologies, I have no idea what is really changed, except that I see that this seems to run using the "servlet" system instead of using a HTTP server and serving "old fashioned" HTML/CSS/JS. It has a bunch of dependencies and drags some huge libs in that isn't necessary today, to support the "servlet" system I presume. New dependencies, especially big ones (which probably have a lot of dependencies themselves), can mean many problems on other areas when it comes to dependency resolution etc., and it will also negatively impact the JAR size. As it is now, it also bumps the required Java version to Java 8 (it failed to build for me since I use Java 7 for development).

So, I see many cons, but I assume there are some pros that I don't understand. I have no idea what these different JS "frameworks" does. To me, JS is JS, an unstructured, type-unsafe mess of a "script language" that's also incredibly slow and doesn't work equally in different browsers (and browser versions). There is already JS in the current solution, what new possibilities does AngularJS provide? Is it faster, and if so is that significant? Can it do "new things" not possible today? Does it allow to move more of the logic to Java code? Are there other advantages that I haven't even thought of?

mpiggott:
> I gave this a quick test earlier, and to me it looked pretty much as before (I assume that was the purpose). What I'm wondering is what's the gain by replacing "one way" with "another way" that produce the same result.

Basically I was attempting to separate the concerns of the media server with the presentation details of the UI. With the current code making changes or writing tests is for the logic of UI is more difficult due to their intermingled nature. Also while refactoring the templating it didn't appear that all values were encoded before being placed in the HTML.

> Since I'm a complete noob when it comes to web-technologies, I have no idea what is really changed, except that I see that this seems to run using the "servlet" system instead of using a HTTP server and serving "old fashioned" HTML/CSS/JS. It has a bunch of dependencies and drags some huge libs in that isn't necessary today, to support the "servlet" system I presume. 

The existing code isn't using Java's servlet APIs, its using discouraged classes in the Oracle JRE.  The code in the PR is using [JAX-RS](https://en.wikipedia.org/wiki/Java_API_for_RESTful_Web_Services) API with RESTeasy for the implementation and Jetty as the web server. I also used Guice for [dependency injection](https://en.wikipedia.org/wiki/Dependency_injection).

> New dependencies, especially big ones (which probably have a lot of dependencies themselves), can mean many problems on other areas when it comes to dependency resolution etc., and it will also negatively impact the JAR size. 

/shrug it can happen but in my experience its generally not a large or common problem, though given lower test coverage(?) it might be more difficult to notice early in the project.

> As it is now, it also bumps the required Java version to Java 8 (it failed to build for me since I use Java 7 for development).

Ah, I don't have Java 7 installed at this point and didn't check what CI was using. It should be possible to use the [animal-sniffer](https://www.mojohaus.org/animal-sniffer/animal-sniffer-maven-plugin/) maven plugin to catch this on CI if java 7 isn't available there.

mpiggott:
> So, I see many cons, but I assume there are some pros that I don't understand. I have no idea what these different JS "frameworks" does. To me, JS is JS, an unstructured, type-unsafe mess of a "script language" that's also incredibly slow and doesn't work equally in different browsers (and browser versions). There is already JS in the current solution, what new possibilities does AngularJS provide? Is it faster, and if so is that significant? Can it do "new things" not possible today? Does it allow to move more of the logic to Java code? Are there other advantages that I haven't even thought of?

I used Angular mainly because it was already in use in `subs.html` so I assumed it was already a known framework, though coincidentally its also one I know well and its easy to use without resorting to adding an npm build. Generally the modern frameworks give y

Nadahar:
@mpiggott Thank you for your answers. I'm not sure I'm that more wiser, but I hope some seeped through :wink: Is jetty an important part of this, or is it simply a "well known" web server? UMS already use a number of different web servers in the code base, and the `subs.html` stuff is a recent contribution that is quite separated from the rest.

From the top of my head, UMS use:
* The (discouoraged) JRE-internal HTTP server
* [Apache HTTP components](https://hc.apache.org/)
* [Netty](https://netty.io/)

Would one of these be able to do it just as well, or is jetty essential to using `JAX-RS` or other parts of your "solution". The reason I'm asking is that there's quite a line walking to get the existing UMS dependencies to play together (because you can't have to dependencies rely on the same sub-dependency but with different versions) since there are already hundreds of dependencies in the "dependency tree". As such, I would see it as beneficial if it was possible to stick with the current dependencies as much as possible.

When you mention writing tests, I assume you mean tests written in JS? Or does this somehow allow more tests to be written in Java?

Again, being a noob, I don't quite understand how it's possible to run JS UI tests given the inconsistent behavior of browsers, but maybe there is some "mock up browsers" or a compilation of rules as to what they can or cannot do? Or, are the tests limited to things that aren't browser dependent?

SubJunk:
I'm still waiting for this to compile before spending time looking into it. The bare minimum for a PR is not to break compilation IMO (I'm aware it's doable if I just do the Java part).

I've had a quick scroll through the code being added and I'm hesitant about the approach, I see controllers being called via routes which is a `pre-1.5` way to organise `AngularJS` code. I would be much more in favour of *at least* a component API implementation, but even more in favour of an `Angular 8` implementation using `Angular CLI`.

That way our tooling is largely taken care of by just keeping in sync with `Angular CLI` updates instead of needing to update little things ourselves, plus then we get to write in `TypeScript` instead of `JavaScript`, and `TypeScript` is super similar to `Java/C#` so will be a nicer transition for us and give us type safety if we want it.

SubJunk:
I've had a quick scroll through the code being added and I'm hesitant about the approach, I see controllers being called via routes which is a `pre-1.5` way to organise `AngularJS` code. I would be much more in favour of *at least* a component API implementation, but even more in favour of an `Angular 8` implementation using `Angular CLI`.

That way our tooling is largely taken care of by just keeping in sync with `Angular CLI` updates instead of needing to update little things ourselves, plus then we get to write in `TypeScript` instead of `JavaScript`, and `TypeScript` is super similar to `Java/C#` so will be a nicer transition for us and give us type safety if we want it.

SubJunk:
@mpiggott my real job is working with AngularJS and Angular every day so I would be able to assist you by answering questions and give fairly good code review if you are interested in trying that.
If you really think AngularJS is the way to go with it, please explain why - I am not saying no. I am just hesitant because it is currently on the slow march to EOL.

SubJunk:
@mpiggott is this dead?

mpiggott:
> @mpiggott is this dead?

Sorry, I had done some experimentation with an integration that RestEasy provides to support Netty and pushed it a little while ago. There was a bit more cleanup I have to tackle for it I believe.

I also noticed an occasional exception that appeared on occasion that I need to track down why it occurs.

> @mpiggott my real job is working with AngularJS and Angular every day so I would be able to assist you by answering questions and give fairly good code review if you are interested in trying that.

Yea, I had maintained an AngularJS application for ~5-years but haven't touched it much i a couple years. This was basically do something simple to get it out there.

>  If you really think AngularJS is the way to go with it, please explain why - I am not saying no. I am just hesitant because it is currently on the slow march to EOL.

For me it just seemed an obvious choice as it was already present. Also compared to other frameworks (e.g. react) it doesn't require introducing npm to build it.

From a longer term perspective to me the current web-ui is simple enough that there isn't a big investment in this Angular code.

SubJunk:
@mpiggott those are good points, it is refreshing to not have to compile the code from TypeScript to JavaScript.

Let me know if you want testing or code review done, or if I should wait until it is closer to finished

Yooooomi:
Hello, any news about this ? It clearly feels like the front has to be refreshed :)

SubJunk:
@Yooooomi unfortunately this looks abandoned :/

PR #1807, opened 2019-06-13
Support newer Java versions

SubJunk:
Our use of the deprecated Vector API is breaking support on newer Java so we should fix that, and update the versions on Travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ matrix:
     - os: linux
       jdk: oraclejdk9
     - os: linux
-      jdk: oraclejdk13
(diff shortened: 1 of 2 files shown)


SubJunk:
@valib I have made some good progress towards getting rid of Vectors but have run out of time for now. Feel free to pick it up if you want.

SubJunk:
This is still potentially a good thing since in our version of Java, we are relying on a deprecated import `Vector`, but I don't plan to prioritise it now that we are including JRE 8 in the bundles. Feel free to open it if you want to continue work on it.

PR #1811, opened 2019-06-28
Standalone builds for all OS

SubJunk:


SubJunk:
@valib I think this is working for all operating systems now - I have tested on Windows and macOS, and it just needs testing on Linux. I am thinking of releasing it as UMS 9.0.0-b1 to get testers.

valib:
@SubJunk it is not wrking for me because here https://github.com/UniversalMediaServer/UniversalMediaServer/blob/598c43e127b4e612fd713901327f1da1e53a583b/pom.xml#L1162 are missing definitions to make dir and subdirs for `jre-x64` and `jre-x86`

SubJunk:
@valib ah I had manually copied the JRE files over which made me not notice that problem. I'm fixing it now

SubJunk:
@valib should be fixed now

VaiTon:
Has this been merged? @SubJunk

SubJunk:
@VaiTon not yet, I will release a public beta before merging, to make sure it works well for everyone

SubJunk:
Ok I have been doing some testing and it's working well for me so will release the beta soon unless you object @valib

pom.xml
```diff
@@ -1091,12 +1091,6 @@
 		-->
 		<profile>
 			<id>windows</id>
-			<activation>
-				<activeByDefault>false</activeByDefault>
-				<os>
-					<family>windows</family>
-				</os>
-			</activation>
```
valib: @SubJunk this deletion stopped making the "UMS-setup.exe". Do not ask me why. I realized it when I deleted the Eclipse "target" folder and tried to make distributable files.

SubJunk: Which command do you use? I do `mvn -P windows package`

valib:
I make it in Eclipse and use the standard setting for build from Wiki https://github.com/UniversalMediaServer/UniversalMediaServer/wiki/Development

SubJunk:
@valib before this change, I was unable to build only for Linux via Windows - if I ran `mvn -P linux package` it would build for both Windows and Linux, so that was making the linux builds take twice as long to build. Maybe there's a way to get Eclipse to specify a profile?

PR #1812, opened 2019-07-03
Implemented latest JNA, Chromecast, OSHI and other dependencies

valib:
Need testing on MAC OS because of changes in the Nadahar's code.
I think it should work because all JNA versions since 5.0.0 removed `Pointer.SIZE` and use `Native.POINTER_SIZE` for all calculation of the arrays size and memory location.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -98,7 +98,7 @@
 			Check the org.slf4j:slf4j-api version of the other libraries before upgrading
 		-->
 		<logback-version>1.2.3</logback-version>
-		<surefire-version>2.20</surefire-version>
+		<surefire-version>2.22.2</surefire-version>
 
 		<!--
 			net.java.dev.jna:jna-platform is shared with
@@ -106,7 +106,7 @@
 
 			Check the net.java.dev.jna:jna-platform version of the other library before upgrading
 		-->
-		<jna-version>4.4.0</jna-version>
+		<jna-version>5.3.1</jna-version>
 
 		<!--
 			 use the Windows makensis.exe for Windows builds unless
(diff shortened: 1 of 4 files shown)


SubJunk:
@valib can you please let me know what I should test on macOS? Then I will do it before merging

valib:
@SubJunk actually the testing in macOS is not needed because in the JNA version 4.4.0 the Pointer.SIZE is defined as `Pointer.SIZE=Native.POINTER_SIZE` plus check if is not null to ensure that JNA is properly initialized. This is removed in the later versions so my implementation should work. You can test if in the macOS are corectly called an external dll's like MediaInfo or FFMpeg.

Nadahar:
JNA isn't used for FFMpeg or any other "engine", only for MediaInfo. But, "my" macOS/JNA code is unrelated to that as well. It is used for various information exchange with macOS itself, much like you use for example `kernel32.dll` (and others) to access the Windows API.

I don't remember exactly where it's used, except that I know that it's used by the sleep prevention code.

valib:
@Nadahar thanks for the reply. You are correct that JNA is used only for MediaInfo. My bad that I mentioned other external engines. I have not full view what you implemented for macOS but my changes should be correct and not affect your code as I explained.

SubJunk:
Ok great I'll test and merge.
I'm also putting some time into https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1811 and I think both Linux and Windows are working now.

@valib I'm interested to know your opinion on it - it means that we include JRE in all our releases, so that users are always using the version we included. I think it will solve some of the issues we see on the forum.

SubJunk:
I did some brief testing and it worked. Nice one

PR #1813, opened 2019-07-06
attempt to fix PrependTrackNumbers not working

tcely:
I believe this was an ordering problem.
setTrack() needing to happen before getTrack()

While I'm changing this, check that `renderer` is not `null` before trying to call a function with it.


SubJunk:
@tcely very nice, thanks

tcely:
@SubJunk You're welcome! I tried this against my audio book with `9.0.0-b1` and found that for an as yet unknown reason, it still doesn't work when "Enable the Media Library" is checked.

PR #1822, opened 2019-08-05
PrependTrackNumbers: prepend with media library

### LINKED ISSUE #1821: With "Enable the Media Library" checked track numbers are not prepended
In #1813 I managed to get audio track numbers prepended, but this only works without the "Media Library" enabled. I don't know why that makes a difference currently. Ideally the track number should be stored in the library too and used when appropriate.

This looks like a likely cause for this issue:
https://github.com/UniversalMediaServer/UniversalMediaServer/blob/b97e1b3aba1205405125fad5844a6647ccf79e2f/src/main/java/net/pms/dlna/DLNAResource.java#L2279-L2293

tcely:
Fixes #1821

### DIFF
--- a/src/main/java/net/pms/dlna/DLNAMediaInfo.java
+++ b/src/main/java/net/pms/dlna/DLNAMediaInfo.java
@@ -875,11 +875,7 @@ public class DLNAMediaInfo implements Cloneable {
 						}
 					}
 
-					if (StringUtils.isNotBlank(audio.getSongname())) {
-						if (renderer != null && renderer.isPrependTrackNumbers() && audio.getTrack() > 0) {
-							audio.setSongname(audio.getTrack() + ": " + audio.getSongname());
-						}
-					} else {
+					if (StringUtils.isBlank(audio.getSongname())) {
 						audio.setSongname(file.getName());
 					}
 
--- a/src/main/java/net/pms/dlna/DLNAResource.java
+++ b/src/main/java/net/pms/dlna/DLNAResource.java
@@ -2285,7 +2285,12 @@ public abstract class DLNAResource extends HTTPResource implements Cloneable, Ru
 			media.isAudio() &&
 			StringUtils.isNotBlank(firstAudioTrack.getSongname())
 		) {
-			title = firstAudioTrack.getSongname();
+			title = "";
+			if (mediaRenderer != null && mediaRenderer.isPrependTrackNumbers() && firstAudioTrack.getTrack() > 0) {
+				// zero pad for proper numeric sorting on all devices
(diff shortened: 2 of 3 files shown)


tcely:
@valib I Incorporated your padding change into the first commit.

I don't think this requires a new function. Also, I routinely listen to books with more than 99 tracks so I chose to pad to three digits instead. FYI, my Samsung device does sort correctly without zero padding. I do understand that might just be a matter of luck with my particular model though.

src/main/java/net/pms/dlna/DLNAMediaInfo.java
```diff
@@ -875,11 +875,7 @@ public void parse(InputFile inputFile, Format ext, int type, boolean thumbOnly,
 						}
 					}
 
-					if (StringUtils.isNotBlank(audio.getSongname())) {
-						if (renderer != null && renderer.isPrependTrackNumbers() && audio.getTrack() > 0) {
-							audio.setSongname(audio.getTrack() + ": " + audio.getSongname());
-						}
-					} else {
+					if (!StringUtils.isNotBlank(audio.getSongname())) {
```
SubJunk:
```suggestion
					if (StringUtils.isBlank(audio.getSongname())) {
```

src/main/java/net/pms/dlna/DLNAResource.java
```diff
@@ -2285,7 +2285,12 @@ public final String getDidlString(RendererConfiguration mediaRenderer) {
 			media.isAudio() &&
 			StringUtils.isNotBlank(firstAudioTrack.getSongname())
 		) {
-			title = firstAudioTrack.getSongname();
+			title = "";
+			if (mediaRenderer != null && mediaRenderer.isPrependTrackNumbers() && firstAudioTrack.getTrack() > 0) {
+				// zero pad for proper numeric sorting on all devices
+				title += String.format("%03d: ", firstAudioTrack.getTrack());
```
SubJunk: I think some devices don't accept `:` in filenames (like my TV) so it would be more safe to do like `"%03d - "` (it would be a current bug that we can fix here)

tcely: That seems reasonable to me.

SubJunk:
@tcely thanks!

tcely:
@SubJunk You're welcome! Is a `9.0.0-b3` likely or is the next one supposed to be final?

SubJunk:
@tcely I'm not sure yet, I was hoping to get some feedback on b2 but no one has commented... Maybe no news is good news?

tcely:
b1/b2 has been working well for me

SubJunk:
@tcely thanks!

PR #1823, opened 2019-08-08
Add UMS to Windows Firewall exceptions on install/uninstall

SubJunk:


SubJunk:
@valib I have tested this on Windows 7 and it works. I think it is ready to merge if you agree

valib:
@SubJunk what about to use `PowerShell` instead of `netsh` when Microsoft says    

> In future versions of Windows, Microsoft might remove the Netsh functionality
> for Windows Firewall with Advanced Security.
> 
> Microsoft recommends that you transition to Windows PowerShell if you currently
> use netsh to configure and manage Windows Firewall with Advanced Security.
> 
> Type Get-Command -Module NetSecurity at the Windows PowerShell prompt to view
> a list of commands to manage Windows Firewall with Advanced Security.
> 
> Visit https://go.microsoft.com/fwlink/?LinkId=217627 for additional information
> about PowerShell commands for Windows Firewall with Advanced Security.
> netsh advfirewall

SubJunk:
@valib I looked into this and it seems that might not be supported on Windows 7 since it only has PowerShell 2.0 and that requires PowerShell 4.0 based on the comments after the answer here https://superuser.com/questions/1231205/create-and-toggle-windows-firewall-rule

Maybe I need to look more but I suggest we merge this since it is working on Windows 7 and 10 at least

SubJunk:
@valib do you agree or still want a change to be made here?

SubJunk:
@valib thanks

PR #1824, opened 2019-08-11
Attempt to fix broken startup since 7.6.0

SubJunk:


SubJunk:
This isn't ready yet, I am just trying things out and getting them tested

PR #1826, opened 2019-08-13
Change method name 'find' to 'searchRendererConfiguration'

pdhung3012:
This class is used to represent RenderConfiguration.  This method named 'find' is to search for a RenderConfiguration. Thus, the method name 'searchRendererConfiguration' is more intuitive than 'find'.


SubJunk:
@pdhung3012 hi and welcome to UMS :)
I have to disagree with this approach though. I don't think the name of the class needs to be appended to method names. You already do `RendererConfiguration.find` so it is redundant to add another reference to that.
Please continue to contribute, and if you want to run an idea past us before writing the code it might save time

PR #1846, opened 2019-09-24
Fixed bug reported in

valib:
https://www.universalmediaserver.com/forum/viewtopic.php?f=9&p=40144&sid=aa8911e4a3d3bc307724b316f5ac07de#p40144
When are used hyphens in the filename like `tvr-criminal-minds-s09e07-720p.mkv` instead of periods or underscores then the exception is raised.


valib:
@SubJunk this is only the quick fix and should be solved properly. I see a lot of users using hyphens in the filenames whose make a problem in pretyfying..

PR #1847, opened 2019-09-29
Fixed duplicate logic in OpenSubs validation + logging

SubJunk:


SubJunk:
@valib it doesn't show up nicely in the diff but the main thing this PR fixes is the duplication of https://github.com/UniversalMediaServer/UniversalMediaServer/blob/master/src/main/java/net/pms/util/OpenSubtitle.java#L4849-L4858 below it in https://github.com/UniversalMediaServer/UniversalMediaServer/blob/master/src/main/java/net/pms/util/OpenSubtitle.java#L4870-L4879
Must have been a merge conflict at some point

Issue #1879, opened 2020-01-01
java.lang.IndexOutOfBoundsException: Index: 0, Size: 0


SubJunk:
I started getting this error in the logs after https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1828 was merged @valib 

```
10:46:24 ERROR Error adding child: "file.ts"
java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
	at java.util.ArrayList.rangeCheck(Unknown Source)
	at java.util.ArrayList.get(Unknown Source)
	at net.pms.dlna.RealFile.getFile(RealFile.java:127)
	at net.pms.dlna.RealFile.checkThumbnail(RealFile.java:324)
	at net.pms.dlna.RealFile.isValid(RealFile.java:90)
	at net.pms.dlna.DLNAResource.addChild(DLNAResource.java:625)
	at net.pms.dlna.MapFile.manageFile(MapFile.java:279)
	at net.pms.dlna.MapFile.analyzeChildren(MapFile.java:347)
	at net.pms.dlna.RootFolder.scan(RootFolder.java:293)
	at net.pms.dlna.RootFolder.scan(RootFolder.java:251)
	at net.pms.dlna.DLNAMediaDatabase.run(DLNAMediaDatabase.java:1539)
	at java.lang.Thread.run(Unknown Source)
```

It seems to only happen with broken files (that particular file is 0 bytes) so it is good it is not added to the library, but we shouldn't get that exception

PR #1880, opened 2020-01-01
Do not add corrupted file to the cache

valib:
fix the #1879

### DIFF
--- a/src/main/java/net/pms/dlna/MapFile.java
+++ b/src/main/java/net/pms/dlna/MapFile.java
@@ -271,7 +271,7 @@ public class MapFile extends DLNAResource {
 					} else {
 						// Otherwise add the file
 						RealFile rf = new RealFile(f);
-						if (rf.length() == 0  && !f.isDirectory()) {
+						if (rf.length() == 0  && !rf.isFolder()) {
 							LOGGER.debug("Ignoring {} because it seems corrupted when the lenght of the file is 0", f.getName());
 							return;
 						}
--- a/src/main/java/net/pms/dlna/MapFile.java
+++ b/src/main/java/net/pms/dlna/MapFile.java
@@ -271,7 +271,7 @@ public class MapFile extends DLNAResource {
 					} else {
 						// Otherwise add the file
 						RealFile rf = new RealFile(f);
-						if (rf.length() == 0  && !rf.isFolder()) {
+						if (rf.length() == 0  && !f.isDirectory()) {
 							LOGGER.debug("Ignoring {} because it seems corrupted when the lenght of the file is 0", f.getName());
 							return;
 						}
--- a/src/main/java/net/pms/dlna/MapFile.java
+++ b/src/main/java/net/pms/dlna/MapFile.java
@@ -271,7 +271,12 @@ public class MapFile extends DLNAResource {
 					} else {
 						// Otherwise add the file
 						RealFile rf = new RealFile(f);
-						//we need to propagate the flag in ordet to make all hierarchy stay outside the media library if needed
+						if (rf.length() == 0  && !rf.isFolder()) {
+							LOGGER.debug("Ignoring {} because it seems corrupted when the lenght of the file is 0", f.getName());
+							return;
+						}
+
+						//we need to propagate the flag in order to make all hierarchy stay outside the media library if needed
 						rf.getConf().setAddToMediaLibrary(this.getConf().isAddToMediaLibrary());
 						if (searchList != null) {
 							searchList.add(rf);
--- a/src/main/java/net/pms/dlna/MapFile.java
+++ b/src/main/java/net/pms/dlna/MapFile.java
@@ -272,7 +272,7 @@ public class MapFile extends DLNAResource {
 						// Otherwise add the file
 						RealFile rf = new RealFile(f);
 						if (rf.length() == 0  && !rf.isFolder()) {
-							LOGGER.debug("Ignoring {} because it seems corrupted when the lenght of the file is 0", f.getName());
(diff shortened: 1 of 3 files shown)


valib:
@SubJunk why do we need to check the Travis with the Open JDK 8 when we provide the Java 8 with the UMS as the official working Java version?

SubJunk:
@valib quick fix, thanks! I agree about OpenJDK but it is still good to run tests on Linux too so I might try fixing the Travis failure a different way, by running Oracle JDK on Linux

Nadahar:
Since OpenJDK is more or less the standard on Linux, and increasingly so on other platforms too since the OracleJDK has become payware, I think it's strange to simply dismiss it as "not official". I doubt there are very many UMS users that use commercial Oracle Java.

I'm not even convinced the issue is due to the JDK, it seems like a Maven issue to me. There are probably different Maven versions running on the Linux and macOS VMs.

SubJunk:
From UMS 9 we bundle Oracle JDK 8 with UMS releases, so all users are using that unless they delete it

src/main/java/net/pms/dlna/MapFile.java
```diff
@@ -271,7 +271,12 @@ private void manageFile(File f, boolean isAddGlobally) {
 					} else {
 						// Otherwise add the file
 						RealFile rf = new RealFile(f);
-						//we need to propagate the flag in ordet to make all hierarchy stay outside the media library if needed
+						if (rf.length() == 0  && !rf.isFolder()) {
+							LOGGER.debug("Ignoring {} because it seems corrupted when the lenght of the file is 0", f.getName());
```
SubJunk:
```suggestion
							LOGGER.debug("Ignoring {} because it seems corrupted when the length of the file is 0", f.getName());
```

SubJunk:
I've tested this fix and it works, but I have the same error caused by another problematic file so I wonder if this is the best way to fix it:

```
15:24:09 INFO  The file song.mp3 could not be parsed. It will be hidden
15:24:09 ERROR Error adding child: "song.mp3"
java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
	at java.util.ArrayList.rangeCheck(Unknown Source)
	at java.util.ArrayList.get(Unknown Source)
	at net.pms.dlna.RealFile.getFile(RealFile.java:127)
	at net.pms.dlna.RealFile.checkThumbnail(RealFile.java:324)
	at net.pms.dlna.RealFile.isValid(RealFile.java:90)
	at net.pms.dlna.DLNAResource.addChild(DLNAResource.java:625)
	at net.pms.dlna.MapFile.manageFile(MapFile.java:284)
	at net.pms.dlna.MapFile.analyzeChildren(MapFile.java:352)
	at net.pms.dlna.RootFolder.scan(RootFolder.java:293)
	at net.pms.dlna.RootFolder.scan(RootFolder.java:303)
	at net.pms.dlna.RootFolder.scan(RootFolder.java:303)
	at net.pms.dlna.RootFolder.scan(RootFolder.java:303)
	at net.pms.dlna.RootFolder.scan(RootFolder.java:303)
	at net.pms.dlna.RootFolder.scan(RootFolder.java:251)
	at net.pms.dlna.DLNAMediaDatabase.run(DLNAMediaDatabase.java:1539)
```

We already encounter a problem and log "could not be parsed. It will be hidden", so maybe we can use that information to fix this, instead of adding exceptions into this `manageFile` function

What do you think @valib ?

valib:
@SubJunk I think that this is the best way to deal with the files with zero length. This is done before this file is added to the cache and will be parsed which take some time. For other problematic files with non zero lehgth I will try to find propper place to fix it.

SubJunk:
Ok cool. The other problematic file seems to have a "Could not parse bitrate" error since MediaInfo can't read it, but the file plays in VLC. Let me know if you want me to upload the file for you

valib:
@SubJunk yes, it will be nice you upload the problematic file for testing.

SubJunk:
@valib [badbitrate.zip](https://github.com/UniversalMediaServer/UniversalMediaServer/files/4050417/badbitrate.zip)

SubJunk:
@valib could you reproduce with that file?

valib:
@SubJunk yes I can confirm that this file has a problem but how to fix it? This file seems to have broken metadata and MediaInfo gives only `Complete name                            : E:\Filmy\Audio and picture\Bro Safari & UFO - Burn The Block.mp3
File size                                : 25.1 MiB` and that is all. For me this file also causes problem in WIN 10 because the right click on it takes almost a minute to show the context menu. I think that the OS is trying to read the broken metadata and it takes long time. This is not the task for the UMS to solve it and this media file must be repared before it will be used by the UMS.

valib:
@SubJunk this is the file like it shoud be 
[repared media.zip](https://github.com/UniversalMediaServer/UniversalMediaServer/files/4162145/Bro.Safari.UFO.-.Burn.The.Block.zip)

SubJunk:
@valib users are going to have problematic files, it's how we handle them that matters to me

valib:
@SubJunk this fix the problem with the file you provided. The only problem is that it takes longer time to parse it.

SubJunk:
@valib nice work, how long does it take? Is it an option to just hide the file instead of trying to parse it?

valib:
@SubJunk it takes longer time because after false MediaInfo parsing it is trying to parse it in other way. If the media is parsed than it is added as any other media parsed by MediaInfo. If it fails in other parsing it is ignored like before.

valib:
@SubJunk do you think that this fix is worth to be parsed?

SubJunk:
@valib yes it seems like a good thing that hopefully won't make it slower for most users. Thanks for your work on it

PR #1883, opened 2020-01-08
Update latest_version.properties

xaitax:
Switch to HTTPS


SubJunk:
Tested and it works, thanks!

PR #1665, opened 2018-11-12
mediainfo and  ffmpeg tests

stasinos:
Towards issue #1651

Issue #1856, opened 2019-10-12
AVC High@L4 are recognized as High@L5, starts transcoding


ltGuillaume:
I found that UMS started transcoding several files for my TV (`samsung-notcd.conf` renderer profile and supports L4.1), even though these files can be played back perfectly fine in untouched form.

The debug log shows
```
TRACE 2019-10-12 22:24:03.970 [HTTPv2 Request Worker 4] MediaInfo parsing results for "file.mkv":
  Format: "Matroska"            Duration: 0:46:37.962         Overall Bitrate: 11751383     Cover: No
  FPS: 23.976
  - Video track ID: 1
    Format: "AVC"                 Profile: "High@L5"            CodecID: "V_MPEG4/ISO/AVC"    Duration: 0:46:37.965
    Bitrate: 10877530             Width: 1920                   Height: 1080                  Colorimetry: 4:2:0
    Chroma: 4:2:0                 Matrix Co: BT.709             PAR: 1.000                    DAR: 16:9
    Scan Type: Progressive        FPS: 23.976                   Framerate Mode: CFR           RefFrames: 5
    Bit Depth: 8                  Delay: 0                      Delay Source: Container
  - Audio track ID: 2
    Format: "E-AC-3"              CodecID: "A_EAC3"             Duration: 0:46:37.962         BitRate Mode: CBR
    Bitrate: 640000               Language: "en"                Channel(s): 6                 Samplerate: 48000
    Delay: 0                      Delay Source: Container
  - Subtitle ID: 3, Title: "English"
    Format: "UTF-8"               Language: "en"
  - Subtitle ID: 4, Title: "English (SDH)"
    Format: "UTF-8"               Language: "en"

[...]

PR #1889, opened 2020-01-15
1651 mediainfo ffmpeg tests

SubJunk:
This is a version of https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1665/commits on our repo so we can work on it


SubJunk:
@valib this is passing on Travis now which is good progress

SubJunk:
Thanks to @JeromeMartinez the framerate parsing is working in the MKV file now.

It seems we have discussed that setting before https://github.com/UniversalMediaServer/UniversalMediaServer/commit/d9198bf02223077911d4127a4619d4ea92df1c6f#commitcomment-9061386 so I wonder if it is ok to leave it at default or if it needs a new value.

Opinions?

ExSport:
When I see posts about missing/incorrect GOP, framerate, bitrate, etc. it always come to my mind "parsespeed" :) There are lot of pros/cons to set it to 0 or 0.1 but it seems the behaviour changes in time (if framerate was shown with parsespeed=0 but newer versions of mediainfo stopped to show it? It is silly or like breaking change) :)
If some people have files in GB sizes like 15-20GB, parsing 10% of every file (0.1) will take tenth of seconds or minutes instead of milliseconds. More such files for scan? Lot of renderers will go into problems. I remember lot of timeouts on Panasonic TV when I tried to open folder from HD camcorder with lot of very big TS files which led to unresponsive UMS and renderer timeouts. Definitely the parsespeed value should be as low as possible (for speed but high enough to get relevant/required data).
https://github.com/UniversalMediaServer/UniversalMediaServer/issues/1856#issuecomment-541830066

valib:
This change is working and I tested the speed with the parsespeed=0 and without it and the speed seems the same on my testing files. Seems like MediaInfo changed its behaviour but as @ExSport said it needs more testing on a large files.

JeromeMartinez:
> it seems the behaviour changes in time

I don't remind about a specific change about that, I just try to fix some issues when I have reports of problems, but the idea is still same.

> If some people have files in GB sizes like 15-20GB, parsing 10% of every file (0.1) will take tenth of seconds or minutes instead of milliseconds.

0.1 does not mean 10% of the file here, this is more a "level of parsing". It would be few first frames.
0.5 (the default) is about 10 seconds of content (trying to detect frame rate and GOP).

> When I see posts about missing/incorrect GOP, framerate, bitrate, etc.

Issue here is lack of global metadata in MKV. For MKV created by mkvtoonix, it is fine as there are statistics metadata (including frame rate), but for MKV created by FFmpeg there is no such metadata (and for a reason I ignore there is also no frame rate metadata in AC or HEVC stream... :( ) so guessing the frame rate requires lot of frames (29.97 is not same as 30000/1001 for some people so I need lot of frames for detecting the difference). I could maybe try to read less content (I need only the frame header so I could seek to the next frame instead of reading the full frame by default) but it is some dev time on my side, and I currently don't have such time (and not a priority for my sponsors, please remind that MediaInfo is what pays my bills).
Similar for GOP.

> with lot of very big TS files 

I see how I could optimize and still compute info for MKV, but for TS I see no solution. It would be less problematic as usually frame rate metadata is in the stream so what you lose is just the GOP.
There could be a small optimization for getting the duration, by trying by default less data at the end (smarter seek), but same it is a problem of development time :(.

If there is a difference between MI versions for ParseSpeed = 0 and frame rate, I am interested in quickly check why there was a change and if it is easy to revert without introducing another issue.

ExSport:
@JeromeMartinez Hello. Sorry for my misleading info. I copy pasted info about percents from this post: https://github.com/UniversalMediaServer/UniversalMediaServer/issues/1856#issuecomment-541939442
Thanks for your detailed info what it is happening here :+1: 
Btw. in past ( few years ago) I did some benchmarks on parsespeed 0 and 0.5 with TS files about 100-200MB in size (folder with hundred of files) and parsing with the value 0 was about 100ms but 0.5 took up to 2 seconds so it was easy to wrongly think 0.5 really means 50% :)
I remember I wanted to donate to your great project in past (many years ago) but if I remember well there was no donation link (e.g. Paypal). Only possibility was to open some special portal, register bug there and wait for other sponsors to push bug into your interest. But the easy way to sponsor "without obligation" was missing. Anything changed here? :)

JeromeMartinez:
> but if I remember well there was no donation link (e.g. Paypal). Only possibility was to open some special portal, register bug there and wait for other sponsors to push bug into your interest. But the easy way to sponsor "without obligation" was missing. Anything changed here? :)

[Voting](https://mediaarea.net/Vote) is optional, no obligation, and the [sponsoring page](https://mediaarea.net/SupportUs/Individual) permits you to donate (Paypal, credit card, transfer) without being registered (anonymous donation).

SubJunk:
@valib what do you think about merging this? It can be improved upon in more branches but there are some good improvements in here already

PR #1897, opened 2020-02-02
Removed unnecessary test

SubJunk:
We don't need Java to be on the PATH anymore


Nadahar:
Tests aren't run by users, they are run when building. Don't you still need Java to build? I guess you never actually needed Java to be on the PATH per se, you can probably get around it using environmental variables and specifying paths. My point is that I can't see how bundling the JRE makes a difference to the build requirements.

SubJunk:
Yep that's a valid point, the test probably never needed to be there

Nadahar:
The purpose of the test wasn't to make sure Java was available - the test was to test the OS path resolve function, and Java was the only executable I could think of that "always could be found on any system". I guess one could specify some standard system file per platform instead, if one figured out one that would always be present on the OS path.

PR #1902, opened 2020-02-05
Bump dependency plugin

SubJunk:


SubJunk:
Not sure why this is failing on Travis still, since it has worked for me on two computers, maybe just a matter of waiting for a cache to expire

Nadahar:
@SubJunk Whatever you have done, version `0.3` isn't available on Maven central: https://search.maven.org/search?q=g:com.universalmediaserver%20AND%20a:external-maven-plugin%20AND%20v:0.3

It seems like you have tried to link to the snapshot, which is very bad practice. You should release it so that it's available on central.

SubJunk:
@Nadahar can you test this branch locally? It works for me and has worked on another person's machine too

SubJunk:
@Nadahar can you see it on this page? https://oss.sonatype.org/#nexus-search;quick~com.universalmediaserver

Nadahar:
The artifact has at least been released now :+1: 

Just make sure that you revert the SNAPSHOT changes in the POM, so that is uses the "official" version.

SubJunk:
@Nadahar I don't think I released the snapshot version - the `pom` on central has the correct version https://search.maven.org/remotecontent?filepath=com/universalmediaserver/external-maven-plugin/0.3/external-maven-plugin-0.3.pom

Nadahar:
@SubJunk Sorry, it seems I jumped to conclusions *and* expressed myself inaccurately. I just took a quick peek at the failed Travis build, including this line:
```powershell
[ERROR] No plugin found for prefix 'external' in the current project and in the plugin groups [org.apache.maven.plugins, org.codehaus.mojo] available from the repositories [local (/home/travis/.m2/repository), ossrh (https://oss.sonatype.org/content/repositories/snapshots), google-maven-central (https://maven-central.storage-download.googleapis.com/maven2/)] -> [Help 1]
```

Since I couldn't see why `https://oss.sonatype.org/content/repositories/snapshots` was in the "Maven search path", I assumed that you had uploaded to sonatype only relied on fetching it from there. I further assumed that you had made some modifications to UMS' POM to make it fetch it from there.

When looking at the commits now, I see that this was all in my imagination :wink: With my "snapshot" reference, I didn't actually mean that you had pushed a snapshot version, but that you used the "snapshot repository" instead of Maven central. This is where artifacts go when you "release" them, before you promote them to go to central.

I'm pretty sure it's impossible to actually deploy a snapshot version to Maven central, so that idea never entered my thoughts.

Anyway, since the artifact is available on Maven central now, all should be OK - regardless of what the original confusion was caused by.

SubJunk:
It's a confusing process but I'm glad we got to the end of it, thanks again for your help!

Issue #1901, opened 2020-02-05
Change fully played action option from video to media in GUI


polygirl:
On the `navigation settings` tab there's a dropdown for `fully played action`.

One option is `hide video` 

For consistency with `mark media` and for relevance to other filetypes this should be updated to `hide media`

- [ ] option should say `hide media`

valib:
@polygirl good idea. Thanks

SubJunk:
This is merged and I made a followup story to make it actually do that

PR #1903, opened 2020-02-05
Change GUI recommended in issue #1901

valib:


Nadahar:
@valib Have you tested this? I think you also have to create the code to hide other media types, I think it's only implemented for video.

SubJunk:
I'm not sure how complete the implementation is but there is code that does the "marking" regardless of format https://github.com/UniversalMediaServer/UniversalMediaServer/blob/master/src/main/java/net/pms/dlna/DLNAResource.java#L2291

It could probably use some more testing with audio files to find any gaps with the feature but that doesn't need to hold up this PR.

Thanks @valib

SubJunk:
@Nadahar yes I think you're right that only video files are hidden at the moment. That should probably be worked on, I'll make a separate issue

Nadahar:
As I remember it, it was per (your) design - which is why we kept "video" in the text as well.

Nadahar:
When you think about it, it's not really that strange. How likely is it that you want music hidden once it has been played *once*, or images once they have been viewed *once*? It seems to me a bit hard to see in which situations users would actually want this functionality for anything but videos. I guess the same argument could be made for some of the other options too though.

SubJunk:
Yeah I agree it's not likely to be something anyone has missed

PR #1906, opened 2020-02-07
Support for symlinks

SubJunk:


SubJunk:
This approach works perfectly on Oracle Java because it has JavaFX, and it also makes the file chooser much nicer (exactly like the native one) but I'm currently working on a different implementation so that we can continue to support OpenJDK users. I tried to do a "best of both" situation where we use JavaFX if it exists, otherwise use Swing, but my efforts stalled.

Maybe this will get resurrected in the future - if OpenJDK start including JavaFX or we can start including it as a third party, it would be a good improvement to make.

In the meantime I'm closing it.

PR #1933, opened 2020-02-27
Fix PS3 WAV

SubJunk:


src/main/external-resources/renderers/Sony-PlayStation3.conf
```diff
@@ -44,8 +44,7 @@ Supported = f:lpcm    a:lpcm         n:6   s:48000   m:audio/L16
 Supported = f:m4a     a:(?!alac).+   n:2             m:audio/x-m4a
 Supported = f:m4a                    n:2   s:48000   m:audio/x-m4a
 Supported = f:mp3                    n:2             m:audio/mpeg
-Supported = f:wav     a:dts          n:6   s:48000   m:audio/wav
-Supported = f:wav     a:lpcm         n:6   s:48000   m:audio/L16
+Supported = f:wav     a:dts|lpcm     n:6   s:48000   m:audio/wav
 Supported = f:wav                    n:2   s:48000   m:audio/wav
```
Nadahar: @SubJunk I'm pretty sure this last line is pretty pointless/wrong. I'm curious why it's been put here in the first place, but it's ancient history and we'll probably never know. The point is that according to the PS3 manual, only LPCM based WAVEs are supported. Since DTS is explicitly listed, I assume that some have found that to work as well. But, this means that I can't see what those "remaining" WAVEs that haven't matched the above line could be that is still supported by the PS3. It would have to mean "some other codec", and when Sony explicitly lists `WAVE(Linear PCM)`, why would we think that other codecs are supported?

SubJunk: PS3 manual doesn't match what they support, it was all based on thorough testing originally to discover what they supported. Subsequent updates have likely changed the support as they have fixed bugs along the way.

Nadahar: @SubJunk I know this, I've heard it many times before. At the same time, I've discovered lots and lots of things that it's correct, also for the PS3. I think the "thoroughness" of the testing is over-estimated, it isn't easy to do after all, considering the amount of potential variations needed.

While I'm aware that the specs aren't always correct, I think they are generally a good guideline for what they have intended to support. As such, I can't imagine that the PS3 supports WAV with MP3 payload for example, which is what this line effectively claims.

The attached files should be supported according to this specification, can you get them to stream to the PS3?

Issue #1939, opened 2020-03-08
Recent external subtitles bug


SubJunk:
@valib please see https://www.universalmediaserver.com/forum/viewtopic.php?f=9&t=14007 it may be related to the recent work on subtitles support. The user attached a nice trace log

PR #1940, opened 2020-03-08
Fix Samsung-9series.conf

valib:
It may fix the issue reported in  #1939
There was implemented the `fps:` which is not supported by the UMS but the main problem was using `v:` instead of the `w:` for the video width which in fact overrides the video format value with the not acceptable value like `1920` instead of e.g. `h264` parsed before.


SubJunk:
@valib I'm posting this to the user, but he says that 9.0.1 works for him, so I'm not sure if this will be the right fix. We will see when he tests :) thanks

SubJunk:
I'm merging this because it is still a correct change even if that user still has problems

PR #1946, opened 2020-03-20
Linux Service file

luca-vercelli:
This file is required for running UMS at startup as a daemon


SubJunk:
@luca-vercelli thanks for this contribution and welcome!

Since this file requires so much customization to be useful (the path and username need to change) I think it would be better as an article in our Wiki, like `How to start UMS as a service`.

We are attempting to consolidate all of our help documentation into one place soon, but that is not ready yet, so until then the best place is the wiki.

Issue #1886, opened 2020-01-08
Error when transcoding with current FFmpeg, fixed in new ones


SubJunk:
When using a command like:

```
ffmpeg -y -loglevel info -i video.mkv -filter_complex [0:v][0:s:0]overlay,pad=iw:iw/(16/9):0:(oh-ih)/2,scale=1920:1080 -bufsize 31250k -maxrate 31250k -crf 19 -ac 6 -ab 640k -c:a ac3 -c:v libx264 -tune zerolatency -level 31 -pix_fmt yuv420p -f mpegts -preset superfast file.ts
```

I get the error:

```
Unable to parse option value "(null)" as sample format
```

and the transcoding fails.
I have confirmed it is fixed with newer FFmpeg versions, with no changes to the commands, so we should try to update to a newer version.

valib:
@SubJunk have you tried to use example from  https://trac.ffmpeg.org/wiki/HowToBurnSubtitlesIntoVideo especially `"[0:v][0:s]overlay[v]" -map "[v]" -map 0:a` ?

SubJunk:
@valib I have referenced that article before but not this time

SubJunk:
Fixed in #1951

PR #1951, opened 2020-03-28
Use system FFmpeg on Linux

SubJunk:
Now we will always use the system FFmpeg if it exists, and otherwise we use our one, which will be platform-specific as we will now offer 5 different Linux packages

Also bumps versions on macOS and Windows


SubJunk:
This also fixes https://github.com/UniversalMediaServer/UniversalMediaServer/issues/1886

Issue #1894, opened 2020-01-28
Make Watched and Unwatched folders nested


SubJunk:
Before:
Media Library > Videos > Unwatched > TV Shows

After:
Media Library > Videos > TV Shows > Unwatched

etc.

PR #1918, opened 2020-02-19
v10, API integration and web interface improvements

SubJunk:

PR #1961, opened 2020-04-07
Nested media library

SubJunk:
Implements #1894 
Also makes the Unwatched and Watched folders dynamic
Fixes ordering bug


SubJunk:
I'm closing this so it can be bundled with https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1918 targeting the 10.0.0-b1 milestone

PR #1965, opened 2020-04-08
Fix table CoverArtArchive to work with h2db version 1.4.200

valib:
The latest version of h2database is more strict to the SQL syntax.


SubJunk:
@valib nice catch, did you check the other tables to see if this could affect them too?

valib:
@SubJunk yes. The rest is OK.

PR #1970, opened 2020-04-11
Test more video formats

SubJunk:


SubJunk:
@valib I added this line https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1970/files#diff-44d24470f2cc9c487cc39ffbefbdf41bR1469 to get the right MIME type on an MP4 file, but do you remember if there is a reason we don't already do that?

There is a lot of logic below to send that MIME type based on the video codec, but it seems more correct to me to do it this new way, what do you think?

valib:
@SubJunk sems to me that the Travis for windows JDK (but actually running on Linux) is giving different media bitrate for video-h265.mp4. For me the MediaInfo returns media bitrate 404647 but Travis is expecting 404646. On the OSMac it works. What is funny that only this media has a problem.

SubJunk:
@valib I noticed that too with newer versions of MediaInfo - it's important to use 18.12.

My goal is that when we have enough files tested like this (I will keep adding more videos and some audio files) then we can update to the latest MediaInfo 20.03 with confidence that it won't break things.

src/test/java/net/pms/dlna/DLNAMediaInfoTest.java
```diff
@@ -91,67 +91,67 @@ public void testContainerProperties() throws Exception {
...
 		);
 
 		DLNAResource mkvHevc10bitVideo = PMS.getGlobalRepo().get(test_content[2]);
 		DLNAMediaInfo mkvHevc10bitVideoMediaInfo = mkvHevc10bitVideo.getMedia();
 		assertThat(mkvHevc10bitVideoMediaInfo.toString()).isEqualTo(
-			"Container: MP4, Size: 706007, Video Bitrate: 404646, Video Tracks: 1, Video Codec: h265, Duration: 0:00:13.958, Video Resolution: 1920 x 800, Display Aspect Ratio: 2.39:1, Frame Rate: 24.000, Frame Rate Mode: CFR (CFR), Frame Rate Mode Raw: CFR, Video Track Title from Metadata: hevc:fps=24@GPAC0.5.1-DEV-rev4807, Mime Type: video/mp4"
+			"Container: MP4, Size: 706007, Overall Bitrate: 404647, Video Tracks: 1, Video Codec: h265, Duration: 0:00:13.958, Video Resolution: 1920 x 800, Display Aspect Ratio: 2.39:1, Frame Rate: 24.000, Frame Rate Mode: CFR (CFR), Frame Rate Mode Raw: CFR, Video Track Title from Metadata: hevc:fps=24@GPAC0.5.1-DEV-rev4807, Mime Type: video/mp4"
```
SubJunk, line 106:
```suggestion
			"Container: MP4, Size: 706007, Overall Bitrate: 404646, Video Tracks: 1, Video Codec: h265, Duration: 0:00:13.958, Video Resolution: 1920 x 800, Display Aspect Ratio: 2.39:1, Frame Rate: 24.000, Frame Rate Mode: CFR (CFR), Frame Rate Mode Raw: CFR, Video Track Title from Metadata: hevc:fps=24@GPAC0.5.1-DEV-rev4807, Mime Type: video/mp4"
```

src/main/java/net/pms/dlna/DLNAMediaInfo.java
```diff
@@ -1766,7 +1772,7 @@ public String toString() {
 		}
 		result.append("Size: ").append(getSize());
```
SubJunk, line 1773:
```suggestion
		result.append("Size: ").append(getSize());
		result.append(", Overall Bitrate: ").append(getBitrate());
```

src/main/java/net/pms/dlna/DLNAMediaInfo.java
```diff
@@ -1766,7 +1772,7 @@ public String toString() {
...
 		if (isVideo()) {
-			result.append(", Video Bitrate: ").append(getBitrate());
+			result.append(", Overall Bitrate: ").append(getBitrate());
```
SubJunk, line 1775:
```suggestion
```

valib:
@SubJunk for me also version 18.12 gives the same error (different bitrate) for the file video-h265.mp4.

SubJunk:
@valib that's weird about the version... Have you tried printing like `mediainfo -v` in the test file to make sure? I had a similar bug recently. I can't think of anything else that could cause that bitrate problem

valib:
@SubJunk sorry I was too fast. After retesting the persists.

SubJunk:
@valib it looks like Maven isn't running the tests on our Windows build on Travis - it just builds the project. It would be good if we can fix that to try to reproduce this. Are you interested in trying that?

PR #1972, opened 2020-04-14
More mediainfo testing, continue  #1970

valib:
It continues the  #1970


valib:
@SubJunk it seems that not only Windows but alo OSX skip the Mediainfo test. Either the MediaInfo is not installed in the Travis environment or there is missing the path for it.

SubJunk:
@valib that makes sense, I will have broken that on macOS when I was fixing it on Linux :p
So MediaInfo is missing for macOS and Windows. I can work on that later if you haven't fixed it already

SubJunk:
We should also try to make it so that the test fails if MediaInfo is not found, instead of just silently passing

SubJunk:
@valib ok so they are all doing their jobs now, and Windows fails as you said. So we have two options - try to update to a newer version of MediaInfo and see if that is fixed, or find another file that doesn't have that bug. I'm leaning towards the second option because a very tiny bitrate difference isn't going to cause us any problems, while an update could.
I want to update MediaInfo when we have more files being tested here. So I'll try a new file now

valib:
@SubJunk tested and works. You can merge it.

SubJunk:
@valib do you think there are enough sample files here now to be confident with upgrading MediaInfo or should I keep adding more? Are there any formats you think we should prioritize?

PR #1973, opened 2020-04-16
Improve SQL creation by partial automatisation

valib:


SubJunk:
@valib thanks this is a good improvement! Can I please ask that if you have more improvements to database stuff, could you put it in the `api-3` branch? That way we don't create conflicts for that branch

Issue #1172, opened 2017-01-30
Server name is not XML escaped


Goten87:
i have gave my profile the server name Movies & Series and now its not showing up on my net work at all
when i rename it again to Movies and Series it works
can u please fix the & symbol in the Server name input

or do i need to do something special for it ?

Nadahar:
This sounds like a bug - the server name string should probably be XML escaped. As a workaround you can try to substitute ```&amp;``` for ```&```. If that works, please report back as it also confirms the bug.

Goten87:
yep the `&amp;` works like a charm

PR #1979, opened 2020-04-19
Proper escaping of friendly server names

snicket2100:
Without it, using "&" in the friendly server name did result in UMS producing invalid XML documents, which then allegedly used to break some clients.

This should fix https://github.com/UniversalMediaServer/UniversalMediaServer/issues/1172. I wasn't able to verify it, as my tv is handling these invalid XML documents properly anyway, but I was able to verify that produced XML documents became valid and that it didn't break anything for my tv either.


SubJunk:
@snicket2100 thank you!

PR #1981, opened 2020-04-20
Add AV1 video format

valib:


SubJunk:
@valib can we find a small sample file to add for testing? I think it would be good to add them whenever we add detection for a new format

src/main/java/net/pms/dlna/LibMediaInfoParser.java
```diff
@@ -696,6 +696,8 @@ protected static void setFormat(StreamType streamType, DLNAMediaInfo media, DLNA
 				format = FormatConfiguration.VC1;
 		} else if (value.equals("au") || value.equals("uLaw/AU Audio File")) {
 			format = FormatConfiguration.AU;
+		} else if (value.equals("av01")) {
```
SubJunk, line 699: The string I can see in the MediaInfo output is `av1` not `av01`, is there a reason to have it like this?

valib: @SubJunk `AV1` is a Format and Codec ID is `av01`. I updated the parsing

SubJunk:
Nice work!

SubJunk:
@valib it would be good to add that sample you mentioned that gets identified differently too. If the file is too big I can try splitting it

valib:
@SubJunk done in #1983

SubJunk:
@valib thanks!

PR #1982, opened 2020-04-21
Fix NPE in FFmpegVideo for media containing only video stream

valib:


valib:
@SubJunk do you know how to restart the Travis when the test fails?

SubJunk:
@valib I have this restart button on the right side
image
Has it disappeared for you?

valib:
@SubJunk yes it disappeared. I have this
screen
without t

SubJunk:
@valib it looks like there is an active forum post about it at https://travis-ci.community/t/job-restart-button-are-not-always-appearing-on-build-page/5811/14
A couple of users found it was fixed after logging out and in

valib:
@SubJunk thanks for info. Logging out and in works. But this is a Travis problem which should be solved.

PR #1985, opened 2020-04-23
Implement AC-4 audio format

valib:


src/test/java/net/pms/dlna/DLNAMediaInfoTest.java
```diff
@@ -231,6 +231,11 @@ public void testContainerProperties() throws Exception {
 						"Container: MKV, Size: 7799945, Overall Bitrate: 10372267, Video Tracks: 1, Video Codec: h265, Duration: 0:00:06.016, Video Resolution: 1920 x 1080, Display Aspect Ratio: 16:9, Frame Rate: 59.940, Frame Rate Mode: CFR (CFR), Frame Rate Mode Raw: CFR, Matrix Coefficients: BT.2020 non-constant, Video Bit Depth: 10, Audio Tracks: 1 [Audio Codec: Enhanced AC-3, Bitrate: 640000, Channels: 6, Sample Frequency: 48000 Hz], Mime Type: video/x-matroska"
 					);
 					break;
+				case 24:
+					assertThat(mediaInfo.toString()).isEqualTo(
+						"Container: MP4, Size: 23449234, Overall Bitrate: 2608913, Video Tracks: 1, Video Codec: h264, Duration: 0:01:11.905, Video Resolution: 1280 x 720, Display Aspect Ratio: 16:9, Scan Type: Progressive, Frame Rate: 29.970, Frame Rate Mode: CFR (CFR), Frame Rate Mode Raw: CFR, Reference Frame Count: 2, AVC Level: 4, AVC Profile: main, Audio Tracks: 1 [Audio Codec: ac4, Bitrate: 128000, Channels: 2, Sample Frequency: 48000 Hz], Mime Type: video/mp4"
```
SubJunk, line 236: This shows that we are incorrectly detecting it as 2 channels. I can see in MediaInfo that it correctly detects 5.1 channels so this is our bug

SubJunk (changes requested):

src/test/java/net/pms/dlna/DLNAMediaInfoTest.java
valib: @SubJunk I checked the code and it seems to me that this is a MediaInfo bug not ours. For this particular media it is not returning channels number only channel mode (5.1) and channel layout. When the MediaInfo returns an empty string for "Channel(s)" the UMS sets the default number to 2. When I introduced this PR the test was working because there was a previous MediaInfo version

SubJunk: @valib I did a bit of reading online and I think that is the way AC-4 works. It can have "substreams" so I think one audio track can have different channels. I'm not exactly sure of the details. There are some more samples here but I don't think they have multiple substreams either https://trac.ffmpeg.org/ticket/8349
Maybe we should parse that "channel mode" instead for this

valib: @SubJunk it seems to me that there is not such parameter to get this value. It works for audio but there is not parameter for substreams. Tested and it returns empty string.

valib:
Latest MediaInfo detects audio channels correctly. It also detects substreams but we don't use it. I have no idea how to offer it to renderers.

SubJunk:
@valib what is the goal of this PR? To support renderer configs having "ac4" in the "supported" lines?

SubJunk:
@valib what is this waiting on?

valib:
@SubJunk maybe it could be merged. It allows to play this format on supporting renderers like my Samsung TV but the Ffmpeg is not able to transcode it and returned error.

SubJunk:
I see. I guess in main now, those files would still fail to play, so this would not makes things worse for players that do not support ac4. So yeah let’s merge it

src/main/external-resources/renderers/Samsung-OLED.conf
```diff
@@ -1,64 +0,0 @@
-#----------------------------------------------------------------------------
```
SubJunk, line 1: Can you restore this before merge? You can use `LoadingPriority` to make yours load first

SubJunk (changes requested):

src/main/external-resources/renderers/Samsung-OLED.conf
valib: Just when I will return from holiday abroad. 😎

SubJunk: Cool where are you visiting?

valib: In Bulgaria

SubJunk (approved):

SubJunk:
Nice one!

Issue #1984, opened 2020-04-23
Update to latest Java


SubJunk:
Eclipse is soon to drop support for Java 8, and require 11+ https://www.eclipse.org/lists/eclipse-pmc/msg03821.html

Which means so is VS Code https://github.com/redhat-developer/vscode-java/wiki/JDK-Requirements#jdk11.requirement

That's two of the most popular IDEs for our developers so maybe it is a good time to do the same.

As I understand it, there are no licensing concerns for us with the upgrade, especially if we switch to using AdoptOpenJDK instead of Oracle, which I have already been considering regardless of this.

Our test suite already uses a newer JDK on macOS and the tests pass, so that is an indication that things will work, but manual testing will be necessary, and we will need to decide which version we use. It is up to version 14 now, so we could update straight to that, or just go to 11.

Any thoughts @UniversalMediaServer/developers ?

ExSport:
Few days ago I was forced to update to the newer Java JRE version where v1.8 was running. As most recent updates and versions can't be used without licensing it (non-personal use)  I moved out of v8 completely and tested OpenJDK v14, shrinked to "JRE binaries" only thanks to comparing it with AdoptOpenJRE what files are needed and what can be removed. Right now it is used for troubleshooting purposes only but it seems it is quite stable 😏
First time I tested AdoptOpenJre8 and v14 but as I don't know how these binaries are trustful I replaced it with OpenJDK14 from Oracle.
As you mentioned it it seems it should be quite trustful 😎

valib:
@SubJunk I support to go straight to the version 14.

SubJunk:
It seems we all agree to do this so I have changed it from a question to an enhancement

SubJunk:
Done in https://github.com/UniversalMediaServer/UniversalMediaServer/pull/2067

PR #2007, opened 2020-04-30
Disabled flaky test environment

SubJunk:


js-kyle:
Do we need a story to investigate & renable @SubJunk ?

SubJunk:
@js-kyle good question, I have mentioned in the story for this issue that it should be addressed again in https://github.com/UniversalMediaServer/UniversalMediaServer/issues/1984

PR #2014, opened 2020-05-01
fix maven lifecycle mapping in Eclipse

valib:
broken in ab709cac650532a07a5799c756c9b881d38450f6


SubJunk:
@valib it looks like all builds are failing after this merged, I don't know why, do you?

PR #2038, opened 2020-05-07
Revert the maven checkstyle plugin

valib:
The checkstyle version 3.1.0 is the last version which works with the maven site plugin 3.9.0 (tested)


SubJunk:
@valib whoops I already merged but then I thought it would be good to add a comment in the pom about that version dependency, otherwise we might forget and try to bump it again. Could you add it please?

PR #2047, opened 2020-05-16
Added support for Amazon Fire TV Stick

SubJunk:


Nadahar:
This doesn't look likely to work IMO - maybe you should test instead of rushing to merge?

PR #2051, opened 2020-05-20
Fix thumbnail IDs not getting written to the FILES table

SubJunk:
This should improve general performance noticeably


SubJunk:
@valib can I please get a review of this soon? It is quite a big bug because it means thumbnails are regenerated on every folder scan, so I am hoping to release it when it merges. If https://github.com/UniversalMediaServer/UniversalMediaServer/pull/2045 could also merge that would give us a significant performance and bugfix release that should make users happy

SubJunk:
@valib thanks :)

PR #2055, opened 2020-05-22
Update deprecated methods and remove unused imports

valib:


src/main/java/net/pms/dlna/RootFolder.java
```diff
@@ -726,7 +728,12 @@ private static DLNAResource getiPhotoFolder() {
 			} catch (XmlParseException | URISyntaxException | IOException e) {
 				LOGGER.error("Something went wrong with the iPhoto Library scan: ", e);
 			} finally {
-				IOUtils.closeQuietly(inputStream);
+				try {
+					inputStream.close();
```
SubJunk, line 732: I agree with this change, but instead of doing it like this, let's use try-with-resources so we can remove the `finally` block completely, as this answer suggests https://stackoverflow.com/a/49251654/2049714

src/main/java/net/pms/dlna/RootFolder.java
```diff
@@ -60,6 +61,7 @@
 import net.pms.util.FileUtil;
 import net.pms.util.FileWatcher;
 import net.pms.util.ProcessUtil;
+import net.pms.util.Rational;
```
SubJunk, line 64: Is this used?

src/main/java/net/pms/io/OutputTextConsumer.java
```diff
@@ -67,7 +67,12 @@ public void run() {
 		} catch (IllegalStateException ise) {
 			LOGGER.debug("Error reading from closed input stream: {}", ise.getMessage());
 		} finally {
-			LineIterator.closeQuietly(it); // clean up all associated resources
+			try {
```
SubJunk, line 70: Same comment about using try-with-resources instead

src/main/java/net/pms/io/OutputTextLogger.java
```diff
@@ -60,7 +60,12 @@ public void run() {
 		} catch (IllegalStateException ise) {
 			LOGGER.debug("Error reading from closed input stream: {}", ise.getMessage());
 		} finally {
-			LineIterator.closeQuietly(it); // clean up all associated resources
+			try {
```
SubJunk, line 63: Same comment about using try-with-resources instead

src/main/java/net/pms/service/ProcessManager.java
```diff
@@ -840,9 +839,16 @@ protected void destroyProcess(@Nullable Process process) {
 			if (process == null) {
 				return;
 			}
-			IOUtils.closeQuietly(process.getInputStream());
-			IOUtils.closeQuietly(process.getErrorStream());
-			IOUtils.closeQuietly(process.getOutputStream());
+
+			try {
```
SubJunk, line 843: Same comment about using try-with-resources instead

src/main/java/net/pms/util/ImdbUtil.java
```diff
@@ -99,8 +99,9 @@ public boolean accept(Path entry) throws IOException {
 						if (isBlank(entryName)) {
 							continue;
 						}
-						double score = StringUtils.getJaroWinklerDistance(nfoFileName, entryName);
-						if (score >= 0.85) {
+						double score = new JaroWinklerDistance().apply(nfoFileName, entryName);
+						if (score <= 0.85) { // threshold to add candidate to map. XXX it should be adjusted.
+											// The value 0.5 should be enough to insert relevant candidate.
```
SubJunk, line 104: Can you change the value now instead of adding that comment?

valib: @SubJunk do you think that value 0.5 is OK or it could be less?

src/main/java/net/pms/service/ProcessManager.java
```diff
@@ -840,9 +840,11 @@ protected void destroyProcess(@Nullable Process process) {
 			if (process == null) {
 				return;
 			}
-			IOUtils.closeQuietly(process.getInputStream());
-			IOUtils.closeQuietly(process.getErrorStream());
-			IOUtils.closeQuietly(process.getOutputStream());
+
+			IOUtils.closeSilently(process.getInputStream());
```
SubJunk, line 844: Why do we want `closeSilently` here, but in the other files we changed to `close`? It doesn't need to change, I just want to understand

src/main/java/net/pms/util/ImdbUtil.java
SubJunk: @valib I don't know, it would probably need some testing with real files to get an idea of that. I don't know if there was much testing for our implementation of Jaro-Winkler. I tested it with dozens of files for the validation of prettifying, but not for choosing subtitles or NFO parsing.
I'm happy for you to change it to what you think is best.

Nadahar: I wrote the above code, so yes it was quite a lot of testing and tweaking to try to find sane values. That said, you merged it before I finished it, so I'm not saying that those values are "ideal", just that I spent a lot of time trying to find a system that would be reasonably good at making the correct decisions.

src/main/java/net/pms/service/ProcessManager.java
valib: @SubJunk frankly I didn't find the way how to use try-with-resources instead. 😞

SubJunk: Ah cool, I think you can just separate them by semicolons or something, I'll try to do it

SubJunk: Oh I see, it is all split up into different methods, so yeah this is probably the best way unless we refactor it. Nice work

src/main/java/net/pms/util/ImdbUtil.java
SubJunk: @valib the rest of the changes look good, are you happy with this change now or are you doing testing on it?

valib:
Jaro-winkler distace should be approed in the PR #1989

SubJunk:
Cool so we should probably revert the dependency updates in here too?

valib:
@SubJunk I force-pushed this branch because before I made a mistake when I merged branch `dependabot/maven/org.apache.commons-commons-text-1.8` into this branch. I also revert the changes in the Jaro-Winkler logic.

valib:
@SubJunk I merged this branch into `dependabot/maven/org.apache.commons-commons-text-1.8` and we should continue in PR #1989

SubJunk:
Good idea

PR #2066, opened 2020-05-30
Try to fix the resource leaks when streams are not closed properly

valib:


SubJunk:
@valib I've tried searching for answers about these kinds of changes and I am still confused. My concern is that returning the resource within a try-with-resources block will return a closed resource.

I found https://stackoverflow.com/questions/22947755/try-with-resources-and-return-statements-in-java and of course https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

It is probably best to design our code to return a result of a closeable stream, rather than the stream itself, but we have a lot of code that already does this pattern so maybe it is best for now to make sure the consumers of these methods which return autocloseable resources are closing them when they are finished using them. What do you think?

src/main/java/net/pms/dlna/DLNAResource.java
```diff
@@ -3460,7 +3460,9 @@ protected DLNAThumbnailInputStream getGenericThumbnailInputStreamInternal(String
 		if (thumb != null && isCodeValid(this)) {
 			// A local file
 			if (new File(thumb).exists()) {
-				return DLNAThumbnailInputStream.toThumbnailInputStream(new FileInputStream(thumb));
+				try (FileInputStream fis = new FileInputStream(thumb)) {
+					return DLNAThumbnailInputStream.toThumbnailInputStream(fis);
```
SubJunk, line 3464: I am unsure about this one. Will DLNAThumbnailInputStream.toThumbnailInputStream still get what it wants after this change?

src/main/java/net/pms/dlna/WebStream.java
```diff
@@ -92,7 +92,9 @@ public DLNAThumbnailInputStream getThumbnailInputStream() throws IOException {
 				FileUtil.isUrl(thumbURL) ? downloadAndSend(thumbURL, true) : new FileInputStream(thumbURL)
 			);
 		} else {
-			return super.getThumbnailInputStream();
+			try (DLNAThumbnailInputStream dtis = super.getThumbnailInputStream()) {
+				return dtis;
```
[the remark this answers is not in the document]
valib: @SubJunk there are Eclipse warnings `Resource leak: '<unassigned Closeable value>' is not closed at this location` so I made this change. It works for me. I tested other way and it could be written like 

> DLNAThumbnailInputStream dtis = super.getThumbnailInputStream();
> 			return dtis;

Both works and the the Eclipse warnings are eliminated.

src/main/java/net/pms/dlna/PlaylistFolder.java
```diff
@@ -92,7 +92,9 @@ private BufferedReader getBufferedReader() throws IOException {
 		} else {
 			File playlistfile = new File(uri);
 			if (playlistfile.length() < 10000000) {
-				return new BufferedReader(new InputStreamReader(new BOMInputStream(new FileInputStream(playlistfile)), charset));
+				try (FileInputStream fis = new FileInputStream(playlistfile)) {
+					return new BufferedReader(new InputStreamReader(new BOMInputStream(fis), charset));
```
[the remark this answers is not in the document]
valib: @SubJunk the same as above

src/main/java/net/pms/dlna/DVDISOTitle.java
```diff
@@ -380,9 +380,13 @@ public DLNAThumbnailInputStream getThumbnailInputStream() throws IOException {
 		if (cachedThumbnail != null) {
 			return DLNAThumbnailInputStream.toThumbnailInputStream(new FileInputStream(cachedThumbnail));
 		} else if (getMedia() != null && getMedia().getThumb() != null) {
-			return getMedia().getThumbnailInputStream();
+			try (DLNAThumbnailInputStream dtis = getMedia().getThumbnailInputStream()) {
+				return dtis;
```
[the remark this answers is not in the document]
valib: @SubJunk the same as above

src/main/java/net/pms/dlna/DVDISOTitle.java
```diff
@@ -380,9 +380,13 @@ public DLNAThumbnailInputStream getThumbnailInputStream() throws IOException {
...
+			}
 		} else {
-			return getGenericThumbnailInputStream(null);
+			try (DLNAThumbnailInputStream dtis = getGenericThumbnailInputStream(null)) {
+				return dtis;
```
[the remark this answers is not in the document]
valib: @SubJunk the same as above

Nadahar:
You usually ignore my warnings, but I still can't not say anything about this one. It will completely break everything, `try with resources` *will* close the stream when it is exited. The Eclipse warnings aren't divine, there's a reason it's possible to suppress them.

Streams are often opened and consumed within the same method, and when that's the case, using `try with resources` is a nice way to protect against unforeseen events that break the normal logic so that the resource isn't closed, but it cannot be used if the stream is to ever leave the method.

For input streams it would theoretically be possible to consume them locally, close the stream and pass on the content, but that requires *reading the entire resource into memory*. This isn't always a good idea, especially when we're dealing with media files. For output streams it's simply impossible to get around this unless you already have all the data you with to write at the time you open the stream.

In short, in more "complex" situations where you work with a stream over time or using multiple methods, you can't use `try with resources`. You must instead be vigorous and making sure that there are no paths, regardless of how unlikely, where the resource isn't closed - including when exceptions are thrown at any location where that is possible. If you fail to do that, you have a leak...

src/main/java/net/pms/dlna/WebStream.java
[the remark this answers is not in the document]
SubJunk: I don't see how it can still do what it did before, since now it is returning a closed stream. How can the consumer use a closed stream? As I said in https://github.com/UniversalMediaServer/UniversalMediaServer/pull/2066#issuecomment-636412907 I think we should instead make sure that the consumers close the streams they receive.
I have seen recommendations for naming conventions too, where any method that returns a closeable stream should be prefixed with `output` or something, and then we can add a note to the JavaDoc for those methods to make sure. Something like:

```
/**
 * Method description.
 *
 * Note: If you use this method, you must close the resource you receive after you are finished with it.
 */

**Comment by valib** [comment]
@SubJunk @Nadahar if the try-with-resources returns closed resource or not it depends on the Java implementation. E.g. the InputStream.close() called by the try-with-resources to close does nothink see the comment in the Java source

> /**
     * Closes this input stream and releases any system resources associated
     * with the stream.
     * <p> The <code>close</code> method of <code>InputStream</code> does
     * nothing.
     *
     * @exception  IOException  if an I/O error occurs.
     */
    public void close() throws IOException {}

Maybe that is why it works for me. Anyway I rewrite some changes.

**Comment by Nadahar** [comment]
@valib Yes, it's true that not all implementations implements `close()`, but most do. `InputStream` is an abstract class, so methods are often overridden in implementations. File-based streams will certainly fail if you try to access them after closure, but memory-based streams probably don't care (like `ByteArrayInputStream` for example).

File: src/main/java/net/pms/dlna/DVDISOTitle.java [hunk 6]
```diff
@@ -380,9 +380,11 @@ public DLNAThumbnailInputStream getThumbnailInputStream() throws IOException {
 		if (cachedThumbnail != null) {
 			return DLNAThumbnailInputStream.toThumbnailInputStream(new FileInputStream(cachedThumbnail));
 		} else if (getMedia() != null && getMedia().getThumb() != null) {
-			return getMedia().getThumbnailInputStream();
+			DLNAThumbnailInputStream dtis = getMedia().getThumbnailInputStream();
+			return dtis;
```
SubJunk, line 384: Is this different to before?

src/main/java/net/pms/dlna/PlaylistFolder.java
```diff
@@ -92,7 +92,8 @@ private BufferedReader getBufferedReader() throws IOException {
 		} else {
 			File playlistfile = new File(uri);
 			if (playlistfile.length() < 10000000) {
-				return new BufferedReader(new InputStreamReader(new BOMInputStream(new FileInputStream(playlistfile)), charset));
+				FileInputStream fis = new FileInputStream(playlistfile);
+				return new BufferedReader(new InputStreamReader(new BOMInputStream(fis), charset));
```
SubJunk, line 96: Is this different to before?

src/main/java/net/pms/dlna/WebStream.java
```diff
@@ -92,7 +92,8 @@ public DLNAThumbnailInputStream getThumbnailInputStream() throws IOException {
 				FileUtil.isUrl(thumbURL) ? downloadAndSend(thumbURL, true) : new FileInputStream(thumbURL)
 			);
 		} else {
-			return super.getThumbnailInputStream();
+			DLNAThumbnailInputStream dtis = super.getThumbnailInputStream();
```
SubJunk, line 95: Is this different to before?

src/main/java/net/pms/network/RequestV2.java
```diff
@@ -238,6 +238,12 @@ public void setTextContent(String content) {
 		this.content = content;
 	}
 
+	/**
+	 * The InputStream for the response to a received request.
+	 * Defining it here make it hopefully automatically closable.
```
SubJunk, line 243: @valib have you tested that it does get closed automatically?

src/main/java/net/pms/network/RequestV2.java
```diff
@@ -241,6 +241,12 @@ public void setTextContent(String content) {
...
+	 */
+	private InputStream responseInputStream = null;
```
SubJunk, line 248: I'm not sure about this without testing, since maybe an instance of RequestV2 should be able to send multiple inputStreams. Will have to test

SubJunk:
This got stale, closing

PR #2067, opened 2020-06-01
JRE 14

SubJunk:
Implements https://github.com/UniversalMediaServer/UniversalMediaServer/issues/1984


valib:
@SubJunk when I build the `UMS-setup.exe` and installed on the WIN 10 the and tried to run it an error returns.
UMS error

SubJunk:
That's a good point @valib I will add that to the installer, we already do that for some files.
I'll also look into that startup failure

SubJunk:
I have tested it on a fresh install and on top of 9.5.0, and it seems to work now. Can you please confirm @valib ?

valib:
@SubJunk I think that the same should be done in the workspace otherwise the wrong build could be possibly made.

valib:
@SubJunk I can confirm that your last update works. But before testing I deleted JRE folders in the `target/bin/win32` to obtain the clean version of the Java.

SubJunk:
@valib are you suggesting we add a deletion step to the Maven process as well? I can see how the current way can cause problems but I also don't want to have to maintain a huge list of files that used to exist in pom.xml.

One way to get around maintaining a list of old JRE files is to change the folder in this release. I can call it like `jre14-x64`. Would that be a good way to fix it for developers who are building before and after this change?

valib:
@SubJunk you make it as I mentioned. Simply delete the old folders.

PR #2077, opened 2020-06-06
Update the supported formats in web renderers

valib:
It can partially solve the problem reported in https://www.universalmediaserver.com/forum/viewtopic.php?f=9&t=14171 where the m4a audio is transcoded for the Firefox browser and it finish with the error at the end of the file. This format is actually supported by all relevant browsers. There still persists the problem to auto skip to the next file when the file is fully played. Also files do not start playing automatically when selected. There must be problem in the web impementation.


SubJunk:
Nice work. I have a push coming to the api-3 branch soon that adds a lot of functionality to the web interface so this is really good timing

Issue #2074, opened 2020-06-05
"Force networking on interface" does not persist


ITGuyDave:
As I perform development work on my PC, I am running HyperV and Docker on the machine - which has the unfortunate effect of creating numerous network interfaces. Due to UMS often locking to the wrong default interface (see logs), I frequently have to change this back to the correct physical NIC for my local network, rather than a random interface such as for Hyper V. As I don't use UMS every day I have not clearly determined what is causing it, but I suspect it is caused by updates to either Windows (and Hyper V consequently) or Docker.

The actual name of the interface is not changing, but the order appears to be changing for whatever reason, which seems to be breaking the "tie" between UMS and the NICs.

I would like to see UMS tying to the name of the forced network interface (or some other variable that doesn't change) to ensure this setting does not get broken as interfaces are added or removed.
[ums_dbg.zip](https://github.com/UniversalMediaServer/UniversalMediaServer/files/4733212/ums_dbg.zip)

valib:
@ITGuyDave if the interface you want to use for the UMS have static IP address you can try to set it like I made an example here 
![network setting](https://user-images.githubusercontent.com/1472678/83878831-0364c200-a73d-11ea-9aff-cc71d9c95c69.jpg)

ITGuyDave:
Hi @valib I do use a dynamic address typically, thus why the interface is preferred. I might simply set up a static DHCP reservation in my case, but it's not ideal.

If UMS is relying on the ifindex, which it seems to be doing - this is never going to be reliable as a means to bind to an interface, as the returned values from most W10 interface commands will return the interfaces in an order that reflects the interface preference; which in Hyper-V systems is typically a NAT interface. It would be far better to simply key off the interface alias, rather than ifindex or similar.

SubJunk:
I agree it would be good to make it match the name

valib:
@ITGuyDave thanks for the input. It seems that we use default interface not the one selected which is a bug.

ITGuyDave:
Hi @valib, sorry for the delay - it's been a manic week. I've attached an updated log as requested with the static IP set.
[ums_dbg - static IP.zip](https://github.com/UniversalMediaServer/UniversalMediaServer/files/4786870/ums_dbg.-.static.IP.zip)

ITGuyDave:
[...]

PR #2080, opened 2020-06-08
Use network interface names instead of their hardware description

valib:
An attempt to make enhancement asked in the issue  #2074


SubJunk:
@valib do you want me to review this or is it still in progress?

SubJunk:
The user has reported it is working for them, and I have verified it works for me when I switch between network interfaces.

However it would be nice if it could be backwards-compatible because if users already have a network interface selected, this clears that selection. Otherwise we may have some confused users

valib:
@SubJunk done

valib:
@SubJunk I think now it is ready to merge

SubJunk:
Great, I have retested it with an old saved option and it retains that option, so I'll merge

PR #2081, opened 2020-06-08
Fix the auto update for Windows

valib:
caused by the commit e0ab226. This bug is reported in https://www.universalmediaserver.com/forum/viewtopic.php?f=9&t=14142

### DIFF
--- a/src/main/java/net/pms/update/AutoUpdaterServerProperties.java
+++ b/src/main/java/net/pms/update/AutoUpdaterServerProperties.java
@@ -47,6 +47,7 @@ public class AutoUpdaterServerProperties {
 
 	private String getPlatformSpecificKey(String key) {
 		String os = operatingSystem.toString();
+		 //we use only one windows version so the exact windows version is not relevant
 		if (os.startsWith("windows")){
 			os = "windows";
 		}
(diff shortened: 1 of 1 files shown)


valib:
@SubJunk this should be relased as soon as possible

src/main/java/net/pms/update/AutoUpdaterServerProperties.java
```diff
@@ -46,6 +46,12 @@ private String getStringWithDefault(String key, String defaultValue) {
 	}
 
 	private String getPlatformSpecificKey(String key) {
-		return key + "." + operatingSystem.toString();
+		String os = operatingSystem.toString();
+		 //we use only the one UMS Windows version so the exact OS Windows version is not relevant
+		if (os.startsWith("windows")){
+			os = "windows";
+		}
```
SubJunk, line 53: This should make it work on macOS too

```suggestion
		if (!os.startsWith("linux")) {
			os = operatingSystem.getPlatformName();
		}
```

SubJunk:
@valib thanks for the fix, I've modified it to also fix macOS. Please let me know if you approve of that change, then I will merge if you do

valib:
@SubJunk I agree

PR #2086, opened 2020-06-10
fix checking the internal UMS Cling requests

valib:
This fix the Netty4 PR and allow to run locally e.g. `cling-workbench-2.1.0-standalone.jar` for testing or to run other external cling-based client running on the same host.

### DIFF
--- a/src/main/java/net/pms/network/RequestHandlerV2.java
+++ b/src/main/java/net/pms/network/RequestHandlerV2.java
@@ -420,8 +420,9 @@ public class RequestHandlerV2 extends SimpleChannelInboundHandler<FullHttpReques
 	}
 
 	private boolean isLocalClingRequest(InetAddress ia, String userAgent) {
-		// FIXME: this would also block an external cling-based client running on the same host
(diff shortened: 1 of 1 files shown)


SubJunk:
Feel free to share...

Nadahar:
Isn't it obvious?

Other software use Cling, especially Android "app"s like BubbleUPnP. I don't have any overview over to which extent, but I know I've seen "Cling" in the user agent many times. UMS will now ignore all requests from these renderers.

SubJunk:
That is already in `master`. This change only added `UMS/`

SubJunk:
However I see that was only added 12 days ago in the netty 4 branch, and I tend to agree with you that we shouldn't ignore those user-agents. I'll make a followup fix to restore the functionality as it was 13 days ago

SubJunk:
Done in https://github.com/UniversalMediaServer/UniversalMediaServer/pull/2090

Nadahar:
Ok, I'm sorry that I misread the diff. I just saw `Cling` added to ignore and reacted to that.

PR #2089, opened 2020-06-11
Fixed support for external subtitles with the se syntax

SubJunk:
Also removes `VideoFormatsSupportingStreamedExternalSubtitles` since that is no longer needed, and renames some things to be more readable.
It makes another change that we had just been discussing on the forum, which is that a file only needs one supported audio format to pass the match, while currently all audio codecs must be supported to match. In my testing, that has been a good change, as the renderers ignore streams they can't read.
It is possible that some renderers will not accept that but I think that will be more rare.

I have tested it with both syntaxed (`SupportedExternalSubtitlesFormats` and `se:`) on an external SRT file and it works with both, while before it only worked with `SupportedExternalSubtitlesFormats`.

### DIFF
--- a/src/main/external-resources/renderers/DefaultRenderer.conf
+++ b/src/main/external-resources/renderers/DefaultRenderer.conf
@@ -716,20 +716,6 @@ CharMap =
 # Default: ""
 SupportedExternalSubtitlesFormats =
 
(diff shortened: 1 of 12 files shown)


SubJunk:
I think this will give us a lot of happy users so I hope to release it soon @valib

PR #2095, opened 2020-06-11
Fixed responsive behavior of browser interface

J-Puls:
I noticed that the responsive classes did not behave properly when viewing on a mobile device. This was due to a missing <meta> tag in the HTML.

Hope this helps! 😃


SubJunk:
@J-Puls thanks for this!

PR #2104, opened 2020-06-14
Fixed bugs "Boxing/unboxing to parse a primitive" reported by SpotBugs

valib:
This can a little bit speed up the UMS performance. See https://spotbugs.readthedocs.io/en/latest/bugDescriptions.html#DM_BOXED_PRIMITIVE_FOR_PARSING


src/main/java/net/pms/configuration/RendererConfiguration.java
```diff
@@ -2398,7 +2398,7 @@ public boolean isOmitDcDate() {
 
 	public static int getIntAt(String s, String key, int fallback) {
 		try {
-			return Integer.valueOf((s + " ").split(key)[1].split("\\D")[0]);
+			return Integer.parseInt((s + " ").split(key)[1].split("\\D")[0]);
 		} catch (Exception e) {
```
valib, line 2402: @SubJunk when I change this general `Exception` catch to `NumberFormatException` which should be the relevant exception of the `Integer.parseInt` the test fails. There must be any other hidden problem.

SubJunk: Yeah it looks like an unsafe line to me. It is making a lot of assumptions about what `s` is

valib: @SubJunk when I run the UMS with the `NumberFormatException` in this method I found that the error is hidden here

> java.lang.NoSuchMethodError: java.nio.ByteBuffer.flip()Ljava/nio/ByteBuffer;
	at net.pms.newgui.components.WindowProperties$WindowPropertiesConfiguration.readConfiguration(WindowProperties.java:520)
	at net.pms.newgui.components.WindowProperties$WindowPropertiesConfiguration.<init>(WindowProperties.java:481)
	at net.pms.PMS.init(PMS.java:419)
	at net.pms.PMS.createInstance(PMS.java:916)
	at net.pms.PMS.main(PMS.java:1086)
	at org.codehaus.mojo.exec.ExecJavaMojo$1.run(ExecJavaMojo.java:254)
	at java.lang.Thread.run(Thread.java:748)

SubJunk: What is wrong with just catching all exceptions and returning a fallback, like this already does? That seems like a valid strategy to me, instead of doing different things with different exceptions

SubJunk: Ok cool. Will you fix that in this branch?

Issue #2105, opened 2020-06-15
9.6.1 browsing regression


SubJunk:
We have had some users report that they can no longer see media on their devices after upgrading to 9.6.1. I can't reproduce on any of my devices. For now I am going to try to get people to test a reversion of the Netty upgrade as I think that is the most likely thing.

Open to suggestions.

SubJunk:
I have posted test builds of this for a few users now so I'm just waiting for one to get back to me about whether it fixes it. If it does I'll rush through a release.

SubJunk:
Fixed in 9.6.2

PR #2107, opened 2020-06-15
Revert netty 2

SubJunk:


SubJunk:
Got confirmation from pipin on the forum that this fixes the problem so I'm merging and releasing https://github.com/UniversalMediaServer/UniversalMediaServer/issues/2105

Issue #2136, opened 2020-06-24
Check for unindexed queries


SubJunk:
It would be good to use an automatic H2database profiling tool to make sure all of our queries are using indexes for performance

SubJunk:
I'm reopening this because now we have the ability to do it, but the other part still needs to be done - we should use the tool to optimize the queries

PR #2137, opened 2020-06-24
Profiling the h2 database

valib:
This is the implementation of #2136.


SubJunk:
@valib this works great, thanks! It will hopefully give me the ability to keep optimizing performance

PR #2156, opened 2020-07-01
Revert "Use network interface names instead of their hardware description"

SubJunk:
Reverts UniversalMediaServer/UniversalMediaServer#2080

@valib unfortunately I noticed a bug with this so I will re-merge this if that gets fixed - the backwards-compatible change isn't working. I have my wired network selected in the GUI so it appears to work, but then it binds to my wireless interface instead.


valib:
@SubJunk I try to fix it in #2159

PR #2161, opened 2020-07-03
Stop logging responses twice

SubJunk:


Nadahar:
The two aren't the same. The "unescaped" logging is just a part of the "full" logging, it's only the part inside `<Result>`. This means you'll lose headers, status ++ when doing this.

SubJunk:
@Nadahar isn't that information still logged at https://github.com/UniversalMediaServer/UniversalMediaServer/pull/2161/files#diff-0a144e1c8e5f70b6279fa87ed5f90f83R1001 ?

Nadahar:
Well, yes, some of it is at least. I didn't notice that you had added it there, but you still lose the message content that's outside `<Result>`. As I see it, this makes the log hard to understand, but I guess you just want to get rid of the extra hits when searching for a file?

I always just saw the unescaping as a "helper function", while the first one was the "real logging". Before I made the unescaping/XML pretty printing, I had to paste this into Notepad++, unescape and then format each time I wanted to study it. Reading it in the "escaped" form is too much of a hassle for me.

I guess the "ideal" solution would be to "unescape" and then replace the "escaped" part of the content with the "unescaped" while keeping the rest of the message. Still, it could mask errors in the escaping itself, which is why I like the log to contain the content exactly as it's being sent to the renderer. But, I guess that's all about preferance...

Issue #2138, opened 2020-06-25
Re-add macOS minimizing on startup


SubJunk:
I removed the ability to start UMS minimized in https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1570/files#diff-9108fb19deecfa93b485b94496039b3dR1817 because of a bug.
Would be good to re-implement that.

PR #2162, opened 2020-07-04
minimize on MacOS

js-kyle:
https://github.com/UniversalMediaServer/UniversalMediaServer/issues/2138


SubJunk:
Nice work. I tested and it works. It's a bit weird how it is visible for a sec and then minimizes, ideally it would not display at all, but it's still much better than what we have so I'll merge

PR #2170, opened 2020-07-09
Added support for Sony STR-DN1080 AV Receiver

GurliGebis:
I have created a renderer for the Sony STR-DN1080 AV receiver and tested it with flac files (I haven't tried anything else, since all my music is in flac format).
I have based it on the specs here: https://helpguide.sony.net/ha/strdn108/v1/en/contents/TP0001221399.html


Nadahar:
There is a nice collection of samples here: https://samples.mplayerhq.hu/

Experience shows that basing it solely on the specs often isn't sufficient, it would be smart to at least test the most common format/codecs.

In addition, the image is usually included in two versions, one ending its filename with `@2x.png` which will be used on high resolution displays on macOS. This image should simply be twice the width and height of the "normal" one.

GurliGebis:
Okay, I'll see if I can get some spare time tonight to try those, and create a higher res image 🙂

Nadahar:
These are mere suggestions from my side, but I know it will increase the likelihood of this getting merged :wink:

GurliGebis:
There we go :)
I have tested the different codecs, and added the ones I have been able to get working.

So it is a good starting point, other people with the same receiver is welcome to extend the list, if more formats is working for them.

SubJunk:
Thanks for this. By the way we also have common media files for testing in our repository that we use for regression testing https://github.com/UniversalMediaServer/UniversalMediaServer/tree/master/src/test/resources/net/pms/dlna

Issue #2177, opened 2020-07-11
Test DLNAMediaInfoTest.testContainerProperties:104 fails on Manjaro Linux 20.0.3


outnos:
Im running Manjaro Linux 20.0.3 and a DLNAMediaInfoTest fails with Overall Bitrate being 1 less on my machine than in the test.

> [ERROR] Failures:
> [ERROR]   DLNAMediaInfoTest.testContainerProperties:104
> Expecting:
>  <"Container: MKV, Size: 5291494, Overall Bitrate: 2619551, Video Tracks: 1, Video Codec: h265, Duration: 0:00:16.160, Video Resolution: 1920 x 960, Display Aspect Ratio: 2.00:1, Frame Rate: 25.000, Frame Rate Mode: CFR (CFR), Frame Rate Mode Raw: CFR, Matrix Coefficients: BT.709, Audio Tracks: 1 [Id: 0, Language Code: eng, Audio Track Title From Metadata: Stereo, Audio Codec: AAC-LC, Bitrate: 0, Channels: 2, Sample Frequency: 48000 Hz], Mime Type: video/x-matroska">
> to be equal to:
>  <"Container: MKV, Size: 5291494, Overall Bitrate: 2619552, Video Tracks: 1, Video Codec: h265, Duration: 0:00:16.160, Video Resolution: 1920 x 960, Display Aspect Ratio: 2.00:1, Frame Rate: 25.000, Frame Rate Mode: CFR (CFR), Frame Rate Mode Raw: CFR, Matrix Coefficients: BT.709, Audio Tracks: 1 [Id: 0, Language Code: eng, Audio Track Title From Metadata: Stereo, Audio Codec: AAC-LC, Bitrate: 0, Channels: 2, Sample Frequency: 48000 Hz], Mime Type: video/x-matroska">
> but was not.

js-kyle:
Note - I've seen this on Ubuntu 18.04 with the same error

SubJunk:
This is caused by a different MediaInfo version. The version we rely on in UMS is 18.12 so if you run that version it should fix it. Other versions report a slightly different bitrate for some files, maybe they changed the rounding or something. We have an issue to update MediaInfo and just need to do it.

Issue #2152, opened 2020-06-29
Try GitHub Actions for CI


SubJunk:
We use Travis at the moment and that gives us the ability to have broad coverage, but it seems like GitHub Actions offers the same things with matrixes, so it would be good to try it
https://github.com/features/actions

PR #2178, opened 2020-07-11
feat: Github Actions for CI

js-kyle:
https://github.com/UniversalMediaServer/UniversalMediaServer/issues/2152

Ubuntu 18.04 is failing on the same test mentioned here, https://github.com/UniversalMediaServer/UniversalMediaServer/issues/2177, however that doesn't run on Travis at the moment anyway


valib:
@js-kyle travis is still waiting for the status to be reported.

js-kyle:
@valib The intention is to replace Travis, so the job won't run there

valib:
@js-kyle so what will be the next step?

js-kyle:
@valib I think the next steps for this would be

- general agreement that we want to switch CI platforms
- agree on the actual changes I have made (e.g. do we want to add more build combinations?)
- disable Travis as a required repository check

Then we could merge

SubJunk:
I think it makes sense to switch to this from Travis.

> Ubuntu 18.04 is failing on the same test mentioned here, #2177, however that doesn't run on Travis at the moment anyway
If we want to use that version instead of 16 here, changing MediaInfo to 18.12 should fix that test

Also we probably don't need to support Javas 9-13 IMO; 8 and 14 are the currently supported ones. So I would do 8 on Linux just as a canary for us breaking compatibility, and then 14 on Windows and macOS

One improvement would be to use AdoptOpenJDK instead of whatever these Java distributions are. Eventually I want to use the actual JRE we supply via pom.xml but that could be its own story

js-kyle:
I've made some tweaks based on the above comment

PR #2179, opened 2020-07-11
Remove crossorigin from webui

outnos:
The html tag "crossorigin" means that the file is retrieved without user authentication, this means that the configuration option "web_authentication=true" should cause constant popups for browsers retrieving a file.
In practice Firefox and Chrome ignores this tag, and in my testing only Safari complies with it, meaning that for each refresh it request a user logon, which is annoying and unpractical.

### DIFF
--- a/src/main/external-resources/web/browse.html
+++ b/src/main/external-resources/web/browse.html
@@ -34,7 +34,7 @@
 		<script src="/files/util/keycode.js"></script>
 		<script src="/files/util/keycontroller.js"></script>
 		<link rel="stylesheet" href="/files/util/remoteui.css" type="text/css" media="screen">
-		<link rel="stylesheet" href="/files/util/fontawesome/css/all.min.css" crossorigin="anonymous">
+		<link rel="stylesheet" href="/files/util/fontawesome/css/all.min.css">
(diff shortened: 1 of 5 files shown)


SubJunk:
This looks good and I'll merge it. However in v10 pre-release (the branch is `api-3`) we rely on crossorigin requests to access movie and TV series posters so we may need a different approach there if that causes problems. Would you have a look at what we have done there? Without that attribute, we can't use that HTML5 canvas to work out the average color of the image (we alter the page colors based on that)

Here is where we use that attribute https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1918/files#diff-08b53e22e7ace3f0d5a524d37b2ca43bR111
Here is the error we encounter if that attribute is not set https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1918/files#diff-4b8bdeaf7ed8ce328b419650f26ed739R410

outnos:
In that commit crossorigin is appropriate, as the poster is from an external source and thus must be marked as a foreign source with the tag crossorigin.
But when retrieving from our own files, which are user protected, its inappropriate as you might need authentication from that source.

PR #2183, opened 2020-07-13
New macOS binaries

SubJunk:
Apple doesn't let us use our perfectly good 32-bit MPlayer/MEncoder anymore so this is for new ones


valib:
@SubJunk I cant review it because I don't have the MAC so you should merge this if you are sure  that it works.

SubJunk:
Ok thanks. It can't get much worse than how it is currently - every user of the latest macOS can't use our MEncoder haha :D

PR #2184, opened 2020-07-13
Added lazy-loading to WebUI images

outnos:
So if you create some, i don't know, 5446 images in one folder, loading because staggeringly slow. Because browsers attempt to download as much as their cache will allow them, and this causes the UMS server to not respond for some 50 seconds.
Instead, what adding loading="lazy" does, is it tells the client, load these when needed on the screen.

So this cuts the loading times from 50 to 5 seconds (in my testing anyway), as the images are just loading in when they appear on-screen.

Note:
It also causes some cover images to appear as black-boxes when you scroll before being loaded (could be lessened by HTTP/2.0), but only if you use firefox (thanks mozilla), chrome and safari doesn't exhibit this problem.


SubJunk:
@outnos thanks, good idea.

PR #2185, opened 2020-07-14
Updates

valib:
Updates of some renderers for better recognition. To be correctly recognized the Samsung-CD.conf and Samsung-NotCD.conf should be disabled because they use very general recognition `(Samsung|AllShare)` which is nonsens.
Selected renderers are now sorted before they are saved to the UMS.conf.


valib:
@SubJunk should not be Travis integration deleted here in the GitHub?

SubJunk:
@valib there is good reason for general detection - it means that when the new TVs come out in 2021 that we don't have a config for yet, we will still have approximate support for them until a specific config can be made. Then the more specific ones use `LoadingPriority` to make sure they get priority over the generic ones.

SubJunk:
Please submit a new branch with LoadingPriority added back in for these, otherwise it may conflict with the generic configs

valib:
@SubJunk set the `LoadingPriority=1` is not working properly. If I set it for my renderesrs then the result is 
Bez názvu.
When the `LoadingPriority` is not set the result is correct 

correct

It seems that the renderer with the  `LoadingPriority=1` which is recognized first overrides the second renderer. Maybe it is only  the thumbnail problem but who knows.

SubJunk:
I have made a bug for that at https://github.com/UniversalMediaServer/UniversalMediaServer/issues/2187
Maybe since you have a good way to reproduce it, you could work on it?

valib:
@ @SubJunk I will try.

valib:
@SubJunk all it was only about wrong setting of the `UpnpDetailsSearch`

SubJunk:
@valib do you mean we can close #2187 ?

PR #2192, opened 2020-07-15
Implemented byte-range support and sustained rendering

### LINKED ISSUE #977: Web interface doesn't work in Safari on iOS
Videos don't play for me

outnos:
This was done to try and fix #977 , but it doesn't fix Safari support.
What i've done is add byte-range support to RemoteMediaHandler.java. 
However, the default implementation stopped and started ffmpeg with each refresh, which is unsustainable for byte-range requests.
So i created a media resource dictionary, which has the last access time, and is scanned by a timer every 2 minutes to check if it has been left alone for 5 minutes, and then deletes it.
I also set the java version to 8 instead of 7 to use lambdas, i don't quite get why it was version 7, since i didn't notice anything breaking, and most operating systems use atleast java 8.

The reason this doesn't fix Safari support is because of the content length, when supplied with a fixed length via Content-Range, Safari starts by scanning this range to find the end, and then sets what it thinks the video length is, meaning playback stops unless you refresh.
To actually support Safari i would recommend using HLS instead, as it can be used by ffmpeg to generate a "playlist" file and a vod file. That is however not supported with the current setup, as this requires safari to access atleast 2 files, which is possible using "hls_flags single_file". But it does work, as i have tried serving via nginx while transcoding, which Safari sees as live.

The does also need more extensive testing.
Dont merge as there are playback bugs, that i have yet to track down.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -848,10 +848,10 @@
 				<artifactId>maven-compiler-plugin</artifactId>
 				<version>3.8.1</version>
 				<configuration>
-					<source>1.7</source>
-					<target>1.7</target>
-					<testSource>1.7</testSource>
-					<testTarget>1.7</testTarget>
+					<source>1.8</source>
+					<target>1.8</target>
+					<testSource>1.8</testSource>
+					<testTarget>1.8</testTarget>
 					<compilerArgument>${compilerArgument}</compilerArgument>
 				</configuration>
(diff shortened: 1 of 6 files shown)


SubJunk:
Nice start, looking forward to seeing how this ends up

SubJunk:
?

PR #2200, opened 2020-07-19
Mux with FFmpeg when the input codec is the same as the transcoding target

SubJunk:


SubJunk:
@valib this has been verified as working on the forum, could it be reviewed when you have a chance please?

PR #2201, opened 2020-07-19
Added more DLNA Org.PN values

SubJunk:
Also makes some other fixes to DLNA values and minor renderer support improvements


SubJunk:
@valib thanks for the fix and approval

PR #2112, opened 2020-06-16
Update the deprecated multicast joinGroup(InetAddress) in the Java 14

valib:

PR #2204, opened 2020-07-21
Respond to notify

SubJunk:
In my testing, renderers are sending a NOTIFY when they start up, and we don't respond so they don't see us until we send an ALIVE.
With this change, we are displayed straight away because we respond. Tested on VLC for iOS and my Panasonic TV.

I am interested in feedback about this; is there a good reason we ignored NOTIFY before?


src/main/java/net/pms/network/UPNPHelper.java
```diff
@@ -477,12 +477,18 @@ public void run() {
...
 									int remotePort = receivePacket.getPort();
 									if (!redundant && LOGGER.isTraceEnabled()) {
-										LOGGER.trace("Received a M-SEARCH from [{}:{}]: {}", remoteAddr, remotePort, s);
+										String requestType = "unreognized";
+										if (packetType == M_SEARCH) {
+											requestType = "M-SEARCH";
+										} else if (packetType == NOTIFY) {
+											requestType = "NOTIFY";
+										}
+										LOGGER.trace("Received a " + requestType + " from [{}:{}]: {}", remoteAddr, remotePort, s);
```
Nadahar, line 491: Why would you use both `+ type +` and `{}` in the same log entry? It means that two `StringBuilder`s have to be created when running this one line in trace mode, and one when not in trace.

Nadahar:
The whole UPnP discovery implementation is very far from spec. It seems like whoever wrote it originally, and those that have modified it subsequently, have never peeked the slightest at the actual standard. Instead, it seems like a "tweak until something works" approach, while probably trying to mimic some other device/software using a packet sniffer.

My guess is that there's a lot more wrong there, but that the reason why there's no response to NOTIFY is caused by the simple fact that the PS3 doesn't send NOTIFY. In fact, it doesn't send anything related to discovery at all, but apparently it does listen (or it would never have found the server). So, it's most likely a consequence of the shortcomings of the PS3 UPnP implementation.

SubJunk:
I have been doing a lot of changes lately to bring us more in line with DLNA and UPnP specs, including using golden devices/renderers, and this is part of that.

Nadahar:
Golden devices?

SubJunk:
See the bottom half of this page https://spirespark.com/dlna/test-tools/

Nadahar:
That doesn't look free to me. Have you convinced them to give you access to these?

valib:
@SubJunk in this case I agree with Nadahar. We still use the old PS3 code but after that there was implemented the Cling which should be able to response automatically to the NOTIFY or M_SEARCH and also do the ALIVE without any other code.

Nadahar:
@valib The "server" (ContentDirectory) isn't using Cling, Cling is only used for remote control of renderers (AVTransport + RenderingService), plus renderer discovery as a kind of "bonus". Since both these services are utilized as a "client", Cling doesn't actually announce anything (UMS has no services from Cling's POV).

SubJunk:
They kindly provided me with a trial.

Yeah that historical reasoning might be true, we may never know the reason for the current state of the code but that seems as likely as any. We can just base our decisions on what works now and this seems to be an improvement for those devices and more closely matches the UPnP spec than what we have. 

I will test it on my PS3 too

SubJunk:
It looks like PS3 uses M-SEARCH instead of NOTIFY so this doesn't affect that renderer - we already show up immediately for that AFAICT

Nadahar:
Yeah, as I said, the PS3 only implement UPnP "half way". It never presents itself, but it does listen and react to media servers' advertisements.

SubJunk:
That doesn't match what I said at all - it presents itself with M-SEARCH immediately

SubJunk:
It should be noted that the PS3 still receives regular software updates, so what has been observed in the past isn't necessarily true now.

Nadahar:
It's been a while since I read about the discovery mechanism now, and I don't really feel like refreshing it just for dialog, but isn't M-SEARCH just a way to find other devices (that matches the search)? I don't see that as advertising itself or "participating", it doesn't tell others that "here I am and these are my capabilities". Instead it just shouts out "Has anyone seen a media server?", and that's all you get out of it. I either remember this completely wrong, or it's very much in line with what I tried to say.

I don't think this is a "recent" update to the PS3, I think you can trigger the M-SEARCH from the menu on the PS3 with "Look for Media Servers" or something like that. Most of the time my PS3 will find xMS pretty quickly, but some times it doesn't. I assume this has to do with the order/timing of when things are started. I will then use that menu option, and within a couple of seconds or so it will usually pop up. While that works OK, if the PS3 actually implemented UPnP properly, no manual action should be necessary.

Mik-S-UMS:
@SubJunk The PS3 has stopped receiving updates for a while now as it has officially reached end of life, the last one was a very minor update to remove support for PS3 <> PS4 messaging. They are not even patching the exploits to install CFW or run homebrew any more. I doubt there has been any change to the internal media player in years.

This might be relevant. When using Movian on my CFW PS3, sometimes UMS will not appear straight away even though UMS sees it and I have to restart Movian till it does. Recently this has stopped working but have found restarting the UMS server makes it appear straight away.

SubJunk:
@Nadahar that's strange as that's not what I experience with PS3, it shows up immediately with no action (I don't even turn on my TV, I just push the power button on PS3 and then PS3 and UMS see each other)

@Mik-S-UMS we are talking about potentially years agon - some parts of this program were written even before PS3 v4 was released :D
It would be great to get your input on testing this branch, I wonder if it will improve that Movian detection?

Mik-S-UMS:
yeah no problem.

Nadahar:
@SubJunk My PS3 will usually have detected the server if it's running when I start the PS3. What I have noticed is that some times when I'm testing things it can't find it, and I have to use the "manual scan". I think that might be after restarting the server. It might be that it would have shown up if I had some more patience, but I've just experiences that doing the "manual scan" solves it right away, so there's really no reason to wait.

My guess is that the next ALIVE broadcast would have straightened it out.

src/main/java/net/pms/network/UPNPHelper.java
```diff
@@ -477,12 +477,18 @@ public void run() {
 
 							boolean redundant = address.equals(lastAddress) && packetType == lastPacketType;
 
-							if (packetType == M_SEARCH) {
-								if (configuration.getIpFiltering().allowed(address)) {
-									String remoteAddr = address.getHostAddress();
-									int remotePort = receivePacket.getPort();
+							if (configuration.getIpFiltering().allowed(address) && getDevice(address) != null) {
```
SubJunk, line 480: @valib this change breaks the functionality - `getDevice` means it will only work for devices we know about, but this is useful for discovering _unknown_ devices

valib:
@SubJunk it is not 100% working because it respons to the first request of the network devices but the next requests are ignored.

SubJunk:
@valib is the logic here to not respond to devices we do not recognize?

Also I'm getting an error:
```
14:36:19 INFO  Exception in thread "UPNPHelper" java.lang.StringIndexOutOfBoundsException: begin -1, end 25, length 171
14:36:19 INFO  	at java.base/java.lang.String.checkBoundsBeginEnd(Unknown Source)
14:36:19 INFO  	at java.base/java.lang.String.substring(Unknown Source)
14:36:19 INFO  	at net.pms.network.UPNPHelper$2.run(UPNPHelper.java:487)
14:36:19 INFO  	at java.base/java.lang.Thread.run(Unknown Source)
```
which is this line:
```
UDN udn = UDN.valueOf(s.substring(uuidPosition, s.indexOf(":", uuidPosition + UUID.length())));
```

valib:
@SubJunk thanks for the comment. This code doesn't affect current logic. It only check the already ignored devices and not send the response to them.

valib:
@SubJunk I checked the code and found that there is a method which do the same as method I made. Also I found the problem in the implementation of checking if the Device is in the ignored devices list when the Device ís null. Now it should works as expected.

SubJunk:
@valib yes that's a problem I have too. The codebase is also just so big it's hard to find things.

After these recent commits it seems to take UMS longer to show up on my TV but I need to test more. I will do that soon

valib:
@SubJunk I made a wrong assumption about the message string received and  I didn't realised that it is line oriented. I made changes which refect this and now it works as expected. Can you checked it?

SubJunk:
@valib I have tested it and it detects VLC on iOS correctly, but how can I test the ignoring part?

valib:
@SubJunk you need to have on the net the device which send NOTIFY but is not a renderer see the part of my log:

```
DEBUG 2020-07-29 01:16:12.182 [pool-10-thread-1] Estimated speed from ICMP packet size 64000 in 43 fragment(s) is 3.374658064516129E7 bit/s
INFO  2020-07-29 01:16:12.183 [pool-10-thread-1] Renderer [TV] Samsung Q9 Series (55) has an estimated network speed of 17.633849323837453 Mb/s
DEBUG 2020-07-29 01:16:12.394 [SpeedStats Failsafe] Stopping process: ping-11
DEBUG 2020-07-29 01:16:12.397 [pool-10-thread-2] Estimated speed from ICMP packet size 64000 in 43 fragment(s) is 4.0236307692307696E7 bit/s
INFO  2020-07-29 01:16:12.397 [pool-10-thread-2] Renderer [AV] Samsung Soundbar MS750 has an estimated network speed of 17.463914005994006 Mb/s
TRACE 2020-07-29 01:16:13.197 [UPNPHelper] IP Filter: No IP filter specified, access granted to 10.0.0.14
TRACE 2020-07-29 01:16:13.197 [UPNPHelper] Ignoring request from device with UUID: [uuid:upnp-InternetGatewayDevice-100014345678900001]
TRACE 2020-07-29 01:16:13.215 [UPNPHelper] Ignoring request from device with UUID: [uuid:upnp-InternetGatewayDevice-100014345678900001]
TRACE 2020-07-29 01:16:13.231 [UPNPHelper] Ignoring request from device with UUID: [uuid:upnp-InternetGatewayDevice-100014345678900001]
TRACE 2020-07-29 01:16:13.247 [UPNPHelper] Ignoring request from device with UUID: [uuid:upnp-InternetGatewayDevice-100014345678900001]
TRACE 2020-07-29 01:16:13.266 [UPNPHelper] Ignoring request from device with UUID: [uuid:upnp-WANDevice-100014345678900001]
TRACE 2020-07-29 01:16:13.282 [UPNPHelper] Ignoring request from device with UUID: [uuid:upnp-WANDevice-100014345678900001]
TRACE 2020-07-29 01:16:13.297 [UPNPHelper] Ignoring request from device with UUID: [uuid:upnp-WANDevice-100014345678900001]
TRACE 2020-07-29 01:16:13.314 [UPNPHelper] Ignoring request from device with UUID: [uuid:upnp-WANConnectionDevice-100014345678900001]
TRACE 2020-07-29 01:16:13.330 [UPNPHelper] Ignoring request from device with UUID: [uuid:upnp-WANConnectionDevice-100014345678900001]
TRACE 2020-07-29 01:16:13.347 [UPNPHelper] Ignoring request from device with UUID: [uuid:upnp-WANConnectionDevice-100014345678900001]
DEBUG 2020-07-29 01:16:27.397 [UPNP-AliveMessageSender] Sending ALIVE...
TRACE 2020-07-29 01:16:27.397 [UPNP-AliveMessageSender] Setting SSDP network interface: name:wlan2 (Intel(R) Dual Band Wireless-AC 3160)
TRACE 2020-07-29 01:16:27.480 [UPNPHelper] IP Filter: No IP filter specified, access granted to 10.0.0.10
TRACE 2020-07-29 01:16:33.197 [UPNPHelper] Ignoring request from device with UUID: [uuid:upnp-InternetGatewayDevice-100014345678900001]
TRACE 2020-07-29 01:16:33.214 [UPNPHelper] Ignoring request from device with UUID: [uuid:upnp-InternetGatewayDevice-100014345678900001]
TRACE 2020-07-29 01:16:33.230 [UPNPHelper] Ignoring request from device with UUID: [uuid:upnp-InternetGatewayDevice-100014345678900001]
TRACE 2020-07-29 01:16:33.247 [UPNPHelper] Ignoring request from device with UUID: [uuid:upnp-InternetGatewayDevice-100014345678900001]
TRACE 2020-07-29 01:16:33.265 [UPNPHelper] Ignoring request from device with UUID: [uuid:upnp-WANDevice-100014345678900001]
TRACE 2020-07-29 01:16:33.282 [UPNPHelper] Ignoring request from device with UUID: [uuid:upnp-WANDevice-100014345678900001]
TRACE 2020-07-29 01:16:33.297 [UPNPHelper] Ignoring request from device with UUID: [uuid:upnp-WANDevice-100014345678900001]
TRACE 2020-07-29 01:16:33.314 [UPNPHelper] Ignoring request from device with UUID: [uuid:upnp-WANConnectionDevice-100014345678900001]
TRACE 2020-07-29 01:16:33.332 [UPNPHelper] Ignoring request from device with UUID: [uuid:upnp-WANConnectionDevice-100014345678900001]
TRACE 2020-07-29 01:16:33.347 [UPNPHelper] Ignoring request from device with UUID: [uuid:upnp-WANConnectionDevice-100014345678900001]
TRACE 2020-07-29 01:16:53.198 [UPNPHelper] Ignoring request from device with UUID: [uuid:upnp-InternetGatewayDevice-100014345678900001]
TRACE 2020-07-29 01:16:53.214 [UPNPHelper] Ignoring request from device with UUID: [uuid:upnp-InternetGatewayDevice-100014345678900001]
TRACE 2020-07-29 01:16:53.231 [UPNPHelper] Ignoring request from device with UUID: [uuid:upnp-InternetGatewayDevice-100014345678900001]
TRACE 2020-07-29 01:16:53.248 [UPNPHelper] Ignoring request from device with UUID: [uuid:upnp-InternetGatewayDevice-100014345678900001]
TRACE 2020-07-29 01:16:53.271 [UPNPHelper] Ignoring request from device with UUID: [uuid:upnp-WANDevice-100014345678900001]
TRACE 2020-07-29 01:16:53.280 [UPNPHelper] Ignoring request from device with UUID: [uuid:upnp-WANDevice-100014345678900001]
TRACE 2020-07-29 01:16:53.296 [UPNPHelper] Ignoring request from device with UUID: [uuid:upnp-WANDevice-100014345678900001]
TRACE 2020-07-29 01:16:53.313 [UPNPHelper] Ignoring request from device with UUID: [uuid:upnp-WANConnectionDevice-100014345678900001]
TRACE 2020-07-29 01:16:53.331 [UPNPHelper] Ignoring request from device with UUID: [uuid:upnp-WANConnectionDevice-100014345678900001]
TRACE 2020-07-29 01:16:53.347 [UPNPHelper] Ignoring request from device with UUID: [uuid:upnp-WANConnectionDevice-100014345678900001]
DEBUG 2020-07-29 01:16:58.993 [UPNP-AliveMessageSender] Sending ALIVE...

SubJunk:
Wow that's a lot of spam, nice work ignoring it. I will see if I have anything similar, probably on my wireless interface which I usually don't use

SubJunk:
This is waiting on me to test it more with my TV

SubJunk:
I have tested `9e4d97f` but not `8212a46` so I need to do that

valib:
merged in #2112

Mik-S-UMS:
I have noticed that recently that Movian is getting added to the ignored list and not showing if it is restarted on the PS3 after having been successfully detected the first time. same for my Samsung TV (not all services but maybe one is needed that is ignored?), but restarting the server allows UMS to appear on both fine.

[ums_dbg_2020-09-11-12-14.zip](https://github.com/UniversalMediaServer/UniversalMediaServer/files/5248404/ums_dbg_2020-09-11-12-14.zip)

sorry for the long log. I don't turn off my pc at night, just put it to sleep and usually watch something before bed before turning of the PS3 and TV. It will be the next day when UMS won't show when I go to watch something else. UMS was running for about a week and I don't exactly know why it is happening to recreate just this issue.

valib:
@Mik-S-UMS why do you set `UpnpAllow = false` to your PS3-Movian.conf?

Mik-S-UMS:
Not sure, I think I assumed it was just for remote control when I wrote the conf so set it to false as Movian doesn't support that.
I changed it back to true and after restarting the PS3 a few times it appears to be working as it should. Thanks.

valib:
@Mik-S-UMS the UPnP is also used for recognition of the renderer so disabling it make the recognition not working properly.

Issue #2197, opened 2020-07-17
Getting multiples of same device


purevertigo:
I am getting weird multiple entries for same device and some are showing 0% instead of IP.

Image is here:

![Annotation 2020-07-17 173351](https://user-images.githubusercontent.com/6506461/87760952-fc5cc380-c853-11ea-843e-92890c765211.jpg)

Logs attached

[ums_dbg_2020-07-17-14-20.zip](https://github.com/UniversalMediaServer/UniversalMediaServer/files/4936369/ums_dbg_2020-07-17-14-20.zip)

valib:
@purevertigo this renderer is not properly recognized. The User Agent send by TV is `USER-AGENT: Linux/4.1.20-1.9, UPnP/1.0, DLNADOC/1.50` but the .conf has `UserAgentSearch = FetchTV`. You can try to change it to something like `UserAgentSearch = Linux/4.1.20-1.9` but it is so general that other .confs can match. In that case go to the General Setting Tab and select only renderers presented on your net. The rest disable.

purevertigo:
OK. Will give it a try and report back @valib

SubJunk:
I think I noticed something similar lately too, the 0% thing. I haven't looked into it. Are there any previous versions of UMS that doesn't happen on?

purevertigo:
This (7.6.2) is the first version it is occurring on. Never had this before.

purevertigo:
@valib That workaround has solved the issue BUT is there a fix coming for this issue so I don't have to 'select' the conf everytime a new device comes onto the network (as I mentioned to @SubJunk this only started happening in 7.6.2).

Also unrelated I have an unknown renderer on my PC apparently, what is this? (attached)

![Annotation 2020-07-18 200405](https://user-images.githubusercontent.com/6506461/87850267-31d7de80-c932-11ea-9963-6cd04b043e38.jpg)

valib:
@purevertigo without log we can't say what is this. I guess that this is the Windows Media Player running on your PC.

valib:
@purevertigo did you try to let all renderers enabled and only change the FetchTV.conf?

SubJunk:
For me it happens if I force networking on my wifi card. It might be a bug with the interface detection change we merged recently

purevertigo:
@SubJunk @valib I am still testing. It seems random.

Also I have done everything I can to disable Media Player from DLNA. Still testing...

[...]

PR #2224, opened 2020-08-13
Don't send the M-SEARCH

valib:
with the address and port of the requesting device. This is a nonsense when the M-SEARCH should be send by the multicast cocket.
Maybe it could fix the issue #2197


valib:
@SubJunk have you tested this PR?

SubJunk:
It looks like I added this in https://github.com/UniversalMediaServer/UniversalMediaServer/commit/98b2e8a1fcce4caf071200a4b1bd6b27cbd50bb2 but I'm not sure why
I'll test now

SubJunk:
I have tested this with my phone and TV and it seems that everything works as it did before, so if you are confident this is the right thing to do @valib we can merge it

SubJunk (approved):

PR #2233, opened 2020-08-28
Update some deprecated methods

valib:


valib:
It was only one wrong change and wrong compilation. Now it is OK.

valib:
@SubJunk we provide the Java 14 with the UMS but the code is compilled as Java 1.8. This is why this PR failed when I changed https://github.com/UniversalMediaServer/UniversalMediaServer/blob/4c823bb84ccf8cee6a4309506b2365867fa28eb4/src/main/java/net/pms/remote/RemoteWeb.java#L429 to 

`String str = bytes.toString(StandardCharsets.UTF_8);` 

which is supported since Java 9 but not in Java 1.8. Do you think that we should still keep the code in the Java 1.8 and not update it to meet the Java 14?

SubJunk:
@valib yes we should update it to compile as Java 14. Feel free to do that in this branch.

valib:
@SubJunk if we are going towards the Java 14 than testing for Java 8 in Github should be removed.

SubJunk:
@valib The tests are there to make sure we don't break that. I think it is fine to optimise for Java 14, but only when it doesn't break support for Java 8.

My opinion is that if Java 8 becomes deprecated we can drop support for that, but for now, Java 8 and 14 are both receiving updates, and some of our users will be using 8.

SubJunk:
@valib these changes are looking good, maybe we should review and merge soon before it gets much bigger?

valib:
@SubJunk I agree.

SubJunk (approved):

PR #1653, opened 2018-11-03
Removing deprecated variables and methods and making public variables private

valib:
This is the first batch and I will continue.

PR #2255, opened 2020-10-02
Partially revert the PR #1653

valib:
This should fix the reported bug

### DIFF
--- a/src/main/java/net/pms/dlna/DLNAResource.java
+++ b/src/main/java/net/pms/dlna/DLNAResource.java
@@ -1443,7 +1443,7 @@ public abstract class DLNAResource extends HTTPResource implements Cloneable, Ru
 	/**
 	 * Use {@link #syncResolve()} instead
 	 */
-	private void resolve() {
+	public void resolve() {
 		if (!resolved) {
 			resolveOnce();
 			// if resolve() isn't overridden, this file/folder is immutable
--- a/src/main/java/net/pms/dlna/FolderLimit.java
+++ b/src/main/java/net/pms/dlna/FolderLimit.java
@@ -63,7 +63,7 @@ public class FolderLimit extends VirtualFolder {
 	}
 
 	@Override
-	public synchronized void syncResolve() {
+	public synchronized void resolve() {
 		this.setDiscovered(false);
 		this.getChildren().clear();
 	}
--- a/src/main/java/net/pms/dlna/FolderLimitLevel.java
+++ b/src/main/java/net/pms/dlna/FolderLimitLevel.java
@@ -22,7 +22,7 @@ public class FolderLimitLevel extends VirtualFolder {
 		} else {
 			start = r.getParent().clone();
 		}
-		syncResolve();
+		resolve();
 	}
 
 	@Override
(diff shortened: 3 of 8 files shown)


valib:
@SubJunk somwhere is the problem. The replace of `resolve()` to `syncResolve()` is not doing what expected. Can you help to find what is wrong?

SubJunk:
@valib maybe it is because of that `synchronized` keyword on `syncResolve`, I'm not sure. I think it's fine to just remove the `@Deprecated` for now, it obviously wasn't ready to be deprecated. Thanks for the fix!

SubJunk (approved):

Nadahar:
You shouldn't assume that `@deprecated` means "can be removed". I put that on `resolve()`, and it was to indicate that it code shouldn't call it, but call `syncResolve()` instead. That doesn't mean that `resolve()` can just be removed. This construct was somewhat of a "hack" to be able to make "resolve" synchronized without making a huge refactoring of it (and breaking plugins). Since `resolve()` is inherited in many subclasses, and that some subclasses call `resolve()` for other classes within the `resolve()` method, I deemed it to messy to be possible to change. Having synchronized methods overridden is a problem as well, as one must make sure that all overrides are synchronized as well. Since the plugins was a big headache at this time, and we had no control over them, making a `final` `syncResolve()` as a "wrapper" that made sure that the whole "resolve process" was serialized seemed to me to be the "best" (or least-worst) solution.

I think hunting for `@deprecated` and removing these is misguided and will cause many problems. The involved code must be fully understood before being changed. I see `@deprecated` as meaning: Still potentially in use, but do not use in new code.

SubJunk:
@valib it also seems like #1653 broke the sleep functionality - my computer doesn't go to sleep automatically now if UMS is open, but it did before

SubJunk:
@valib sorry for my previous post, it must have been something else preventing sleep, it is working correctly for me now

SubJunk:
@valib It looks like I was wrong about being wrong, it is UMS preventing sleep and I didn't experience that before this.

I ran `powercfg -requests` on Windows 10 to verify it, it gave the output:

```
C:\WINDOWS\system32>powercfg -requests
DISPLAY:
None.

SYSTEM:
[PROCESS] \Device\HarddiskVolume6\Program Files (x86)\Universal Media Server\jre15\bin\javaw.exe

AWAYMODE:
None.

EXECUTION:
None.

PERFBOOST:
None.

ACTIVELOCKSCREEN:
None.
```

I have played videos on this instance of UMS but none are playing

valib:
@SubJunk I checked it and it works. The first is when playing the second when UMS is iddle

```
C:\WINDOWS\system32>powercfg -requests
DISPLAY:
None.

SYSTEM:
[PROCESS] \Device\HarddiskVolume5\Program Files (x86)\Universal Media Server\jre15\bin\javaw.exe

AWAYMODE:
None.

EXECUTION:
None.

PERFBOOST:
None.

ACTIVELOCKSCREEN:
None.


C:\WINDOWS\system32>powercfg -requests
DISPLAY:
None.

SYSTEM:
None.

AWAYMODE:
None.

EXECUTION:
None.

PERFBOOST:
None.

ACTIVELOCKSCREEN:
None.
```
What do you have set for "Prevet sleep"? "During playback" or "While running"

SubJunk:
"During playback". I had also been testing on the web interface, but had closed the UMS tab and nothing was playing.

SubJunk:
I just tried and I can't reproduce the problem reliably. I'll keep trying

PR #2257, opened 2020-10-06
Do not use virtual network interface

valib:
This fix the problem reported in http://www.universalmediaserver.com/forum/viewtopic.php?f=9&t=14389&p=42648#p42648
Using the virtual adapter even if it has an asociated address and it is first on the list is nonsense.


Nadahar:
Why is it? What if UMS is running on a VM - then all adapters are "virtual"..?

SubJunk:
@valib did you accidentally commit to master? https://github.com/UniversalMediaServer/UniversalMediaServer/commit/a328bc09712079e207e4e0099a8494eeb2227cf6

PrplHaz4:
> 
> 
> Why is it? What if UMS is running on a VM - then all adapters are "virtual"..?

I believe adapters within VMs are usually presented as physical not virtual (only the host will see it as virtual)

Nadahar:
> I believe adapters within VMs are usually presented as physical not virtual (only the host will see it as virtual)

I don't think there are any "hard rules" for that, I'm pretty sure I've seen "virtual" on many VM NICs, but I'm also sure that varies between virtualization systems and "virtual drivers".

I'm just skeptical on excluding adapters based on their name, but I guess one could make a logic where "virtual" adapters would only be chosen if there are no viable "non-virtual" ones.

PrplHaz4:
> 
> 
> > I believe adapters within VMs are usually presented as physical not virtual (only the host will see it as virtual)
> 
> I don't think there are any "hard rules" for that, I'm pretty sure I've seen "virtual" on many VM NICs, but I'm also sure that varies between virtualization systems and "virtual drivers".
> 
> I'm just skeptical on excluding adapters based on their name, but I guess one could make a logic where "virtual" adapters would only be chosen if there are no viable "non-virtual" ones.
> 
> Another "challenge" with said approach is that Windows (at least) localizes the names of network adapters, so you'd have to cover "virtual" in all the world's languages.

I believe the way the Jellyfin project (which is currently working on the same problem) did it is to prefer adapters that have a gateway defined (as well as ignoring known "bad" strings).  I thought Windows used "vEthernet" - which may not be localized?

You are correct either way, it is a challenge on many levels...hard to get right on any of them...

Nadahar:
Adapters with a gateway isn't a bad idea, and the adapter with the default route might be the "best pick" in most cases where VPN isn't involved. VPN adapters is problematic though, as they will generally "claim" the default route.

All that said, I don't know how much of this information is even available from Java. That's another big problem, Java "abstracts" by hiding most of the actually useful information like this. I don't remember how much of this information is available through Java though, so it might not be an issue. Avoiding VPN adapters is still tricky though, you can't rely on them having "VPN" in their name.

valib:
@SubJunk sorry to commit it to master was an accident. Reverted now.

valib:
@Nadahar as is written in the comment the "virtual" interfaces are still there and can be choozen by user. This change only not choose this vitual interface automatically. I think that users working with VM are not standard users. Standard user can have a problem with virtual interface like reported in http://www.universalmediaserver.com/forum/viewtopic.php?f=9&t=14389&p=42648#p42648

Nadahar:
@valib Yes, I think I understand your logic, but I disagree. It's not about "being advanced" or not, but rather about knowing UMS and how this works. If you don't know that UMS is going to pick a network interface for you if you don't "force" it, you won't get this to work regardless of how "advanced" you are. Also, VM's are becoming quite widespread.

But, you do as you see fit - I just wanted to mention that I think it is somewhat short-sighted and yet another thing that leads to "strange behavior" as seen from a user perspective.

valib:
@Nadahar thanks for the comments. This is only a PR which could be merged or not. It depends on @SubJunk.

valib:
> Another "challenge" with said approach is that Windows (at least) localizes the names of network adapters, so you'd have to cover "virtual" in all the world's languages.

@Nadahar are you sure? In my czech localized Windows interface names are the same as is written in the [Strings] part in their interface.inf and the same should be in other languages.

valib:
> I'm just skeptical on excluding adapters based on their name, but I guess one could make a logic where "virtual" adapters would only be chosen if there are no viable "non-virtual" ones.

@Nadahar this is a good idea and I will try to implement it.

Nadahar:
@valib It's hard to be sure of anything, names are often the same regardless of language. What I do know is that there is a software, I don't remember which now, that fails to work on Norwegian Windows versions because the network connections are localized. The "names" of the connections themselves are clearly in Norwegian, but the "string on the 3rd line" that is for example `TAP-Windows Adapter V9` might be the name gotten from the driver itself as you say. It's hard to say how the internals here work.

SubJunk:
@valib are you wanting this to be merged? I'm not clear based on the discussion

valib:
@SubJunk without testing by users we will never know if it is correct. For me it works.

SubJunk:
I agree with you Valib that it is better to not choose virtual interfaces by default. Ideally I would like us to support multiple interfaces at the same time, but that is really hard to do in Java it seems :(
I will test this on my machine now

Nadahar:
@SubJunk 
>  I would like us to support multiple interfaces at the same time, but that is really hard to do in Java it seems :(

There's no special restriction when it comes to that in Java, but this is complicated to do. UPnP requires that you keep track of which interface traffic arrives on and use the same when replying, as far as I remember. You also can't just "bind" to multiple interfaces (on the OS/network stack level), unless you use ANY/`0.0.0.0`/`::`, so you'd have to actually run multiple "HTTP server instances" in parallel. I can't possibly see how this would be worth it, considering that you'd need a pretty "special" network setup for that to be useful.

SubJunk:
Yeah I have tried to do that before but it got pretty complicated.

I would have thought the situation is common since I have it on my PC, though probably it is becoming less common as people tend to have laptops instead. At home I always need to force the network interface because it has a wired and wireless interface that I use - I use wired for the media network, and wireless for Internet. So if UMS binds to the wireless interface I can't use it. That was the problem I was trying to solve when I tried.

Nadahar:
I thought it was common to let the router(s) handle this, so that all devices have internet access. I have one router handling internet connectivity and firewall and one (transparent/bridged) wireless access point that is wired to the "internet router". That way all devices can see each other, regardless of if they are wired or wireless, and all have internet access.

Why would you have the PC be the "connection point" between the local network and the Internet? That's typically the job of a router. If you wish to restrict internet access to some devices, that can easily be controlled in the router by using (MAC bound) DHCP and then creating firewall rules that block those that shouldn't reach the Internet. The router should allow no incoming access unless specific rules are created to allow specific ports to specific devices, so incoming access is nothing to "worry about" regardless, so the only "point" in not connecting devices to the Internet is to prevent outgoing access.

SubJunk:
There's another case that is maybe more common with the same kind of network configuration that I have - wanting to serve media on both. For example I might want to serve high quality media over the wired network to a 4K TV, so it's not suited to Wi-Fi, but also sometimes watch lower quality things on my phone via Wi-Fi. It would be good to be able to do that with one instance of UMS.

I could probably change my network configuration to work around it but it would be nice if it just worked, since users may not have the ability or interest in learning how to reconfigure their network. The experience I want for our users is that everything just "magically" works, and I think similar products do it. Plex handles serving on multiple interfaces for sure.

Nadahar:
I don't know what/how Plex does things, but are you sure that they doesn't just bind to "ANY"? That can be done without running multiple bindings, but there are some differences between OSes regarding the details AFAICR.

When it comes to handling wired and wireless from the same server instance, I have had no issues with this. That said, I haven't yet watched 4K videos (I have no 4K equipment and not much interest), so maybe that's why I haven't come across it. But, "automatic bitrate" should be able to compensate for this, since it tunes the bandwidth on a per-renderer basis.

When it comes to "just working" I agree that it's desirable, but there are so many ways to configure networks that it might be impossible or at the very least impractical to do in reality. My stance is thus that one should try to make it work on "typical" configurations by default, and then assume that those with the more "creative network setups" have enough knowledge to figure out how to do things. My assumption is that most routers today handle both internet access, wired ports and the wireless network in once box, which make things very straight forward. Alternatively a solution like mine, with one wired router and a wireless access point connected to the internal network. The network topology will be "the same" in both situations, one IP subnet that is shared between the wired and the wireless network.

valib:
Guys this PR will prefere the first non-virtual interface with relevant IPV4 address and if such interface not exists than the first virtual interface with relevant IPV4 address is used. I can add warning to the log that using the virtual interface can cause problems. I know that there are many other interfaces not exactly stated that they are virtual but they obviously dont have a relevant IPV4 address.

SubJunk:
@valib sorry for going off-topic in your PR, we were just discussing the bigger picture. I still think this PR is an improvement on what we have.

@Nadahar I think I tried the "any" binding when I was trying to get that working but I had problems. It was a while ago so I can't remember why.
What programming project/s are you working on these days? Just curious

valib:
@SubJunk fine but what about to review this PR?

SubJunk:
I just tested it and it worked fine, though I didn't have the problem with virtual interfaces so my testing just involved confirming that it still works as it did before.
Will do a code review next

src/main/java/net/pms/network/NetworkConfiguration.java
```diff
@@ -297,15 +298,24 @@ private void checkNetworkInterface(NetworkInterface networkInterface, String par
...
 						final InterfaceAssociation ia = new InterfaceAssociation(address, networkInterface, parentName);
 						interfaces.add(ia);
-						mainAddress.put(networkInterface.getDisplayName(), ia);
+						try {
+							// Add network interface to the list of interfaces with associated address only
+							// if interface is supporting multicast.
+							if (networkInterface.supportsMulticast()) {
+								interfacesWithAssociatedAddress.put(networkInterface.getDisplayName(), ia);
+								LOGGER.trace("added interface {} with associated address {} supporting multicast to the list of relevant addresses", networkInterface.getName(), ia.getAddr());
+							}
```
SubJunk, line 314:
```suggestion
							} else {
								LOGGER.trace("Ignoring interface {} with associated address {} because it does not seem to support multicast", networkInterface.getName(), ia.getAddr());
							}
```

src/main/java/net/pms/network/NetworkConfiguration.java
```diff
@@ -381,16 +391,33 @@ public InterfaceAssociation getDefaultNetworkInterfaceAddress() {
...
+		for (Entry<String, InterfaceAssociation> entry : interfacesWithAssociatedAddress.entrySet()) {
+			if (entry.getValue().getDisplayName().toLowerCase().contains("virtual")) {
+				virtualInterfaces.put(entry.getKey(), entry.getValue());
+				continue;
+			}
+
+			return entry.getValue();
+		}
+
+		// The non-virtual interface was not found so choose the first virtual one if exists
```
SubJunk, line 417:
```suggestion
		// We did not find any non-virtual interfaces, so choose the first virtual one if it exists
```

SubJunk (changes requested):
The logic seems good to me, I just left some comment/logging suggestions before approving

Nadahar:
@SubJunk I'm still working on DMS locally. In addition I maintain some Maven plugins and have been looking a bit at OpenHAB and whatever else seems interesting.

SubJunk:
@Nadahar OpenHAB looks cool, I hadn't heard of that

@valib thanks, will merge

SubJunk (approved):

Nadahar:
@SubJunk I'm already using it for controlling some lights, for smoke detectors, water heating in the stable (so the pipes don't freeze), for the electric fence etc... it's not exactly bug-free, but it works reasonably well once you figure out how to work it. Documentation is somewhat sparse though, so it takes some effort.

SubJunk:
That sounds great. I use a similar product but it's not open source

PR #2259, opened 2020-10-11
JRE 15

SubJunk:


valib:
@SubJunk there are missing files in the https://www.universalmediaserver.com/svn/binaries/linux/x86_64/jre15/lib/ or there are not relevant URLs in the `pom.xml`
I have got ` [get] Error opening connection java.io.FileNotFoundException: https://www.universalmediaserver.com/svn/binaries/linux/x86_64/jre15/lib/libawt.so?p=117`

SubJunk:
Good catch, it's weird because my SVN client ignores `.so` files but catches all other ones. I'll look in the settings and fix this

SubJunk:
@valib that should be fixed now

valib:
@SubJunk it is fixed but why the `linux-bionic` and `macos` are still waiting for the status to be reported?

valib:
@ @SubJunk checking if Java 14 or 11 is working seems to me not relevant when you set java 15 as the default version for the UMS.

SubJunk:
@valib when this merges, I can change the GitHub settings to fix that. It's because we are changing the versions of required pipelines in this branch

valib (approved):

Issue #2247, opened 2020-09-21
Save log zip to desktop by default


SubJunk:
See https://github.com/UniversalMediaServer/UniversalMediaServer/issues/2124#issuecomment-695110583

dotslash21:
Hi @SubJunk, Can I take up this issue?

dotslash21:
Hi @SubJunk, I did some digging through the code and observed that the default log file save location is given by `PmsConfiguration.getDefaultLogFileFolder()`. Currently, the default save location is generated using the following rules:

1. (On Linux only) path to `/var/log/ums/%USERNAME%/`.
2. Path to profile folder (`~/.config/UMS/` on Linux, `%ALLUSERSPROFILE%\UMS` on Windows and `~/Library/Application Support/UMS/` on Mac).
3. Path to the user-defined temporary folder specified by the `temp_directory` parameter in UMS.conf.
4. Path to system temporary folder.
5. Path to the current working directory.

At this point, I have some questions,

1. Do I save the log zip to the desktop for all platforms at first then try all other existing fallback location in the order?
2. If not, then what should be the priority saving to desktop.
3. Should I save to desktop for all the platforms?

SubJunk:
@dotslash21 yes you're very welcome to work on this. Good questions, thanks. I think attempting to save to desktop is good for all operating systems

PR #2267, opened 2020-10-19
Try saving log zip to desktop first

dotslash21:
fixes: #2247

Hi @SubJunk, As per our discussion https://github.com/UniversalMediaServer/UniversalMediaServer/issues/2247#issuecomment-711106316, I've implemented the changes. Please review my PR and confirm the same.

I request you to label this PR as `hacktoberfest-accepted` if you deem this PR as a valid contribution, that would help me very much.

Thanks.

### DIFF
--- a/src/main/java/net/pms/configuration/PmsConfiguration.java
+++ b/src/main/java/net/pms/configuration/PmsConfiguration.java
@@ -379,6 +379,9 @@ public class PmsConfiguration extends RendererConfiguration {
 	// Path to default logfile directory
 	protected String defaultLogFileDir = null;
 
+	// Path to default zipped logfile directory
+	protected String defaultZippedLogFileDir = null;
+
 	public TempFolder tempFolder;
 	@Nonnull
 	protected final PlatformProgramPaths programPaths;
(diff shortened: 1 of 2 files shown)


valib:
@dotslash21 it works but I would recommend to store "debug.log" and "debug.log.prev" back to the UMS defined folder and store in the new folder only zip file. Also create only e.g. "UMS-log" folder instead of create folder "UMS" and subfolder "log".

SubJunk:
@dotslash21 thanks, we will be happy to add the `hacktoberfest-accepted` label when @valib 's suggestions are implemented

dotslash21:
@SubJunk, @valib thanks for reviewing my code. If the requirement is trying to save only the zip file on the desktop first, then I'm thinking of creating a method (say `getDefaultZippedLogFilePath()`) that will first verify the path to the desktop, then return it if ok, otherwise it will fall back to the default path from `PmsConfiguration.getDefaultLogFilePath()`.

At this point, I have a question. where will it be best to create this method?

I have considered 2 possible answers,

1. At `DbgPacker.java`, because all the zip logic seems to be there so creating this method here adheres to the single responsibility principle. But creating the method here would need me to re-implement all the path verification logic already present in `PmsConfiguration`.

2. At `PmsConfiguration.java`, because it is concerned with all the config of UMS. Also, in the future, if zipping log file functionality is introduced in say, a web interface then getting the zip configs from a central config source will reduce code duplication.

dotslash21:
@SubJunk, @valib I assume that you have missed my [last comment](https://github.com/UniversalMediaServer/UniversalMediaServer/pull/2267#issuecomment-712923650) given the load due to `Hacktoberfest`. I've taken the liberty to go ahead with the option [2](https://github.com/UniversalMediaServer/UniversalMediaServer/pull/2267#issuecomment-712923650). Could you please review the code and check if it is in line with the requirements?

Thanks :smiley:

valib (dismissed):

valib (approved):

valib:
I feel that is the time to formate this whole class.

valib:
Merged thanks @dotslash21

dotslash21:
Hi @valib, could you please label this PR as `hacktoberfest-accepted`? That would be very helpful as this will be my 4th PR for hacktoberfest.

Mik-S-UMS:
Had trouble finding the logs, couldn't see them on my desktop after saving them.

After a bit of searching I found them in C:\Users\Mik\Desktop\UMS-log as I guess should be expected

However I use a small boot SSD and most of my user directories like Downloads and Desktop are pointed to D:\Users\Mik\

I would assume %USERPROFILE%\Desktop would point to D: in my case but it appears it does not. Is there a way of using the actual Desktop directory being used or is my case a strange one to detect?

PR #2276, opened 2020-10-29
Add support for webp images

valib:
This is the first part of the implementation and this image format is not ignored and shown. But I have a problem to implement the https://github.com/sejda-pdf/webp-imageio to read those images. It raised an exception when instantiate the class.

EDIT: finally solved


valib:
Now it is working

src/main/java/net/pms/dlna/LibMediaInfoParser.java
```diff
@@ -14,6 +14,7 @@
 import net.pms.formats.Format;
 import net.pms.formats.v2.SubtitleType;
 import net.pms.image.ImageFormat;
+import net.pms.image.ImageInfo;
```
SubJunk: This seems unused?

src/main/java/net/pms/image/ImagesUtil.java
```diff
@@ -1333,7 +1333,7 @@ protected static Image transcodeImage(
 			throw new UnknownFormatException("Unable to read image format", e);
 		}
 
-		if (inputResult.bufferedImage == null || inputResult.imageFormat == null) { // ImageIO doesn't support the image format
+		if (inputResult.bufferedImage == null) { // ImageIO doesn't support the image format
```
SubJunk: Why don't we need to check for that?

valib: @SubJunk this was only for the testing. I returned it back in the final version

src/main/java/net/pms/dlna/LibMediaInfoParser.java
valib: @SubJunk please check the final version

valib:
@SubJunk during testing I have found that the latest implementation of the WIN10 Edge browser is not properly recognized and the icon is obsolete so I fix it.

valib:
@SubJunk I think that we should make the correct description of the new classes or methods to keep the code understandable. This is the main problem of the current code. So we should add the description for the old code whenever is possible.

SubJunk (approved):

SubJunk:
Nice work

PR #2296, opened 2020-11-26
Fix external subtitles streaming if renderers do not have any "se" defined in Supported lines

### LINKED ISSUE #2052: Embedded subtitles used instead of external
User jkwu mentioned a regression in v8+ at https://www.universalmediaserver.com/forum/viewtopic.php?f=9&t=14007&start=10#p41144

@valib

SubJunk:
Fixes #2052

### DIFF
--- a/src/main/java/net/pms/configuration/FormatConfiguration.java
+++ b/src/main/java/net/pms/configuration/FormatConfiguration.java
@@ -198,9 +198,7 @@ public class FormatConfiguration {
 		private String mimeType;
 		private String videoCodec;
 		private String supportLine;
-		// List of embedded subs supported by renderer defined in the Supported line in the renderer.conf
 		private String supportedEmbeddedSubtitlesFormats;
-		// List of external subs supported by renderer defined in the Supported line in the renderer.conf
 		private String supportedExternalSubtitlesFormats;
 
 		SupportSpec() {
--- a/src/main/java/net/pms/configuration/FormatConfiguration.java
+++ b/src/main/java/net/pms/configuration/FormatConfiguration.java
@@ -198,10 +198,10 @@ public class FormatConfiguration {
 		private String mimeType;
 		private String videoCodec;
 		private String supportLine;
-		/** List of embedded subs supported by renderer defined in the Supported line in the renderer.conf */
-		private String embeddedSubs;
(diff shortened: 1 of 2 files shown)


src/main/java/net/pms/configuration/FormatConfiguration.java
```diff
@@ -198,10 +198,10 @@
 		private String mimeType;
 		private String videoCodec;
 		private String supportLine;
-		/** List of embedded subs supported by renderer defined in the Supported line in the renderer.conf */
-		private String embeddedSubs;
-		/** List of external subs supported by renderer defined in the Supported line in the renderer.conf */
-		private String externalSubs;
+		// List of embedded subs supported by renderer defined in the Supported line in the renderer.conf
+		private String supportedEmbeddedSubtitlesFormats;
+		// List of external subs supported by renderer defined in the Supported line in the renderer.conf
```
Nadahar, line 203: This change means it's no longer a JavaDoc, and thus won't show up when you hover over the variable in IDE's etc.

SubJunk: That's true, but I almost removed those comments since the new variable name already describes it IMO

src/main/java/net/pms/configuration/FormatConfiguration.java
```diff
@@ -198,10 +198,10 @@
 		private String mimeType;
 		private String videoCodec;
 		private String supportLine;
-		/** List of embedded subs supported by renderer defined in the Supported line in the renderer.conf */
-		private String embeddedSubs;
-		/** List of external subs supported by renderer defined in the Supported line in the renderer.conf */
-		private String externalSubs;
+		// List of embedded subs supported by renderer defined in the Supported line in the renderer.conf
```
SubJunk, line 201:
```suggestion
```

src/main/java/net/pms/configuration/FormatConfiguration.java
SubJunk, line 203:
```suggestion
```

SubJunk:
@valib the user has reported that this fixed their bug. They have also reported another bug it did not fix, but I would like to merge this and then work on the other one separately, to maintain focus. Please review

valib (approved):

PR #2301, opened 2020-12-02
Fix force external when both types are present

SubJunk:
User confirmed the fix works at https://www.universalmediaserver.com/forum/viewtopic.php?f=9&t=14007&start=30


valib (approved):

valib:
@SubJunk do you know why the `linux-trusty` is not passing the test?

SubJunk:
@valib good point, it was an old setting in GitHub, it's fixed now. Thanks :)

PR #2306, opened 2020-12-06
Added ability to overlay texts under the renderer icon.

SurfaceS:
A small added option on the renderer configuration file that allow underlaying some text on the renderer icon.
It draw the text/s directly on the BufferedImage at icon creation, so no undesired side effect.

This can avoid using alot of base renderer's png image in the jar file.
It allow the renderer icon to be more precise, by adding a year, "UHD", wathever you want directly on the icon. 

The option on the renderer configuration file is **`RendererIconOverlays`**.
The value is formated **`TEXT@X,Y,FONTSIZE`** and multiple entries can be used by separate them with a **`|`**.

Exemple :
`RendererIconOverlays = SAMSUNG@35,65,24|UHD@132,25,17`

DefaultRenderer.conf was updated to show users the option exists and how to use it.

### DIFF
--- a/src/main/external-resources/renderers/DefaultRenderer.conf
+++ b/src/main/external-resources/renderers/DefaultRenderer.conf
@@ -30,6 +30,12 @@ RendererName =
 # Default: unknown.png
 RendererIcon = 
 
+# Determines the text/s that will overlaying the renderer icon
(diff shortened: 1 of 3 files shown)


SubJunk:
@SurfaceS seems cool, can you resolve the conflict please?

SubJunk:
@SurfaceS ?

PR #2302, opened 2020-12-04
Added option for symlinked files

SurfaceS:
Added an option to allow symlinked files to be treat as their real target file.
That avoid multiple entries in the database for the same real file if you symlink file on different folders (genre, actors, etc for exemple).

PR #2308, opened 2020-12-06
Added option for symlinked files

SurfaceS:
Added an option to allow symlinked files to be treat as their real target file.

**Goal :** Avoid multiple entries in the database for the same real file if you symlink files on different folders (genre, actors, etc for exemple).

First started at #2302, which I closed because I created a new branch for it and fixed my IDE tab effect.

https://github.com/UniversalMediaServer/UniversalMediaServer/pull/2302#issuecomment-739085138


SubJunk (approved):

SubJunk:
Thanks!

PR #2309, opened 2020-12-06
Fixed FileWatcher not working for Device Configuration file

fu2x:
**Issue:**
Device Configuration files have the following name: for me, e.g. [TV]UE55JU6450-c02aa3.conf
Changes in the file will be ignored until next restart of UMS, although a Filewatcher entry is added for the file.

**Reason:**
Filewatcher is creating a matcher for the file path. But because of the special chars `[` and `]` the `getPathMatcher()` methods interpretes them as glob commands and transforms them to regex expressions, which is leading to a misformed matcher. 

**Solution:**
I changed the `init() `method of the Filewatcher, where the matcher is created. 
The `getPathMatcher()` methods expects a glob or regex pattern string. The existing coding assumed a valid glob syntax and added the respective syntax prefix to the file path. 
So now, instead of assuming valid input, we are defaulting to glob with the prefix, but also escaping all glob specific special characters that could occur in the file path.

This is my first pull request on github and I am not a "productive" Java developer. So if you have any feedback, improvements, better ideas to solve it or just questions I would gladly hear it. Let me hear/see what you think about this.

### DIFF
--- a/src/main/java/net/pms/util/FileWatcher.java
+++ b/src/main/java/net/pms/util/FileWatcher.java
@@ -99,13 +99,14 @@ public class FileWatcher {
 			String match;
 			String globChar;
 			final String[] globCharsToEscape = {  
-					"[",
-					"]",
-					"{",
-					"}",
-					"?",
-					"*",
-					"," };
+				"[",
+				"]",
+				"{",
+				"}",
+				"?",
+				"*",
+				","
+			};
 
 			if (fspec.startsWith("glob:") || fspec.startsWith("regex:")) {
 				match = fspec; //assume fspec is valid glob or regex pattern, not just a file path
--- a/src/main/java/net/pms/util/FileWatcher.java
+++ b/src/main/java/net/pms/util/FileWatcher.java
@@ -100,13 +100,12 @@ public class FileWatcher {
 			String globChar;
 			final String[] globCharsToEscape = {  
 					"[",
-                    "]",
-                    "{",
-                    "}",
-                    "?",
-                    "*",
-                    "," };
-
+					"]",
+					"{",
+					"}",
+					"?",
+					"*",
+					"," };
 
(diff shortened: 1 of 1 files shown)


src/main/java/net/pms/util/FileWatcher.java
```diff
@@ -96,8 +96,28 @@ public Watch(String fspec, Listener listener, Object item, int flag) {
...
+			String globChar;
+			final String[] globCharsToEscape = {  
+					"[",
+                    "]",
+                    "{",
+                    "}",
+                    "?",
+                    "*",
+                    "," };
+
```
SubJunk, line 109:
```suggestion
					"]",
					"{",
					"}",
					"?",
					"*",
					"," };
```

SubJunk (changes requested):
Looks good, just minor formatting to commit

src/main/java/net/pms/util/FileWatcher.java
```diff
@@ -96,8 +96,27 @@ public Watch(String fspec, Listener listener, Object item, int flag) {
...
+			String match;
+			String globChar;
+			final String[] globCharsToEscape = {  
+					"[",
+					"]",
+					"{",
+					"}",
+					"?",
+					"*",
+					"," };
```
SubJunk, line 108:
```suggestion
			final String[] globCharsToEscape = { "[", "]", "{", "}", "?", "*", "," };
```

SubJunk, line 108:
```suggestion
			final String[] globCharsToEscape = {  
				"[",
				"]",
				"{",
				"}",
				"?",
				"*",
				","
			};
```

SubJunk (approved):

SubJunk:
This looks good, thank you!

SubJunk:
@fu2x this seems to have broken watching of real files for me so I will revert it. Feel free to submit a different implementation

PR #2311, opened 2020-12-08
Solved errors on remote web

SurfaceS:
### Fact :
- `Align` attribute and element are obsolete in HTML4 and not supported in HTML5,  then we should use the `style` attribute instead.
- `Content-Type=text` is been interpreted as xml by jquery ajax, and result in a background error if empty on some browser.
- A character encoding declaration is required (either in the Content-Type metadata or explicitly in the file) even if the encoding is US-ASCII.
- The `title` element is a required child of `head`.
- The `html` element represents the root of an HTML document.
- Cookies with `SameSite=None` without `Secure` attribute will soon be rejected.
- `xmp` attribute is deprecated since HTML3.2 and completely removed from the language in HTML5.
- bootstrap need jquery
- `css sourceMappingURL` should point to a valid url or be disable

### Updated js.cookie.min from 2.1.0 to 2.2.1 :
Version 2.2.1 allow setting cookie's `SameSite` value.
Renamed to `js.cookie.min`.

### On files main\external-resources\web\play.html, flow.html and browse.html :
- Load `js.cookie.min` before `jquery.ums.js`, as it need it.

### On file main\external-resources\web\browse.html :
- Set `document.cookie` with `SameSite=Strict`.

### On file main\external-resources\web\doc.html :
- Added `jquery` as `bootstrap` and `keycontroller` need it.
- Added `<meta charset="utf-8"> `

### File main\external-resources\web\util\log.html :
- Replaced obsolete `xmp` element with `pre` element.
 
### File main\external-resources\web\util\logviewer.html :
- Added `<html>`
- Added `<meta charset="utf-8"> `
- Added `<title>Log viewer</title>`
- Replaced obsolete `align` element with `style` attribute.
- Format tabulation.

### On file main\external-resources\web\util\bootstrap.min.css :
- Comment `sourceMappingURL` as there is no source map file (`bootstrap.min.css.map`).

### On class `net.pms.configuration.WebRender` :
- Modified function `getPushData()` to return a real empty json if `push` is empty.

### On file main\external-resources\web\jquery.ums.js :
- Add a check for `jQuery` scripts loaded
- Add a check for `js.cookie` scripts loaded
- Modified `poll` function to handle json automatically

### On class `net.pms.remote.RemoteWeb` :
- Set response `Set-Cookie SameSite` to `Strict`

### On class `net.pms.remote.RemoteWeb.RemotePollHandler` :
- Set response mime type to `application/json`

### On class `net.pms.network.HTMLConsole` :
- Replaced obsolete `align` attribute with `style` attribute.
  By the way, their was an error in `align=center`, should be `align="center"`
- Added `<meta charset="utf-8">`.


SubJunk (approved):

SubJunk:
Looks good, thanks!

PR #2312, opened 2020-12-09
Changed upnp:class for PlaylistFolder from storageFolder to playlistContainer

ik666:
Give control points the ability to identify a DLNAResource PlaylistFolder as upnp:class of type object.container.playlistContainer.


SubJunk:
@ik666 thanks for this. What is the likely effect of this change on devices?

ik666:
This is a good question. I'm coming from an audio perspective to UMS. Since m3u/pls playlists are not directly streamed/played by a device there should be no difference for them. From a control point of view, which is responsible for "entering" or "displaying" a "playlist folder", a correct reported type, gives the ability to inform the user about this fact, or give the user a better user experience, i.e. by displaying playlist before other storage folders. For users with large audio libraries this is a really nice feature (or at least what I use it for).

SubJunk:
I'm convinced. Thanks!

PR #2313, opened 2020-12-09
Added support for providing control points additional metadata

ik666:
If available within the songs audiotag section, MusicBrainz-ReleaseID… will be included in a DIDL response as descMetadata. With this information a control point can clearly identify a song for additional services.


Nadahar:
Does there exist a control point/renderer that actually utilize `descMetadata`? I seems moot to me to "invent" fields unless they are implemented in some way.

SubJunk:
@ik666 hi, thanks for this! It seems like a good change, and I agree with Nadahar that we would normally only merge code that is being used, rather than for future use

ik666:
This control point (https://github.com/sf666/nextcp2) uses descMetadata to clearly identify songs. It's still under active development and not "finished" yet, but working. It has already the ability to display "star ratings" if songs can be clearly identified. Ratings can be export into a users private MusicBrainz account or be stored as Audiotag within the song file itself. 

Current showcase for this feature:
The CP added support for this media renderer : https://github.com/PeteManchester/MediaPlayer . The renderer (current master) reports musicBrainzID as descMetadata (if MPD is used as a backend and songs have Audiotags set). The CP activates star rating in the "now listening" dialog. 

Showcase for this pull request:
Right now, the song has to be played for being rated. I intend to add star rating support also for content browsing. That's why I need those metadata. 

This meta-info might also be useful for other CP's but the source of initiative should be the CDS. If not, I believe we'll have a "hen's egg problem".

src/main/java/net/pms/dlna/DLNAMediaDatabase.java
```diff
@@ -105,8 +105,9 @@
 	 *       to 1.4.196 because 1.4.197 broke audio metadata being
 	 *       inserted/updated
 	 * - 23: Store aspect ratios as strings again
+	 * - 24: added MusicBrainzID to audio tracks
 	 */
-	private final int latestVersion = 23;
+	private final int latestVersion = 24;
```
SubJunk, line 110: Do you think it would be possible to do this without bumping the database version? I am hesitant to do that since it is quite an expensive operation to happen on people's machines.

Maybe you can use the existing `MusicBrainzReleases` table and add a `AudioTrackID` column?

SubJunk:
Thanks for explaining. In that case I think this is great, I am very happy to support these kinds of additions. I left a comment in the code

src/main/java/net/pms/dlna/DLNAMediaDatabase.java
ik666: Sure, I'll have a look at it ... there seems to be a method for this use case.

ik666:
I have moved the table handling code into a "table" class, which should do the upgrade now, avoiding the "latestVersion" update to 24 within the DLNAMediaDatabase class.

valib:
@sf666 this PR makes errors like 
`org.h2.jdbc.JdbcSQLDataException: Parameter "#20" is not set; SQL statement:
INSERT INTO AUDIOTRACKS (FILEID, ID, LANG, TITLE, NRAUDIOCHANNELS, SAMPLEFREQ, CODECA, BITSPERSAMPLE, ALBUM, ARTIST, ALBUMARTIST, SONGNAME, GENRE, YEAR, TRACK, DELAY, MUXINGMODE, BITRATE, MBID_RECORD, MBID_TRACK) VALUES (?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?) [90012-199]
	at org.h2.message.DbException.getJdbcSQLException(DbException.java:587)
	at org.h2.message.DbException.getJdbcSQLException(DbException.java:427)
	at org.h2.message.DbException.get(DbException.java:205)
	at org.h2.message.DbException.get(DbException.java:181)
	at org.h2.expression.Parameter.checkSet(Parameter.java:94)
	at org.h2.command.Prepared.checkParameters(Prepared.java:180)
	at org.h2.command.CommandContainer.update(CommandContainer.java:132)
	at org.h2.command.Command.executeUpdate(Command.java:267)
	at org.h2.jdbc.JdbcPreparedStatement.executeUpdateInternal(JdbcPreparedStatement.java:200)
	at org.h2.jdbc.JdbcPreparedStatement.executeUpdate(JdbcPreparedStatement.java:154)
	at net.pms.dlna.DLNAMediaDatabase.insertOrUpdateAudioTracks(DLNAMediaDatabase.java:823)
	at net.pms.dlna.DLNAMediaDatabase.insertOrUpdateData(DLNAMediaDatabase.java:1071)
	at net.pms.dlna.DLNAResource.registerExternalSubtitles(DLNAResource.java:3894)
	at net.pms.dlna.RealFile.resolve(RealFile.java:253)
	at net.pms.dlna.DLNAResource.syncResolve(DLNAResource.java:1436)
	at net.pms.dlna.DLNAResource.run(DLNAResource.java:1241)
	at net.pms.dlna.RealFile.isValid(RealFile.java:80)
	at net.pms.dlna.DLNAResource.addChild(DLNAResource.java:516)
	at net.pms.dlna.DLNAResource.addChild(DLNAResource.java:478)
	at net.pms.dlna.Playlist.discoverChildren(Playlist.java:117)
	at net.pms.dlna.DLNAResource.discoverChildren(DLNAResource.java:1332)
	at net.pms.dlna.DLNAResource.discoverWithRenderer(DLNAResource.java:1195)
	at net.pms.dlna.DLNAResource.getDLNAResources(DLNAResource.java:1117)
	at net.pms.dlna.DLNAResource.getDLNAResources(DLNAResource.java:1076)
	at net.pms.remote.RemoteBrowseHandler.mkBrowsePage(RemoteBrowseHandler.java:308)
	at net.pms.remote.RemoteBrowseHandler.handle(RemoteBrowseHandler.java:479)
	at jdk.httpserver/com.sun.net.httpserver.Filter$Chain.doFilter(Filter.java:77)
	at jdk.httpserver/sun.net.httpserver.AuthFilter.doFilter(AuthFilter.java:82)
	at jdk.httpserver/com.sun.net.httpserver.Filter$Chain.doFilter(Filter.java:80)
	at jdk.httpserver/sun.net.httpserver.ServerImpl$Exchange$LinkHandler.handle(ServerImpl.java:692)
	at jdk.httpserver/com.sun.net.httpserver.Filter$Chain.doFilter(Filter.java:77)
	at jdk.httpserver/sun.net.httpserver.ServerImpl$Exchange.run(ServerImpl.java:664)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1130)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:630)
	at java.base/java.lang.Thread.run(Thread.java:832)
ERROR 2020-12-15 23:51:44.866 [pool-8-thread-30] Database error while trying to add parsed information for "E:\Filmy\Bourne Supremacy CD2.avi" to the cache: Parameter "#20" is not set; SQL statement:` 

when the database is deleted and the UMS is making new.

ik666:
OMG, thanks and sorry. I should have found this one myself :) With latest commit this should be fixed now. 

I just wonder, why the tests above fail. The "database" symbol look actually ok.

ik666:
OK, fixed it, maybe a concurrent change.

SubJunk:
@ik666 thanks for the work on this. Is there a reason you didn't go with my suggestion of using the existing musicbrainz table? To me it would make more sense to use the existing table rather than create a new one, but I am happy to be convinced otherwise :)

ik666:
Hi SubJunk,

one thing is, that I've to dig into (for me new) UMS code, which I may still do not understand ( how things are being done), that's true. However, I try to offer the least invasive change for a feature.

If you look closely at the PR, I haven't "invented" a new table, but modified the existing "Audiotrack" table adding two columns (one for the trackID and one for the releaseID). In the current api-3 branch, altering the "Audiotrack" table was done by dropping and recreating it. You proposed not to drop it. Therefore, I have just added "Update" logic to the "Audiotrack" table.

**Where to store the IDs - Background**:
A MusicBrainzRelease is actually some kind of an album, or a CD which has an 1:n relation to MusicBrainzTrackID, which identifies a track on the CD. Therefore we would need at least two tables - the release table and the track table. 

1. Since the "Audiotrack" table is the table which actually contains physical tracks, the MusicBrainz_TrackID should also be stored there. 

2. In relational database systems foreign keys are usually stored on the many side for an 1:n relation. The many side is in this case is the "Audiotrack" table. This means the foreign key to the release table (MBID_RELEASE) should also be stored in the "Audiotrack" table.

**Additional notice**:

I haven't added a referential integrity check to the Music_Brainz_Release table, because it looks SubOptimal (OMG) for me. I do not understand, why the table is being populated with tracks informations, rather than release informations. The "MusicBrainzRelease" table consists of doubled columns from the Audiotrack table (Artist, Album, Title, Year) which makes it somehow useless?!

I personally think the Music_Brainz_Release table (and linked logic) needs some rework for the intended behaviour (which I haven't identified yet).

SubJunk:
@ik666 ok, thank you for explaining. I am happy with your logic. It looks like it has failed our lint check, are you able to fix that and let us know if you want it to be merged?

Nadahar:
The reason why you don't understand the existing tables, is that they aren't supposed to store data about the current media. Instead, they are "cache-tables" that caches the replies from MusicBrainz so that UMS doesn't unnecessarily hammer their API. If you look at the code, you will see that it first looks things up locally and only contacts their API if nothing is cached for the given "query".

ik666:
@SubJunk Sure, I'll integrate, if the cleanup und checkstyle eclipse profiles. I hope they'll do the job :)

@Nadahar OK, thanks for the explanation! I didn't want to offend anyone, I was just on the wrong path, because of the table naming. I personally think it's a good idea to store (or cache) MusicBrainz Track and Release Informations. Since it's not trivial to query the correct release/track info I'm just asking myself, if this will end up in a second sonkong or MusicBrainz Picard implementation. May I ask if you intend to use those information beyond the UPnP entities?

ik666:
@SubJunk Can you tell me, which CheckStyle.xml ruleset is being used by your linter. The CheckStyle.xml file provided at root level seems to deliver different results ...

Nadahar:
> Since it's not trivial to query the correct release/track info I'm just asking myself, if this will end up in a second sonkong or MusicBrainz Picard implementation. May I ask if you intend to use those information beyond the UPnP entities?

@ik666 I'm not sure what you're asking. I wrote that code back in 2016 I believe, and I haven't thought much about it since. I'm pretty sure it's not perfect, but focus has shifted, shifted and shifted again - as usual. I have no further "plans" with it, I just wanted to explain what it did and why it might not seemed to make sense.

SubJunk:
@ik666 it should be using the CheckStyle.xml in the root

ik666:
@SubJunk I'm a little confused. I could fix one file, but on the `DLNAResource.java` one I'm stuck. Maybe you can help.

From what I see checkstyle version 8.29 is used, maven is 3.6.3

I had to add this config properties element to the root pom for the provided config to be read by `mvn checkstyle:checkstyle`:

```
<properties>
    <checkstyle.config.location>CheckStyle.xml</checkstyle.config.location>
</properties>
```

I fix a configuration error,  by moving the `lineLength` module config element within the CheckStyle.xml from the Treewalker section to the top like this :

```
<module name="Checker">
  <module name="LineLength">
    <property name="max" value="140"/>
    <property name="tabWidth" value="4"/>
  </module>
```

After running `mvn checkstyle:checkstyle` the generated `checkstyle-result.xml` file doesn't contain the errors reported by your linter.

Github linter report some "tab indention" errors like this one :

`Error:  D:\a\UniversalMediaServer\UniversalMediaServer\src\main\java\net\pms\dlna\DLNAResource.java:2353: Indent must use tab characters [RegexpSinglelineJava]`

With a hex editor moving to the reported source line looks like this:

checkstyle_error

As far as I see, there are no spaces, only tabs (0x09) within the source. That's most likely the reason why I cannot reproduce the checkstyle result. I see no way how to fix this.

Maybe you can just merge this PR and have a look yourself ?

ik666:
@SubJunk I think I know what happens. Those errors seems to occur after the "automated merge" by git adding spaces instead of tabs. If so, there is no way to fix this on this branch. 

Maybe the github build pipeline has to be configured (if possible) to match the CheckStyle.xml rules set ... or revert the current logic (convert tabs to spaces).

SubJunk:
Ah I see, yeah I think you were running a different command to run checkstyle than I do. Can you also push your fixes for checkstyle?

ik666:
sure ... I'll commit it in this branch.

ik666:
@SubJunk Ups ... you deleted the original api-3 branch ... do you still want to merge it?

SubJunk:
@ik666 I merged `api-3` to master, please submit this against master. Sorry for the trouble

Nadahar:
There's no reason to resubmit anything. Just open the PR before pushing to it, rebase the source branch on master and then retarget the PR. Note that pushing to a closed PR might "break" the recovery.

PR #2314, opened 2020-12-10
Display thumbnails for playlists.

ik666:
Display thumbnails for playlists. The thumbnail image has to be named as the playlist file with replaced extension to "png" or "jpg". If an individual image is not found, a general folder image named "folder.png" or "folder.jpg" is searched for.

So if we have a playlist named "myPlaylist.m3u8" an image placed in the same directory named "myPlaylist.png" will be used as thumbnail, instead of the default one.


SubJunk (approved):

SubJunk:
Nice contribution, thank you!

Issue #2323, opened 2020-12-21
DLNAMediaInfoTest crashes build


anthraxn8b:
Building on Fedora 31a test crashes:

[ERROR] Failures: 
[ERROR]   DLNAMediaInfoTest.testContainerProperties:95 
Expecting:
 <"Size: 1325017, Overall Bitrate: 0, Mime Type: video/mpeg">
to be equal to:
 <"Container: MP4, Size: 1325017, Overall Bitrate: 676979, Video Tracks: 1, Video Codec: h264, Duration: 0:00:15.658, Video Resolution: 640 x 360, Display Aspect Ratio: 16:9, Scan Type: Progressive, Frame Rate: 23.976, Frame Rate Mode: CFR (CFR), Frame Rate Mode Raw: CFR, Matrix Coefficients: BT.601, Reference Frame Count: 4, AVC Level: 3, AVC Profile: high, Audio Tracks: 1 [Audio Codec: AAC-LC, Bitrate: 125547, Channels: 2, Sample Frequency: 48000 Hz], Mime Type: video/mp4">
but was not.

valib:
@anthraxn8b it seems that the MediaInfo library is not properly installed or initialized.

anthraxn8b:
@valib maybe, but IMHO everything that is used within a build should be provided and set up correctly by the build.

SubJunk:
Unfortunately it is tricky on Linux because there are so many different distros to consider. We do include MediaInfo for Windows and macOS, but on Linux we leave that to the users themselves. I just noticed that MediaInfo was missing from our prerequisites on our Build wiki so I have added it https://github.com/UniversalMediaServer/UniversalMediaServer/wiki/How-to-build-UMS

valib:
@SubJunk I think you should add to Wiki at least basic instructions how to install the MediaInfo on Linux.

SubJunk:
I have added a link to their download page which contains links to instructions for many different operating systems - it's their job to document their instructions IMO, and they have already put a lot of work into that

PR #2325, opened 2020-12-21
Add mediainfo check

valib:
This is a reaction on #2323 and can show the wrong MediaInfo instalation.


SubJunk (approved):

SubJunk:
Good idea, thanks. I wonder if we could do an improvement to make it fail with a readable message like "You do not appear to have MediaInfo installed on your machine, please install it before running this test" by wrapping the assertion in a try...catch and rethrowing that error. What do you think?

valib:
@SubJunk done in #2326

Issue #2307, opened 2020-12-06
Failed to send stream to [LG] webOS TV


pablocarle:
Hi,
Very similar to #1877 

UMS detects my LG Web OS TV correctly, when trying to play a video with this renderer it fails with a Http 500 error, I'm interpreting as LG TV replying this.

- WebOS version 05.00.02

Attached to the issue my current UMS.conf and log with startup and attempt at playing video in LG renderer.
192.168.1.10 is the ip address of my UMS headless server running under Arch Linux ARM.
192.168.0.154 is the ip address of my LG TV. They can see each other.

please let me know if you wish me to try something else.

[ums_dbg_2020-12-06-12-17.zip](https://github.com/UniversalMediaServer/UniversalMediaServer/files/5648669/ums_dbg_2020-12-06-12-17.zip)

romario74:
I have the same issue.

UMS 9.8.2, jre 15
Ubuntu 20.10

- LG TV Softwareversion 5.05.70
- webOS TV Version 1.4.0-2532 (afro-ashley)

```
13:09:33 INFO  Started playing Elementary S02_D4-12.m4v on your LGTV
13:09:33 ERROR Failed to send action "SetAVTransportURI" to LGTV: Error: Invalid Args (HTTP response was: 500 Internal Server Error)
13:09:33 ERROR Failed to send action "Play" to LGTV: Error: Transition not available (HTTP response was: 500 Internal Server Error)
```

This works fine on an old Samsung TV but not with this newer LG.
Maybe LG TVs do not support this?

romario74:
UMS 9.8.3 and UMS 10.0.0 a1 
show same behaviour.

SubJunk:
Thanks for creating the issue. It looks to me like that 500 error is coming from the TV, and we are just logging it. I'm not sure if that is ultimately our fault or the TV's. Please let me know exactly which TV you have @romario74 and I will look into it

romario74:
Thanks for your answer. The model is:

**LG 42LB630V-ZA**

I'm here in my vacation home until 6th of january, until I can test. Then around in 3 months again.

valib:
@SubJunk my Panasonic TV also respods with error 500 when I try to push media from the UMS control panel to my TV. I think they did that also from the web interface like is descripted in https://github.com/UniversalMediaServer/UniversalMediaServer/wiki/Pushing-Media-to-a-Renderer.
[...]

PR #2335, opened 2021-01-05
Disable metadata when pushing media by default

SubJunk:
This may fix https://github.com/UniversalMediaServer/UniversalMediaServer/issues/2307

### DIFF
--- a/src/main/external-resources/renderers/DefaultRenderer.conf
+++ b/src/main/external-resources/renderers/DefaultRenderer.conf
@@ -812,10 +812,9 @@ Output3DFormat =
 # Default: true
 UpnpAllow = 
 
-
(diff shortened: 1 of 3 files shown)


valib:
@SubJunk tested and it works.

SubJunk:
Which device did you test with? We may need to change the setting in the renderer configs instead since that user reported it didn’t work

valib:
@SubJunk tested on the Samsung Q9 TV.

SubJunk:
Ok cool. We had a post that it worked on an "old samsung TV" so I guess we shouldn't apply to all Samsung TVs, so I'll put together a PR for that one, and all Panasonics. It can be at least a temporary fix as we might find a better one later

PR #2337, opened 2021-01-07
Fix push didl

valib:
This is for testing


valib:
@SubJunk it works on my Samsung Q9 TV and the Samsung Soundbar MS750

SubJunk:
@valib doesn't work for my Panasonic sorry. I tried with .mpg and .mkv files. So far the only thing that works for my Panasonic is to send no metadata.

valib:
@SubJunk can you try this and provide the log?

SubJunk:
Yes but I'm away from the Panasonic for a couple of days, will after

SubJunk:
@valib here are the logs
[ums_dbg_2021-01-28-10-13.zip](https://github.com/UniversalMediaServer/UniversalMediaServer/files/5883106/ums_dbg_2021-01-28-10-13.zip)

valib:
@SubJunk I see the problem here `D:\a\AC3AlcornTest.mpg` in the `ERROR 2021-01-28 10:11:18.065 [AWT-EventQueue-0] Failed to send action "SetAVTransportURI" to VIERA VT60 Series: Error: Resource not found (HTTP response was: 500 Internal Server Error)` which is the original resource path but it should be something like `http://10.0.0.10:5001/get/50$Temp/E%3A%5CFilmy%5C20170105-WA0005.mp4` which is the response of my UMS. I have no idea why your UMS is sending this.

SubJunk:
@valib there is some related work in an old PR from @skeptical particularly this line https://github.com/UniversalMediaServer/UniversalMediaServer/pull/899/files#diff-4d1450a90dc93ab1a920c1f54df745e91987a102212d931b7f6c3b4afbdec240R788 but he was trying some of the things we have been trying too

SubJunk:
Same thing happens with a fresh database, and we can see that this bug existed a long time ago too in https://github.com/UniversalMediaServer/UniversalMediaServer/pull/899/files#diff-4d1450a90dc93ab1a920c1f54df745e91987a102212d931b7f6c3b4afbdec240R791 so I think we need to fix the wrong URI bug

valib:
@SubJunk I see only one difference in the `Panasonic-VieraET60.conf` which can cause problems. It is the `SendDateMetadata = false`.  Can you try to delete this setting and test it? I made in this PR something similar to the change the @skeptical made in the https://github.com/UniversalMediaServer/UniversalMediaServer/pull/899/files#diff-4d1450a90dc93ab1a920c1f54df745e91987a102212d931b7f6c3b4afbdec240R791 but in the different way.

SubJunk:
I don't have that TV anymore so I don't think this will see any more progress

PR #2339, opened 2021-01-09
Add the network interface updating

valib:
It may fix the issue  #2334


SubJunk:
Closing since it didn't work

PR #2345, opened 2021-01-13
Avoid spamming log

valib:


SubJunk:
@valib we can just remove those lines I think

valib:
@SubJunk I agree

SubJunk (approved):

PR #2347, opened 2021-01-16
Resolves #1891

### LINKED ISSUE #1891: Add option to avoid first-run wizard by just choosing default settings
Users shouldn't be forced to do that

ahmedadan:
Adding functionality to resolve Issue #1891


src/main/resources/i18n/messages.properties
```diff
@@ -682,3 +682,4 @@ Wizard.9=Wired (100 Megabit)
 Wizard.10=Wireless
 Wizard.11=Should UMS hide its advanced options? They can be enabled later
 Wizard.12=Finally, choose a folder to share. You can choose more later on the Navigation/Share Settings tab.\n\nIf you don't choose a folder, we'll choose a default based on your operating system.
+Wizard.13=Would you like to configure UMS with the default configuration?
```
SubJunk, line 685: I think it would be good to reverse the logic of the question so it is like:

```suggestion
Wizard.13=<html><strong>Would you like to run the first-time configuration wizard?</strong><br>This is a quick way to get the best experience from UMS.</html>
```

and the the buttons could use custom labels:
`Yes, run the wizard" and "No, use the defaults"

What do you think?
Do any of the @UniversalMediaServer/developers have input about that?

SubJunk (commented):
Thanks for this, I have made a suggestion and am interested in opinions about it

src/main/java/net/pms/newgui/Wizard.java
```diff
@@ -78,116 +77,137 @@ public static void run(final PmsConfiguration configuration) {
...
-			String.format(status, currentQuestionNumber++),
-			JOptionPane.YES_NO_OPTION,
-			JOptionPane.QUESTION_MESSAGE,
-			null,
-			yesNoOptions,
-			yesNoOptions[0]
-		);
-
-		if (whetherToHideAdvancedOptions == JOptionPane.YES_OPTION) {
+			configuration.setMaximumBitrate("90");
```
valib, line 82: I would recommend to use `automatic maximum bandwidth` instead of this. The default setting anticipate that the connection is wired but the wireles can be used. So checking the maximum bandwidth could avoid problems users should have with sending the transcoded files exceeding the network bandwidt.

ahmedadan:
```suggestion
			configuration.setAutomaticMaximumBitrate(true);
```

src/main/resources/i18n/messages.properties
ahmedadan: I like the presentation you have here, should I implement the change or wait for other input?

valib (dismissed):

ahmedadan:
@SubJunk @valib Anything I need to complete to have the PR merged?

SubJunk:
@ahmedadan it is just waiting on me to test it first unless @valib has tested it?

ahmedadan:
Thanks for that @SubJunk I just wanted to make sure I wasn't holding it up in any way.

src/main/resources/i18n/messages.properties
```diff
@@ -95,6 +95,8 @@ Dialog.Restart=%s needs to be restarted to apply the changes. Do you want to res
 Dialog.Select=Select
 Dialog.Warning=Warning
 Dialog.YES=Yes
+Dialog.DefaultsYes=Yes, run the wizard
```
SubJunk, line 98: These new options should have Wizard keys instead of Dialog keys, as they are very specific to wizard functionality

src/main/resources/i18n/messages.properties
```diff
@@ -95,6 +95,8 @@ Dialog.Restart=%s needs to be restarted to apply the changes. Do you want to res
...
+Dialog.DefaultsNo=No, use the defaults
```
SubJunk, line 99: These new options should have Wizard keys instead of Dialog keys, as they are very specific to wizard functionality

SubJunk (changes requested):
I have tested and it works well, great job! I just left one change request and then it can merge

src/main/resources/i18n/messages.properties
ahmedadan: All done!

SubJunk (approved):

SubJunk:
@ahmedadan thank you!

PR #2357, opened 2021-01-22
LG OLED UpnpDetailsSearch regex fix

tkabala:
Extending LG OLED UpnpDetailsSearch regex so it should match all LG OLED 2016-2021.


SubJunk:
@tkabala this looks good thanks, I have made a version of it with tests at https://github.com/UniversalMediaServer/UniversalMediaServer/pull/2359 since I noticed we were missing tests for LG OLED detection. I'm closing this and will merge that one instead. Thank you again

PR #2360, opened 2021-01-24
Set maximum bitrate for MPEG-2

valib:
This could fix problems when users set the maximal net throughput higher than the MPEG-2 is capable to do. This could fix the problem reported in http://www.universalmediaserver.com/forum/viewtopic.php?f=9&p=43217#p43217 where the FFmpeg is producing the transcoded video with the maxbitrate 19 Mbit/sec.


SubJunk:
@valib I don't think it's true that MPEG-2 is only allowed 10mbps, since that is only a limit for DVDs. On other media it can be much higher, for example this article says 300mbps https://en.wikipedia.org/wiki/H.262/MPEG-2_Part_2
I think it's best that we limit it per renderer when applicable, and per user network. For example when a user chooses the wifi option in the first-time wizard, their network is capped to 30mbps afaicr

valib:
@SubJunk thanks for info

Issue #2338, opened 2021-01-08
GetProtocolInfo not being recieved


Mik-S-UMS:
Going back to 9.7.2 and looking at the logs I noticed that in newer versions UMS is not receiving the GetProtocolInfo data from my TV. Also  UpnpDetailsSearch does not seem to work any more as UMS has gone back to detecting my TV as a Samsung CD
 series instead of using my F series conf.

This is the bit that is missing from newer versions, full log is in issue #2334

```
TRACE 2021-01-07 10:29:00.019 [UPNPHelper] Received a NOTIFY from [192.168.1.64:24234]: NOTIFY * HTTP/1.1
HOST: 239.255.255.250:1900
CACHE-CONTROL: max-age=1800
LOCATION: http://192.168.1.64:7676/smp_6_
NT: upnp:rootdevice
NTS: ssdp:alive
SERVER: SHP, UPnP/1.0, Samsung UPnP SDK/1.0
USN: uuid:0c845880-00d2-1000-ac78-f47b5ebe1def::upnp:rootdevice


TRACE 2021-01-07 10:29:00.020 [UPNPHelper] Resending last discovery [192.168.1.64:24234]
TRACE 2021-01-07 10:29:00.507 [cling-147] Ignoring remote device: RemoteControlReceiver (RemoteDevice) Identity: (RemoteDeviceIdentity) UDN: uuid:0c845880-00d2-1000-ac78-f47b5ebe1def, Descriptor: http://192.168.1.64:7676/smp_6_, Root: true
TRACE 2021-01-07 10:29:00.507 [cling-147] This device was added to the list of ignored devices.
TRACE 2021-01-07 10:29:00.629 [cling-147] Ignoring remote device: dialreceiver (RemoteDevice) Identity: (RemoteDeviceIdentity) UDN: uuid:0d1cef00-00dc-1000-9130-f47b5ebe1def, Descriptor: http://192.168.1.64:7676/smp_24_, Root: true
TRACE 2021-01-07 10:29:00.629 [cling-147] This device was added to the list of ignored devices.
DEBUG 2021-01-07 10:29:01.056 [UPNP-AliveMessageSender] Sending ALIVE...
TRACE 2021-01-07 10:29:04.910 [UPNPHelper] Received a NOTIFY from [192.168.1.64:24234]: NOTIFY * HTTP/1.1
HOST: 239.255.255.250:1900
CACHE-CONTROL: max-age=1800
[...]

PR #2366, opened 2021-01-25
Don't log the GetPositionInfo failure

valib:
and not set the renderer as false in this case. This fix the implementation of the GetPositionInfo which could be wrong in the renderer like the Movian.


SubJunk:
@valib I agree it is spammy at INFO level, but maybe it would be better to change the level to DEBUG or TRACE, since it might be useful for debugging renderer capabilities

valib:
@SubJunk this could be the final version

valib:
@SubJunk in the `UPNPControl `is implemented the `Device` and the `Renderer` whose in fact should be the same. Maybe we could rewrite it to use only one of those.

SubJunk:
How can I test this?

SubJunk (commented):
Why is ps3slim.png in here?

valib:
> Why is ps3slim.png in here?

Because during testing I that the PS3Slim icon is much bigger than others. So I take the opportunity to fix it in one PR.

valib:
@SubJunk you can test by installing Movian on PC or Android phone. The UMS shoud show the error only once.

valib:
Maybe we can try to send the GetPositionInfo two or three times before we set that the renderer is not capable to respond. What do you think?

valib:
@SubJunk what do you think about the several checking if the GetPositionInfo is implemented? I think that it could be done only once and if renderer doesn't respond than it could be marked as not implemented this feature.

SubJunk:
@valib yes that seems like a good idea. Has @Mik-S-UMS tested this fix to see if it fixes https://github.com/UniversalMediaServer/UniversalMediaServer/issues/2338 ?

Mik-S-UMS:
If you can do me a build I'll give it a test but my network setup has changed to a more normal setup, with the router connected to the internet instead of using a second wireless connection for the internet. I have no way of replicating my original setup for that issue.

I should be able to check that Movian doesn't spam the log with errors though, and maybe my Samsung TV too as it creates that error when it's turned off.

SubJunk:
I have 3 variations of that error that I get a lot on my Panasonic TV.

### 1) The error happens after my TV has been turned off, and UMS knows it is off

```
DEBUG 2021-04-06 08:48:24.320 [cling-359] VIERA VT60 Series is now offline.
DEBUG 2021-04-06 08:48:24.320 [cling-361] Subscription cancelled: ConnectionManager on uuid:4D454930-0100-1000-8001-000B977EA155: DEVICE_WAS_REMOVED
DEBUG 2021-04-06 08:48:38.732 [UPNP-AliveMessageSender] Sending ALIVE...
ERROR 2021-04-06 08:48:45.440 [UPNP-VIERA VT60 Series] Failed to send action "GetPositionInfo" to VIERA VT60 Series: Error: Current state of service prevents invoking that action. Connection error or no response received.
```
That is our fault, because a TV can't be playing, recording or transitioning when it is powered off, so we should add logic to not send GetPositionInfo when a device is offline.

### 2) TV is telling UMS that it does not support that feature with "Not implemented":

```
Failed to send action "GetPositionInfo" to VIERA VT60 Series: Error: Current state of service prevents invoking that action. Non-recoverable remote execution failure: 501 Not Implemented. (HTTP response was: 501 Not Implemented)
```
So again that is our fault for not gracefully handling that situation, when the renderer has specifically told us that it does not support what we are asking from it.

### 3) TV is on, but reports that the state of service prevents invoking the action.

Have we considered that the error is correct? For example in the UPnP spec it says:

> a control point can invoke GetPositionInfo every second *when the TransportState is
PLAYING, RECORDING or TRANSITIONING*. This is more efficient and flexible than requiring event
notifications to be sent to all subscribing control points, in all cases.
from http://upnp.org/specs/av/UPnP-av-AVTransport-v1-Service.pdf

So maybe our TransportState isn't correct in this case. Do we have logic to verify that the state is ok before we send GetPositionInfo?

valib:
> So maybe our TransportState isn't correct in this case. Do we have logic to verify that the state is ok before we send GetPositionInfo?

It is here:
https://github.com/UniversalMediaServer/UniversalMediaServer/blob/589692cb390c19e70c3d1fa913185fc45aab5830/src/main/java/net/pms/network/UPNPControl.java#L204

valib:
> 1) The error happens after my TV has been turned off, and UMS knows it is off

I see a way how to fix it. Do not mark the renderer as not active but remove them from the renderers pool and let the UMS recognize them again when it responds to the UMS ALIVE message. This may

SubJunk:
Maybe that logic should be changed to match what the spec says instead of just saying not stopped?

SubJunk:
@valib I pushed a commit to show what I mean, do you think that's a good idea? Feel free to revert if not

SubJunk:
@valib I think cases 1 and 2 from https://github.com/UniversalMediaServer/UniversalMediaServer/pull/2366#issuecomment-816338332 are still problems, do you think we should try to fix them here or in another branch?

valib:
@SubJunk the case 2 is fixed in
https://github.com/UniversalMediaServer/UniversalMediaServer/blob/e809aa533c7e6350aacc33f079672097c3916b43/src/main/java/net/pms/network/UPNPControl.java#L781-L783

The only question if this should be shown like error here
https://github.com/UniversalMediaServer/UniversalMediaServer/blob/e809aa533c7e6350aacc33f079672097c3916b43/src/main/java/net/pms/network/UPNPControl.java#L770-L773

SubJunk:
Maybe we should not treat it as an error if the renderer has responded correctly like that, what do you think?

valib:
@SubJunk why only MacOS (push) is not working? The (pull_request) is OK. There should be somethink wrong in the checks setup.

SubJunk:
@valib there is a PR that might fix it, I ran it 4 times without failure. I think it is just a caching problem

valib:
@SubJunk the commit e19b8d6 should also fix the case 1.

valib:
@SubJunk the MacOS problem is here again. Maybe the Github has a problem with cache when it is trying to download artifacts from the URL `http://0.0.0.0/com/...`  which is nonsense. What about to not use the cache for MacOS?

SubJunk:
Hmm yeah it's weird. I don't think our use of the cache will affect it, it's more like an internal cache at GitHub. I'll try that though.

SubJunk:
Oh I think I see the problem. They have updated to Maven 3.8.1 and that has disabled access to HTTP URLs. Will try to fix.

Maybe the inconsistency is that sometimes we get an instance with Maven 3.6 as they roll out the update

SubJunk (approved):

SubJunk:
I have been testing this with my TV and it seems to work

PR #2369, opened 2021-01-30
Resolves #1663

### LINKED ISSUE #1663: Make the "default" part of tooltips automatic
At the moment we have a convention with tooltips like:

```
<html><strong>Default:</strong> Disabled<br><strong>Notes:</strong> Some notes</html>
```

but instead we should only have to do:

```
Some notes
```

and have a translatable "shell" for the tooltip that only needs to be written once for each language

ahmedadan:
Added `toolTipsMessage()` method, next step is to edit `newgui`


ahmedadan:
@SubJunk Just wondering if I am on the right track here, also would love some clarification on the use in the `newgui` folder

SubJunk:
Yes you're on the right track. Next step would be to make the default value automatic since right now it is hardcoded to false.
It might also make sense to send in a `notesKey` and do the translation within that function too.

Then in the GUI you can change e.g. https://github.com/UniversalMediaServer/UniversalMediaServer/blob/master/src/main/java/net/pms/newgui/NavigationShareTab.java#L572 to `ignorethewordthe.setToolTipText(toolTipMessage("isIgnoreTheWordAandThe", "FoldTab.44"));`

valib:
@ahmedadan I made some chages to fix it. It seems that you not fully undersand what the @SubJunk means. This is the update to show how the tooltip could be implemented.

valib:
@ahmedadan I think that the convention to use the original key in the `messages.properties` as the default value and add the new line with append the `.Note` to the original key could work. If you are willing to help us can you update the rest of this "Default" messages?

ahmedadan:
@valib I will take care of it.

src/main/java/net/pms/newgui/NavigationShareTab.java
```diff
@@ -264,7 +265,7 @@ public void keyReleased(KeyEvent e) {
 
 		// Use MPlayer for video thumbnails
 		mplayerThumb = new JCheckBox(Messages.getString("FoldTab.14"), configuration.isUseMplayerForVideoThumbs());
-		mplayerThumb.setToolTipText(Messages.getString("FoldTab.61"));
+		mplayerThumb.setToolTipText(toolTipMessage("FoldTab.61", "FoldTab.61Note"));
```
valib, line 268: @ahmedadan why do you not use the `Disabled` here as I introduced? Using this can simplify the message.properties to not translate this general setting for all relevant notes.

src/main/java/net/pms/newgui/NavigationShareTab.java
```diff
@@ -383,7 +384,7 @@ public void itemStateChanged(ItemEvent e) {
 
 		// Enable the Media Library
 		cacheenable = new JCheckBox(Messages.getString("NetworkTab.EnableMediaLibrary"), configuration.getUseCache());
-		cacheenable.setToolTipText(Messages.getString("FoldTab.ShowMediaLibraryFolderTooltip"));
+		cacheenable.setToolTipText(toolTipMessage("FoldTab.ShowMediaLibraryFolderTooltip", "FoldTab.ShowMediaLibraryFolderTooltipNote"));
```
valib, line 387: @ahmedadan why do you not use the `Enabled` here as I introduced? Using this can simplify the message.properties to not translate this general setting for all relevant notes again.

src/main/resources/i18n/messages.properties
```diff
@@ -155,23 +155,30 @@ FoldTab.39=Ignore articles "a" and "the" while sorting
...
-FoldTab.47=<html><strong>Default:</strong> Disabled<br><strong>Notes:</strong> If enabled, 3 new virtual folders will be displayed on your device/s to allow you to access your iTunes music:<ul><li>Browse by Artist</li><li>Browse by Album</li><li>Browse by Genre</li></ul></html>
-FoldTab.49=<html><strong>Default:</strong> 10000<br><strong>Notes:</strong> If the number of items in a folder exceeds this number, they will be divided into virtual subfolders by their first letter/number</html>
+FoldTab.44=Enabled
+FoldTab.44Note=If enabled, files will be ordered like this:<ol><li>Apple</li><li>The Banana</li><li>Carrot</li><li>A Date</li></ol>If disabled, files will be ordered like this:<ol><li>Apple</li><li>A Date</li><li>Carrot</li><li>The Banana</li></ol>
+FoldTab.45=Disabled
```
SubJunk, line 160: These defaults should not be in the translations file at all - that is what the story is meant to be about. We should automatically show the default.

SubJunk (changes requested):

valib:
@ahmedadan there are more general settings whose can be translated only once.

src/main/java/net/pms/newgui/NavigationShareTab.java
SubJunk: Sorry for double-reviewing, I hadn't seen yours :)

src/main/resources/i18n/messages.properties
valib: @SubJunk we need to translate `Enabled` and `Disabled`. What do you mean to a make it automatically? Do you mean to read the UMS.conf and look for the default value stored in it?

ahmedadan: My understanding was also that `Enabled` and `Disabled` had to be translated. 

I think there were two expectations of what the story needed. If we could all align before I make anymore changes that would be great.

src/main/java/net/pms/newgui/NavigationShareTab.java
ahmedadan: I see, I can simplify this as well.

valib: @SubJunk I was accidentally faster than you 😃

src/main/resources/i18n/messages.properties
SubJunk: Yes "Enabled" and "Disabled" need to have one entry each in this file, and that can be reused.
So for booleans we detect whether the default value is `true` or `false`, and then display the correct value: if the default is `true`, display the translation for `Enabled`, if it is false, display the translation for `Disabled`.
Does that clarify it?

ahmedadan: Thanks for that @SubJunk Where would I be able to see the boolean value for the default? As an example would I check the `configuration.isUseSymlinksTargetFile()` for the default of `useSymlinksTargetFile`?

valib: The goal is to not repeat the translalation for values and make it only once.

SubJunk: Yes you could start a fresh `configuration` instance so it is all defaults (since the existing one will return the values the user has changed too) and then do what you suggested

SubJunk:
@ahmedadan is this branch abandoned or will you be back to it?

ahmedadan:
@SubJunk I'll be back!

SubJunk:
Closing this for now, feel free to reopen if works resumes

PR #2375, opened 2021-02-02
Audio items transcoding setting

ik666:
This PR adds the ability to natively serve audio files, without transcoding.

- If the renderer (Renderer Configuration file) supports native playback of the streamed audio file, the native file is being streamed instead of the transcoded file.
- If `TranscodeAudio` is set to `NO_TRANS` the native file is always being streamed.

Fixed issues, if native playback is used (not transcoded):
- MPD based UPnP renderer will show/update the played/duration time and progress bar.
- Some renderer extract cover art and other information from the metadata section of the streamed file instead of the UPnP resource information. Transcoded files however lose all metadata.
- Audiophile listeners don't like their files to get touched. The NO_TRANS flag leaves the audio file in any case untouched.

### DIFF
--- a/src/main/external-resources/renderers/DefaultRenderer.conf
+++ b/src/main/external-resources/renderers/DefaultRenderer.conf
@@ -252,7 +252,7 @@ CreateDLNATreeFaster =
 TranscodeVideo = 
 
 # Profile to use for transcoding audio files
-# Options: "LPCM", "MP3" or "WAV".
+# Options: "LPCM", "MP3" or "WAV" or "NO_TRANS". "NO_TRANS" disables audio file transcoding.
 # Default: LPCM
 TranscodeAudio = 
 
--- a/src/main/java/net/pms/configuration/RendererConfiguration.java
+++ b/src/main/java/net/pms/configuration/RendererConfiguration.java
@@ -1346,14 +1346,14 @@ public class RendererConfiguration extends UPNPHelper.Renderer {
 					matchedMimeType = HTTPResource.MPEG_TYPEMIME;
 				}
 			} else if (HTTPResource.AUDIO_TRANSCODE.equals(mimeType)) {
-				if (isTranscodeToWAV()) {
-					matchedMimeType = HTTPResource.AUDIO_WAV_TYPEMIME;
+				if (isMediaTypeSupported(media.getMimeType())) {
+					matchedMimeType = media.getMimeType();
 				} else if (isTranscodeToMP3()) {
 					matchedMimeType = HTTPResource.AUDIO_MP3_TYPEMIME;
 				} else if (isTranscodeToLPCM()) {
 					matchedMimeType = defaultAudioTranscoding(media);
-				} else if (isMediaTypeSupported(media.getMimeType())) {
-					matchedMimeType = media.getMimeType();
+				} else if (isTranscodeToWAV()) {
+					matchedMimeType = HTTPResource.AUDIO_WAV_TYPEMIME;
 				} else {
 					matchedMimeType = defaultAudioTranscoding(media);
 				}
@@ -1823,7 +1823,7 @@ public class RendererConfiguration extends UPNPHelper.Renderer {
 	 * @return The codec name.
 	 */
 	public String getAudioTranscode() {
-		return getString(TRANSCODE_AUDIO, "");
+		return getString(TRANSCODE_AUDIO, LPCM);
 	}
 
 	/**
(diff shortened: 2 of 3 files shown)


SubJunk:
@ik666 I can see the importance of not transcoding audio, but I don't see why this setting is needed. People can already use the `Supported` lines like this https://github.com/UniversalMediaServer/UniversalMediaServer/blob/master/src/main/external-resources/renderers/Yamaha-RXA2050.conf#L33-L40 or `StreamExtensions` like this https://github.com/UniversalMediaServer/UniversalMediaServer/blob/master/src/main/external-resources/renderers/Yamaha-RXV671.conf#L17

Am I missing something?

ik666:
@SubJunk This is actually the issue. In current UMS master the source code doesn't check the `Supported` line (at least for audio playback). It always transcodes the audio file to one of the configured `TranscodeAudio` format: MP3, WAV, WMV, LPCM. Default is LPCM, if `TranscodeAudio` is UNSET. You can check that by entering an audio URL delivered by UMS into your browser. The downloaded content is always LPCM (or one of the configured formats, losing all metadata in the stream). With this PR applied, you get the untranscoded file (FLAC or whatever) if the renderer supports the mime-type, or `TranscodeAudio` is set to NO_TRANS.

I've added the `Supported` line check (see RendererConfiguration around 2949) but it's hard to say what the desired function should be, if the `TranscodeAudio` format ist set to a value. Should it override the `Supported` line and transcode it anyway or not? I've put the `Supported` line check at the end. This means a user has to leave the `TranscodeAudio` UNSET for activating `Supported` line check, otherwise audio will be transcoded to the configured `TranscodeAudio` format. 

For people who don't want to hassle around with mime-types, configuring the correct `Supported` line the `NO_TRANS` option just disables (audio file) transcoding. It's actually not "needed" but a convenience feature.

I haven checked the `StreamExtensions` yet, and I'm not sure if this is necessary?

SubJunk:
It sounds like that is a bug with our supported lines that needs to be fixed, rather than a new setting implemented. IMO the best thing would be to fix the supported lines not being used for audio.

We already have the user option to "skip transcoding for file extensions" in the GUI for users who don't want to edit the renderer configs too

Nadahar:
There's also a setting to skip transcoding altogether somewhere.

I think this is based on a misunderstanding though. "Everybody" wants to have the best quality possible, so the idea has never been to transcode content that can be played as-is. Transcoding only comes in when the renderer can't play the media unless it's transcoded. So, it's not a choice between "good" and "bad" quality, it's between playing the media or not. Every time I read about "no transcoding", "disable transcoding" etc., I'm thinking to myself "here we go again". The only thing you actually achieve by disabling transcoding is that those media files that would otherwise have been transcoded isn't playable.

If UMS transcodes a file when the renderer can play the source media as-is, it's either a misconfiguration or a bug.

SubJunk:
Yes, our goal is to *always* stream all files supported by the renderer. Transcoding should only be done when the device will not play the raw file.

ik666:
@SubJunk , @Nadahar : OK, I agree with your position. So your proposal is to remove the "TranscodeAudio" NO_TRANS option, but leave the rest as is? As I mentioned, the `support line` is interpreted for audio playback in this PR. If the renderer is able to play the file ,no transcoding will happen.

Should I also interpret the "skip transcoding for file extensions" (is this equal to `StreamExtensions` ?), or where do I get it from?

SubJunk:
Yes I think `StreamExtensions` is the same. I'll have a look at your code soon to see what has been done but your description seems like something we would want. It is a bug if the supported lines are ignored.

ik666:
All right. I've removed the NO_TRANS option.

If the renderer has a `Supported` line that matches the streamed resource mime-type, no transcoding will be applied.

I think this should be ok now.

valib:
@ik666 I retested the UMS with my renderer and the supported audio is streamed not transcoded. Which renderer is transcoding all audio for you? Can you provide the log in the TRACE mode for the master branch?

ik666:
@valib you might be right. I've digged through the renderer configuration code and found what went wrong. What worked for me is to set `MediaInfo` Option to `true`. Second, on my system the `mediainfo` native libraries were not found. After I fixed this, the file was streamed instead of transcoded.

Setting `StreamExtensions` (forced streaming) didn't work right away, since this line is split around "," and not ", ", meaning that the "flac" extension was compared against  the " flac" string ... I'll close this PR and give a fix for this in a separate PR

valib:
@ik666 to set `MediaInfo=true` is necessary to get working the `Supported` lines.

ik666:
@valib Yes, I haven't seen this dependency. Thanks a lot for your patience :)

PR #2376, opened 2021-02-02
upnp search (audio)

ik666:
This PR implements a basic UPnP search. The result structure is compatible to https://minimserver.com/ search implementation.

Implemented features (tested for audio files only, yet). Search for objects of type :
- Song
- Album
- Artist
- Playlists
only one type (upnp:class) is supported per search request.

Additional parsed properties:
- upnp:artist
- upnp:genre
- dc:title (this attribute is mapped to a database column, which correlates to the requested object type)


valib:
@ik666 it looks good but how to test it? My renderers don't send the Search request?

ik666:
@valib Good question ;) I think BubbleUPnP has search support, but you'll need an android device.

Alternatively, I've written a control point with search support (optimized for audio) which can be downloaded as binaries here : https://github.com/sf666/nextcp2/releases/tag/v2-20210203. Build from the source code is also available. It runs on a regular server and can be access by a Chromium-Browser (mobile supported).

Nadahar:
@ik666 I'm always looking for more software renderers and control points to use when testing. However, it seems you've made this Java 11 only? Do you have a version that will at least run on Java 8? I don't have Java 11 installed as I still develop with Java 7 support.

ik666:
@Nadahar Sure, I've switched the java compiler level to JDK8 within the pom.xml file and have updated the binaries.

Nadahar:
@ik666 Great, thanks.

Nadahar:
@ik666 It's kind of off topic in this thread, but I see that you're sending `Windows7/6.1 UPnP/1.0 Cling/2.0` as your User-Agent. It would be helpful if you'd use a User-Agent as specified by UPnP, like for example `OS/OS-version UPnP/1.0 nextCP/2.0`.

sf666:
@Nadahar You're right. I'll have a look at it. 

PS: On current master: I've added JDK11+ and JDK8 support into the maven build profile. You can now initiate a JDK 8 build by calling the `build_jdk8.sh` script , or manually be calling `mvn clean package -P jdk8`.

SubJunk:
@ik666 this looks great to me, thanks!

@valib are you testing it?

valib:
@ik666 I tried the BubbleUPnP but have no idea how to invoke the search. I also tried the Cling Workbench vhere is possible to invoke the search but I have no idea how to properly set the SearchCriteria.

valib:
@ @ik666 I tried the BubbleUPnP and it has a lot of errors. See the log.
[debug.log](https://github.com/UniversalMediaServer/UniversalMediaServer/files/5928919/debug.log)
You implemented only audio files search but the renderer is searching for all media with the requested pattern.
Also there is a wrong implementation like https://github.com/UniversalMediaServer/UniversalMediaServer/blob/2e198535c8f6f905c8b548eafa4795dd75a250c8/src/main/java/net/pms/network/SearchRequestHandler.java#L230 which should be `object.container.album.musicAlbum` or the different checking. The same is in the https://github.com/UniversalMediaServer/UniversalMediaServer/blob/2e198535c8f6f905c8b548eafa4795dd75a250c8/src/main/java/net/pms/network/SearchRequestHandler.java#L228 Did you test this PR? Which renderer do you use for testing?

ik666:
@valib Thanks for your comments and your effort to test this PR! Since I don't have an Android device, I couldn't check myself with BubbleUPnP. Maybe I get a hand on it. I think you have identified the issue. I've modified the `upnp:class` checks to identify derived object type as you proposed.

I've just tested this feature with my own control point (and yes, I know this is not a sufficient test :), so again thanks for helping out with your BubbleUPnP logs!

ik666:
@Nadahar The user agent issue: I think I got it and updated a fix on the master branch. I'm not sure, how you've tested it, so you might take a look if you like. If there is still an issue, I'd propose to open a ticket on the nextcp/2 project, for not messing up this task.

valib:
@ik666 there are still errors. See 
[debug.log](https://github.com/UniversalMediaServer/UniversalMediaServer/files/5935622/debug.log)

ik666:
OK, I'll get a hand on an android device test bubbleupnp and will implement the video search functionality, too.

valib:
@ik666 not only video name search. The bubbleupnp is also searching for the `object.item.audioitem` as filename which is also not implemented.

ik666:
Yes sure, I'll make BubbleUPnP work as expected. I got an Android device today ...

ik666:
@valib BubbleUPnP is now working without exceptions and given search results, also video items but here is work to do.

But I think, I've one issue with building the result for albums. Sometimes I've to search twice before I can browse the folder. Is there a caching issue? I'm not sure if adding the search result to virtual folders is the best approach ...  @Nadahar maybe you have a hint for me?

Nadahar:
@ik666 I don't have much here, as I'd have to study it a bit more first. I've never really looked into the UPnP search functionality except that I know that it's something that's missing, and that I seem to recall that it's more like a filter than a search..?

When it comes to "caching issues" the UMS code is "deeply infected" with many layers of this. In short, most of the code have never been written to handle changes at all, but assumes that everything is static and can be cached without any mechanisms to control modifications at all. Some "hacks" have then been made to try to force updates in certain situations, which usually leads to a lot of unnecessary overhead because the approach is often crude by forcing large parts of the code to be rerun.

Except behavior to be inconsistent because the large number of hacks means that the tiniest details can determine actual behavior. That said, UMS builds the "tree" in memory per renderer, and the state of this "tree" is what is returned when UMS received browse requests. Rescanning will probably lead to parts of the tree being overwritten most of the time, which results in updated results on subsequent browse requests. On top of that, you have the database ("media library"). It must also be updated if you intend to rely on that for your results.

After taking a quick look at your search implementation, it seems that you use the database to do the filtering and that you then build a new "tree" of those search results in memory before returning the results. This can be a very slow process, which probably includes disk activity because of the above mentioned chaos of hacks/workarounds. If multiple threads are involved, you might end up with results being ready after the response have been sent to the renderer. Because of disk/OS buffering, doing this a second time might provide faster results, which means that it "works" the second time. The whole process is probably highly volatile/depending on the system, so it might work the first time on some systems and never on others.

I would start by looking into `folder.discoverChildren()` and see if this spawns a new thread somewhere. If it does, you have to find a way to wait for this to finish before proceeding. Thread safety is another huge problem in all of this, most of this is done on the "we hope things happen in the order we imagined" basis.

ik666:
@Nadahar OK, thanks! The threading issue might be interesting to look at. Maybe I can join the running thread and wait until it finishes ...

ik666:
@Nadahar I haven seen any threads open and close, but what I've seen is, that a new ObjectID is given for a "found" file, even on further searches, discovering the same file. This might be because the search container is build dynamically and the ID's are generated relative to the parent container. 

I'll have a look, if I can use static ID's , let's say to use the FILES.ID column as ObjectID ... I think this should come out to the functionality but is more predictable.

Nadahar:
@ik666 My reason for suggesting a thread issue is that when something fails the first time but works the second, there must be some kind of a timing issue. It doesn't have to be threading, but it's the easiest way to imagine how a timing issue can occur IMO. Please note that it's not enough to look for new threads spawning, because thread pools are in use (where threads lie idle waiting for a task, performs the task and then goes back to the pool) to avoid the overhead of constantly spawning threads. I'll have to follow the path of the code, preferably in the debugger, to really find out I'm afraid.

The way the "logic" currently works in UMS, ID's can't be the same. Each "instance" of a resource must have a unique ID, even if the same underlying resource found in a different "folder" (virtual or not). The reason is that the information about how to process a resource (transcode or not, MIME type, subtitles, audio language++) is all stored in a `DLNAResource` instance, and the ID's refer to this particular `DLNAResource` instance, not the underlying media source. When requests appear via HTTP, the ID is needed to look up the exact `DLNAResource` in question to retrieve the correct set of decisions that have already been made.

On top of that, each `DLNAResource` can only have one parent, so even if you built your search results in a much more complicated way and found the `DLNAResource` instances from the "browse tree", you couldn't make the "search result folder" the parent.

There are many problems/flaws in the current logic. In addition, neither UPnP AV nor DLNA is very well thought through in many situations. It's often very clear for me to see that the standards were written by people concerned with making the renderers, not servers, so that every inconvenience is left for the server to solve.

ik666:
I see, I'll think about it a little ... theoretically speaking, it would be sufficient "to create the DLNAResource" just for accessing (streaming) the resource (accessed by the media renderer), not for "discovering / displaying" the resource (control point).

Nadahar:
@ik666 If it only were that simple. Unfortunately it isn't, since the server has to make all the detailed decisions before responding to a request. If you look at the content of the DIDL-Lite, you'll see that it must all have been figured out by the time it is being generated. I agree that the standard would be must more useful if it was possible to merely list the content during browse, and then supply the details at a later stage - but as I said, the standards bears witness of whose concerns have been "the priority". It's easiest to make a renderer if you get all the details up front, and can create buffers, ready decoders and whatnot only using the information received in the browse response, and thus this is what was chosen.

Transcoding doesn't seem to have been much in their thoughts when the made the standards, if your server doesn't transcode, it's very easy to come up with the details up front. There are no decisions to make, you just report the properties of the file, and if the renderer can read it, it works, if not it doesn't. This is how WMP's "server" works for example.

The 3-box model (when using a separate control point as opposed to the renderer and the control point being one and the same) makes all this even more challenging, because the server have no way to know what the renderer can and cannot play when the control point is doing the browsing. As such, you will find many "challenges" when using UMS with this approach.

ik666:
@valib Have you had a chance to have a look at the search implementation. BubbleUPnP should work now.

SubJunk:
@valib this seems to be waiting on you, do you have time to test it? If not, I can jump in

ik666:
@Nadahar Actually the "cache-issue" is/was quite annoying. Obviously, repetitive searches lead to cache misses. With the lastest commit, I've written a "databased resource locator" and integrated it into the search handler and RequestV2 resolver. This means audio (& videos) files, playlists, album and persons can now be dynamically "browsed". No need to cache DLNAResources or add them to a global repository. Seems to work quite well and uses less memory. 

@valib Maybe the video presentation part is not implemented well. Maybe you can give a hint, how is should be presented to the user using BubbleUPnP.

ik666:
@SubJunk Just FYI. macos build seems to fail.

SubJunk:
@ik666 we hopefully fixed that failure now, if you can sync with master

ik666:
@SubJunk Worked .... Thx!

SubJunk:
Is this still waiting on testing or is development still happening? I'm interested to get this progressed if I can

ik666:
From my point of view it is "mergeable" because it does what it should and there are not so many users using this feature, yet, though admitting, this feature probably needs some more testing i.e. with BubbleUPnP.

There are some "handling issues" with Album and Person searches, since both classes cannot be uniquely identified (only by name) leading to collisions if i.e. albums have same names. I think I can improve this by using MusicBrainz Album-ID if the ID is available.

SubJunk:
@ik666 are you going to do that extra testing that is needed? that would seem ideal since you know the feature you want to create

ik666:
@SubJunk I have tested the search feature again on an Android system with BubbleUPnP. It seems to work quite well.

Tested cases:
- local video playback (playback with locally installed VLC on android device)
- video playback on LG television (webos, LG device selected as player). 
- audio: titles & person & album discovery and playback on local device (internal audio player)

Open test case transcoding: 
I believe my home player were able to stream the content and therefore I'm uncertain if any transcoding has happened.

ik666:
@SubJunk I did some more testing. With the latest commit, there is no need to manually force server side search within BubbleUPnP. UMS UPnP "search feature" is being detected and used without further configuration.

valib (approved):

SubJunk:
@ik666 we have some users reporting problems after 10.6.0 which I suspect points to this merge, since we didn't have many other changes, and the logs show an error from the request.

Are you still around and able to look into this please? https://www.universalmediaserver.com/forum/viewtopic.php?f=9&t=14807&p=44337#p44337

SubJunk:
@ik666 it seems the problem is that some renderers like Sony TVs are using search instead of browse when it is offered, so they send a request for `object.container.storageFolder` which isn't implemented by us.

ik666:
Hi SubJunk. I can supply a fix for this. Since the TV want's a "regular browsing" result, I can revert to browse logic in case of any search errors.

SubJunk:
@ik666 thank you. I'll upload a build of your fix for the user to test when you do it

SubJunk:
I posted it for the user to try at https://www.universalmediaserver.com/forum/viewtopic.php?f=9&t=14807&p=44349#p44349

ik666:
ok, I'll also monitor the board for comments from the user.

PR #2379, opened 2021-02-05
Added bitsPerSample to DIDL result

ik666:
Browse requests do not populate `bitsPerSample` attribute in the DIDL result string (at leaset for audio items). This fixes it.


SubJunk (approved):

SubJunk:
Nice one, thanks

Issue #2258, opened 2020-10-09
Sometimes the web interface shows unclickable videos


SubJunk:
Usually I can play all videos on the web interface but it seems there is some race condition that makes it use a different renderer configuration (that's a guess) which makes it think some files can't be played

SubJunk:
I got this again today and if I click on a clickable video, I can see in the logs:

```
15:00:04 ERROR Unexpected error in RemotePlayHandler.handle(): class net.pms.configuration.DeviceConfiguration cannot be cast to class net.pms.configuration.WebRender (net.pms.configuration.DeviceConfiguration and net.pms.configuration.WebRender are in unnamed module of loader 'app')
```

which is likely an important clue

PR #2392, opened 2021-02-10
Fix CI

SubJunk:
A playground to see how to fix our Bionic build

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -45,6 +45,35 @@ jobs:
       - name: Run tests
         run: mvn verify -P testing-windows
 
+  linux-focal:
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        # test against multiple major Java versions
+        java: [ 8, 15 ]
+    steps:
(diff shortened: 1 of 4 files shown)


SubJunk:
Well this is passing now so it was probably a bug on GitHub that has been fixed

SubJunk:
It passed and I closed the PR, then it failed when I opened it

SubJunk:
Same error on Linux 20.04

valib:
@SubJunk as you can see only the `linux-focal` and the `linux-bionic` have a problem to make a new RealFile even if I used the real file not the fake one.

SubJunk:
Ok so we have a "fix" now, the configuration was coming through `null`, but I will spend more time on it to figure out why that is the case, as I suspect it may be causing other bugs in the program, like maybe (hopefully) this https://github.com/UniversalMediaServer/UniversalMediaServer/issues/2258

SubJunk:
@valib I think this is an acceptable fix - the PMS instance wasn't initiated properly by the test

valib (approved):

PR #2394, opened 2021-02-11
Implemented simple external API

ik666:
This PR implements a small API stub for external programs to access UMS functionalities. 

Motivation: Sometimes it might be necessary to tick an UMS functionality by an external application. 

my current case: If the media library content changes, the new content is not discovered until you browse to the folder containing the new content. This means the "search implementation" I'm working on will not find new content unless you browse to it, or restart the server (which makes a full scan). Therefore, it would be nice, if an external program is able to initiate a new scan.

This PR add's a new API path at "/api" to control some UMS functionality. Right now the "rescan" path is interpreted.

How to use? Just make HTTP GET request to the external api like this :

`curl http://localhost:5001/api/rescan`

What do you think?


Nadahar:
This isn't up to me, but think this is a bad idea for several reasons.

First of all, UMS already has the possibility to trigger a scan by calling `/console/scan` (not web inferface port, but UPnP HTTP server port - 5001 by default).

Second, this is a terrible idea because a scan can be very slow for large libraries, and users will get very frustrated of why UMS constantly starts chewing through the whole library.

Third, there is already defined ways to handle this in the UPnP AV spec, although this isn't supported by UMS. `ContentDirectory` is supposed to inform subscribing control points/renderers of any changes using the `LastChange` mechanism (as far as I remember, if not, it's via "normal UPnP eventing"). So, your control point could merely describe to `ContentDirectory` and UMS would send events as they happened. That would mean that this had to be implemented in UMS of course, but the benefit of doing it this way would be that any other compliant renderer would be informed as well.

UMS has a `FileWatcher` service, that although a bit buggy, either is supposed to or could be used to monitor shared folders. Per now I suspect that it's only monitoring the folders configured as `monitored`, but regardless. The correct process would be to use events from `FileWatcher` to trigger UPnP events and update renderers with the information. The road there is long though, so be prepared for a lot of work if you want to make it. Still, it's the only sensible solution that I can see that would actually be usable without causing massive disk scans.

SubJunk:
Thanks for your thoughts @Nadahar that is useful.

I wouldn't say it's a bad idea but I think we need to be very careful to be safe especially with the recent major exploits in Plex, we need to make sure we don't open ourselves up to those kinds of gaps, given that we are looking to add the ability to access the server over the Internet in the future.

I do like this concept though and I can see it being useful for doing actions via the web interface too if we can make it secure. Maybe it should be behind some authentication mechanism so the user has to do a one-time approval for it to work?

Nadahar:
I'd like to add, if it wasn't obvious, that I wasn't commenting on a HTTP API in general, but the issue of updating/rescanning resources.

SubJunk:
I think if the initial startup scan has been run, which is the default, subsequent scans barely touch the hard drive; if I do a rescan after that has completed, it is finished in only a few seconds, while the initial scan takes minutes. I think it doesn't re-parse files which is what takes a long time, it only checks the modified date of the file, then moves on unless that modified date has changed since we parsed.

Nadahar:
@SubJunk I don't think that's the reason. The initial scan should only check the modified date as well (unless the files in the shared folders have actually changed). I think the reason you see this is buffering, potentially both in your drive(s) and OS. The queries for the content of the shared folders are probably cached somewhere for a while, which makes it fast. That won't be true for everybody though, and it depends on so many details both regarding hardware and operating system that you can't rely on this behavior.

SubJunk:
You might be right, I'll have a look in the code so I can gain a better understanding

ik666:
@Nadahar Thanks for your comments, you're right. With a `curl http://localhost:5001/console/scan` you get the same result. However, your other thoughts would actually also apply to using the web console (`/console/scan`), don't they ?!

I personally think it's up to the user to initiate a full scan, and if he initiates a full scan, he'll be aware of any inconveniences. Right now I restart the server. This is even more resource consuming. Generally speaking, I agree rescanning the whole library is kind of dumb, if you already know which resource exactly changed. Therefore, i added a second API call `rescanFileOrFolder` which does a partial tree scan, if the given path is served by the server library. 

Initiate a POST request to `/api/rescanFileOrFolder` with the file or folder to rescan as body content. From command line with curl :

`curl -d "/Volumes/music/Pop/Rammstein/" -X POST http://10.0.9.139:5001/api/rescanFileOrFolder`

Fast, easy effective even for large libraries. What do you think about that?

Your UPnP idea: Even though the UPnP spec says the server can post an event in case of any content changes using `ContainerUpdateIDs`, that's not the intention of this PR. The intention is just to recognize a change and to update UMS internal library.

My other ideas:
The second intention of this PR is to have an api (I think on that point you're with me), that goes beyond the limited UPnP spec. I'd, like i.e. to implement a search function for a musicbrainz.org track-id to see, if a song is already in the database, or retrieve an Object-ID for a given musicbrainz.org track-id. This is useful for playing playlists hosted at https://listenbrainz.org/.
I think there are more use cases I'll come up with :)

@SubJunk Security: I think adding an API key is a good start. I'll have a look at it.

ik666:
@SubJunk I've added a simple api-key authorization to the API. 

Each API call must supply an api-key in the request header. The key must match the configured api_key value in UMS.conf.

Usage:

Set in `UMS.conf` : `api_key=123456789012`

do the call

`curl -d "/Volumes/music/Ava Max/" -w "%{http_code}\n" -H "api-key: 123456789012" -X POST http://localhost:5001/api/rescanFileOrFolder`

you get in case of a success "204" HTTP code.
401 for an invalid api-key
503 if the api_key is unavailable or misconfigured.

Nadahar:
@ik666 
> However, your other thoughts would actually also apply to using the web console (/console/scan), don't they ?!

Yes. I meant to say that triggering a rescan is a bad idea regardless of which method is used to achieve it, while at the same time pointing out that this exact functionality already was there.

> I personally think it's up to the user to initiate a full scan, and if he initiates a full scan, he'll be aware of any inconveniences. Right now I restart the server.

Sure, the user must be free to initiate a scan as he/she wishes. This wasn't how I perceived the suggestion though, I thought you were thinking of doing this automatically either on a timer or when certain events, like a search, took place.  If this is only meant as a way to let the user initialize a manual scan I don't really see a problem with it, but I'm not sure I think it's very logical/intuitive either. Your control point would get very tied in to UMS specifically, and I'd see it as a mixing of roles as well. It might be just as good if the user simply had UMS' "web console" open in another tab and initiated scan that way. The advantage of this is that it would also indicate to the user when the scan was actually finished (I think the text changes from "start scan" to "stop scan" while the scan is in progress). Either way, I don't have any "strong views" on this, I only strongly suggest to avoid "automatic" rescans.

> Fast, easy effective even for large libraries. What do you think about that?

A partial scan is obviously better than a full scan, that said, I have a network drive with a lot of media files where simply enumerating the files takes a long time, I haven't measured it, but it often feels like 30 seconds or so. From my point of view, using OS native file watching and just rescanning exactly what changes, automatically and immediately is a vastly better solution. That said, I'm not sure how well that works on "foreign file systems" like network drives. The really bad thing about Java's "file watching" implementation is that it automatically falls back to polling if the underlying system doesn't support change notification. That could end up being very "expensive" on slow network drives. I am however just speculating now, to know for sure, these things would have to be studied in detail.

Regarding the UPnP route, I again thought you intended to make changes automatically reflect, in which case I think it would make a lot more sense to implemented it in a way that other control points and renderers could benefit from as well. This becomes moot when it's about manually triggering a scan, there's no mechanism for that in UPnP that I'm aware of anyway, as it's supposed to be "live".

Regarding the general API part of it, I see that it might be useful and have no problem with that except that you're tying your control point very much to a specific server, which might reduce its usefulness to users.

SubJunk:
Thank you, I will test that auth soon, sorry for the delay. I have been busy working on the API server architecture since it was overloading the existing web server

ik666:
ok. I see you sometimes writing about this: can you tell, what the aim of the server API architecture is ?

SubJunk:
Prior to UMS v10, we would do lookups for video information on the OpenSubtitles API, for things like TV episode titles. However their service is very popular and as a result it is often unavailable and rate-limited. So in v10 we added our own API as a cache for those lookups, so that each lookup only needs to succeed one time for all users instead of once per user. However we get a lot of lookups over 17 million requests per day) so our existing web server struggled to handle all of that traffic and it was making other things slow, like our website.

To fix it I have set up a different web server to handle only the API requests, so it has no effect on the website. So far it seems to have worked well.

ik666:
Ah, ok. This makes sense. Maybe I can be of any help ... at the moment I'm trying to get some audio stuff running :)

SubJunk:
Cool, you are doing important work already :) there is never a shortage of important things to do on UMS haha

SubJunk:
@ik666 this looks good to me and I think is ready to merge.
We have also been working on a new support website to improve our documentation, I wonder if you would be interested in making a page on there for this feature, what do you think?
https://support.universalmediaserver.com

src/main/java/net/pms/network/api/ApiHandler.java
```diff
@@ -0,0 +1,136 @@
...
+
+		String serverApiKey = PMS.getConfiguration().getApiKey();
+		if (serverApiKey.length() < 12) {
+			LOGGER.warn("Weak server API key configured. UMS.conf api_key should have at least 12 digests.");
+			output.setStatus(HttpResponseStatus.SERVICE_UNAVAILABLE);
+			return;
+		}
+
+		String givenApiKey = extractApiKey(event);
+		if (serverApiKey.equals(givenApiKey)) {
```
js-kyle, line 52: If this API handler is used for more important things in the future, there could be a benefit of making this API authentication more robust - thought it might also be overkill for UMS. Using the string comparison can be used for timing attacks especially since this is a cleartext comparison (String.equals(String) returns as soon as a single character doesn't match so it can be used to brute force passwords)

Could maybe do something like 

```
public static boolean arePasswordsEqual(String a, String b) {
    boolean equal = true;
    if ( a.length != b.length ) {
       equal = false;
    }

    for ( int i = 0; i < MAX_PASSWORD_LENGTH; i++ ) {
        if ( a.charAt(i%a.length()) != b.charAt(i%b.length()) ) {
            equal = false;
        }
    }
    return equal;
 }
```

Nadahar: Might as well do it right the first time, instead of having to remember to "upgrade" such logic later should the need arise.

I have a couple of questions to your approach though:

- Why would you want to do the initial length comparison? That will allow a time attack to easily determine the password length.
- Why the strange constructs with `MAX_PASSWORD_LENGTH` and `i%x.length()`? Isn't it easier to just do `i < Math.min(a.length, b.length)` in the loop definition and then do the length comparison as the very last step before returning?

js-kyle: I'm sure the brains here can come up with a much better implementation - I was more keen to point out the timing attack possibility with `String.equals(`. Your suggestions sound good

ik666:
@SubJunk Sure. I would add a section for this feature on the support site. How can I contribute?

src/main/java/net/pms/network/api/ApiHandler.java
ik666: Hi @js-kyle . Thanks for pointing out this issue. I've updated the code to prevent timing attacks by hashing the "given" and "server" api keys and always comparing the complete byte array.

SubJunk (approved):

SubJunk:
@ik666 if you let me know which email address you'd like me to use I can invite you to join the support site at https://support.universalmediaserver.com/

ik666:
ok, please use this email : ikrahne[at]hotmail.com

SubJunk:
@ik666 hopefully you just received an invite

ik666:
@SubJunk unfortunally not :| Nothing in spam folder either ... Maybe you can write me an test mail.

SubJunk:
@ik666 looks like it's a bug with that server config not sending emails. To workaround it until it's fixed, I set the password on your account to your username twice. Please change it ASAP

ik666:
This worked. Thanks.

SubJunk:
Great. I've enabled editor permissions for your account

PR #2396, opened 2021-02-17
Added renderer support

ik666:
I've added support for 
- https://github.com/PeteManchester/MediaPlayer (renderer)
- https://github.com/sf666/nextcp2 (control point)


src/main/external-resources/renderers/mediaplayer.conf
```diff
@@ -0,0 +1,41 @@
...
+# See DefaultRenderer.conf for descriptions of all possible renderer options
+# and UMS.conf for program options.
+
+RendererName = MediaPlayer
+
+UpnpDetailsSearch = Made in Manchester , Open Home Java Renderer: v0.0.9.8
+
+UserAgentSearch: Music Player Daemon
+
+StreamExtensions = wav, flac, mp3, wma, aac
```
valib, line 12: @ik666 why there is that line when the `MediaInfo=true` and the same extensions are defined in the "Supported"? It is overcome when the MediaInfo is used.

ik666: @valib Thanks for the hint. I'm not sure if I understand the way UMS uses the configuration, since some options seem to do the same stuff for me. First I had some audio native streaming issues. I finally came up with this configuration, but please feel free to adjust it, as you think it should be correct. I'll then test it here.

I actually asked myself, why the supported protocol types are not read from the `SinkProtocolInfo` of the ConnectionManagerService. Or is this what `MediaInfo = true` does?

valib: > I actually asked myself, why the supported protocol types are not read from the `SinkProtocolInfo` of the ConnectionManagerService. Or is this what `MediaInfo = true` does?

Yes this is the shortage of the UMS and should be updated to read the ProtocolInfo.

Nadahar: @ik666 I think @valib sometimes forgets that not everybody is familiar with the history of the project.

In the beginning, there was no MediaInfo. Parsing was very rudimentary, and transcoding decisions were made based on file extension only. When MediaInfo was introduced, a more detailed way to specify format/media support was needed since UMS now had much more detailed information available about the media.

To avoid breaking the already existing, working configurations, `MediaInfo=` was introduced as a switch for whether the "new configuration rules" should apply. To this day, many configurations exists that use the "old" system without "supported" lines, only being matched on file extension. It's very "coarse" and makes a lot of wrong decisions, but it seems to work well enough for some renderers. If the "new system" was applied to all, it would break a lot of the existing "working" configurations.

When it comes to the `GetProtocolInfo` information, I started working on that at one stage. You will find that the information is actually parsed into "profiles" by UMS, but the information is never used. The reason it stopped where it did is complex, other problems got priority, interest among other developers seems very low, and there are some real challenges in how to "translate" the information gained into detailed format/codec information that UMS can use to make decisions. `GetProtocolInfo` often provides too little information to be properly useful, and the "quality" of the information given by renderers varies a lot.

On top of that, early DLNA devices (DLNA 1.0) often didn't even implement `GetProtocolInfo`, including the PS3 which is what was the original target for this project. So, back when many of these decisions were made, it might not even have made sense to try to use it.

It might see easy and straight-forward for audio media, but it isn't that easy and straight-forward for video media in many situations, which has always been the "main" focus of the project.

ik666: @Nadahar 👍  Thanks for this information. This makes my understand of UMS config much better ... And I find it somehow cool, that that you already parsed the `ProtocolInfo`. I think at least audio renderer (stand alone commercial products like a music box) provide correct data here. Maybe I work on this for the audio part. I've already noticed that video streaming seems to be the main focus, maybe I might help to "adjust" the focus a little ;)

Nadahar: @ik666 
> maybe I might help to "adjust" the focus a little ;)

I tried back in the day, but time convinced me that it was never going to happen. The problem is that while there is audio (and image) users, video is the concern of the vast majority of the users. Good luck ;)

SubJunk: That would be great to see that area get more focus @ik666 :) that is pretty high on my list of desired features in UMS

ik666: Thy, I'll do my best.

PR #2399, opened 2021-02-18
Set automatic maximum bandwidth enabled as default.

valib:


valib:
I would like to try to recognize wired an wireless connections to set the automatic transcoding setting acordingly to each renderer. @SubJunk do you think that the actual network speed (bandwidth) can be the threshold? E.g. less than 30 Mbit/s for wireless and higher for wired?

src/main/resources/i18n/messages.properties
```diff
@@ -210,6 +210,7 @@ GeneralTab.9=Run the configuration wizard (needs application restart)
 GeneralTab.10=Only run a single instance of UMS
 GeneralTab.11=<html><strong>UMS must be run as administrator for single instance to work.</strong></html>
 GeneralTab.12=Use automatic maximum bandwidth
+GeneralTab.12.Tooltip=<html><strong>Default:</strong> Enabled<br><strong>Notes:</strong> It sets the optimal bandwidth not exceeded the actual limit of each renderer capability.<br>When disabled the maximum bandwidth can be set manually but generally for all renderers (wired and wireless).</html>
```
SubJunk, line 213:
```suggestion
GeneralTab.12.Tooltip=<html><strong>Default:</strong> Enabled<br><strong>Notes:</strong> It sets the optimal bandwidth for each renderer.<br>When disabled the maximum bandwidth can be set manually, but it applies to all renderers (wired and wireless).</html> 
```

SubJunk (changes requested):
This looks good, just a minor language suggestion

SubJunk:
@valib Yeah I don't feel good about `HalveBitrate` either. I think it is better to send a lower bitrate than higher in general, because having slightly lower quality is a better experience for the user than having choppy playback, so I would be worried about removing that setting.
I think it would be good to not have it in new configs though, and I generally remove that setting when I develop configs with users.

Some TVs have very specific limits on bitrates, the Samsungs in particular are very picky.

valib:
@SubJunk I think that we can remove all the HalveBitrate implementation when this PR will be approved. The last commit I made is only workaround for testing that the HalveBitrate implementaion is wrong.

src/main/java/net/pms/encoders/MEncoderVideo.java
```diff
@@ -1302,22 +1302,24 @@ public ProcessWrapper launchTranscode(
...
-						// It has been reported that non-PS3 renderers prefer keyint 5 but prefer it for PS3 because it lowers the average bitrate
-						if (params.getMediaRenderer().isPS3()) {
-							mpeg2Options = "keyint=25:vqscale=1:vqmin=2:vqmax=3";
-						}
+					// when the automatic bandwidth is used than use the proper automatic MPEG2 setting
+					if (configuration.isAutomaticMaximumBitrate()) {
+						mpeg2Options = params.getMediaRenderer().getAutomaticVideoQuality();
+					}
 
-						if (mpeg2Options.contains("Wireless") || maximumBitrate < 70) {
```
valib, line 1314: This was in fact the bug. It never checks the `Wireless` because the `mpeg2Options` was set to `"keyint=5:vqscale=1:vqmin=2:vqmax=3"` at the line 1307

Nadahar: These limits are old and honestly quite arbitrary. I don't think there ever was a WiFi standard that was 30 Mbps. The biggest "flaw" in the logic IMO is forgetting that the bandwidth is shared between all devices on the network, and forgetting that bandwidth often varies a lot over time on wireless connections. Since bandwidth will vary, it's impossible to actually achieve the "average" on a constant basis. The peaks will be above, and the bottoms will be below, so if you try to send something, you must make sure that you don't surpass the lowest available bandwidth you will experience. Receiver buffering will help to smooth this out, but we don't control how much renderers buffer so we can't really count on that.

So, the actual bandwidth used must be quite a bit lower than what you "measure" if you want to achieve non-interrupted playback. Guessing the delta between the minimum and the average bandwidth is no easy feat, and my guess is that the failure to handle/understand these things is what led to the whole "half bitrate" thing in the first place.

SubJunk: I agree that it is outdated now since modern Wi-Fi can handle very high bitrates. At the time, there was a clear speed difference between Ethernet and Wi-Fi, but in everyday terms now with good equipment there doesn't need to be.

So "wireless vs wired" is a concept we might not need anymore - we should probably just make sure our automatic speed measuring good. However the one difference between wired and wireless is that wired is far more stable, as wireless can drop low temporarily for a number of reasons (maybe a new phone joined, or someone's computer starts doing updates in the background, etc.).

The best tools constantly alter bitrate because of this inconsistency, by having the server offer several different bitrates and the client can choose the right one based on how well it is keeping up with buffering, but we are really limited by DLNA in this regard. We can be smarter with our web interface but probably the best thing to do via DLNA is to measure the network speed directly before attempting to play a video, instead of only doing it at startup.

Just some quick thoughts.

src/main/java/net/pms/configuration/RendererConfiguration.java
```diff
@@ -1829,23 +1829,29 @@ public boolean isDefaultVBVSize() {
...
 	 */
-	// TODO this should return an integer and the units should be bits-per-second
-	public String getMaxVideoBitrate() {
+	public int getMaxVideoBitrate() {
 		if (PMS.getConfiguration().isAutomaticMaximumBitrate()) {
 			try {
-				return calculatedSpeed();
+				int calculatedSpeed = calculatedSpeed();
+				if (calculatedSpeed >= 70) { // this should be a wired connection
+					setAutomaticVideoQuality("Automatic (Wired)");
```
SubJunk, line 1841: If we are changing this setting automatically now, we should also remove it from the first-run wizard, as it would be a waste of the user's time to set it

SubJunk (changes requested):
The code looks good otherwise and I will start testing it

src/main/java/net/pms/configuration/RendererConfiguration.java
SubJunk: Also maybe we should disable that input box with a note that says something like "Video quality is handled automatically when automatic maximum bandwidth is enabled"

valib:
@SubJunk I implemented what you requested. Can you review it?

src/main/java/net/pms/newgui/Wizard.java
```diff
@@ -108,6 +108,7 @@ public static void run(final PmsConfiguration configuration) {
 				configuration.setMinimized(false);
 			}
 
+/*
```
SubJunk, line 111: Can we remove this commented chunk instead of commenting? We can always retrieve it from the git history if we need it again

SubJunk (changes requested):
Nice one, I will test soon too but wanted to leave this comment while I saw it

SubJunk (changes requested):
Language tweaks

src/main/java/net/pms/newgui/Wizard.java
valib: No problem

SubJunk:
@valib thanks for making the changes. I'm doing testing on it today and will merge if that goes well.

SubJunk:
It works well for me after testing. I had another idea too. It would be cool if existing users got a message when they open UMS to offer once to change the setting, if they have disabled it. Something like:
"We have improved our automatic video quality setting, would you like to enable it?"

What do you think @valib ?

valib:
@SubJunk good idea but how to make it?

SubJunk:
@valib good question :) it would be a new concept for us, but maybe it can be a pattern that we follow in the future when we change defaults. I can try making it if you want?

valib:
@SubJunk I am working on it and will introduce it soon.

valib:
@SubJunk I simply hardcoded it but it could be made more sophisticated to be used whenever we need it. Do you have any idea how to do it?

src/main/resources/i18n/messages.properties
```diff
@@ -212,13 +212,16 @@ GeneralTab.9=Run the configuration wizard (needs application restart)
...
+GeneralTab.12.Tooltip=<html><strong>Default:</strong> Enabled<br><strong>Notes:</strong> It sets the optimal bandwidth for each renderer.<br>When disabled the maximum bandwidth can be set manually, but it applies to all renderers (wired and wireless).</html> 
 GeneralTab.13=Select/deselect all renderers
 GeneralTab.14=Language:
 GeneralTab.ForceDefaultRenderer=Force default renderer
 GeneralTab.ForceDefaultRendererTooltip=<html><strong>Default:</strong> Disabled<br><strong>Notes:</strong> Disables automatic detection</html>
 GeneralTab.StartWithWindows=Start with Windows
 Generic.Folder=Folder
 GenericIcons.DVDVideo=DVD Video
+InfoMessage = We have improved our automatic video quality setting, would you like to enable it?
+InfoMessage.Title = The UMS information for users
```
SubJunk, line 224:
```suggestion
ImprovedFeatureOptIn.AutomaticVideoQuality = We have improved our automatic video quality setting, would you like to enable it?
ImprovedFeatureOptIn.Title = Improved Feature
```

src/main/java/net/pms/PMS.java
```diff
@@ -475,6 +475,34 @@ private boolean init() throws Exception {
...
+			configuration.setShowInfoAboutVideoAutomaticSetting(false);
+			Object[] yesNoOptions = {
+					Messages.getString("Dialog.YES"),
+					Messages.getString("Dialog.NO")
+			};
+			// Ask if user wants to use automatic maximum bitrate
+			int whetherToUseAutomaticMaximumBitrate = JOptionPane.showOptionDialog(
+				null,
+				Messages.getString("InfoMessage"),
+				Messages.getString("InfoMessage.Title"),
```
SubJunk, line 491:
```suggestion
				Messages.getString("ImprovedFeatureOptIn.AutomaticVideoQuality"),
				Messages.getString("ImprovedFeatureOptIn.Title"),
```

SubJunk (commented):
Nice one, just language tweaks to consider

SubJunk:
@valib I think it's ok for now to hardcode it. What you've done looks good. If we keep doing this kind of thing it could be refactored in the future but we have more important things to do IMO

PR #2417, opened 2021-03-09
Improved support for x264 via MEncoder

SubJunk:
This makes MEncoder use x264 directly instead of via lavc, which runs faster for me. Needs testing though.


SubJunk:
I have tested this update with a few files including ones with subtitles. I haven't noticed any problems and audio is still in sync so I think it's good to merge

valib (approved):

PR #2434, opened 2021-03-20
Fix TsMuxeR HEVC implementation

### LINKED ISSUE #2433: tsmuxer fails with HEVC
Ubuntu 20.10 / 10.3.0:

I have a HEVC encoded video, mediainfo shows HEVC (Main@L3.1@Main) video with HE-AAC audio. Nevertheless, when transconding with tsmuxer the UMS will create a .meta file that contains "V_MPEG4/ISO/AVC", which is IMHO wrong. Correct would be "V_MPEGH/ISO/HEVC". The not-so-surprising result is that the H264 code of tsmuxer will fail to accept the transconding:

[tsMuxeR-6-2] tsMuxeR: ../tsMuxer/h264StreamReader.cpp:486: virtual int H264StreamReader::getTSDescriptor(uint8_t*, bool, bool): Assertion `0' failed.

[tsmuxer.zip](https://github.com/UniversalMediaServer/UniversalMediaServer/files/6175142/tsmuxer.zip)

valib:
Fix #2433


src/main/java/net/pms/encoders/TsMuxeRVideo.java
```diff
@@ -159,8 +160,13 @@ public ProcessWrapper launchTranscode(
 		}
 
 		String videoType = "V_MPEG4/ISO/AVC";
-		if (media.getCodecV() != null && media.getCodecV().startsWith("mpeg2")) {
-			videoType = "V_MPEG-2";
+		String codecV = media.getCodecV();
+		if (codecV != null && configuration.isUseMediaInfo()) {
```
SubJunk, line 164: Why do we need `isUseMediaInfo` here? That seems unrelated to the logic

valib: Sorry. Wrong assumption.

SubJunk (approved):

Issue #2435, opened 2021-03-21
ffmpeg transcoded files are no longer playing as of 10.3.0


mbentley:
## Problem
I am running UMS headless inside of a Docker container.  With what I assume is the new capability of "Improved automatic maximum bandwidth setting and enabled it by default", it looks like this is causing videos that were previously rendering fine to fail transcoding with `ffmpeg`.  In taking a look at the differences, I found that it worked as of `10.2.0` but stopped working in `10.3.0`.  My best guess is that this problem came from #2399.

I'm not sure if it is just because I am running headless or not but it is passing in invalid argument to `ffmpeg`.  In digging into the change a bit, it seems like there must be something in the handling of `Automatic (Wireless)` that isn't working in [src/main/java/net/pms/encoders/FFMpegVideo.java](https://github.com/UniversalMediaServer/UniversalMediaServer/pull/2399/files#diff-ba7a7eef129f4ceba66a325e08707372dc35544a5db48afc9277c28fe16c1dbb) + some settings I have set in my config (see below for my config) that are causing the value to be set in the parameter to not be returned properly.

## Workaround:
Set `automatic_maximum_bitrate = false` in `UMS.conf`

## Problem Details
I was digging through the logs and found where it is complaining about an invalid argument in `10.3.0`:

```
DEBUG 11:58:53.612 [ffmpeg-12-2] ffmpeg version N-56320-ge937457b7b-static https://johnvansickle.com/ffmpeg/  Copyright (c) 2000-2021 the FFmpeg developers
DEBUG 11:58:53.612 [ffmpeg-12-2]   built with gcc 8 (Debian 8.3.0-6)
[...]

PR #2436, opened 2021-03-21
Fix issue #2435

valib:


src/main/java/net/pms/network/SpeedStats.java
```diff
@@ -150,17 +150,24 @@ private Integer doCall() throws Exception {
 				}
 			}
 			double speedInMbits1 = bps / (cnt * 1000000);
-			LOGGER.info("Renderer {} has an estimated network speed of {} Mb/s", rendererName, speedInMbits1);
-			int speedInMbits = (int) speedInMbits1;
-			if (speedInMbits1 < 1.0) {
+			int speedInMbits;
+			if (speedInMbits1 == Double.NaN) {
```
SubJunk, line 154: Instead of catching it here when the `NaN` has occurred, which means that `doPing` has likely run 5 times, it might be better to move the check higher up.

Looking at https://github.com/UniversalMediaServer/UniversalMediaServer/blob/master/src/main/java/net/pms/network/SpeedStats.java#L199 maybe you could remove that `try...catch` and instead allow that method to throw the error, and then we catch it here?

What do you think?

src/main/java/net/pms/network/SpeedStats.java
```diff
@@ -150,17 +150,24 @@ private Integer doCall() throws Exception {
...
-			if (speedInMbits1 < 1.0) {
-				speedInMbits = -1;
-			}
-			synchronized (speedStats) {
-				CompletedFuture<Integer> result = new CompletedFuture<>(speedInMbits);
-				// update the statistics with a computed future value
-				speedStats.put(ip, result);
-				speedStats.put(hostname, result);
+			int speedInMbits;
+			if (speedInMbits1 == Double.NaN) {
```
Nadahar, line 154: This never yields true

valib:
@SubJunk, @Nadahar I made completely different fix. Can you evaluate it?

valib:
Maybe some delay could be implemented before the `pw.isSuccess` is used to do some time to wrapper to finish?

valib:
This need more invertigation

Nadahar:
@valib I haven't looked that deeply at it, I just noticed that you did `== Double.NaN`, which is *never* true because nothing, including `NaN` itself is equal to `NaN` per definition. That's why you have `isNaN()` for testing for `NaN`.

SubJunk:
@valib are you still working on this issue or do you want me to try?

valib:
@SubJunk I have no idea how to fix it.

PR #2442, opened 2021-03-23
Fix bug in automatic video setting.

valib:


SubJunk (approved):

SubJunk:
@valib which bug does this fix? Is it the one in this comment? https://github.com/UniversalMediaServer/UniversalMediaServer/issues/2435#issuecomment-803666980
How can I test it?

valib:
@SubJunk this fix the to set the quality because I forgot to set it for below 720+ content. Simply try to transcode the video with width less than 720p.

SubJunk (approved):

valib:
@SubJunk what about to resing on the quality settings and let the FFmpeg do the best only restricted by the maximum bitrate? Automatic or user set.

SubJunk:
@valib if that is supported by FFmpeg then we can try it but it would need a lot of testing for video quality. All those years ago I spent a long time tweaking things for quality.

valib:
@SubJunk I know but this should simplify the UMS code. And the FFmpeg changed since that time.

SubJunk:
@valib I think it's a good thing to try. I will be surprised if FFmpeg changed that much. They are very active but not usually in that way.

valib:
@SubJunk tested and the actual approach is better than the one I suggested.

PR #2443, opened 2021-03-23
Fix locks

### LINKED ISSUE #2374: Raise limit on queue of API lookups
At the moment we use ThreadPoolExecutor to queue API requests, but there seems to be some upper limit on the queue since if I do a full library scan, it stops adding them after a time. Memory use at that point is not noticeable so I think the limit can be increased.

I'm not really familiar with it so help would be appreciated.

SubJunk:
Fixes #2374

### DIFF
--- a/src/main/java/net/pms/dlna/DLNAMediaDatabase.java
+++ b/src/main/java/net/pms/dlna/DLNAMediaDatabase.java
@@ -536,6 +536,7 @@ public class DLNAMediaDatabase implements Runnable {
 	 */
 	public DLNAMediaInfo getData(String name, long modified) throws IOException, SQLException {
 		DLNAMediaInfo media = null;
+		ArrayList<String> externalFileReferencesToRemove = new ArrayList();
 		try (Connection conn = getConnection()) {
 			TABLE_LOCK.readLock().lock();
 			try (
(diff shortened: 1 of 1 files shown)


SubJunk:
This also fixes the issue reported at https://www.universalmediaserver.com/forum/viewtopic.php?f=9&t=14577&p=43533#p43533

valib (approved):

Nadahar:
@SubJunk I'm not sure what the problem is, but you can't just move some things outside locks. That breaks the entire system, and you might as well just remove all locks and go completely unsynchronized.

You need to think about the order in which things are done to "make it work" with the locks. A writelock is exclusive, so it "counts" as read lock as well. You should automatically be granted a readlock from the thread that holds the writelock (for the same lock), so calling code that does readlocks inside it writelock should be ok. What you can't do is "upgrade" from a readlock to a writelock. That will result in a deadlock, as the readlock is shared while the writelock is exclusive. The readlock will never be released so that the writelock can never be be acquired, Therefore, if you need a writelock at some stage during an operation, you should acquire the writelock from the beginning.

I don't know if the above is the problem, but it's usually the only "trap" you can get yourself into unless there are multiple locks in play. With multiple locks one thread can acquire lock a and try to get lock b while another thread acquires lock b and then tries to get lock a. Neither will ever success and you have the "classical" deadlock.

SubJunk:
@Nadahar with this solution, we still aren't doing any database actions outside of locks

PR #2457, opened 2021-04-10
Detect more filename formats for episodes

SubJunk:


valib:
@SubJunk do you know why MacOS checks fail?

SubJunk:
I think it started happening after a dependency update. Possibly a caching problem. Annoying

valib (approved):

PR #2464, opened 2021-04-19
Use jupnp

valib:


valib:
Partially working. The DeviceProtocolInfo is not working.

valib:
@SubJunk any help is appreciated.

valib:
It seems to me that adding `RegistryListener` in https://github.com/UniversalMediaServer/UniversalMediaServer/blob/6928849063dbf3ca4f6a3a4b335da649c18518e7/src/main/java/net/pms/network/UPNPControl.java#L404 doesn't work and devices are managed by internal procedures out of our implementation.

valib:
@SubJunk I think that the main problem are conflicts between the Jetty and our UPNPHelper. Both are creating multicast socket at the same interface and are listening the incoming messages. Who will respond first is a question. Why do we send `Alive` and `ByeBye` from UPNPHelper when it is done automaticaly by JUPNP?

SubJunk:
@valib that seems like good investigation. I think with the alive/byebye part, we had to do that manually because Cling only does it once instead of at intervals. Maybe JUPnP is different?

valib:
@SubJunk I tried to use Jetty with Cling as recommended in http://4thline.org/projects/cling/core/manual/cling-core-manual.xhtml#section.ConfiguringTransports but it also has problems with our implementation. See [debug.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/6354358/debug.txt). So using Jetty needs more investigation and refactoring.

valib:
@SubJunk it seems that Jupnp is using the Jetty as default. Maybe changing it back to Java built-in webserver could fix it.

SubJunk:
@valib sure let's try it, if you feel like working on that :)

SubJunk:
It would be a nice way to move on from that old version of Netty too since our upgrades to it haven't gone well

valib:
@SubJunk sorry but I am not capable to completely refactor the `RequestV2` and `RequestHandlerV2` to use Jetty. Change the UMS `HTTPServer` implementation is quite simple. See https://www.eclipse.org/jetty/documentation/jetty-10/programming-guide/index.html#pg-server-http but to make the handler is other thing.

SubJunk:
@valib ok, thanks for the update. no problem

valib:
@SubJunk it seems that they change Cling quite a lot and simple changing to the Jupnp is not working. There must be some significant changes which must be implemented.

SubJunk:
Closing this for now just to clear up the PR list. Reopen if work continues

PR #2472, opened 2021-04-27
Fix LibmediaInfoParser for JNA 5.7+

valib:
I localized the problem. The MediaInfo `Count_Get(StreamKind, -1)` for JNA 5.7+ returns always 0 but the `Get(StreamKind, 0, "StreamCount")` returns correct number or empty string.


SubJunk (approved):
Nice work!

Issue #2466, opened 2021-04-24
Multiple Errrors Hitting Log - Duplicated or Same Issue Scannning Target Folders


marksjc:
[ums_dbg_2021-04-24-16-42.zip](https://github.com/UniversalMediaServer/UniversalMediaServer/files/6370935/ums_dbg_2021-04-24-16-42.zip)

I see multiple issues upon runnihg 14.1 for first time.  I'm a dumb user so I would appreciate a simple yes/we're workihg oh it or no/it's not a problem.  I think I provide attached the correct data as specified.  Of course if you need more ihforamatioh simply say what you need and I'll forward it along.  If it's likely noh-programmers would not know how to access/collect please provide explicit instructions (for collecting the data, only, to keep focused on this possible issue :-)  _Long live emoticohs, were emoji found in Cheop's pyramid?_

SubJunk:
Why was this closed? Is it working now?

marksjc:
I closed this accidentally and it's still broken  I deleted the old access path to the files (out of hundreds, 5 were actually displaying in the UI).  Got the same 5 and a lot of log errors consistent with the log file attached.  Sorry I closed it, I use Github infrequently, but will be more careful in the future.

valib:
@marksjc please try to delete the database at `/Users/mac/Library/Application Support/UMS/database` and start the UMS again. It seems like the wrong database version.

valib:
Fixed in  #2482

valib:
@marksjc thanks for the report.

PR #2482, opened 2021-05-02
Fix updating AUDIOTRACKS Table version

valib:
This fix the issue #2466

### DIFF
--- a/src/main/java/net/pms/database/TableAudiotracks.java
+++ b/src/main/java/net/pms/database/TableAudiotracks.java
@@ -32,7 +32,7 @@ public class TableAudiotracks extends Tables {
 	 * definition. Table upgrade SQL must also be added to
 	 * {@link #upgradeTable(Connection, int)}
 	 */
-	private static final int TABLE_VERSION = 1;
+	private static final int TABLE_VERSION = 2;
 
 	/**
 	 * Checks and creates or upgrades the table as needed.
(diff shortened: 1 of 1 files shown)


valib:
@SubJunk I have one question? I find out that the table `AUDIOTRACKS` has an `ID` column which is always `0` fo each entry. Is it used somewhere and do we need them?

valib:
This code in the master is nonsens https://github.com/UniversalMediaServer/UniversalMediaServer/blob/34978d9071747fddd675cfe0ec3cfdaf9329ef03/src/main/java/net/pms/database/TableAudiotracks.java#L70-L81
Seems that somebody tried to update the table for the version with `MBID_RECORD` but not finish it.

valib:
@SubJunk can you check it and fix it?

SubJunk:
@valib I think the idea in https://github.com/UniversalMediaServer/UniversalMediaServer/pull/2341/files#diff-25449232e5cc977f814fb9fd3fc79a6710ed9b5f5a3792128fa64a4e41b1d338L378 was that there was no previous version of the table, so if version is null, then we set it to `v1` and then perform those updates in the `upgradeTable` method https://github.com/UniversalMediaServer/UniversalMediaServer/pull/2341/files#diff-c2f72cd835e967741e30cece3df9e580abed68f987b0cf27cc58086a5d6806d0R69

Maybe that logic didn't work properly

valib:
@SubJunk the logic seems OK but how do you explain that in the log of the issue #2466 the table version is `1` but column `MBID_RECORD` is missing and I am sure that `MBID_TRACK` too. From this perspective this PR can be closed and we can only recommend users when they have such problem to delete the database and recreate new one because database upgrading failed.

valib:
@SubJunk this could warn users that something went wrong.

SubJunk:
@valib yeah I have seen that error before too, so something has gone wrong.

Maybe in this branch we can bump the table version, and for all existing tables, check if those columns exist, and if not we create them.

valib:
@SubJunk this Gihub checks are annoying. Because they are not able to receive dependencies they failed.

SubJunk:
Ah yeah that's the same thing as last time, where they updated Maven which doesn't allow HTTP anymore. I'll fix it

src/main/java/net/pms/database/TableAudiotracks.java
```diff
@@ -55,34 +56,41 @@ protected static void checkTable(final Connection connection) throws SQLExceptio
 				} else {
 					// Moving sql from DLNAMediaDatabase to this class.
 					upgradeTable(connection, null);
-					setTableVersion(connection, TABLE_NAME, TABLE_VERSION);
```
SubJunk, line 58: Why remove this?

SubJunk (changes requested):

SubJunk:
@valib this looks good to me now, can you check my changes before I merge?

SubJunk:
Just failed on lint, otherwise we can merge

valib:
@SubJunk now it is ready to be merged.

valib:
@SubJunk this is the last change before merging.

PR #2485, opened 2021-05-04
Use JRE 8u292

SubJunk:
This is to fix Cling bugs as it doesn't fully support the JRE we are using. When we replace Cling we can go back up


SubJunk:
@valib I have been running this at home for a few days and it works well

valib (approved):

PR #2500, opened 2021-05-24
Fix wrong logic in the XXE implementation

valib:


SubJunk:
@valib can you please explain why this is good? also it might be good to add tests for it if you want

valib:
The default value of `configuration.disableExternalEntities()` is `true` see https://github.com/UniversalMediaServer/UniversalMediaServer/blob/e0accbe51df159295dbb5186faa0cdf138bdb53b/src/main/java/net/pms/configuration/PmsConfiguration.java#L4990-L4997
so the code was wrong from the begining when I introduced it. I have no idea what was the reason I made it in that way.

SubJunk (approved):

SubJunk:
@valib I'm not familiar with where that code is used, can you explain the effects of this change so I can explain it in the changelog? Or add it to the changelog file directly

valib:
@SubJunk The main effect is that we better maintain the web security vulnerability explained e.g. here https://portswigger.net/web-security/xxe

Nadahar:
As I've mentioned before when we've discussed this, UPnP doesn't seem to "work" without `DTD`s according to my testing. So, you'd better hope that this doesn't impact any of the UPnP parsing. It probably doesn't though, since Cling use its own XML parsers that is still vulnerable, as also explained before.

The only way that I can think of to actually protect against XXE is to dig so deep into UPnP that one understand exactly what goes wrong and why when disabling `DTD`s, and then find a way to allow whatever functionality that is required while preventing XXE attacks. I'm not sure it's possible though.

valib:
@Nadahar I know that we discussed it before. But this is about our own implementation not the Cling itself. It could be still vurnelable by Cling but not for us.

Nadahar:
@valib I know that, but as I said, it breaks parsing of UPnP XML documents - so you should make sure that this any parsers created using this factory doesn't parse UPnP documents. It's fine for other uses, but it seems that UPnP is "based on" DTDs, which has later shown itself to be "dangerous".

I think the real problem is that you can't tell the parsers to simply ignore DTDs (don't look them up), which I think would work fine with UPnP. Instead, what this does is that the parsers reject documents with DTDs (or external references), which means rejecting XML documents. I think the reason for this is that UPnP decided to use DTDs as a "namespace", so the URL's aren't expected to actually be used. Thus, UPnP documents will have these "links" in them that really are just name space indicators, but they result in "secured" parsers rejecting the documents. I'm not sure about these details though, and I'm not sure how to solve it either, as you would need to find a parser that you could make "ignore" DTDs instead of rejecting them.

PR #2507, opened 2021-05-28
Sony XBR OLED

SubJunk:
Confirmed working on forum


Mik-S-UMS:
I think you should remove the "truehd" that I added to the conf as it turns out he was using ARC to pass the audio to his receiver that supported it and I don't think his TV did, I couldn't see it mentioned anywhere in the specs.

SubJunk:
Thanks, done

valib (approved):

PR #2518, opened 2021-06-08
Match more anime 2

SubJunk:


valib (approved):

valib:
@SubJunk this is not about this PR but there is an issue https://www.universalmediaserver.com/forum/viewtopic.php?f=9&t=14744&p=43924#p43924 when not correct video name like `Assassins.Creed.2016` fails. I tested it and the `FileUtilTest.testGetFileNameMetadata` for this name returns `null` and `FileUtilTest.testGetFileNameWithRewriting` returns `UMS 2016`. Using `Assassins.Creed.(2016)` fix it.

SubJunk:
Yeah we should definitely fix it. Users shouldn't need to use filename conventions to avoid our bugs. I'll look into it

valib:
@SubJunk one comment. The `2016` is correctly parsed as the year and added to media metadata but there is a problem in parsing the name.

SubJunk:
Ok cool. Thanks. I am focusing on this area lately so it's great to have the user feedback. For me my library gets about 90% matches and I'll try to keep improving until it's 100%

valib:
@SubJunk to test it you can add to `prettified_filenames_metadata.json`
```
},
		{
		"comment": "Video of a movie with a year at the end of title",
		"filename": "UMS.2018.mp4",
		"prettified": "UMS (2018)",
		"metadata": {
			"type": "movie",
			"title": "UMS",
			"released": { "year": 2018 }
		}
	}
```

PR #2521, opened 2021-06-10
Match more movies and anime episodes

SubJunk:
Hopefully fixes the problem reported at https://www.universalmediaserver.com/forum/viewtopic.php?f=9&t=14744


valib (approved):

valib:
@SubJunk there are still problems reported in https://www.universalmediaserver.com/forum/viewtopic.php?f=9&t=14744&start=10 which seems to be problems in API.

Issue #2533, opened 2021-06-22
Extract internal subtitles during initial scan


SubJunk:
This should allow us to workaround FFmpeg taking a long time to parse subtitles before it starts transcoding.

It could also mean we get to remove the code that makes FFmpeg defer to MEncoder for those subtitles, which would also fix bugs related to MEncoder like the audio getting out of sync.

Mik-S-UMS:
Could that also be used for streaming the subtitles for transcoded videos? similar to what I suggested in #2123

valib:
The subs extraction is implemented in https://github.com/UniversalMediaServer/UniversalMediaServer/blob/72521029a629b2cbc69e6ccdfb02a2927f6b92b7/src/main/java/net/pms/util/SubtitleUtils.java#L201-L220 or https://github.com/UniversalMediaServer/UniversalMediaServer/blob/72521029a629b2cbc69e6ccdfb02a2927f6b92b7/src/main/java/net/pms/util/SubtitleUtils.java#L377-L394
Subs must be extracted only after the media is fully parsed.

SubJunk:
@Mik-S-UMS yes it could be a step closer to that

@valib cool thanks for the tips, I will work on this soon

valib:
@SubJunk I am not sure if this will speed up the the FFmpeg before playing media. The main problem is the PC power. If it is low than the FFmpeg is producing the video stream slowly and the renderer is buffering the data before it start to play. For example on my notebook I use for testing it can take 10 seconds before the renderer starts to play transcoded media. The time for FFmpeg parsing external subtitles is about 100 ms which not a problem.

SubJunk:
My understanding is that for some internal subtitles, FFmpeg needs to parse the whole file before it starts burning them, so it creates a delay that is sometimes more than 1 minute while it parses. When a file is parsed once, it will start immediately on a the next play. It has been an issue in FFmpeg for about 8 years and it is on their bugtracker but no one has worked on it yet. 
That's why we have the "defer to MEncoder" setting in UMS to workaround this, because MEncoder can burn subtitles without parsing the whole file

SubJunk:
I've got this mostly finished but one thing I noticed is that UMS is clearing the "subs" folder when it closes, do you know where that is being done @valib ?

valib:
> do you know where that is being done

[...]

Issue #1721, opened 2018-12-15
FFmpeg timeout on burning subtitles


SubJunk:
On a lot of my files, FFmpeg times out or takes 2+ minutes before outputting anything (in its initial parsing of subtitles
My TV doesn't wait for 2+ minutes so it fails playback.

There are a few different bugs at play here but my goal is to successfully transcode the files with FFmpeg, and I'm working on that

MadokaAyukawa:
Is this why the following files don't play (you only need on episode). Both shows give an error with FFmpeg on my PS3, but will play by manually selecting MEncoder.

https://nyaa.si/view/1062197

https://nyaa.si/view/1103308

Sami32:
Yes.
https://trac.ffmpeg.org/ticket/4499
FFmpeg doesn't handle natively the subtitles from the libavfilter library, so that is why this delay and why it fail following the UPnP/DLNA standard.
A cheap trick could be done but with a limited effect, and that will not remove the lowering of video quality that you can see when FFmpeg transcode video with subtitles other than ASS/SSA du to the FFmpeg filter trick used for the subtitles sync.
https://trac.ffmpeg.org/ticket/2235

Its why such media are automatically transfered to MEncoder, for Windows users at least.
Of course, when selecting the video engine from the #-TRANSCODE-# folder it doesn't do such automatic engine triggering.

MadokaAyukawa:
Currently, even if "Defer to MEncoder" checked, UMS does not use MEncoder for second set of files (I no longer have the first set handy to test). In fact, I haven't seen UMS defer to MEncoder in quite some time now.

Sami32:
Only media with ASS or Vobsub subtitles are deferred to MEncoder if the deferring FFmpeg option has not been disabled.

That is said, I don't use any UMS version after 6.7.4 so i cannot say if something have been broken sine then.

SubJunk:
MEncoder should kick in with that file, I just tested and it does for me. Can you post logs?

SubJunk:
> Currently, even if "Defer to MEncoder" checked, UMS does not use MEncoder for second set of files (I no longer have the first set handy to test). In fact, I haven't seen UMS defer to MEncoder in quite some time now.

@MadokaAyukawa MEncoder should kick in with that file, I just tested and it does for me. Can you post logs?

MadokaAyukawa:
[...]

PR #2537, opened 2021-06-25
Extract subs for FFmpeg

SubJunk:
Closes https://github.com/UniversalMediaServer/UniversalMediaServer/issues/2533
Fixes https://github.com/UniversalMediaServer/UniversalMediaServer/issues/1123
Fixes https://github.com/UniversalMediaServer/UniversalMediaServer/issues/1721


SubJunk:
I tested it last night and it seems to work - videos that timeout with internal subtitles burning via ffmpeg in master work in this branch.
One thing it still needs is to add the cleanup step so that subtitles for removed files get deleted

SubJunk:
One note for testing is that FFmpeg caches the internal subtitles after the first run, so it can be annoying to reproduce - the bug this fixes only happens once per file. So if you do any testing on it you will likely need to find two similar files (I use two episodes of the same anime), verify the bug on one episode in master, then verify it is fixed on the other episode in this branch.

SubJunk:
There also seems to be a memory leak in this branch, which I think likely exposes an existing memory leak. I'll look into it

valib:
@SubJunk why we use linux-focal for testing? Linux is only 2,38% of the OS used by users. It seems that in the linux-focal they made some changes that destroyed the Java implementation.

SubJunk:
I'm not sure why it's failing. It seems like some stuck cache problem like usual, because adding that logging cleared it in that branch. But then the problem came back.
Maybe it's not important to test two versions of Linux though.

valib:
@SubJunk agree that testing Linux only in the most importand implementation could solve this problem.

SubJunk:
An update on the memory leak is that I can't reproduce it but I think it is unrelated to this - there was a stuck transcoding process that was taking up memory, and then I was also transcoding another thing, and extracting subtitles. So the real bug is the stuck transcoding process which I have seen in other branches. We could also try increasing the maximum memory limit on supported Windows systems, given that it is at 1280 now while my web browser is using like 5GB+

valib:
@SubJunk for me this implementation is not acceptable. Please try to delete the database, disable scanning folders and start browsing without starting scanning folders. For me it is extremly slow and works only in the web interface where there is no timeout like in renderers. In the web interface some media has no thumbnails because of timing.

SubJunk:
Yeah I did the comparison and you're right, this is slower. I guess even when FFmpeg runs non-blocking, it still makes the CPU go slower for other things.

Maybe we can do some manual locking where this new extraction thread is not allowed to run while a browse request is running

SubJunk:
I will try to make a lighter implementation of the extraction that only extracts the original format instead of converting to ASS, and will test again

Nadahar:
I'm pretty sure that CPU isn't the bottleneck here - IO is. I recently tested extracting subtitles from a file that is on a network share via a WiFi (n) connection, and while I didn't measure the time, it was in the range of 20-30 minutes.

I don't know how FFmpeg does it internally, but I suspect that in many containers, the "subtitles chunks" are spread throughout the file (to supply them at an appropriate time for the decoder) so that the whole file actually has to be scanned. If the location of the subtitles chunks aren't "indexed" in some way, there's no way to jump to their location. Also, not all types of I/O allows "jumping"/skipping, so the file has to be read sequentially.

The file I did the test on was a 6.5GB MKV with embedded SRT. Doing it on a local disk or even better, SSD, would obviously make it much quicker, but it could still be enough to pretty much saturate I/O for a little while. This will make all other "competing" I/O very slow. It's not possible as far as I know to give I/O "lower priority" like you can with CPU time.

SubJunk:
Yeah you're likely right. There could still be some manual locking solution I guess, where we make our known heavy IO methods use the lock. I'll try

SubJunk:
I did that locking approach and it seems to work. Can you test @valib ?

One note for testing is that for me, hard drive (or something outside of UMS) caching seems to play a big role: if I let some subtitles get extracted between tests, then this branch is slower than master, even though the extraction is started after browsing has completed. However this test where we delete the database each time is only useful when we assume this is the user's first time browsing that folder, because the next times will be cached in our database on either branch, and those subsequent browse requests take 1 second on both branches for me no matter what.
So for testing, to rule out hard drive caching, it is important to close UMS as soon as the browse request completes; before the extraction process starts.
I hope that makes sense.

SubJunk:
Also CI is failing because our SVN server is down, I'm working on it

SubJunk:
SVN is fixed

Nadahar:
@SubJunk If you want to test without the caching interfering, simply create a network share on another computer and put the media there. I don't think there is any caching on network drives.

SubJunk:
@Nadahar that's a good idea. I have tested it anyway and I am satisfied that browsing is the same speed as it is in master. So now just waiting for @valib to comment

SubJunk:
Actually I think there is another enhancement I'll do before this is ready - I will restore the previous functionality for when the subtitles are not extracted for a file that is being played yet. Because in that case, this branch will be worse for a user currently. I will comment here when that change is done

SubJunk:
@valib this is ready for re-review
I recommend using the "Hide whitespace changes" functionality in GitHub because most of the line changes are just indentation

image

valib:
@SubJunk I am now on holiday. I will test it when I return home.

SubJunk:
@valib that's great, enjoy your holiday!

valib:
Balaton. Hungary

SubJunk:
@valib fixed this blocking the shutdown process

valib:
@SubJunk can you check if I merged the master correctly? Than this PR could be merged.

SubJunk:
This is a good idea but needed a better implementation, possibly a separate subtitle extraction process in the future. It was too flaky this way

PR #2539, opened 2021-06-26
Remove filename from entries inside the TRANSCODE folder

SubJunk:


src/main/java/net/pms/dlna/DLNAResource.java
```diff
@@ -1604,6 +1604,10 @@ public String getDisplayName(RendererConfiguration mediaRenderer) {
 		return getDisplayName(mediaRenderer, true);
 	}
 
+	public boolean isInsideTranscodeFolder() {
+		return parent != null && parent instanceof FileTranscodeVirtualFolder;
```
Nadahar, line 1608: `parent instanceof FileTranscodeVirtualFolder` also checks that it's non-null. ´null´ is never the instance of anything.

SubJunk: Thanks, I removed that check

SubJunk:
@valib this is passing now

SubJunk:
@valib can this be merged?

valib (approved):

PR #2547, opened 2021-07-05
Fix bump skin to work with jquery 3.5.1

Pikarobbi:
**Problem:**
In the web view renderers can no longer be selected. `bump`s media control buttons appear, but not the renderer selection. (tested with UMS 10.6.0 and Vivaldi Browser 4.0, but problem already exists longer, since transition to jquery 3.5.1)

**Solution:**
Use proper end tags in the `bump` skin file.

**Context:**
In HTML5 self-closing tags are only allowed for the defined void elements (and foreign elements).
https://html.spec.whatwg.org/multipage/syntax.html#start-tags

It seems jquery changed the way it recovers from self-closing non-void elements.
In jquery 3.5.1 the `bplaylistctrl` and `brenderers` are removed in the process, breaking `bump`'s renderer selection.

### DIFF
--- a/src/main/external-resources/web/bump/bump.html
+++ b/src/main/external-resources/web/bump/bump.html
@@ -1,7 +1,7 @@
 <html>
 	<head>
 		<title>bump</title>
-		<link rel="stylesheet" href="/files/web.css" type="text/css"> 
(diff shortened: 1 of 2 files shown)


SubJunk (approved):

SubJunk:
Thank you!

Issue #1868, opened 2019-11-03
Managing web content is unreliable


SubJunk:
This was reported in https://www.universalmediaserver.com/forum/viewtopic.php?f=3&t=13823#p40362

I haven't confirmed this but it makes sense since I have tried to make the interface sync with the filesystem and there might be race conditions there depending on hardware configurations

SubJunk:
I have confirmed this on macOS, needs some work for sure

SubJunk:
Fixed in #2550

Issue #1732, opened 2018-12-21
Removing a link from the web content panel should maintain the selection


polygirl:
In Version 8.0.0.0 alpha

- [ ] Select and remove a link from the list
- [ ] See that you are scrolled to the top of the list
- [ ] Should maintain selection so that the previous item in the list is selected and the user remains at the same place in the list.

Issue #1604, opened 2018-10-05
Make the sample web feeds/streams work on non-Windows by default


SubJunk:
On Windows we copy WEB.conf to the profile directory on install which is good, but on macOS at least, that doesn't happen.
Since the "installer" isn't really an installer, we probably need to have the samples stored in our Java code instead, so that we can create the file when it doesn't exist.

SubJunk:
This is hopefully implemented in https://github.com/UniversalMediaServer/UniversalMediaServer/pull/1603

SubJunk:
Fixed in #2550

PR #2550, opened 2021-07-11
Fix YouTube support

### LINKED ISSUE #2128: Fix YouTube support
It doesn't work

SubJunk:
Fixes #2128, #1868, #1732 and #1604

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -875,7 +875,7 @@
 
 			<plugin>
 				<artifactId>maven-idea-plugin</artifactId>
-				<version>2.3-atlassian-10</version>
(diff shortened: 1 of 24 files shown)


SubJunk:
I have tested this on macOS via web interface, will also test on Windows via web interface and DLNA before removing the `awaiting testing` label

valib (approved):

SubJunk:
@UniversalMediaServer/developers this is working well for the web interface, where it starts playing quickly, but for DLNA it is downloading the whole video before it starts playing, which could break playback for big videos. Can you see why that would be?

PR #2576, opened 2021-07-24
Fix chapters

SubJunk:


SubJunk:
@valib I have confirmed this fixes the bug. It doesn't work in the "No transcoding" folder but that was existing behaviour. I don't think we have ever supported sending a raw file with an offset except as a response. I hope to work on that feature soon because it would make resuming much nicer.

valib (approved):

Issue #1864, opened 2019-10-30
Thread exhaustion, UMS created 32k threads


ChrisLundquist:
I don't have a reproducible problem yet, but am reasonably sure I've tracked down the problem.
I was unable to ssh to one of my servers last night (I'd auth and the connection would close, logs later revealed fork was failing). After some debugging I saw that when the server became unresponsive, there were about 32k java threads running. UMS is the only java process.

My configuration is:
UMS is running in an LXD container using bridged networking. nginx is fronting UMS to serve static assets, and forward requests upstream. (java doesn't run as root this way, port 80/443 work)

Log details:
```
INFO  2019-10-29 12:35:55.684 [cling-2] Exception in thread "cling-2" java.lang.OutOfMemoryError: unable to create new native thread
INFO  2019-10-29 12:35:55.684 [cling-2]         at java.lang.Thread.start0(Native Method)
INFO  2019-10-29 12:35:55.684 [cling-2]         at java.lang.Thread.start(Thread.java:717)
INFO  2019-10-29 12:35:55.685 [cling-2]         at java.util.concurrent.ThreadPoolExecutor.addWorker(ThreadPoolExecutor.java:957)
INFO  2019-10-29 12:35:55.685 [cling-2]         at java.util.concurrent.ThreadPoolExecutor.processWorkerExit(ThreadPoolExecutor.java:1025)
INFO  2019-10-29 12:35:55.685 [cling-2]         at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
INFO  2019-10-29 12:35:55.685 [cling-2]         at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
INFO  2019-10-29 12:35:55.685 [cling-2]         at java.lang.Thread.run(Thread.java:748)
...
DEBUG 2019-10-29 17:52:27.959 [UPNP-AliveMessageSender] Sending ALIVE...
...
DEBUG 2019-10-29 18:52:05.432 [HTTPv2 Request Worker 6] Finishing configuration for Unknown renderer
DEBUG 2019-10-29 18:52:05.432 [HTTPv2 Request Worker 6] Reading MediaInfo: true (default: false)
DEBUG 2019-10-29 18:52:05.432 [HTTPv2 Request Worker 6] Updating status button for Unknown renderer
[...]

PR #2588, opened 2021-07-28
Try limiting Cling's thread pool

SubJunk:
Related to discussion in #1864


valib:
@SubJunk it stopped showing the current state of playing the streaming media in the UMS.

SubJunk:
I guess that indicates this is the wrong fix. I may have to build a customised Cling after all. Thanks for testing

PR #2589, opened 2021-07-29
Shutdown more reliably

SubJunk:


src/main/java/net/pms/PMS.java
```diff
@@ -788,6 +788,11 @@ public void run() {
 							"-script", database.getDatabasePath()  + File.separator + "logging_report.txt");
 					} catch (SQLException e) {}
 				}
+
+				// Shut down library scanner
+				if (getConfiguration().getUseCache()) {
+					getDatabase().stopScanLibrary();
+				}
```
valib, line 795: @SubJunk it seems that the database is already closed when the shutdown hook is processed.

SubJunk:
@valib I made this because I noticed the scan was still running after closing UMS, but maybe that isn't reliably happening

SubJunk:
@UniversalMediaServer/developers I have identified that on my machine the UMS shutdown process tends to hang at this line https://github.com/UniversalMediaServer/UniversalMediaServer/pull/2589/files#diff-ac160cd181c0a9603f909975d43756470fd0091a46a422f498ac57a7eb166accR208 but I don't know why. Any ideas?

src/main/java/net/pms/network/HTTPServer.java
```diff
@@ -195,14 +194,19 @@ public void stop() {
...
+			if (allChannels != null) {
+				allChannels.close().awaitUninterruptibly();
 			}
+			LOGGER.info("Confirm allChannels is empty: " + allChannels.toString());
 
-			if (factory != null) {
-				factory.releaseExternalResources();
-			}
+			LOGGER.info("This log happens");
+			bootstrap.releaseExternalResources();
```
valib, line 208: @SubJunk what about to add some delay before executing this to let the UMS to stop properly all classes?

SubJunk: @valib that could work. I would also like to understand why it hangs though since their manual says it should be safe to do in this way. I don't think we are creating new channels in the meantime but maybe I'm wrong.

SubJunk:
@valib Rolling back h2database seems to be the key to the clean shutdown, but this branch could improve some edge cases too

valib (approved):

PR #2599, opened 2021-08-08
Improved UpdateID handling

SubJunk:
We were using individual UpdateIDs because a previous spec said to do that, but in the latest one it says to always send the SystemUpdateID

This also ensures that the SystemUpdateID gets bumped for more events like when a fully played action happens, and filesystem changes


SubJunk:
This has been verified to fix a problem with thumbnails not updating at https://www.universalmediaserver.com/forum/viewtopic.php?p=44273#p44273

valib (approved):

Mik-S-UMS:
The thumbnail updating has been working great now. I have found it takes browsing several folders (about 5) on PS4 before it will reload that folder with the updated thumbnail. If a new file is added to that folder it is updated on the next load.
I'm assuming there is a way of telling the renderer that the folder has changed, if so then it should also be invoked when the thumbnail is changed too.

SubJunk:
@Mik-S-UMS that's good feedback, I'll look into it more so it can change every time

PR #2600, opened 2021-08-08
Less IO spam

### LINKED ISSUE #2465: Javaw.exe opening files and stopping parent folders being moved/deleted (Win 10 permission denied error)
this is a very random bug, maybe once every 24 fours, sometimes a little more.

I tried to move and delete a folder in a directory that was being scanned by UMS as a shared folder. 

In windows 10 pro I got a permission denied error when trying to move/delete this directory.

Looking into it the files were open in another process.

I used process explorer to ascertain that javaw.exe was opening these files and not closing them during the current UMS session. 

Quitting UMS freed javaw.exe hold(open file handle but never closed) on these files.

I was then able to delete the folders/files that javaw.exe had opened for UMS but not closed. Exiting UMS closed said files. 

Javaw.exe is not closing all the files it accesses while in memory.

SubJunk:
This might fix #2465 but needs testing


valib:
@SubJunk what about to filter files during discovery?

SubJunk:
@valib yes I think that seems like a good improvement. We still need the check during parseFileForDatabase too since that happens outside of the normal discovery system. If you're happy with it we can merge

valib (approved):

valib (approved):

PR #2611, opened 2021-08-19
Performance improvement

ik666:
On large item sets (in my case browsing the media library to audio -> all audio files, 15.000 elements) the `DLNAMediaDatabase.getStrings()`method is not very performant. It took over 4 seconds to compute the data. With this PR it takes only 1 second (on my machine).


valib (approved):

SubJunk:
This is great, thanks!

Issue #2612, opened 2021-08-20
Denon AVC-X6700H cannot see files on network


pyorkie:
I have had UMS working for a number of years all OK with my ONKYO TX-NR5008 & Windows 10 ... but just bought a Denon AVC-X6700H and now I can no longer see my files that are on the computer. It will sometimes show the drive and sometimes the folders but cannot see any files (Music). I've tried everything.
Any and all help would be gratefully accepted.
And thanks in advance.

pyorkie.

[ums_dbg_2021-08-21-00-59.zip](https://github.com/UniversalMediaServer/UniversalMediaServer/files/7022391/ums_dbg_2021-08-21-00-59.zip)

valib:
@pyorkie please use HTTP engine V2 and send the log again.

pyorkie:
[ums_dbg_2021-08-21-13-06.zip](https://github.com/UniversalMediaServer/UniversalMediaServer/files/7024773/ums_dbg_2021-08-21-13-06.zip)

I've tried what you said ... but still comes up saying that no files .. even though it finds the folders. I am using UMS v10.0.
Thanks for the reply.
pyorkie

SubJunk:
@pyorkie this looks like the same problem we have seen on some other renderers after version 10.6.0.
Please confirm this by installing 10.5.0 from https://github.com/UniversalMediaServer/UniversalMediaServer/releases/download/10.5.0/UMS-10.5.0.exe and let us know

SubJunk:
Also please try enabling the setting to scan shared folders on startup

pyorkie:
Hi

That seems to fixed the problem. I went back to 10.5 ... now works.
Thanks for all your help.
pyorkie

SubJunk:
@pyorkie if you don't mind could you try this version? https://www.universalmediaserver.com/uploads/UMS-538d288.exe
Hopefully we can fix the problem so you're not stuck on an old version forever :)

pyorkie:
Hi
Thanks .... I've installed it and will report when tested .... by the way .. the 10.5 fixed the missing files prob ... but when tested it was passing my FLAC files as MP3 files to the receiver.
pyorkie

SubJunk:
I'm closing this since the user never replied, feel free to reopen

PR #2613, opened 2021-08-20
Fix NPE made by PR 8865d4a

valib:
It stops working the HTTP engine V1. Reported in #2612. See the log.

### DIFF
--- a/src/main/java/net/pms/network/RequestHandler.java
+++ b/src/main/java/net/pms/network/RequestHandler.java
@@ -104,17 +104,8 @@ public class RequestHandler implements Runnable {
 			// header matches are attempted and if those fail as well we're stuck with the
 			// default renderer.
 
-			// Attempt 1: If the reguested url contains the no-transcode tag, force
-			// the default streaming-only conf.
-			if (request.getArgument().contains(RendererConfiguration.NOTRANSCODE)) {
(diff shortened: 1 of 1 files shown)


src/main/java/net/pms/network/RequestHandler.java
```diff
@@ -104,17 +104,8 @@ public void run() {
 			// header matches are attempted and if those fail as well we're stuck with the
 			// default renderer.
 
-			// Attempt 1: If the reguested url contains the no-transcode tag, force
-			// the default streaming-only conf.
-			if (request.getArgument().contains(RendererConfiguration.NOTRANSCODE)) {
```
valib, line 109: `request` is always null at this line.

SubJunk: @valib cool, can you do that as part of this branch, or should I?

valib: @SubJunk it would be better if you do it

SubJunk: That's done now, can you review?

SubJunk:
@valib I'm going to do a release in about 12 hours so if this can be merged by then it would be good

SubJunk (approved):

PR #2621, opened 2021-08-23
Set MediaInfo defaults

### LINKED ISSUE #2598: Set MediaInfo=true by default
The default value in RendererConfiguration should be set to `true`, but without changing behaviour for any existing renderers, to avoid breaking compatibility for them.
So existing renderers that do not have that line, or have set it to false, should still have it as false, while those who set it to `true` can have it removed.

The main benefit of this change is that new renderers going forward will not have to have the line at all, because it is much better to use MediaInfo than not.

ahmedadan:
- Any file that had explicitly set false was left as false as well as any file that did not have MediaInfo set was explicitly set to false.
- Changed Default Behaviour in `RendererConfiguration.java`

Closes #2598

### DIFF
--- a/src/main/external-resources/renderers/DirecTV.conf
+++ b/src/main/external-resources/renderers/DirecTV.conf
@@ -21,4 +21,3 @@ CustomFFmpegOptions = -ac 2 -c:a mp2 -ab 192k
 CustomMencoderOptions = -channels 2 -lavcopts acodec=mp2:abitrate=192
 KeepAspectRatio = true
 HalveBitrate = true
-MediaInfo = true
--- a/src/main/external-resources/renderers/DirecTV.conf
+++ b/src/main/external-resources/renderers/DirecTV.conf
@@ -21,3 +21,4 @@ CustomFFmpegOptions = -ac 2 -c:a mp2 -ab 192k
 CustomMencoderOptions = -channels 2 -lavcopts acodec=mp2:abitrate=192
 KeepAspectRatio = true
 HalveBitrate = true
+MediaInfo = true
--- a/src/main/external-resources/renderers/Google-Android.conf
+++ b/src/main/external-resources/renderers/Google-Android.conf
@@ -14,5 +14,4 @@ MuxDTSToMpeg = true
 H264Level41Limited = false
 StreamExtensions = mkv,hdmov,hdm,flac,fla,dts,ogg,asf,asx,m2v,mp4,mpg,mpeg,avi,mov,wmv
 ChunkedTransfer = true
-HalveBitrate = true
-MediaInfo = false
+HalveBitrate = True
--- a/src/main/external-resources/renderers/Google-Android.conf
+++ b/src/main/external-resources/renderers/Google-Android.conf
@@ -15,3 +15,4 @@ H264Level41Limited = false
 StreamExtensions = mkv,hdmov,hdm,flac,fla,dts,ogg,asf,asx,m2v,mp4,mpg,mpeg,avi,mov,wmv
 ChunkedTransfer = true
 HalveBitrate = true
+MediaInfo = false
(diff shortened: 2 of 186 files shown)


src/main/external-resources/renderers/Miracast-M806.conf
```diff
@@ -13,7 +13,7 @@ RendererIcon = miracast.png
 # {friendlyName=LOLLIPOP-1C3A18-DMR, manufacturer=Rockchip, modelName=Rockchip Media Renderer, modelNumber=1.0, modelDescription=Rockchip Media Renderer, DLNA(DMR), manufacturerURL=http://www.rock-chips.com, modelURL=http://www.rock-chips.com}
 # ============================================================================
 
-#MediaInfo = true
+MediaInfo = false
```
ahmedadan, line 16: This was commented out so I set it to false, I can change that if we think true would be good in this situation.

SubJunk:
@ahmedadan thank you for this, I will review later today :)

SubJunk (changes requested):
The default still needs to be changed in `RendererConfiguration.java`

src/main/java/net/pms/configuration/RendererConfiguration.java
```diff
@@ -2119,7 +2119,7 @@ public String toString() {
 	 * @return whether to use MediaInfo
 	 */
 	public boolean isUseMediaInfo() {
-		return getBoolean(MEDIAPARSERV2, false) && LibMediaInfoParser.isValid();
+		return getBoolean(MEDIAPARSERV2, true) && LibMediaInfoParser.isValid();
```
ahmedadan, line 2122: After reading the chained constructors I found what I believe to be the default setting, let me know if I misunderstood or made a bad change @SubJunk @valib

src/main/external-resources/renderers/DirecTV.conf
```diff
@@ -21,3 +21,4 @@ CustomFFmpegOptions = -ac 2 -c:a mp2 -ab 192k
 CustomMencoderOptions = -channels 2 -lavcopts acodec=mp2:abitrate=192
 KeepAspectRatio = true
 HalveBitrate = true
+MediaInfo = true
```
valib, line 24: All `MediaInfo = true` should be removed from all renderers.conf

src/main/external-resources/renderers/Google-Android.conf
```diff
@@ -14,4 +14,4 @@ MuxDTSToMpeg = true
 H264Level41Limited = false
 StreamExtensions = mkv,hdmov,hdm,flac,fla,dts,ogg,asf,asx,m2v,mp4,mpg,mpeg,avi,mov,wmv
 ChunkedTransfer = true
-HalveBitrate = true
+HalveBitrate = True
```
SubJunk, line 17: Can you revert this change please?

src/main/external-resources/renderers/Samsung-BDC6800.conf
```diff
@@ -32,3 +32,4 @@ Supported = f:wmv      v:vc1|wmv            a:wma                    m:video/x-m
 # Supported audio formats:
 Supported = f:mp3   m:audio/mpeg
 Supported = f:wma   m:audio/x-ms-wma
+MediaInfo = false
```
SubJunk, line 35: This is interesting since I think we have found a bug with this configuration - it has the `Supported` lines without the `MediaInfo = true`. In this case let's leave it, so can you please remove this line?

SubJunk (changes requested):
Nice work, just a couple of minor changes to make. Can you also sync with master after making the changes?

SubJunk:
If you sync with master I hope it will pass now

ahmedadan:
> 
> 
> If you sync with master I hope it will pass now

Looks like you fixed it, Let me know if you want me to squash into a single commit but otherwise it's all done on my end

SubJunk:
@ahmedadan nice job, no need to squash, I will squash when it merges. I'm just going to test it tonight before merging, in case there are any unanticipated problems

SubJunk:
This is working well for me so I'll merge

SubJunk (approved):

PR #2634, opened 2021-09-06
Implement forked cuelib

valib:
to avoid not Mavenized dependencies


SubJunk (approved):

SubJunk:
It looks good, but what about this instance of it too? https://github.com/UniversalMediaServer/UniversalMediaServer/blob/29aefa4272be3c28624fedd4328177edd1075cb3/pom.xml#L919-L923

PR #2646, opened 2021-09-14
compact database on shutdown ...

ik666:
With this PR the database will be compacted on shutdown freeing up some space. 

Reproduce:
I deleted my DB, UMS did an initial scan and the DB size was around 1.6 TB. With this PR applied, the DB size shrunk to around 425 MB after a shutdown.

```
root@trans:/home/ums/.config/UMS/database# ls -lt 
-rw-r--r-- 1 ums music 1644015616 Sep 14 14:15 medias.mv.db
-rw-r--r-- 1 ums music        370 Sep 14 10:37 medias.trace.db
root@trans:/home/ums/.config/UMS/database# systemctl stop ums
root@trans:/home/ums/.config/UMS/database# ls -lt 
-rw-r--r-- 1 ums music 425132032 Sep 14 14:16 medias.mv.db
-rw-r--r-- 1 ums music       370 Sep 14 10:37 medias.trace.db
```


valib:
@ik666 this change prevents to remove the shared folder from the list.

ik666:
@valib could you please explain what exactly this means. Sounds like something doesn't work any more, but I am not sure what :)

valib:
@ik666 I am not sure how this is related to this PR but when I delete one shared folder and close the UMS than this folder appears in the shared golders again after the UMS is started. Seems like  the configuration is not saved before the UMS is closed.

ik666:
@valib I have made some small changes. Please try it, however I cannot reproduce your issue on my system. If I delete a shared folder, press "stop" in the UI and start UMS again, the deleted shared folder doesn't appear any more (on UPnP protocol level). Adding a shared folder also worked after restarting UMS.

valib (dismissed):

valib (dismissed):

valib:
@SubJunk now this PR is worth to be merged.

valib:
@SubJunk we should check the code to avoid to use getter and setter in the class for variables defined in the class. This could speed up the code.

src/main/java/net/pms/PMS.java
```diff
@@ -788,6 +791,12 @@ public void run() {
 							"-script", database.getDatabasePath()  + File.separator + "logging_report.txt");
 					} catch (SQLException e) {}
 				}
+
+				try (Statement stmt = database.getConnection().createStatement()) {
```
SubJunk, line 795: Shouldn't we check whether the database is enabled before doing this?

SubJunk (changes requested):

src/main/java/net/pms/PMS.java
ik666: @SubJunk What do you mean with "enabled"? The database doesn't shut down on an HUP any more, because I've added `";DB_CLOSE_ON_EXIT=FALSE";` to the database connect URL. Therefore, the database is still online while entering the UMS shutdown hook, where we actually manually close it by calling `SHUTDOWN COMPACT`.

SubJunk: I just mean the user might have the database/cache disabled, so we should check `configuration.getUseCache()`

ik666: I see. Tracing back, which methods could initiate a database connection, by calling `PMS.getDatabase()` I think it's more save to track the initialization process.

ik666:
@valib Sorry, haven't seen your changes early enough. I hope I corrected the merge conflicts and left your code as expected ...

SubJunk (approved):

SubJunk:
I'm happy with this but will keep an eye on whether it affects hanging java process, because we only just fixed that :)

Issue #2643, opened 2021-09-14
Network speed test only runs once per renderer


SubJunk:
My understanding of how the network speed test is that it is supposed to run periodically, or after certain events. But I can see in my logs that it only runs once for each renderer, which has resulted in some non-ideal behaviour for my instance.

Was my understanding wrong @valib or is this a bug?

valib:
@SubJunk you are correct. The test is running only once during the UMS inizialization. The idea to run it periodically sounds good.

PR #2650, opened 2021-09-17
Implement checking network speed periodically

valib:
This fulfill the issue #2643


valib:
@UniversalMediaServer/developers I set the time to refresh the network speed of the renderers to 1 minute. Do you think that the timer should be configurable by users or set to other value?

SubJunk:
@valib I was thinking more like action-triggered, like maybe it could happen after a browse to the first folder, so then it will retry after the user stopped using the device and came back to it?

I tried it locally being triggered by certain events, like stopping a video, but it made it slow to seek to position of a video. What do you think?

valib:
@SubJunk I don't think that to trigger it after some events is good idea. It takes quite long time and it will not respond in time when it is needed e.g. before the transcoging started. When transcoding started than the previous measured speed is used. There is no way to change parameters when transcoding starts only when user use fastforward ot so. Than the new measured speed is used.

Mik-S-UMS:
I think it would be better to check the speed at regular intervals in the downtime when UMS is not doing anything to minimise any effect to the user. maybe something like after 5-10 min?

Is the plan to use the last speed test to change the transcoding paremeters or would it be better to take an average of all tests to get a better idea or typical speed of the network.

valib:
@Mik-S-UMS it is a good idea to use an average value of tests. I can try to implement it.

SubJunk:
@valib I think the averaging wouldn't achieve what my idea was, which is to allow UMS to recover from very different situations. For example if I start UMS when I'm downloading something and that is maxing out my connection so I get 1mbps, and then the download finishes and I go back to 100mbps. So I think the best thing is to use the latest measurement.

Every 10 minutes sounds good to me. You can also use the REALTIME_LOCK I made, to make sure we do not launch it during an operation that is time-sensitive, see https://github.com/UniversalMediaServer/UniversalMediaServer/pull/2537/files#diff-6b3adc65303f216242be4eb5c2c0e70cf8281bd6dc0f006a510d09bca064f0ecR148

valib:
> You can also use the REALTIME_LOCK I made, to make sure we do not launch it during an operation that is time-sensitive,

@SubJunk I can use it but it will be better to merge the PR #2537 before.

Nadahar:
You should make sure that you don't try to measure available bandwidth while using the same bandwidth for streaming. Otherwise, your results will be way off. The "automatic approach" has its limitations no matter how you do it, because you don't know the network topology of the user. It might be that all renderers shares the bandwidth, and it might be that some have lots of bandwidth while others have little. As such, you might end up getting it wrong anyway, but especially if one is playing while the "measurement" is taking place.

SubJunk:
@Nadahar good point, I agree

Nadahar:
@valib I'm not sure I understand what you mean, but in case you were addressing what I said: You can't assume that every network will behave the same as your network does. If you have a lot of spare capacity when streaming, the difference will be relatively small. If the capacity is low though, the difference will be huge. Not only that, you have to consider how the "measurement" itself is done. I works by sending a series of increasingly large ICMP packets while measuring the time it takes for the response to return. The "result" will then be some variant of an average (somewhat weighted as far as I remember). This means that the goal of the measurement is to consume all available capacity while the packets are in transit, and then use the time they take to reach their destination to figure out how much bandwidth is available. If you have good capacity, the packets will transfer in a short time, and the interruption to streaming will probably be minimal and "covered up" by buffering. If the capacity is low on the other hand, they will take relatively long to transfer, and they will consume bandwidth that the streaming needs in the meanwhile, probably leading to empty buffers and stuttering or other playback issues. 

The result is both that you disturb playback *and* that you end up with a measurement that is wrong. That's not what I would want.

valib:
@Nadahar agree that my network is different than other users have. I implemented the REALTIME_LOCK and now the speed is not measured when the UMS is responding to the renderer. It waits until the media is fully streamed.

SurfaceS:
My 2 cents :
Having an estimate network speed once can be good.
But you will always have an estimate network speed that will not reflect the network speed when client request a stream.
The transfer speed between 2 devices can be changed by anythings at anytime (you have multiple device on the network, waves if wifi, and you are not controlling what they do on the network).

The only way to handle this is to check the transfer speed on the outputstream itself and adapt the bitrate of the transcoding stream. I don't know if it's possible on the UMS, but anything else will be ineffective.

Nadahar:
@SurfaceS The perfect solution would obviously be if you can actually measure what was transferred and then adapt. I can't see how this could be achieved though, for multiple reasons. I think there is some support for MPEG-DASH in more recent versions of DLNA, but I don't know if any renderers actually implement it. Even if they do, it's hard to see how to use this to achieve "adaptable bitrate". The way all the online streaming services handles this is that they encode the content in different qualities in advance, and then offer parts in multiple qualities to the player. The player can then switch to lower quality parts if it senses that the buffer is draining - and vice versa. It is thus controlled by the player/renderer.

To support a scenario like this, UMS would have to transcode to multiple different qualities in parallel and somehow cache all these parts so that the renderer could pick the "ideal" quality. It's not really practical.

There's no way for UMS to "measure" what is actually being transferred in real-time - all renderers have buffers, although the size of the buffers varies a lot. Thus, the transfer speed at any give moment is governed by many other factors than the network bandwidth only. On top of this, even if UMS was able to decide the bandwidth at any specific moment, there's no way to "tune" a transcoding process once it has started. FFmpeg and other transcoding engines will use the parameters you supplied them with from the start, there's no way to change the parameters during transcoding.

This is why the only thing that is possible is to get some rough estimate of the available bandwidth, and do the best you can with that very limited information.

SubJunk:
I think this should be closed. We have HLS support now which gives us a better way to adapt to network changes.
Thanks @valib for developing it!

PR #2659, opened 2021-09-23
Checking for non set variables with better syntax

donut87:
The -z option is widely understood wheras the idea of having a prefix, evaluating the string and then look, if it is just the prefix is a bit harder to understand.


Nadahar:
POSIX shell scripts all look kind off "funny" to me, but it's not a particularly uncommon way to do it apprently: https://stackoverflow.com/questions/1805663/shell-script-purpose-of-x-in-xvariable/

PR #2660, opened 2021-09-23
Update PlaylistFolder content

ik666:
At the moment changes (adding / removing elements) made in playlists-files (.m3u8 , m3u, etc.) are not visible in UMS, until UMS is restarted. 

This PR updates the playlist content each time the playlist is accessed. User's will see changes in playlists without restarting UMS.

### DIFF
--- a/src/main/java/net/pms/network/RequestV2.java
+++ b/src/main/java/net/pms/network/RequestV2.java
@@ -1185,7 +1185,10 @@ public class RequestV2 extends HTTPResource {
 
 		for (DLNAResource dlnaResource : files) {
 			if (dlnaResource instanceof PlaylistFolder) {
-				((PlaylistFolder) dlnaResource).resolve();
+				File f = new File(dlnaResource.getFileName());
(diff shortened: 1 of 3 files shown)


src/main/java/net/pms/network/RequestV2.java
```diff
@@ -1182,6 +1183,12 @@ private StringBuilder browseSearchHandler(BrowseSearchRequest requestMessage) {
 			searchCriteria
 		);
 
+		for (DLNAResource dlnaResource : files) {
```
SubJunk, line 1186: We are already looping through `files` further down on line `1202`, can you add this logic to that existing loop for performance?

ik666: good point. Have done it ...

ik666:
UPDATE: The playlist will not be updated on each access any more, but only if the modified date of the playlist files is updated.

SubJunk (approved):

SubJunk:
@ik666 thanks!

PR #2661, opened 2021-09-24
Use the jaudiotagger fork and not use the local jar which not works.

valib:
@SubJunk it seem that the owner of the`jaudiotagger` is an idiot who is not able to merge the PR made by @Nadahar in the https://bitbucket.org/ijabz/jaudiotagger/pull-requests/67 so we should fork the https://bitbucket.org/ijabz/jaudiotagger/src/master/ to the GitHub and store it in the Maven repository instead of the https://repository.mulesoft.org/nexus/content/repositories/public/ which seems to stop working. You now how to store it in the Maven repository so you should help. To use the Maven repository will be bether than to use the obscure repositories like the `repository.mulesoft.org` or `jitpack.io`


SubJunk:
Ok, I'll merge this now to fix master, and I'll make our fork later

PR #2666, opened 2021-10-03
Caching solution for open subtitles hash

valib:


valib:
The UMS is caching several values in the `HashMap` like renderers, network addresses etc. so this is only new implementation of it.

SubJunk:
@CodeLover254 thanks :D

PR #2668, opened 2021-10-06
simplify if loop

kxrtiswithak:
Inspired by [**DeepSource's suggestion**](https://deepsource.io/gh/UniversalMediaServer/UniversalMediaServer/issue/JS-D009/occurrences), rather than replacing with a ternary operator, I chose to refactor into one line as a return statement. Very small change so unlikely to unforeseen   #@ramifications


SubJunk:
@kxrtiswithak thanks!

PR #2670, opened 2021-10-06
Should not have unused variables

Flonator3000:


SubJunk:
@Flonator3000 thanks!

PR #2673, opened 2021-10-11
Show the folder name in the Status Line when the subtitles API

valib:
is searching for the files subs not the empty line which force the GUI to hide the Status Line and let the Shared folders to grow.


SubJunk:
@valib I have noticed a bug with this - when scanning finishes, it still shows the last folder it scanned at the bottom, but it should disappear

Issue #2553, opened 2021-07-15
Since version 10.2.0 UMS didn´t work on Samsung ks8000 (year 2016) Tvs


ilopezv:
Since version 10.2.0 UMS didn´t work on Samsung ks8000 (year 2016) Tvs

I tried every single version since this release until 10.7.0 and no one work after 10.2.0

SubJunk:
@ilopezv can you please describe the problem more? What does "didn't work" mean?

ilopezv:
sorry for the previous bad quality comment.

All seems to be all right in UMS; the tv appears in the panel; no errors in log, but the dlna icon with UMS logo didn´t appear in the TV, so i can´t use it any more.

Like i said before only ultil UMS 10.2.0 the icons appears in tv and all things go ok.

What kind of debug info do you need?

SubJunk:
@ilopezv can you please post logs from 10.2.0 and 10.3.0 so I can compare what changed? Instructions are at https://www.universalmediaserver.com/forum/viewtopic.php?f=3&t=556&sid=31084e979a309863f7640c2d4ec48f14

valib:
It seems that the main problem is the network setting on your PC. You use (or the UMS choose) the HyperV Virtual interface for your network. Please try to set the network adapter manually.

valib:
The problem is that you have two virtual IP addresses. The `192.168.88.242` on the `eth6, Hyper-V Virtual Ethernet Adapter #2` and `172.22.96.1` on the `eth21, Hyper-V Virtual Ethernet Adapter` and the not working log v10.12 shows that the UMS chooses the IP `172.22.96.1` see 
`INFO  2021-10-17 12:58:56.481 [main] net.pms.network.HTTPServer Using address /172.22.96.1 found on network interface: name:eth21 (Hyper-V Virtual Ethernet Adapter)` which is wrong. It could be a bug in the UMS. We made the change that virtual adapters mostly used for VPN connection are skipped when there is at least one physical adapter with the valid IPV4 address. You don't havy any even if you have  Intel(R) I211 Gigabit Network Connection adapter. Choose manually the  `eth6, Hyper-V Virtual Ethernet Adapter #2` should solve the problem.

PR #2681, opened 2021-10-18
Continue to fix the #2553

valib:


SubJunk:
> This assumes that interfaces order in the OS is made upon their importance.

I'm not sure whether that is a correct assumption but I don't see a problem with the code

PR #2689, opened 2021-10-26
Feature/master/sort music tracks

ik666:
This PR delivers `musictrack:audioitem` objects sorted by track id and disc id (if available), overriding the configured sort algorithm. If `track id` metadata is not available, objects will be sorted by name. 

Why: Sorting audio files are usually handled differently then video files. Most people add either correct metadata to their audio files or have the filenames prefixed with the track number.


valib (dismissed):

ik666:
Audio files from multiple discs can be copies into one folder. Audio files will be sorted and delivered in correct order to the requesting client.

ik666:
@valib thx for checking the code. Instead of opening new PR for added functionality I used this one, because I thought it is closely related. This should be the final state from my point of view.

valib (approved):

SubJunk:
@ik666 we have had a user on the forum report a problem with this, which is that it breaks sorting in mixed folders - a folder that has video, audio and images. https://www.universalmediaserver.com/forum/viewtopic.php?p=45050#p45050

Maybe you could do an enhancement to this feature where it happens in folders that only have audio, what do you think?

ik666:
Hi there, yes I'll provide a fix, right when I'm back from vacation next week. Thanks for reporting.

SubJunk:
@ik666 thanks!

ik666:
Hmmm ... usually people have album art inside the audio folder, therefore having always mixed folders ... 

Maybe adding an additional boolean sort flag like "sort by track if audio is available" might help people disabling or enabling audio files sorting ?

ik666:
I have implemented a more strict check, when track/disc sorting should be done. It is still a heuristic.

ik666:
I added a configuration switch to disable this feature at all, just in case someone has still has issues and doesn't like it:

added `disable_audio_track_sorting = true` to UMS.conf 

Info: The branch has a windows_lastest build issue ...

PR #2691, opened 2021-10-30
Api data versioning

SubJunk:
When our API has bumped a version to indicate that data is stale, progressively refetch the data.

Other changes:
- Protect against uncaught exceptions
- Fix memory leaks
- Improved database thread-safety
- Increased performance of some existing queries
- Fix sending API requests when external network is disabled

Scenarios I tested:
- Database from 10.12.0 upgrades successfully, and drops/updates outdated API info successfully
- Fresh database is created successfully


src/main/java/net/pms/database/TableTVSeries.java
```diff
@@ -163,27 +167,36 @@ public static long set(final HashMap tvSeries, final String seriesName) {
 		return -1;
 	}
 
+	/**
+	 * Get TV series by IMDb ID. If we have an API version
```
js-kyle, line 171: Is this comment finished?

src/main/java/net/pms/dlna/DLNAResource.java
```diff
@@ -4820,77 +4820,77 @@ public String getResolutionForKeepAR(int scaleWidth, int scaleHeight) {
...
-		}
+				if (extraInformationFromFilename != null) {
+					media.setExtraInformation(extraInformationFromFilename);
+				}
 
-		try {
-			if (configuration.getUseCache()) {
-				PMS.get().getDatabase().insertVideoMetadata(file.getAbsolutePath(), file.lastModified(), media);
+				if (configuration.getUseCache()) {
+					// TODO: Make sure this does not happen if ANY version already exists, before doing this
```
js-kyle, line 4874: Just to double check - these are todos not to be addressed in this PR?

js-kyle (approved):

src/main/java/net/pms/dlna/DLNAResource.java
SubJunk: Good question, yes I think it can wait since that's existing code

PR #2699, opened 2021-11-06
HLS on web interface

SubJunk:


SubJunk:
@SurfaceS I have a request for you, and feel free to say no, but can you resolve the conflicts here, since I think they are from your code merging? i

SubJunk:
@SurfaceS just checking you saw this post, since the branch is now in a broken state https://github.com/UniversalMediaServer/UniversalMediaServer/pull/2809#issuecomment-1031022115

SurfaceS:
I think HLS should be implemented at the `net.pms.encoders` level and not in the `net.pms.configuration.WebRender` class.

Mik-S-UMS:
Would it be possible to use HLS on renderers that support it instead of just on the web interface? I don't know if many do but I know Movian does.

Nadahar:
I still don't see the point of using HLS, MPEG-DASH or similar for UMS. What is the benefit? That you can rewind? You can't seek beyond the point the transcoding has "reached" anyway, which assume would be the most relevant use case (you want to seek to where you previously quit - aka continue).

The benefit when using pre-encoded content with chunks in multiple qualities is obvious, but this is so far from the scenario we're in that I just don't see the point.

The "proper" solution is time-based seeking, where the renderer requests a playback position using a time stamp instead of a byte position. This is a part of DLNA, but sadly it seems to many renderer manufacturers are too lazy to implement it. Still, it's the only solution I can see that makes seeking work properly when transcoding. When not transcoding, seeking by byte position already works well.

SurfaceS:
@Nadahar I agree with you.
HLS, MPEG-DASH have some benefit if you have multiples bit rates, then the client can take the stream it want, but UMS need to store transcoded parts in file. It does not have direct stream (via pipe).

@Mik-S-UMS  that what I mean by 
> should be implemented at the net.pms.encoders level

SubJunk:
> I still don't see the point of using HLS, MPEG-DASH or similar for UMS. What is the benefit? That you can rewind? You can't seek beyond the point the transcoding has "reached" anyway, which assume would be the most relevant use case (you want to seek to where you previously quit - aka continue).

Browsers have their own built-in HLS resuming which seems to work pretty well and it's one of the benefits of this branch. With our settings (ultrafast x264 preset) files are transcoded pretty quickly in my experience, on an 8-year-old CPU, so it gives an improved user experience on the web interface. It takes us from not being able to seek in transcoded videos, to being able to.

Nadahar:
@SubJunk It might be possible to implement time-based seeking by "hacking" videojs, but JS isn't my cup of tea so I haven't even attempted. I don't know how much of the handling is done natively in the browser and how much is in videojs, but since plugins/"tech"s can implement support for things like Flash and HLS, I would think it should be possible. The time-based seeking itself is super-simple, you just use some custom HTTP header for the time info, preferably something very similar to the DLNA headers. The playback must be "reinitialized" and the buffer dumped after every seek though, because the different "segments" acquired when using time-based search won't "fit" each other (or at least I think that would be very difficult to accomplish). Apart from that, all you would need to do is to handle the "total time/length" in some way (a HTTP header could be used from the server to videojs for example) so that progress and seek limits could be handled properly.

When it comes to "fast transcoding", I think that will only be practically usable for low-quality video content. My computers at least, still take some time (minutes at least) to transcode a complete video of an hour or more in decent quality with ultrafast, so it would still mean that seeking forward could mean minutes of waiting (and probably a dozen timeout errors in the meanwhile). I think most users would conclude that something was wrong/that it didn't work after 10-20 seconds.

SubJunk:
@Nadahar I'm not sure whether you have tried using it but this branch handles that situation in a user-friendly way in my opinion. The video player interface grows the seekable part as the playlist grows in realtime, with the seekable duration growing along with it, the same way that players show buffered amounts on websites.

Nadahar:
@SubJunk I haven't tried this branch, but I've been "studying the problem" a bit earlier. If the "length" grows you won't get the "waiting with timeouts" problem, but the situation where you have to wait until the transcoding has reached the point you want to seek to still applies.

As I said, it might work fine in some situations, especially for shorter videos in limited quality, but my "goal" when looking at it was to try to find a way that actually worked "universally". I ended up thinking that making a plugin/tech for videojs probably was the best solution, but that's just theory until someone has actually tried it. It might not be possible for some stupid reason, which is so common with both browsers and JS (and especially in combination).

SubJunk:
> the situation where you have to wait until the transcoding has reached the point you want to seek to still applies.

Yes this is true. However it is still better than the current situation, which is that seeking is not possible.

I just tested with a 1080p file and it was gaining about 1 minute per 1-2 seconds which felt fast enough (subjectively for me), but 4K files would likely feel slow

SurfaceS:
I think that this implementation was not taking the good start.

Writing transcoded files onto temp folder is not the good way to think for UMS (need to delete, taking space, etc).
UMS don't want to transcode to file then serve it, it always used pipe to serve stream, and it's I think the way to work.
It is impossible with ffmpeg HLS way that is :
Create a ts file, then create a m3u8 file, then create another ts file, update m3u8 file, etc....

Maybe we could :
Create a m3u8 file with multiple renditions settings of the video (BANDWIDTH, CODECS, RESOLUTION, AUDIO, VIDEO, SUBTITLES, CLOSED-CAPTIONS).
That is an improvement for low bandwidth and bandwidth changing over the time clients, non supported codecs, and may be an answer to "Checking network speed".
**example file 9.m3u8 :**
```
#EXTM3U
#EXT-X-VERSION:3 
#EXT-X-STREAM-INF:BANDWIDTH=800000,RESOLUTION=640x360 
9_360.m3u8 
#EXT-X-STREAM-INF:BANDWIDTH=1400000,RESOLUTION=842x480
9_480.m3u8
#EXT-X-STREAM-INF:BANDWIDTH=2800000,RESOLUTION=1280x720
9_720.m3u8
#EXT-X-STREAM-INF:BANDWIDTH=5000000,RESOLUTION=1920x1080
9_1080.m3u8
```


The HLS client will take the one better for him itself.
Let create it on the fly, by putting all the EXTINF ts files that will be at end (video duration divided by #EXT-X-TARGETDURATION).
On the ts link, UMS should be able to know the file id, the requested rendition setting and the offset.
**example file 9_360.m3u8 :**
```
#EXTM3U
#EXT-X-VERSION:6
#EXT-X-TARGETDURATION:10
#EXT-X-MEDIA-SEQUENCE:0
#EXT-X-PLAYLIST-TYPE:VOD
#EXT-X-INDEPENDENT-SEGMENTS
#EXTINF:10.000000,
/get/ts_9_360_0.ts
#EXTINF:10.000000,
/get/ts_9_360_1.ts
#EXTINF:10.000000,
/get/ts_9_360_2.ts

.... a lot more ....

#EXTINF:10.000000,
/get/ts_9_360_5000.ts
#EXT-X-ENDLIST
```

Then, ffmpeg will then transcode and serve the requested ts on the fly.

Nadahar:
@SurfaceS If you could get that to work, it would probably be a really good solution. It would both allow seeking anywhere at any time, AND allow some bitrate adaption :+1: 

I don't know enough about HLS to know if you can create the segments that way though (I mean as short, standalone video files with headers and all). But if you can, it seems like it would be a really good solution IMO.

SurfaceS:
Something like that :
https://github.com/SurfaceS/UniversalMediaServer/tree/11.0.0_HLS
It need tweaking / reviewing / working as it's just a quick and dirty shoot and that I don't investigate (know) all parts of UMS.
For now, it show a wrong duration on the interface (like 12 hours instead of 40 min) so it can be seek only on the few part from the start. Maybe a ffmpeg param missing.
But it's the idea.

Nadahar:
@SurfaceS I don't understand all you're doing, but I suspect that's because UMS' code has changed quite a bit since I left as a developer. I do understand the principle though, and my big question is: Does this actually work - does it play as it is now? If so, that's really great!

Regarding the total duration, isn't that specified in the `.m3u8` file? Or does the player somehow retrieve that from the video chunks themselves - if so, it should be possible to specify some FFmpeg parameters that make them have the "correct" time codes for the segments they represents. Is the total duration correct in the generated `.m3u8` file, or if there's no total duration there, is the sum of the duration of all the chunks in the `.m3u8` file correct?

SubJunk:
@SurfaceS I agree that would be better to do it the way you propose, if it works. What I have done is only meant as an incremental improvement, not a perfect solution, and yours is better.

If you are willing and motivated to do it that way, I support it. I have a lot of other things I'm working on in UMS at the moment so I am happy to leave this one to you?

SurfaceS:
I don't know if I will have a lot of time next days/weeks. But I'll do my best.

SubJunk:
@SurfaceS ok great. Please keep me updated, if it becomes too much work. We can always do both approaches - this one first since it is close to being done, and then replace it with your one when there is time.

SurfaceS:
@SubJunk 
That was a bug in UMS (#2866).
The [11.0.0_HLS](https://github.com/SurfaceS/UniversalMediaServer/tree/11.0.0_HLS) should work as expected now.

SurfaceS:
Added multi language selection on [11.0.0_HLS](https://github.com/SurfaceS/UniversalMediaServer/tree/11.0.0_HLS)

SubJunk:
@SurfaceS nice work! do you want to make a new PR for it?

SurfaceS:
> @SurfaceS nice work! do you want to make a new PR for it?
done

SubJunk:
This has been superceded by https://github.com/UniversalMediaServer/UniversalMediaServer/pull/2880

PR #2729, opened 2021-12-16
Allow users to have no shared content (both remote or local)

### LINKED ISSUE #2663: Folders being re-added if deleted from shared content.
A user on the forum only wants to use virtual folders and has tried to delete the default folders set in the shared content tab, but when they are deleted they are added again.
https://www.universalmediaserver.com/forum/viewtopic.php?f=4&t=14874

I have noticed similar behaviour if I try and delete all the items in web content.

I think these should only be auto added if it is a new install but not on an existing one.

SubJunk:
Fixes #2663


SubJunk:
This was tested and confirmed by a user so I'll merge

PR #2733, opened 2021-12-19
h2database from 1.4.197 to 2.0.202 upgrade

SurfaceS:
Fixes #2715
- Update h2database dependency from 1.4.197 to 2.0.202.
- Migrate database from h2 format 1 to format 2 on needed.
- set DB_USER / DB_PASSWORD as private static.
- Escape words KEY, VALUE and YEAR in SQL queries. (These are reserved SQL words and will result in query fail with new h2 version.)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -193,8 +193,7 @@
 		<dependency>
 			<groupId>com.h2database</groupId>
 			<artifactId>h2</artifactId>
-			<!-- Versions above 1.4.197 should only be used when https://github.com/UniversalMediaServer/UniversalMediaServer/issues/2030 is done -->
-			<version>1.4.197</version>
+			<version>2.0.202</version>
 		</dependency>
 		<dependency>
 			<groupId>ch.qos.logback</groupId>
--- a/src/main/java/net/pms/database/TableAudiotracks.java
+++ b/src/main/java/net/pms/database/TableAudiotracks.java
@@ -134,7 +134,7 @@ public class TableAudiotracks extends Tables {
 			sb.append(", ALBUMARTIST       VARCHAR2(").append(SIZE_MAX).append(')');
 			sb.append(", SONGNAME          VARCHAR2(").append(SIZE_MAX).append(')');
 			sb.append(", GENRE             VARCHAR2(").append(SIZE_GENRE).append(')');
-			sb.append(", YEAR              INT");
+			sb.append(", `YEAR`              INT");
 			sb.append(", TRACK             INT");
 			sb.append(", DISC              INT");
 			sb.append(", DELAY             INT");
@@ -161,7 +161,7 @@ public class TableAudiotracks extends Tables {
 			executeUpdate(connection, "CREATE INDEX IDXGENRE on AUDIOTRACKS (GENRE asc);");
 
 			LOGGER.trace("Creating index IDXYEAR");
-			executeUpdate(connection, "CREATE INDEX IDXYEAR on AUDIOTRACKS (YEAR asc);");
+			executeUpdate(connection, "CREATE INDEX IDXYEAR on AUDIOTRACKS (`YEAR` asc);");
 		}
 	}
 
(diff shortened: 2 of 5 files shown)


SubJunk:
@SurfaceS this looks promising, thank you.
I'm interested in the upgrade part. Their docs say the best upgrade path is backup and restore (https://h2database.com/html/tutorial.html#upgrade_backup_restore), so I wonder why they wouldn't prefer this way instead. Do you know why?

SurfaceS:
This is exactly what is done by their function.
- Download the old jar (by maven then by http if maven is not on the computer). By the way we need to give the function the old db version used, as their is no way to find it from the db file itself, 197 for us.
- Create a backup of the old db file (copy to medias.mv.db.bak).
- Load the old module and use it to open the old db.
- Create an SQL script backup file of it.
- Unload the old module.
- Use the current version to create a database with the new version 202.
- Run the script to fill and convert data in it.
- Delete the SQL backup file, but leave the medias.mv.db.bak db file (version 197).

SubJunk:
@SurfaceS I see, that's really cool. Thanks for explaining. It would be good if they would link to that utility from their docs since I thought we had to write that ourselves

SubJunk (approved):

SubJunk:
Thanks again for this. It will make database operations faster again since they have been slow since the rollback

SubJunk:
@SurfaceS sorry I have had to revert this because I noticed a problem in the tests, can you please look at it?
https://github.com/UniversalMediaServer/UniversalMediaServer/runs/4611809555?check_suite_focus=true
```
[INFO] Running net.pms.database.TableFilesStatusTest
Error: org.h2.jdbc.JdbcSQLSyntaxErrorException: Constraint "PRIMARY KEY | UNIQUE (FILENAME)" not found; SQL statement:
CREATE TABLE FILES_STATUS(ID            IDENTITY PRIMARY KEY, FILENAME      VARCHAR2(1024)        NOT NULL, MODIFIED      DATETIME, ISFULLYPLAYED BOOLEAN DEFAULT false, CONSTRAINT filename_match FOREIGN KEY(FILENAME) REFERENCES FILES(FILENAME) ON DELETE CASCADE) [90057-202]
```

I'm looking into why tests don't run properly in forks now, since this should have been caught before merge

SubJunk:
Ah, I see that `YEAR` is a reserved keyword in h2database v2 so that is causing the error. We will need to change that in the migration too. Also `KEY` and `VALUE`. The list is at https://h2database.com/html/advanced.html#compatibility

SurfaceS:
> Escape words KEY, VALUE and YEAR in SQL queries. (These are reserved SQL words and will result in query fail with new h2 version.)

That's why I escaped the fields in queries.
`YEAR[*]` mean that the db engine do not escape this one. ([*] at end)

SubJunk:
I think it would be ideal to rename the affected columns during the upgrade, is that something you can implement?

We can change the FILENAME column to be UNIQUE on creation.

It looks like there are quite a few changes, maybe we should collaborate on the branch, or do you want to try again yourself?

SurfaceS:
>I think it would be ideal to rename the affected columns during the upgrade, is that something you can implement?

Sure
`KEY` and `VALUE` are used on the `METADATA` table. That can simply be rename to `M_KEY` and `M_VALUE` and code changed to work with it.
For `YEAR`, it's use in more tables, so I'll need more time.

Also, `TABLES` is a reserved keywords for SQL, not used by H2, but maybe they will some day. It is used for the version table, that could be renamed to `TABLES_VERSIONS`.


I think we should harmonize the database creation / update / usage under `net.pms.database` instead of doing it under `net.pms.dlna`.
Same for the tables `FILES` `METADATA` `SUBTRACKS` and `REGEXP_RULES`.
I'll give you a first shoot on it today.

On the same time, I'll investigate why the `TableFilesStatusTest.testIsFullyPlayed` test fail on the reverted.

SubJunk:
@SurfaceS that sounds like all positive changes, thank you. Happy holidays to you!

PR #2735, opened 2021-12-20
Using TCK Tested JDK builds of OpenJDK 

carldea:
The AdoptOpenJDK has been discontinued since July 2021 (https://adoptopenjdk.net). Switching the distribution to Azul Zulu. When using Zulu you get all the latest updated (TCK Tested) builds for all versions of OpenJDK. 

Please see more details at Foojay.io:
https://foojay.io/today/github-actions-with-java-part-2/

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -33,6 +33,9 @@ jobs:
 
   windows-latest:
     runs-on: windows-latest
+    strategy:
+      matrix:
+        java-version: [ 8 ]
     steps:
       - uses: actions/checkout@v2
       - name: Set up JDK ${{ matrix.java-version }}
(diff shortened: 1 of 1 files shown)


SubJunk:
Thanks for this. We have been using the Adoptium builds to get the latest versions. What's the difference? https://adoptium.net/?variant=openjdk8&jvmVariant=hotspot

carldea:
> Thanks for this. We have been using the Adoptium builds to get the latest versions. What's the difference? https://adoptium.net/?variant=openjdk8&jvmVariant=hotspot

Both are fine, others may not know, but `microsoft` and `liberica` are recently available too. More distros coming soon. 

The only difference is that Temurin does not support many Non-LTS, archived, and early access releases versions.

The original GitHub action `setup-java@v1` was using Azul’s Zulu. 🙂

SubJunk:
@carldea thanks for the explanation, I have also been bundling Liberica for Linux 32-bit since Adoptium doesn't offer that, so it would be nice to use one source for all architectures.
I'll merge this change now, with the future goal of switching to Zulu in our distributions too

SubJunk (approved):

PR #2736, opened 2021-12-21
Fix non utf8 processWrapper

SurfaceS:
**Fact :** logged line from processWrapper on windows show strange characters on the logfile on systems using non ascii char like accent, non latin, etc.
Windows process inputstream connected to the normal output use the windows encoding.
When reading debug files from others, I've seen a lot of them with that issue (on ping output, ffmpeg output, etc).

This fix fill once the static (to avoid checking each time the class is use) charset used by the system on the OutputTextConsumer Class.
On the same time, I also set some variable final for code optimization.

### DIFF
--- a/src/main/java/net/pms/io/OutputTextConsumer.java
+++ b/src/main/java/net/pms/io/OutputTextConsumer.java
@@ -47,7 +47,7 @@ public class OutputTextConsumer extends OutputConsumer {
 			if (Platform.isWindows() && WinUtils.getOEMCharset() != null) {
 				charset = WinUtils.getOEMCharset();
 			} else {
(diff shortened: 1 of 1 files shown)


Nadahar:
Have you tested this on non-Windows? I think using UTF-16 is quite questionable, I'm not sure if that will ever be the console encoding. AFAIK all non-Windows OS'es in use today use UTF-8.

SurfaceS:
> AFAIK all non-Windows OS'es in use today use UTF-8.
It was a typo, changed.

SubJunk (approved):

SubJunk:
Nice one, thanks!

PR #2764, opened 2022-01-01
Media database instance

SurfaceS:
**Database**
- [x] Moved database initiator, opener, closer, under the database class.

- [x] Moved database report creation under the database class.

- [x] Track database status `DatabaseStatus` (opening, opened, closing, closed, ....)

- [x] Uniformized logging texts

- [x] Wait for all connections released  before `shutdown compact` the db (allow functions to finish all the job and release connection)

**MediaDatabase**
- [x] Instantiating MediaDatabase as only one instance is needed (connection pool)

- [x] Uniformized Table creation

- [x] Allow access to tables members only with an existing connection as parameter.

- [x] Try to reuse the same database connection from a function calling other function if possible (avoid connection recreation, costly and wastes resources)

- [x] Moved the library scanner thread away from `MediaDatabase` to the new `dlna.LibraryScanner` class (as it as nothing to do with the db itself)

**licence headers**
- [x] Uniformizing UMS licence headers, as some have `streaming any medias to DLNA` and others `streaming any media to DLNA`.


SubJunk:
@SurfaceS these all look like positive changes, is it ready for review and testing?

SurfaceS:
I think so.

SubJunk:
@SurfaceS ok great. I'm going to do a release now since there's a lot of value sitting in master, and then I'll prioritize this

src/main/java/net/pms/dlna/DLNAResource.java
```diff
@@ -3360,7 +3369,15 @@ protected void checkThumbnail(InputFile inputFile, RendererConfiguration rendere
...
-				MediaTableThumbnails.setThumbnail(media.getThumb(), inputFile.getFile().getAbsolutePath(), -1);
+				Connection connection = null;
+				try {
+					connection = MediaDatabase.getConnectionIfAvailable();
+					if (connection != null) {
+						MediaTableThumbnails.setThumbnail(connection, media.getThumb(), inputFile.getFile().getAbsolutePath(), -1);
+					}
+				} finally {
+					MediaDatabase.close(connection);
+				}
```
SubJunk, line 3380: I wonder if we can have the best of both worlds by _allowing_ an incoming connection, but if one doesn't exist, the method itself (in this case `MediaTableThumbnails.setThumbnail`) makes a new one?
That way if we have a chain of db operations to do, we can use the same connection, but in cases like this one we don't have to create one in the caller for just one operation.
Is that too messy logically? I'm interested in your opinion

src/main/java/net/pms/util/FileUtil.java
```diff
@@ -1646,11 +1656,12 @@ public static Charset getFileCharset(@Nullable File file) throws IOException {
 	}
 
 	/**
-	 * Tries to detect the {@link Charset}/encoding for the specified file. If
-	 * no valid {@link Charset} is detected or the confidence of the best match
-	 * is below the threshold, {@code null} will be returned.
+	 * Tries to detect the {@link Charset}/encoding for the specified file.If
+ no valid {@link Charset} is detected or the confidence of the best match
+ is below the threshold, {@code null}will be returned.
```
SubJunk, line 1661: Formatting

SubJunk (changes requested):
This all looks very positive, I like the refactoring. I found some tiny things, and asked a question to get your thoughts

src/main/java/net/pms/dlna/DLNAResource.java
SurfaceS: That's what I have done before, and then I removed it.
When I removed it, I had not yet implemented all the stuff (db state, ...).

So it was reopening the db on connection creation call, even if we stopped it.

By the way, some functions need to access the db even if its state is not OPENED (like table checks).

But, I'm now able to recreate some function without connection if you want.
I'll do it for `MediaTableThumbnails.setThumbnail`

SurfaceS:
Fixed review
Added a timeout on close (waiting threads) and logging

SubJunk:
@SurfaceS thanks for the changes

SubJunk (approved):

PR #2776, opened 2022-01-06
servlet the webserver

SurfaceS:
Servlet the webserver part.

- Rename the `remote` directory to `webserver` to reflect what is under.
- Moved the handlers classes to `webserver.handlers` package (sun httpserver handlers).
- Moved the `net.pms.network.PlayerControlHandler` to the `webserver` package as it is served under it.
- Added all servlets classes in the `webserver.servlets` package.
- Server is still using the old way (sun `httpserver`) to be sure we can test everything before setting `jetty` by default.
- Added a config key to allow testing jetty (`web_type =jetty`).

This is a first shoot, it may need some improvements.
The good thing is that the default is to run like before (sun httpserver way).

Servlets can be served by many web container (jetty, tomcat, etc).

On this pull, Jetty is on version 9.4 (due to Java 8) and need `javax.servlet` version 3.1.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -227,6 +227,21 @@
 			<version>3.10.6.Final</version>
 			<scope>compile</scope>
 		</dependency>
+		<dependency>
+			<groupId>javax.servlet</groupId>
+			<artifactId>javax.servlet-api</artifactId>
+			<version>3.1.0</version>
+		</dependency>
+		<dependency>
+			<groupId>org.eclipse.jetty</groupId>
+			<artifactId>jetty-server</artifactId>
+			<version>9.4.44.v20210927</version>
+		</dependency>
+		<dependency>
+			<groupId>org.eclipse.jetty</groupId>
+			<artifactId>jetty-servlet</artifactId>
+			<version>9.4.44.v20210927</version>
+		</dependency>
 
 		<!-- maven replacement for nsisant-1.2.jar -->
 		<dependency>
--- a/src/main/java/net/pms/PMS.java
+++ b/src/main/java/net/pms/PMS.java
@@ -23,7 +23,6 @@ import static org.apache.commons.lang3.StringUtils.isNotBlank;
 import ch.qos.logback.classic.Level;
 import ch.qos.logback.classic.LoggerContext;
 import com.sun.jna.Platform;
-import com.sun.net.httpserver.HttpServer;
 import java.awt.*;
 import java.io.*;
 import java.net.BindException;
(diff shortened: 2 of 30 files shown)


Nadahar:
Just out of curiosity, what is the benefit of using a servlet for this? I mean I know that `HttpServer` kind of sucks, but it mostly does what is needed and I assume that a servlet is a much "heavier" solution. I've never worked with jetty or any of the other servlets, but I've always assumed that they were intended for making "advanced Java webserver solutions". Isn't it also some extra hassle with all the changes since Oracle dropped them and all this making it potentially difficult to support all Java versions?

When it comes to `remote` I've never liked the name either, but the fact is that UMS contains two HTTP servers, not one. I assume that whoever chose the name did it to distinguish the two. The "primary" HTTP server uses netty while the web interface uses `HttpServer`. I think `webserver` can be somewhat ambiguous.

SurfaceS:
servlet is a kind of interface that is not linked to any particular server, allowing to move more easily on another solution if needed.

Concerning the CVEs, the major web container providers (servlets holders) update their code even if you are stuck with java 8 (ex : 9.4.44 2021-09-27 or Tomcat v8.5.73 2021-11-17).

For the `webserver` package, I see that there is a dlna server that run on port 5001, and the web server that sit on 9001.
This server is named `web` under the `pms` instance, retrieving it with `getWebServer()`, and "Web" in the debugging texts.

Nadahar:
Whether you call it a DLNA server or a web server is subjective I guess, they are both HTTP servers. The one on 5001 doesn't serve much HTML, it mostly serves XML and binary files. So, in a sense I guess you could argue that the one on 9001 is "more of a web server" because it actually serves HTML files. Technically though, they are both HTTP servers and don't differ that much. My point was only to try to explain what I believe was the rationale behind naming things related to the web interface "remote". I never thought "remote" was a good name, I just thing that one should be aware of the situation and don't make it even more "confusing".

Regarding the CVEs I understand that they still maintain Java 8 versions - but for how long? The "advantage" with the built-in stuff is that exploits can be fixed (at least to some extent) simply by updating the JRE - in addition to the fact that I doubt the search for exploits is anywhere near that intensive for `HttpServer` since I doubt it's in use for Internet facing web servers.

Even though a servlet isn't implementation specific, that doesn't really tell me what the benefit of using one is. It will mean additional dependencies and a larger JAR - which I would consider disadvantages, so what are the advantages that makes up for this and make it worth it in addition to the work of implementing and bugfixing the move itself?

SubJunk:
Very interesting contribution, thank you

I'm ignorant about servlets, I have been reading about them for the last hour to try to understand, and it seems there are some benefits to them instead of our ad-hoc approach, but I am interested to hear what your thoughts are about their advantages for us @SurfaceS 

FYI my full time job is full stack developer working on a hybrid application with some older server-rendered pages, and newer RESTful APIs accompanied by SPAs, so I am generally more knowledgeable about web stuff than Java stuff.

I do wonder whether it might make more sense to use a RESTful approach instead of re-implementing/refactoring the server-rendered approach. I see the same group offers https://projects.eclipse.org/projects/ee4j.jaxrs

Interested in any thoughts

SurfaceS:
@Nadahar  I understand your point of view about adding new dependencies.
@SubJunk  No advantage as UMS webserver do not need more than HttpServer

SubJunk:
@SurfaceS this is a lot of work to throw away, are you sure?
I think we should at least keep the class refactoring, I like what you did with the `remote` directory changes

PR #2607, opened 2021-08-16
fix for reported problem on some SONY TV's while content browsing / searching

ik666:
This might solve the issue reported in: #2376. I hope the TV doesn't make consecutive search requests.

https://www.universalmediaserver.com/forum/viewtopic.php?f=9&t=14807&p=44349#p44349

PR #2781, opened 2022-01-11
fixed thread synchronization issues in MediaDatabase implementation

### LINKED ISSUE #2742: Unstable testIsFullyPlayed
This test is making master unstable and needs to be fixed. Unfortunately I can't reproduce the failure locally, so am debugging it by pushing commits to here. Any ideas would be helpful. https://github.com/UniversalMediaServer/UniversalMediaServer/runs/4611809555?check_suite_focus=true

```
Error:  Failures: 
Error:    TableFilesStatusTest.testIsFullyPlayed:103 
Expecting actual not to be null
```

ik666:
This PR should fix #2742: thread synchronization wasn't correctly implemented.

### DIFF
--- a/src/main/java/net/pms/database/Database.java
+++ b/src/main/java/net/pms/database/Database.java
@@ -137,7 +137,9 @@ public abstract class Database extends DatabaseHelper {
 	 * @throws SQLException
 	 */
 	public Connection getConnection() throws SQLException {
-		return cp.getConnection();
+		Connection c = cp.getConnection();
+		c.setTransactionIsolation(Connection.TRANSACTION_REPEATABLE_READ);
+		return c;
 	}
 
 	public int getActiveConnections() throws SQLException {
--- a/src/main/java/net/pms/database/MediaDatabase.java
+++ b/src/main/java/net/pms/database/MediaDatabase.java
@@ -23,6 +23,7 @@ import java.sql.*;
 import net.pms.PMS;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 import net.pms.dlna.RootFolder;
@@ -36,6 +37,8 @@ import net.pms.dlna.RootFolder;
 public class MediaDatabase extends Database {
 	private static final Logger LOGGER = LoggerFactory.getLogger(MediaDatabase.class);
 	private static final ReadWriteLock DATABASE_LOCK = new ReentrantReadWriteLock(true);
+	private static final Lock WRITE_LOCK = DATABASE_LOCK.writeLock();
+
 	public static final String DATABASE_NAME = "medias";
 	/**
 	 * Pointer to the instanciated MediaDatabase.
@@ -81,7 +84,8 @@ public class MediaDatabase extends Database {
 	 * @throws SQLException
 	 */
 	public final void checkTables(boolean force) throws SQLException {
-		synchronized (DATABASE_LOCK) {
+		try {
+			WRITE_LOCK.lock();
 			if (tablesChecked && !force) {
 				LOGGER.debug("Database tables have already been checked, aborting check");
 			} else {
@@ -122,6 +126,8 @@ public class MediaDatabase extends Database {
 				}
 				tablesChecked = true;
 			}
+		} finally {
+			WRITE_LOCK.unlock();
 		}
 	}
 
@@ -138,38 +144,48 @@ public class MediaDatabase extends Database {
 		checkTables(true);
 	}
 
-	public static synchronized void dropAllTables(Connection connection) {
-		dropAllTablesExceptFilesStatus(connection);
-		dropTableAndConstraint(connection, MediaTableRegexpRules.TABLE_NAME);
-		dropTableAndConstraint(connection, MediaTableFilesStatus.TABLE_NAME);
-		dropTableAndConstraint(connection, MediaTableMetadata.TABLE_NAME);
-		dropTableAndConstraint(connection, MediaTableFiles.TABLE_NAME);
-		dropTableAndConstraint(connection, MediaTableTablesVersions.TABLE_NAME);
+	public static void dropAllTables(Connection connection) {
+		try {
+			WRITE_LOCK.lock();
+			dropAllTablesExceptFilesStatus(connection);
+			dropTableAndConstraint(connection, MediaTableRegexpRules.TABLE_NAME);
+			dropTableAndConstraint(connection, MediaTableFilesStatus.TABLE_NAME);
+			dropTableAndConstraint(connection, MediaTableMetadata.TABLE_NAME);
+			dropTableAndConstraint(connection, MediaTableFiles.TABLE_NAME);
+			dropTableAndConstraint(connection, MediaTableTablesVersions.TABLE_NAME);
+		} finally {
+			WRITE_LOCK.unlock();
+		}
 	}
 
-	public static synchronized void dropAllTablesExceptFilesStatus(Connection connection) {
-		dropTableAndConstraint(connection, MediaTableMusicBrainzReleases.TABLE_NAME);
-		dropTableAndConstraint(connection, MediaTableCoverArtArchive.TABLE_NAME);
-		dropTableAndConstraint(connection, MediaTableThumbnails.TABLE_NAME);
+	public static void dropAllTablesExceptFilesStatus(Connection connection) {
+		try {
+			WRITE_LOCK.lock();
+			dropTableAndConstraint(connection, MediaTableMusicBrainzReleases.TABLE_NAME);
+			dropTableAndConstraint(connection, MediaTableCoverArtArchive.TABLE_NAME);
+			dropTableAndConstraint(connection, MediaTableThumbnails.TABLE_NAME);
 
-		dropTableAndConstraint(connection, MediaTableTVSeries.TABLE_NAME);
-		dropTableAndConstraint(connection, MediaTableFailedLookups.TABLE_NAME);
+			dropTableAndConstraint(connection, MediaTableTVSeries.TABLE_NAME);
+			dropTableAndConstraint(connection, MediaTableFailedLookups.TABLE_NAME);
 
-		// Video metadata tables
-		dropTableAndConstraint(connection, MediaTableVideoMetadataActors.TABLE_NAME);
-		dropTableAndConstraint(connection, MediaTableVideoMetadataAwards.TABLE_NAME);
-		dropTableAndConstraint(connection, MediaTableVideoMetadataCountries.TABLE_NAME);
-		dropTableAndConstraint(connection, MediaTableVideoMetadataDirectors.TABLE_NAME);
-		dropTableAndConstraint(connection, MediaTableVideoMetadataIMDbRating.TABLE_NAME);
-		dropTableAndConstraint(connection, MediaTableVideoMetadataGenres.TABLE_NAME);
-		dropTableAndConstraint(connection, MediaTableVideoMetadataPosters.TABLE_NAME);
-		dropTableAndConstraint(connection, MediaTableVideoMetadataProduction.TABLE_NAME);
-		dropTableAndConstraint(connection, MediaTableVideoMetadataRated.TABLE_NAME);
-		dropTableAndConstraint(connection, MediaTableVideoMetadataRatings.TABLE_NAME);
-		dropTableAndConstraint(connection, MediaTableVideoMetadataReleased.TABLE_NAME);
+			// Video metadata tables
+			dropTableAndConstraint(connection, MediaTableVideoMetadataActors.TABLE_NAME);
+			dropTableAndConstraint(connection, MediaTableVideoMetadataAwards.TABLE_NAME);
+			dropTableAndConstraint(connection, MediaTableVideoMetadataCountries.TABLE_NAME);
+			dropTableAndConstraint(connection, MediaTableVideoMetadataDirectors.TABLE_NAME);
+			dropTableAndConstraint(connection, MediaTableVideoMetadataIMDbRating.TABLE_NAME);
+			dropTableAndConstraint(connection, MediaTableVideoMetadataGenres.TABLE_NAME);
+			dropTableAndConstraint(connection, MediaTableVideoMetadataPosters.TABLE_NAME);
+			dropTableAndConstraint(connection, MediaTableVideoMetadataProduction.TABLE_NAME);
+			dropTableAndConstraint(connection, MediaTableVideoMetadataRated.TABLE_NAME);
+			dropTableAndConstraint(connection, MediaTableVideoMetadataRatings.TABLE_NAME);
+			dropTableAndConstraint(connection, MediaTableVideoMetadataReleased.TABLE_NAME);
 
-		// Audio Metadata
-		dropTableAndConstraint(connection, MediaTableAudiotracks.TABLE_NAME);
+			// Audio Metadata
+			dropTableAndConstraint(connection, MediaTableAudiotracks.TABLE_NAME);
+		} finally {
+			WRITE_LOCK.unlock();
+		}
 	}
 
 	/**
@@ -179,11 +195,14 @@ public class MediaDatabase extends Database {
 	 * @return {@link net.pms.database.MediaDatabase}
 	 */
 	public static MediaDatabase get() {
-		synchronized (DATABASE_LOCK) {
+		try {
+			WRITE_LOCK.lock();
 			if (instance == null) {
 				instance = new MediaDatabase();
 			}
 			return instance;
+		} finally {
+			WRITE_LOCK.unlock();
 		}
 	}
 
@@ -192,8 +211,11 @@ public class MediaDatabase extends Database {
 	 * Will initialize the database instance as needed.
 	 */
 	public static void init() {
-		synchronized (DATABASE_LOCK) {
+		try {
+			WRITE_LOCK.lock();
 			get().init(false);
+		} finally {
+			WRITE_LOCK.unlock();
 		}
 	}
 
@@ -203,8 +225,11 @@ public class MediaDatabase extends Database {
 	 * Will check all tables.
 	 */
 	public static void initForce() {
-		synchronized (DATABASE_LOCK) {
+		try {
+			WRITE_LOCK.lock();
 			get().init(true);
+		} finally {
+			WRITE_LOCK.unlock();
 		}
 	}
 
@@ -252,10 +277,13 @@ public class MediaDatabase extends Database {
 	 * @throws java.sql.SQLException
 	 */
 	public static void resetCache() throws SQLException {
-		synchronized (DATABASE_LOCK) {
+		try {
+			WRITE_LOCK.lock();
 			if (instance != null) {
 				instance.reInitTablesExceptFilesStatus();
 			}
+		} finally {
+			WRITE_LOCK.unlock();
 		}
 	}
 
@@ -274,10 +302,13 @@ public class MediaDatabase extends Database {
 	 * Shutdown the MediaDatabase database.
 	 */
 	public static void shutdown() {
-		synchronized (DATABASE_LOCK) {
+		try {
+			WRITE_LOCK.lock();
 			if (instance != null) {
 				instance.close();
 			}
+		} finally {
+			WRITE_LOCK.unlock();
 		}
 	}
 
--- a/src/main/java/net/pms/database/MediaTableFilesStatus.java
+++ b/src/main/java/net/pms/database/MediaTableFilesStatus.java
@@ -474,8 +474,6 @@ public final class MediaTableFilesStatus extends MediaTable {
 			) {
 				if (resultSet.next()) {
 					result = resultSet.getBoolean("ISFULLYPLAYED");
-				} else {
-					return false;
 				}
 			} finally {
 				TABLE_LOCK.readLock().unlock();
--- a/src/main/java/net/pms/database/MediaTableFilesStatus.java
+++ b/src/main/java/net/pms/database/MediaTableFilesStatus.java
@@ -24,8 +24,6 @@ import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
 import java.sql.Timestamp;
-import java.util.concurrent.locks.ReadWriteLock;
-import java.util.concurrent.locks.ReentrantReadWriteLock;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
@@ -53,7 +51,6 @@ public final class MediaTableFilesStatus extends MediaTable {
 	 * transaction locks. All access to this table must be guarded with this
 	 * lock. The lock allows parallel reads.
 	 */
-	private static final ReadWriteLock TABLE_LOCK = new ReentrantReadWriteLock();
 	private static final Logger LOGGER = LoggerFactory.getLogger(MediaTableFilesStatus.class);
 	public static final String TABLE_NAME = "FILES_STATUS";
 
@@ -72,28 +69,23 @@ public final class MediaTableFilesStatus extends MediaTable {
 	 * @throws SQLException
 	 */
 	protected static void checkTable(final Connection connection) throws SQLException {
-		TABLE_LOCK.writeLock().lock();
-		try {
-			if (tableExists(connection, TABLE_NAME)) {
-				Integer version = MediaTableTablesVersions.getTableVersion(connection, TABLE_NAME);
-				if (version != null) {
-					if (version < TABLE_VERSION) {
-						upgradeTable(connection, version);
-					} else if (version > TABLE_VERSION) {
-						LOGGER.warn(LOG_TABLE_NEWER_VERSION_DELETEDB, DATABASE_NAME, TABLE_NAME, DATABASE.getDatabaseFilename());
-					}
-				} else {
-					LOGGER.warn(LOG_TABLE_UNKNOWN_VERSION_RECREATE, DATABASE_NAME, TABLE_NAME);
-					dropTable(connection, TABLE_NAME);
-					createTable(connection);
-					MediaTableTablesVersions.setTableVersion(connection, TABLE_NAME, TABLE_VERSION);
+		if (tableExists(connection, TABLE_NAME)) {
+			Integer version = MediaTableTablesVersions.getTableVersion(connection, TABLE_NAME);
+			if (version != null) {
+				if (version < TABLE_VERSION) {
+					upgradeTable(connection, version);
+				} else if (version > TABLE_VERSION) {
+					LOGGER.warn(LOG_TABLE_NEWER_VERSION_DELETEDB, DATABASE_NAME, TABLE_NAME, DATABASE.getDatabaseFilename());
 				}
 			} else {
+				LOGGER.warn(LOG_TABLE_UNKNOWN_VERSION_RECREATE, DATABASE_NAME, TABLE_NAME);
+				dropTable(connection, TABLE_NAME);
 				createTable(connection);
 				MediaTableTablesVersions.setTableVersion(connection, TABLE_NAME, TABLE_VERSION);
 			}
-		} finally {
-			TABLE_LOCK.writeLock().unlock();
+		} else {
+			createTable(connection);
+			MediaTableTablesVersions.setTableVersion(connection, TABLE_NAME, TABLE_VERSION);
 		}
 	}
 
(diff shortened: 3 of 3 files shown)


ik666:
@Nadahar I fully agree with your statement, but I did actually exactly what you suggest. Method synchronization is done against the class, the synchronized(obj) is done against the object. I have thrown out all `synchronized` implementation and replaced them by a (from my point of view) correct ReadWriteLock implementation making sure _not_ to mix sync objects.

I have found another place, where thread syncing seems to be broken. Maybe this will help ...

ik666:
@Nadahar Seems like it is now working. Can you confirm? In regard to performance issues the read methods can use a read lock instead of a readWrite lock, but I personally think the are called so seldom, that it will have no effect.

Nadahar:
@ik666 I don't see what your latest change does at all. What is the point of storing the write lock from the `ReentrantReadWriteLock` in a separate field (`WRITE_LOCK`)? As far as I can see, this will do the exact same as before the change, just in a more confusing way.

ik666:
@Nadahar OMG, the the last commit did nothing, you're right. I was in a tunnel ...

ik666:
Strange though the test passed with that "fix" ... ??

ik666:
I see: `isFullyPlayed` seems to return a tri-state boolean on purpose ...

Nadahar:
> Strange though the test passed with that "fix" ... ??

@ik666 Not really. That's the think with concurrency issues - you can't rely on tests to catch them. It all comes down to nanosecond timing on the host that is running the code. It will vary from machine to machine and from run to run, depending on what other things are running on the machine at the exact time when the code is running. A typical "trap" is also that the code is tested using a single core - which means it will never happen. Just using dual cores will also reduce the chance that it will happen etc. Very minor changes to the code that changes the timing can affect whether it works or not at a particular run, but it's also quite random.

To have any decent chance to "catch" synchronization flaws using tests, you must run it with lots of threads spread over multiple cores and repeat the test lots and lots an times. Since "chance" is involved you might still falsely think that the code is safe, but if you run it say a million times with lots of threads and cores, at least you stand a fair chance to produce a situation where it fails.

In short, forget relying on tests for concurrency issues. You must simply make sure that the code is correct and leaves no room for error.

ik666:
@Nadahar  I think the last commit did it.

SubJunk:
@ik666 this is a good idea in general I think, thanks for working on it :) I can see you have identified some unlocked methods that might explain that race-condition.

I wonder if it would still work while using reentrantreadwritelocks, since then performance can be better.
I suggest trying that instead of making a new lock, since using a write lock on a popular SELECT query like isFullyPlayed could impact performance.

SubJunk:
@ik666 also you have been contributing great changes for a while now, would you like to be added to the UniversalMediaServer org so you can branch off master directly?

ik666:
@SubJunk There were no unlocked methods but

- (part 1) synchronization was done against different objects (first commit) -> I believe/hope you wont get the SQL exception you have referred [here](https://github.com/UniversalMediaServer/UniversalMediaServer/runs/4611809555?check_suite_focus=true) any more
- (part 2) switched from ReentrantReadWriteLock.readLock to ReentrantReadWriteLock.writeLock for the read operations. (last commit) -> `TableFilesStatusTest` class line 98 `MediaTableFilesStatus.isFullyPlayed(connection, "FileThatHasBeenPlayed")` should not deliver `NULL` any more.

For (part 2)  - in case this commit provides a stable solution - the good question is why does it work (`isFullyPlayed ` does not return NULL any more) ... My best guess is that changes to memory (RAM) between spawned threads (at least embedded H2DB thread against the other ones) are not synced ending up in a race condition. One can add an explicit memory barrier by entering a critical section (synchronized block) for the failing part. I've exemplary changed the access code (Test code) [in this branch](https://github.com/ik666/UniversalMediaServer/commit/9ce7b31d6cdf95ab7401d4596a5fe7be7f7be5f9). Wrapping the read part of the test code in a `synchronized` block seems also to work (build pipeline runs file). I'm aware that this is just an indication ...

ik666:
@SubJunk Have you been faced with a performance issue in the past for the read methods?

ik666:
@SubJunk Thanks for offering me repository access. It's no hassle for me using a forked repo, though it would be nice being a member :)

SubJunk:
@ik666 just when I thought I understood locks, it looks like I need to have a play around to understand better :) I'll do that in another branch for my own learning.
I don't understand why a writelock would be needed for a read operation unless there are other methods that aren't being accounted for.

Welcome to the team :) @UniversalMediaServer/developers meet our newest member

Nadahar:
@SubJunk "Read" and "write" locks can be a bit misleading. What they really are are shared and exclusive locks. A "read" lock can be acquired by an infinite (or at least huge) number of threads simultaneously and will succeed as long as the corresponding "write" lock isn't held by a thread. The "write" lock is exclusive and can only be held by one thread at a time. A "write" lock will be granted if another thread doesn't hold it already and no threads holds the corresponding "read" lock. Otherwise the acquiring thread will block until no other threads holds either a "write" or a "read" lock. As long as the "write" lock is held by a thread, attempts to acquire the corresponding "read" lock will block unless the thread that asks for the "read" lock also is the thread that holds the "write" lock, in which case it will succeed.

Acquiring a "write" lock while holding the corresponding "read" lock will always block indefinitely.

Using these "advanced" locks for a shared resource like the database might be problematic depending on the details of how the database works. If the database doesn't support multiple threads accessing it at the same time, it might not work to use the shared "read" lock even if you just read from the database, because a read operation might actually write to internal fields in the database classes.

Without knowing the details of the database implementation very well, it's impossible to say if this is "safe" or not. I know that I implemented the "read"/"write" scheme for the database originally, it was a result of trial and error only. It seemed to work when I did it, so I chose to use it to allow concurrent reads. It might still be situations where it doesn't work, or the H2 code might have changed since then so that it no longer works. The only thing one can do to be completely "safe" is to use an exclusive lock for all access to the database. If I were to implement that, I would use `synchronized` and forget the "lock" objects, since `synchronized` is implemented much more effectively in Java (just look at some benchmarks where people compare the locking schemes).

Ideally, the database should handle multithreading and there should be no need to "protect" it with locks. Maybe it does now..? Regardless, that's the way it should work, where the database manages record or table level locks. This is the way it is done with "proper" databases, but from what I remember this was not fully implemented in H2 at the time when I made it.

SubJunk:
@Nadahar that description matches my understanding as well, so either we are both wrong or something isn't working as I expect it to

About a year ago I fixed a longstanding bug where if you browsed a folder during a scan, it would either never succeed or at least take a very long time, and for that I found there were performance benefits to selectively using read locks instead of always using a full lock

ik666:
@Nadahar I think SubJunk would like to keep concurrent read operations.

@SubJunk Even though we seem to share a common understanding of what locks do, we're faced with failing tests ... For me this looks like a "Happens-before Order" problem and that's why I was guessing there is an issue with the readLock implementation.

Could you try to execute the test pipeline with added JVM parameter : `-XX:+UseMembar` (not in this branch)?

Nadahar:
@SubJunk I think I have a pretty good understanding of how the locks work, but I might not explain myself well. I'm not exactly sure where I'm being unclear, so I'll just try to specify some further detail around the things I find the most likely.

Regarding "performance", there are multiple concepts at play. Thread contention is what can really make performance bad, it is a result of threads waiting for each other, potentially creating bottlenecks and drastically reduced throughput.

But, there's also a performance hit associated with the locks themselves in at least two ways. One is the fact that they will introduce memory barriers that forces things to happen in a specific order, which prevents the normal "optimization" taking place where things are executed in the order that is most efficient for the CPU. It also affects complicated CPU stuff like cache hit/misses, cache synchronization etc. that you really must be a CPU expert for the architecture in question to fully understand. This "cost" isn't really related to Java but applied regardless of language, although some languages will allow you to control the details better so that you might limit the impact. Java doesn't give us much low-level control, as always. 

In addition, there's the cost of the lock implementation itself. This is the Java part, and the part I was talking about above. I don't know exactly why, but `synchronized` is more effective than an exclusive `Lock` implementation. I'm not exactly sure why that is, but I've seen benchmarks that show that the difference is relatively substantial. I guess some of the reason is that `synchronized` is "native", it's handled in the JVM - while the `Lock` implementations are Java classes. Another reason is that I think they have done quite a lot of optimization in the JVM for `synchronized` and I've read that if the runtime will skip doing the `synchronized` if it deems that it's not needed at that time. There's also the obvious reason that for example `ReentrantReadWriteLock` is substantially more complex than a simple exclusive lock like `synchronized`. Most of the `Lock` implementations supports "fairness" for example, where the thread that has waited the longest will get the lock if it's contested. `synchronized` does no such thing, it simply give access to one of the waiting threads.

So, what I was trying to say above was that *if* you only use exclusive locks without "fairness" (like you effectively do if you only use the "write" lock from `ReentrantReadWriteLock` without enabling fairness), it is more efficient to use `synchronized`. If there are more reads than writes, so that you reduce the contention by allowing multiple threads to read in parallel, you obviously gain much more by using a shared read lock. If that wasn't the case, there would be no reason to ever use for example `ReentrantReadWriteLock`.

The other point I was trying to make is that we don't really know what happens "behind the scenes" in H2. So, even though you strictly only reads from the database, it doesn't mean that something isn't written when you do that. It could be anything, optimization information, statistics, performance counters - you name it. It's actually quite likely that H2 does some kind of write operation (not necessarily to the database, but to memory) even though you just do a `SELECT`. If this is the case, it might not be "safe" to let multiple threads do multiple `SELECT`s in parallel, so a shared lock might not be safe to use - despite it resulting better performance when you're "lucky" and nothing bad happens as a result of some race condition.

There are other possible explanation for your observation though, it might also be that you experienced a deadlock in some circumstances, and by switching the a shared lock you circumvented the deadlock situation. This would best explain why it would hang "sometimes" IMO.

The bottom line is that what is the best implementation really depends on H2 - and knowing the details of how it works. H2's thread handling has always been a bit obscure to me, some places it says that it's thread-safe, while the opposite is indicated other places. This also depends on the connection mode (embedded or client/server) etc, so I'm just not sure what is and isn't safe. H2 claims that each connection is synchronized internally, which would make concurrent operation impossible even if you use a shared lock "outside". From what I understand, you need to use multiple connections to achieve concurrent operations. H2 might have changed since I originally made the decision to protect access with locks too.

ik666:
> I don't understand why a writelock would be needed for a read operation unless there are other methods that aren't being accounted for.

@SubJunk The answer might a bit tricky. I'm referring to the Java Memory Model [java docs](https://docs.oracle.com/javase/specs/jls/se8/html/jls-17.html) Chapter 17.4.

See 17.4.5:
> More specifically, if two actions share a happens-before relationship, they do not necessarily have to appear to have happened in that order to any code with which they do not share a happens-before relationship. Writes in one thread that are in a data race with reads in another thread may, for example, appear to occur out of order to those reads.

I believe `MediaTableFilesStatus` doesn't share a happens before relationship with the h2db thread, it's creating a new monitor (ReentrantReadWriteLock). 

Since `MediaTableFilesStatus` doesn't have any shared member variables (all member are final) the class itself is already tread safe and needs the lock only for the memory effect (to synchronize memory against other threads). I think the ReentrantReadWriteLock guarantees the memory effect only for `writeLock` operations. 

This would also mean, the read operations do not need to be locked at all.

ik666:
@SubJunk Regarding your performance concerns, I've refactored the `MediaTableFilesStatus ` class:
- Read operations do not block any more
- Write operations do sync memory

SubJunk:
@ik666 if this is correct, it's nice to be able to remove those locks in the table. I'm learning a lot about Java.

I wonder if we should make the same changes to other tables. There isn't testing for other tables, so they might have the same bug it wouldn't get picked up.

ik666:
@SubJunk Please let's further monitor this. I'd like to think more about it and possibly discuss my ideas (maybe with @Nadahar ?) ... 

Actually, I am a little unhappy with my answer, because it somehow implies, the `ReentrantReadWriteLock` implementation is buggy. The `ReentrantReadWriteLock.writeLock` in the read methods seemed to be necessary, only for the "happens before relationship" to work correctly. For my understanding, exiting the `ReentrantReadWriteLock.writeLock` monitor should have already done the memory sync (but it didn't ?). That's why I asked you for applying the `-XX:+UseMembar` JVM parameter, which notifies the compiler to use native memory barriers instead of an "performance optimized" version.

I'm pretty sure though, that `synchronized` sections do have a working memory sync effect.

SubJunk:
@ik666 ok cool, yes I'm happy not to move too quickly on this until we understand it
I'm not sure how to run the tests with extra flags on GitHub Actions, I tried it here but I'm not sure if that's the right way https://github.com/UniversalMediaServer/UniversalMediaServer/pull/2786
I got the solution from https://maven.apache.org/docs/3.3.1/release-notes.html#JVM_and_Command_Line_Options

Nadahar:
@ik666 I don't follow your thinking, I don't think `-XX:+UseMembar` has anything to do with this. As far as I can tell, there was a bug in Java 5 and 6 (when the "optimization" was relatively new) that could make it fail in some situations, and this option lets you fall back to the old, slower non-optimized handling. I don't think this problem exists anymore. In any case, I think this optimization only applies to `synchronized`, not the `Lock` implementations.

I also doubt that there are bugs in `ReentrantReadWriteLock` at this time, it has been thoroughly battle-tested by now (most of the concurrency classes came with Java 5 AFAICR).

I'm not sure exactly what goes wrong here, since I can't reproduce it, but I took a quick look at the H2 code that is called by `isFullyPlayed()` here:
```java
	TABLE_LOCK.readLock().lock();
	try (
		Statement statement = connection.createStatement();
		ResultSet resultSet = statement.executeQuery(query)
	) {
		if (resultSet.next()) {
			result = resultSet.getBoolean("ISFULLYPLAYED");
		}
	} finally {
		TABLE_LOCK.readLock().unlock();
	}
```

Digging into `createStatement()`, the code is:
```java
    public Statement createStatement() throws SQLException {
        try {
            int id = getNextId(TraceObject.STATEMENT);
            if (isDebugEnabled()) {
                debugCodeAssign("Statement", TraceObject.STATEMENT, id, "createStatement()");
            }
            checkClosed();
            return new JdbcStatement(this, id,
                    ResultSet.TYPE_FORWARD_ONLY,
                    Constants.DEFAULT_RESULT_SET_CONCURRENCY, false);
        } catch (Exception e) {
            throw logAndConvert(e);
        }
    }
```

The JavaDoc for `JdbcConnection` states that:
> Thread safety: the connection is thread-safe, because access
> is synchronized. However, for compatibility with other databases, a
> connection should only be used in one thread at any time.

This doesn't ring true to me, I already "smell" a simplified understanding of thread-safety when reading it. Following the very first call to `getNextId()` confirms my suspicion. The method looks like this:
```java
    protected static int getNextId(int type) {
        return ID[type]++;
    }
```
The definition of `ID` is this:
```java
    private static final int[] ID = new int[LAST];
```
This isn't thread-safe. `ID` isn't protected, and it increments the ID counter for the type in question. Even making `ID` `volatile` would have been enough, since the `++` operation is really `n = n + 1` this consists of one read and one write of the same variable. `n` could have been changed by another thread between the point where it is read and where it is written, making the IDs out of order or potentially duplicated (two threads can read the same value before it is incremented so that they end up with the same ID).

This means, as far as I can tell, that using a shared lock is NOT safe, as I suspected. This goes for all database access using H2, such basic concurrency flaws as the above means that none of the code should be trusted to be thread-safe. The solution, as far as I can tell, is to get rid of all `ReentrantReadWriteLock`s guarding database access and use `synchronized` instead. It will essentially have the same effect as "using just the write-lock" from `ReentrantReadWriteLock` but be cleaner and faster.

SubJunk:
Another angle to consider is that we have had this test for almost 4 years and I never saw it fail until the last couple of months, so what changed?

SubJunk:
The first instance I can see of it is at https://github.com/UniversalMediaServer/UniversalMediaServer/runs/4042558266?check_suite_focus=true so I think it appeared shortly before that

Nadahar:
> Another angle to consider is that we have had this test for almost 4 years and I never saw it fail until the last couple of months, so what changed?

Haven't you been updating H2 lately? Maybe something changed internally in H2 that affects the timing and "reveals" the problem? It could also be seemingly unrelated changed in UMS that affects the timing and makes it more likely to go wrong.

SubJunk:
I think it started happening before the h2 upgrades but I’ll have a look. I’ll run some of the old code

SubJunk:
Another unexplained thing is it seems to only happen on the Windows pipeline

Nadahar:
I don't consider these things "unexplained", it's in the nature of thread safety. You can run unsafe code thousands or millions of times with nothing bad happening, or it could fail every other time. It depends on so many things, to really "pin it down" you'd need to study exactly what happens in what order on the CPU in question on a CPU instruction level. It's in the nature of the problem, if multiple threads doesn't happen to manipulate the same variable/field/object at the exact "wrong" time, nothing bad happens. I have a completely different philosophy: Relying on "luck" isn't an option, a potential thread issue/race condition/concurrent access is a bug even if you can never "prove" that it happens. The fact that it CAN happen means that it most likely will at some stage, and it's probably not while you're debugging or giving the issue attention.

This whole "mystery" could be due to changes in the VMs that runs the tests. Maybe the Windows runners are the only ones that use multiple cores - maybe they didn't use to in the past.

Remember, you can never experience a concurrency issue if only one core is in use. Even though we think if threads running in parallel, they aren't actually doing that within the same core. Threads are scheduled time slots in the CPU scheduler and runs only for the time the scheduler has given them. With a single core, no two threads will *actually* run at the same time. I believe this is one reason why many developers didn't seem to grasp why they should care about thread safety some years back. Many had single core Pentium CPU's etc, so they never actually experienced it and probably concluded that it was all about nothing. These days, as the number of cores increase, these problems happen more and more frequently.

Again, to me, the only solution is to make sure that it cannot happen. I don't care to try to figure out "why" it does or doesn't happen in a given scenario if I know that it can. Make it impossible, and it will never happen ;)

SubJunk:
It first appeared here https://github.com/UniversalMediaServer/UniversalMediaServer/pull/2607
I suspect it is related to the changes to the instances in the tests. In my testing I have found there is some leakage between tests, which means the order they run in can affect the outcome, which isn't good.
So this bug may not be one that users can experience - it may only happen in a unit-testing scenario. That's just a maybe though, it needs to be dissected more.

ik666:
>Again, to me, the only solution is to make sure that it cannot happen. I don't care to try to figure out "why" it does or doesn't happen in a given scenario if I know that it can. Make it impossible, and it will never happen ;)

@Nadahar This is the right motivation :) However, finding the original cause of the issue, makes it easier to circumvent it (or not to make it happen). 

I thought about this failing test again and I'm unsure, if we're on the right track, because:
- The test case is single threaded, or at least the same thread that sets `fullyPlayed` also reads `fullyPlayed`. Because of same thread, setting the value happens before reading it.
- Saying the above, any locks (synchronized or other types) in the class `MediaTableFilesStatus` have no effect, since no other threads will acquire them. 
- Even if multiple threads access `MediaTableFilesStatus` still the class is thread save, because no internal member variables are being accessed. I think any synchronization in the class `MediaTableFilesStatus` has finally no effect.
- I assume that the locks were placed to help h2db to be thread save? However, for this to work h2db and `MediaTableFilesStatus` need to share the same lock (java spec 17.4), which is hard to realize (I think, Nadahar, you already mentions this). Again, I think the locks can be removed.

Another possible cause I can think of are database transaction isolation level to be set to optimistic. I think the default isolation level is `ReadCommitted` which is open to "lost updates". Setting isolation level to "Repeatable Read" would prevent lost updates.

Nadahar:
@SubJunk 
> It first appeared here #2607 I suspect it is related to the changes to the instances in the tests. In my testing I have found there is some leakage between tests, which means the order they run in can affect the outcome, which isn't good. So this bug may not be one that users can experience - it may only happen in a unit-testing scenario. That's just a maybe though, it needs to be dissected more.

To me, this just looks more and more like a "traditional" thread safety issue after you found #2607 to be what started it. Remember my comment there? https://github.com/UniversalMediaServer/UniversalMediaServer/pull/2607#issuecomment-944965356

Just tweaking things so that the timing doesn't bite you when you run the test isn't a solution. I'm pretty sure it *can* happen to users too, and I'm pretty sure that it will be very difficult to figure out what's the cause because it won't be reproducible.

@ik666 

> * The test case is single threaded, or at least the same thread that sets fullyPlayed also reads fullyPlayed. Because of same thread, setting the value happens before reading it.
> * Saying the above, any locks (synchronized or other types) in the class MediaTableFilesStatus have no effect, since no other threads will acquire them.

Your logic is sound if it is true. I do however remember having similar issues in the past, and I think it's because Surefire does some strange things. I don't remember the details, but I remember that I had to use some parameter to force Surefire not to "fork" at some point in the past. Surefire won't run the tests in a predictable order, and it will reuse the forks (https://stackoverflow.com/a/2288104) which will mean that sometimes the "state" left by one test will impact another. It might be that this is what happens, not a race condition. There is an explanation of how this is supposed to work here: https://maven.apache.org/surefire/maven-surefire-plugin/examples/fork-options-and-parallel-execution.html

> * Even if multiple threads access MediaTableFilesStatus still the class is thread safe, because no internal member variables are being accessed. I think any synchronization in the class MediaTableFilesStatus has finally no effect.

Here I disagree. It's not just about what happens in that class, it's about what happens in all code that is called by that class too, and since the for example `connection.createStatement()` isn't thread-safe, it must be protected against being run by multiple threads at the same time. A shared lock won't do though, it must be exclusive since class fields are being modified also for "read operations". This might not have to do with why the test fails, but it's still a situations where bugs can occur.

> * I assume that the locks were placed to help h2db to be thread safe? However, for this to work h2db and MediaTableFilesStatus need to share the same lock (java spec 17.4), which is hard to realize (Nadahar, you already mentions this). Again, most likely the locks can be removed.

I think this is a misunderstanding. It is true that you must use the same lock to protect an object in all locations, but when using locks the way it was intended here (except the mistake of using shared locks for reads) it will make sure that no more than one thread will work with the `connection` at any particular time. That will still work to protect H2 from lacking thread safety, simply because we make sure that only one thread at a time can access it. It do however agree that the locking scheme might be pointless, I don't remember the details now, but if there are different locks protecting the same `connection` in different classes, it is of course futile. I think maybe this was made to try to "help" H2 with table transaction protection. This is, as you said, probably pointless, and the locks as they are now probably does nothing good. I wrote what eventually became the code in questions many years ago, and I don't remember all my rationales at the time. What I do remember was that it was very hard to get to the bottom of what was and wasn't safe with H2, both when it came to thread safety and "transaction safety".

> Another possible cause I can think of are database transaction isolation level to be set to optimistic. I think the default isolation level is ReadCommitted which is open to "lost updates". Setting isolation level to "Repeatable Read" would prevent lost updates.

I think most people use `optimistic` because there's quite the performance hit when not using it? Anyway, I don't thing transaction isolation is relevant here, as only one `connection` is in use. I'm pretty sure that H2s transaction stuff only apply between different connections, and that each connection can only handle one thing at a time. That's probably why connections have been "synchronized" by the H2 author, even though there are obvious holes in this protection.

SubJunk:
So I guess we have two separate types of problems to look at now:
- We need more robust thread-safety with database operations
- We need more robust setup/teardown in tests (I'm working on this)

@ik666 I have a copy of this branch that runs tests 100x per commit at https://github.com/UniversalMediaServer/UniversalMediaServer/compare/test-thread-safety-fullyplayed?expand=1 and even after changing the readLocks to writeLocks I can't get it to be stable

ik666:
@Nadahar Are you sure, you're having analyzed the correct h2db driver? I've looked at the current code in jar `h2-2.0.206.jar` 

```

      @Override
    public Statement createStatement() throws SQLException {
        try {
            int id = getNextId(TraceObject.STATEMENT);
            debugCodeAssign("Statement", TraceObject.STATEMENT, id, "createStatement()");
            checkClosed();
            return new JdbcStatement(this, id, ResultSet.TYPE_FORWARD_ONLY, Constants.DEFAULT_RESULT_SET_CONCURRENCY);
        } catch (Exception e) {
            throw logAndConvert(e);
        }
    }
```

and 

```

    protected static int getNextId(int type) {
        return ID.getAndIncrement(type);
    }
```

ID is an AtomicInteger ... so it is thread safe?

ik666:
@SubJunk That's what I meant. I believe the whole locking thing has absolutely no effect (or at least not reproducible side effects) und can be removed.

By now I think all code runs as expected. The test case fails because of the too optimistic insolation level. I changed the DB transaction level in your `test-thread-safety-fullyplayed` branch. Looks stable now.

ik666:
I removed all locks, branch still seems to pass all 100x test ...

ik666:
Last commit in the other branch was fixing an initialization glitch of singleton member in class MediaDatabase

SubJunk:
@ik666 nice work, this is a good discovery! Although it seems that `TRANSACTION_REPEATABLE_READ` flag might not be needed if we fix the instantiation problems. So if I/we can fix that, we can then decide whether it's better to keep the flag anyway

@Nadahar this is the branch we have been playing with most recently https://github.com/UniversalMediaServer/UniversalMediaServer/compare/test-thread-safety-fullyplayed?expand=1

ik666:
>I'm still not convinced that the optimistic locking is the cause though. Are you sure it wasn't caused by the static initialization problem instead?

@Nadahar Could you pinpoint the static code, you have in mind? However, I am pretty sure it's a database isolation (locking) issue. I've committed an alternative SQL row locking in my last commit in the other test branch. The manual row locking should also do the desired transaction isolation, let's see ... the test runs around 1 hour :)

ik666:
The "alternative implementation" didn't work. Test failed. I reverted back ...

Nadahar:
@ik666 
> Could you pinpoint the static code, you have in mind?

I have nothing in particular in mind, I just know from experience that there are situations where the order of the tests matter because of static "state" that is "inherited" from one test to another because the same thread is re-used. UMS has "a lot of things" that must be initialized in the "normal order" before everything works as intended, and some times the unpredictable order in which tests are run can bite you in this regard. The `PMS` instance is static, the `PmsConfiguration` is static, and lots of other things are as well.

ik666:
@Nadahar ok, I see what you mean. However one have to keep in mind that the static PMS instance is actually the real world scenario. IMO the test could have more significance if they were executed in parallel imitating multiple client access. 

One thought : Maybe, a test order that would guarantee a "first test" to run which "sets up" a specific PMS configuration that fits the later called tests might help?

ik666:
@SubJunk I wish us the best, I hope you're not chasing a ghost.

Nadahar:
@ik666 I'm not saying how it should or should not be, the way it works today has some obvious challenges - but restructuring it to make testing "easier" isn't trivial either. It isn't straight-forward for tests to "share" the static instances, tests can not rely on any particular configuration setting, and they should work with "mock" configuration instances so that they're completely independent from configuration files on the computer running the tests. Some tests modify the configuration to have a predictable situation as far as I remember.

The reason something fails on CIs but works locally could for example be that something relies on some configuration setting. The VMs running the tests won't have a configuration file, while development computers usually do. But, it could also be due to the order in which they are run and the way that affects the "state" the previous test leaves the static resources.

Although the `PmsConfiguration` is the most "obvious" potential issue, there are plenty of other static resources that could cause problems. I have no idea if this is the case here though, I haven't dug into this exact failure. I'm just saying that I've had these problems before.

When things are inconsistent and not reproducible on local development computers, it does give you some hints about the most likely culprit though. Things like concurrency and locking are obvious suspects, which is why we have discussed them. If that turns out not to be the problem, the static resources are my "second best guess" - but it's still a guess.

I'm not saying that the database transaction control can be ruled out, I just find it a bit strange if this should turn out to be the problem. The reason is that UMS' database use is extremely simple compared to what databases often do, and my impression is that optimistic locking usually works well in much more complex scenarios. It's still obviously a possibility though. Another thing is that I didn't think UMS used transactions. Isn't "autocommit" usually left on so that every operation commits? Is rollbacks implemented at all? I would think that such simple circumstances wouldn't be much of a "challenge" for the default locking scheme. On top of that, there are locks (both in UMS and H2) that protects against many concurrent operations. Does UMS use multiple connections at all?

SubJunk:
@ik666 thanks for the luck! I have some ideas to try, I just need to make the time to do them now. I'll keep this thread updated

SubJunk:
I did the setup/teardown change in https://github.com/UniversalMediaServer/UniversalMediaServer/pull/2803 and that seems to have fixed it, It passed 601 times in a row

There might still be improvements from these other branches we should keep though

SubJunk:
I have merged #2803 so now the test is stable

Now we need to decide whether we should we remove the ReentrantReadWriteLock implementations from database tables. Any thoughts?

Nadahar:
> Now we need to decide whether we should we remove the ReentrantReadWriteLock implementations from database tables. Any thoughts?

I'd say it's hard to see how they could be useful if H2 has now fixed their implementation. Shared locks didn't really cut it with the old H2 implementation either because "read" operations caused "writes" to the ID's for example. All in all, I can't really see what usefulness they could represent as the situation is now.

ik666:
I agree with Nadahar.

SubJunk:
That's done in https://github.com/UniversalMediaServer/UniversalMediaServer/pull/2804
Can you please review @ik666 ?

ik666:
@SubJunk  Yes, done, though I still have an uncomfortable feeling (code smells) I'd like to share with you guys ... 

I'm not a real fan of killing the running process in #2803 just for getting the test green:

```
	@AfterAll
	public static final void tearDown() throws ConfigurationException, InterruptedException {
		if (PMS.getConfiguration().isRunSingleInstance()) {
			PMS.killOld();
		}
	}
```

As much as I can follow the strategy a test should not interface with other tests, forcing a process kill is a big club, making them somewhat artificial. Some test do a more complex setup, but I don't see any real differences between them yet. Therefore, I question myself why or why not should a test fail with the given configuration below?

Actually, some PMS methods seem to serve only tests like `PMS.getNewInstance()` and `PMS.killOld()` (I'm not a fan of this either). And regarding the usage: Why are those methods called in the last block? Won't this revert all of the configuration setup that has been made before?

```
	@BeforeAll
	public static final void setUp() throws ConfigurationException, InterruptedException {
		LoggerContext context = (LoggerContext) LoggerFactory.getILoggerFactory();
		context.getLogger(Logger.ROOT_LOGGER_NAME).setLevel(Level.INFO);
		PMS.forceHeadless();
		try {
			PMS.setConfiguration(new PmsConfiguration(false));
		} catch (Exception ex) {
			throw new AssertionError(ex);
		}
		assert PMS.getConfiguration() != null;
		PMS.getConfiguration().setAutomaticMaximumBitrate(false); // do not test the network speed.
		PMS.getConfiguration().setSharedFolders(null);
		PMS.getConfiguration().setScanSharedFoldersOnStartup(false);
		PMS.getConfiguration().setUseCache(false);

		Services.create();

		try {
			PMS.getConfiguration().initCred();
		} catch (Exception ex) {
			LOG.warn("Failed to write credentials configuration", ex);
		}

		if (PMS.getConfiguration().isRunSingleInstance()) {
			PMS.killOld();
		}

		// Create a new PMS instance
		PMS.getNewInstance();
	}
```

And even if this is all correct, why should this help to solve the database concurrency issue other than side effects?

Regarding the question, which database isolation level to use, the proposed `TRANSACTION_REPEATABLE_READ` will prevent "lost updates". It is of course arguable if the UMS systems need this guarantee.

SubJunk:
@ik666 good points.
One thing I realised while looking at this is that our tests folder has a mix of integration tests and unit tests, while usually they would be split into different folders and have different runners (`surefire` would run unit tests, and `failsafe` would run integration tests).
If we do make that split, we can have before and after hooks in failsafe that do the setup and teardown before and after the whole integration suite instead of every test. That's more efficient and removes noise from test files.

And yes potentially that does remove the configurations passed in. I only focused on test stability here, not on the setup steps which are the same as they were before, I just merged them together.

SubJunk:
Can this be closed now @ik666 ?

ik666:
Yes, let's close it ...

PR #2784, opened 2022-01-13
Playing around

SubJunk:


SubJunk:
I reran this second commit over ten times and had no failures so it seems that logging itself causes it to succeed

PR #2789, opened 2022-01-15
Cling to JUPnP

### LINKED ISSUE #2462: Try JUPnP
Cling is unsupported and doesn't work well with JRE 11+. JUPnP has some fixes related to that, but still only claims to support JRE 8. It may be closer to what we want though, and because it is maintained, we can submit code changes to them instead of creating our own fork of Cling, which is what I've been doing.

This is a path to being able to fix bugs related to UPnP, so I am marking it as a bug instead of enhancement

SurfaceS:
This PR should fix #2462.


SurfaceS:
Normally, Cling and JUPnP are build to run the all UPnP services on it.
Here, the config is really special. So it may need tweak.

This PR will allow to remove cling dependency, that's the first goal that allow UMS to move to another java version >8.

But in my mind, I think that JUPnP should at least be set to handle all the datagrams messages by creating LocalDevices and Services that notify the url of the actual http MediaServer (remove UMS own datagram service (alive/notify)).
And maybe it can handle the whole services (http/serving) on it's own engine, but that another goal.
A lot of things to do...

SubJunk:
This is great, thank you. I’ll do some testing over the next couple of days.
The comments about using more of the cling/jupnp features are great too. I support that.

There was some reason we did our own discovery messages, I can’t remember since it was so long ago but it involved reading through Cling GitHub and finding they didn’t do something we wanted. I guess we will either rediscover that or find it fixed if we try again.

SubJunk:
@SurfaceS I've been testing this and can see a lot of errors in the console:

```
Subscription failed, invalid or missing (SID, Timeout) response headers
```

SurfaceS:
Seems that JUPnP is more verbose that Cling.
They change from java.util.logging to org.slf4j API, that's maybe the reason.
Also they changed alot things under the wood.

Our Cling implementation is not compatible with java 8+ as it use the hacked (by Cling) JDK httpclient.
I think that replacing this for the apache httpclient will allow us to not be linked to java 8, but we still be linked to the old Cling maven.

SubJunk:
@SurfaceS no problem, do you want to keep working on JUPnP or should we close this? What would you like to do?

SurfaceS:
After tests, current cling version does not log on our slf4j logger.
That's why you see a lot message now, whit JUPnP that use slf4j as logger.

If you want, I will push on another pr a cling that use apache HttpClient instead of the hacked HttpConnection (that doesn't work on java >8). That can be a first step. I'll not use slf4j logger for the apache impl, but the java one like cling did.

wborn:
> Seems that JUPnP is more verbose that Cling.
> They change from java.util.logging to org.slf4j API, that's maybe the reason.

jUPnP was using both JUL and SLF4J. :upside_down_face: 
Now it is using only SLF4J after https://github.com/jupnp/jupnp/pull/162.

SubJunk:
@wborn great to see some active development on that project, nice work!

wborn:
I now see this project did switch over to jupnp after another try with #2806. :slightly_smiling_face:

SubJunk:
Yes it's working well. Excited to see where you are taking it in v3

PR #2792, opened 2022-01-17
Fix NPE for possibly slow remote urls in DLNAResource.

Premik:
For me this NPE was manifesting when using Web content or `m3u` playlists with urls to my `tvheadned`. 
Randomly some tv channels were not showing up in the `dlna` clients giving only empty folders. While it was working ok over the web. After this change it seems to be stable so far.


SubJunk (approved):

SubJunk:
@Premik thank you!

PR #2793, opened 2022-01-17
Sse implementation on web interface

SurfaceS:
Implemented Server-sent events to the web interface instead of ajax pull if the browser support it.
In short : ajax = client request, sse = server sent

With SSE, the server keep the connection stream opened to send data when needed, instead of pulling each seconds (meaning recreate an http request handle each seconds).

Also, the web interface media player could be ajaxed to avoid changing page on browse.


SubJunk:
@SurfaceS another great contribution, you're on a roll. Thanks! I'll do some testing ASAP

src/main/java/net/pms/network/webinterfaceserver/ServerSentEvents.java
```diff
@@ -0,0 +1,107 @@
...
+	public ServerSentEvents() {
+	}
+
+	public ServerSentEvents(OutputStream os) {
+		addEventStream(os);
+	}
+
+	OutputStream os;
+	public final void addEventStream(OutputStream os) {
+		//clean current OutputStream in case of....
```
SubJunk, line 24: Is this comment incomplete?

SurfaceS: It's a reminder note for me.
In the previous implementation, pull allow only one control connection (it's normal).
If you open 2 tabs on your explorer, then 2 connections call pull, the first that call will have the data, the other call will be empty.
I'll need to manage that to allow only 1 WebRender by explorer, at least for sse.
For now, it work like before.

SurfaceS: I added a message when multiple tabs browse the interface to let user now it.
fixed some html issues that was present.

SubJunk:
Good idea. I'm hoping to test this today

src/main/java/net/pms/configuration/WebRender.java
```diff
@@ -541,21 +542,44 @@ public void setSubLang(String s) {
...
+				push.clear();
+			}
 		}
 		return json;
 	}
 
+	private ServerSentEvents sse;
+	public void addServerSentEvents(ServerSentEvents sse) {
+		if (this.sse != null && this.sse.isOpened()) {
+			this.sse.sendMessage(gson.toJson(new String[] {"close", "warn", "Another tab is browsing the interface"}));
```
SubJunk, line 569: @SurfaceS The message "another tab is browsing the web interface" works but clicking "OK" closes the current tab, that might be unexpected for a user because the text doesn't mention that.

Let's change it to e.g.
`The web interface is open in another tab which can cause problems. We recommend you close this tab" then the button says "Close this tab" and then another button says "Ignore"

Also let's make this translatable because it is user-facing

SurfaceS: Translatable now.
I removed the `windows.close()` call as it doesn't work on many web browser (modern ones won't allow scripts to close tab).

SubJunk:
@SurfaceS those changes look good, thanks. What are the problems that can happen if a user is running two tabs? Are those problems introduced in this branch or is it an existing issue?

SurfaceS:
@SubJunk 
_Are those problems introduced in this branch or is it an existing issue?_
It's an existing one. SSE handle better that issue by showing it to the user.

`What are the problems that can happen if a user is running two tabs?`
**Short answer :**
In SSE mode, the last tab that browse take the control, the others one left it.
In pull mode, all tabs share the control, but only one have it at a time, causing mistakes as you never know the one that will handle the request (depend on the one that ask first after UMS store event).

**Long answer :**
When a web browser reach the server, UMS create a WebRenderer for it, and give him an id stored in a cookie (HTTP Header).
But cookies are shared between tabs/windows of a web browser.
So, when you open a new tab, UMS think you are the same renderer, and link you to the other tab WebRenderer.
With pull, you cannot know the tab that request the pull.
With SSE mode, you have a socket opened between the webpage and UMS.
So if it's already opened for a WebRenderer, you know that's not a normal situation, and let the current socket know it should close.

SubJunk:
Thanks for explaining

`In pull mode, all tabs share the control, but only one have it at a time, causing mistakes`
Can you give me an example of a mistake? It will help my understanding as well as the changelog

SurfaceS:
The goal of pull, and now sse, is to remote control the WebPlayer from the app gui.
From the server side, as the WebPlayer is recognized from ip/cookie, you can't know if it come from tab1 or tab2 as they will be both the same WebPlayer.

This PR mainly add the SSE (server-sent events). It's real time, and works on most browser.
Only few web browsers (Internet Explorer and Deno) doesn't support it. They will use the old pull mode.

From the gui, when you remote control a WebPlayer : 
- In SSE mode, the command is sent in real time to the opened socket.
- In pull mode, the command is stored on a list, then the first call to pull will give that list.

If you open 2 tabs on the browser : 
- In SSE mode, the server will inform the tab that have the SSE socket opened (tab1) that another tab take the SSE socket (tab2).
No more remote control for tab1, tab2 is remote controlled. So no mistake.
- In pull mode, each tab run a timer that call the http server each second to check if a command is waiting for them.
So you can't predict the one that will have the command. The first tab that call after the command was stored will have it, it can be tab1 or tab2. And that reproduce on each command. So basically you control 2 tabs that will have only part of the commands.
eg : play a film handled by tab1, pause handled by tab2, ....

SubJunk:
Cool, I understand now. In that case it might be good in the future to make all tabs share the state of that player instead, so we can remove this tab limitation

SubJunk (approved):

SurfaceS:
@SubJunk  _In that case it might be good in the future to make all tabs share the state of that player instead, so we can remove this tab limitation_
No, this one (`PollHandler` and `EventStreamHandler`) is application gui -> control -> web browser. So it's should control only one tab (or you will play the same video on all tabs, etc).

The other direction is under `ControlHandler` : web browser -> control -> upnp devices.
`ControlHandler` may need some improvements as well, but I did not look at it.
For now I'm working on `MediaServer`, trying to clean a bit and handle better all stuff in there.

SubJunk:
In that case I don't see why we can't identify each tab uniquely, to allow the use of multiple tabs. Can't we just generate an ID in each tab, then attach it to requests so that the server knows which tab the request came from?

SurfaceS:
Generating an id is not the problem. You need to send it back to the server.
When you change page, the web client request send all cookies to the server, so the server know who you are.
But cookies are common for all your tabs, so you can't.
It need a complete rewrite of the web device identify, not using id on cookie.

ik666:
@SurfaceS would it be a huge rewrite to send the ID as a - in example - header parameter instead of a header cookie?

SurfaceS:
Sure, everything can be done. The link between the cookie and `WebRender` seems to sit under `WebInterfaceServerHttpServer.getRoot()`.

But what is the goal to control 2 tabs from the application gui ?
I mean, as I understand, `PollHandler` was made to remote control your web browser from the java application gui.

First... who really use it ?
Ok, let say that I open the `WebInterface` on a computer and instead of browsing / playing from the web browser, I want to back on the java application gui to play a film on it. It's ok... I can understand.

But who really need to have 2 tabs on the same web browser, same session and same computer, and want to control both of them from the application gui ?

SubJunk:
> But who really need to have 2 tabs on the same web browser, same session and same computer, and want to control both of them from the application gui ?

Yeah maybe no one does that, but this change has introduced a popup that warns about having two tabs open, regardless of what the user was doing. They probably just have one video paused in one tab, and another one playing in another tab. Or they wanted to have the list of episodes open in one tab while another tab plays one.

If we don't think it is likely that a user will encounter that use-case then maybe that warning popup should be removed

SubJunk:
@SurfaceS maybe we should remove that popup warning about multiple tabs? Did you see the comment above?

SurfaceS:
@SubJunk  `They probably wanted to have the list of episodes open in one tab while another tab plays one of them, or something more common like that.`
In both case, the sse or pools will not working as supposed.
In pools, command will be handle by a random tab (the first one that its 1 second timer ask after the command was asked)
In SSE, the last tab that browse the server will handle the command.

@ik666 `would it be a huge rewrite to send the ID as a - in example - header parameter instead of a header cookie?`
The hardest rewrite is :
On browse, each tab will have a unique id, then, UMS will create a new WebRenderer and show it on the renderer list.
A new tab will be handle the same, so UMS will end having 2 renderers.
1: You will not be able to differentiate them (same IP, same device, etc), so you should handle this.
2: When the tab is closed, the renderer is never removed from the renderer's list.
If you open the web interface, and close it, you will see that the renderer is still present.
So UMS will end with alot renderer for each time you browse a new tab (even after closing all of them).

In fact, for this to work properly, it need a big rewrite on how UMS handle the WebRenderers.

I'll remove the warning i na PR by the way.

SubJunk:
@SurfaceS 1. why can’t the tabs be differentiated? I haven’t used SSE before but have used websockets a lot which seems similar
2. I don’t think we remove renderers from the list ever, but we do deactivate them when they say byebye, so we could do the same for tabs on beforeunload

just thoughts, I’m not saying that is the best idea

PR #2804, opened 2022-01-25
Removed ReadWriteLocks

SubJunk:


ik666 (approved):
I think this look good.

PR #2809, opened 2022-01-29
HLS on web interface

SurfaceS:


SubJunk:
Thanks for this. It is mostly working but it needs me, or someone, to make the transcoding process stop if playback stops. At the moment, because it is not piping the output, it's just a detached process that keeps going until it's finished.

SubJunk:
@SurfaceS thanks for that last commit too. You have been contributing a lot lately, would you like to join the project? Assuming that is ok with the other @UniversalMediaServer/developers

SubJunk:
@SurfaceS transcoded videos don't work at all for me on the web interface after https://github.com/UniversalMediaServer/UniversalMediaServer/pull/2809/commits/40e5deb8afee6ffe68d9978880a84fbfca9d3b36
UMS has a failed response instead of the video.
It is working in https://github.com/UniversalMediaServer/UniversalMediaServer/pull/2809/commits/708f255fe2a6b9778cea6f18b0824b5c051a7c99
Can you check that commit?

PR #2822, opened 2022-02-10
Feature/master/like music

ik666:
This PR does

- some API cleanup
- adds `like` API. Music songs (identified by musicBrainz trackID) and albums (identified by musicBrainz releaseID) can be liked / disliked
- adds `rating` API. Songs can be rated from 0-5. Optionally (default enabled) IDv3 Tags in song files are updated.

Documentation will be updated after merge.

Later I'll add virtual "MyMusic" and "MyAlbum" folders for navigating to liked songs and albums.


ik666:
@SurfaceS 
1. From my point of view, I'd like to enforce separation of concerns (media renderer, control point, media server) in separate applications. I know you guys are pushing UMS into one "big pile of UPnP features" able to serve, control and render media from UMS. I think, this leads to issues you have discussed in other threads (identifying renderer, uncertain control point behaviour). It's of course easier to implement advanced features directly in UMS, since you don't have to think about how to communicate with another process, but if you go the other ways, you need the API. That's why I am developing a stand alone [control point](https://sf666.github.io/). It has some features I slowly migrate to UMS-API (star rating of songs, liking songs and albums, etc.) because I think managing internal media content data is more a concern of media servers than of control points, and other control points see the effect of the API (like TV's, etc.). 

Like every API, it has to be used by others ... Since it it very straight forward to use, almost everyone can script a [curl call](https://support.universalmediaserver.com/books/configuration/page/external-api) as documented and use it as one wish ... 

2. This PR is no star rating and not implemented in the API yet. It will migrate star rating later ... This `like` is more how it's implemented in Tidal. In a later implementation, a virtual "liked albums" folder will show liked albums, same to songs.

3. I have no video content, but am willing to help.

SurfaceS:
I better understand your need and the api goal.

I talk about rating, as "like" it is a PERSONAL_RATING of the media with the max value.
If you modify the db with a boolean, you will be linked to like/dislike and need to reversion the db later, when you want to say, "i like this one, but not so much" (like 3.5/5).
Having a float will let you the later ability to do so, and can be use like a boolean for now (ex : `SET PERSONAL_RATING = 5 where...`, `WHERE PERSONAL_RATING > 2.5`, etc).

Also, having the PERSONAL_RATING column on the FILE table will allow your api to retrieve the like on audiotrack, but avoid multiplication of table changes when someone else will want a PERSONAL_RATING on film, then series, photos, etc...

I know it's frustrating to read me :)
I can help you next week on this without breaking your api if you want (I'm away all the WE).

ik666:
Nono ... you have good points and I see the benefit in rating all kinds of media ... I have moved the (personal) rating column to the files table.

ik666:
So I thinks it's done now. File (song) rating support is added. ID3v2 Tags of files are updated (can be disabled).

ik666:
[Documentation](https://support.universalmediaserver.com/books/configuration/page/external-api/edit) is also updated

SubJunk:
I agree that we shouldn't be attempting to edit user files unless the user has opted in, and been warned about the risks

ik666:
@Nadahar Since it's no big deal, I can go along with disabled defaults. However, storing the rating in the file is not only for other software, but also for reading the values after a database reset.

Nadahar:
@ik666 I agree that it's many advantages to storing the rating in the ID3v2 tag. I for example typically rate my audio files using WinAmp (yes, still i 2022 ;) ) - so most of my audio files would already be rated "correctly" and it would be handy for that information to be used.

I am just worried that (maybe unrelated) bugs, now or in the future, could lead to corrupted files. Since the ID3 tags are located at the beginning of the files (at least in MP3 and FLAC, I'm not exactly sure all what other file formats that support them), it doesn't take much to destroy the rest of the information in the file while rewriting it if there's a bug. Imagine a situation where some bug triggers this during a library scan and ends up effectively destroying a user's entire audio library.

I just think that the user should make the decision to allow this. Some may have all the audio files stored somewhere else, so that it wouldn't be a big deal - for them it doesn't matter. For others though, it might be the only copy they have and re-ripping them would be a huge job.

ik666:
@Nadahar I sure comply with your and the others' position a user should opt in this feature, since you never know if there's a bug in jaudiotagger.

With the last commit, your rating information written by one of the greatest audio player (at least one you trust) will be populated to the database as well. :)

ik666:
"My Album" folder is now available for browsing "liked" albums. This feature is pretty cool for people with large audio libraries. Link position of "My Albums" is configurable with `KEY_AUDIO_LIKES_IN_ROOT_FOLDER` key between "root folder" and "Audio" folder (default).

ik666:
@SubJunk This looks OK for me now ... before we drift off to much, review/merge it ?
@Nadahar I hope you don't mind hijacking your coverart table, but it looked like the right place

Nadahar:
@ik666 No worries, it's not "mine" :wink: I haven't looked at the code, but I'm not sure it's a good fit from what was the purpose of the table when I made it. The tables (both for cover art and MBID) was meant to be pure cache tables that cached MBIDs and cover art binaries. They were meant to be "disposable" in that no "important" data was stored there and that they could be cleared with any other consequences than that the information would have to be re-downloaded if it were requested in the future.

As such, these have no "link" to the "media library". I would think one would care a bit more about "like" information which is something the users actually have to spend time entering in some way or another. If this "philosophical difference" has been handled, I can't see any other "negative consequences" of that - except that it might be a bit slow to have to look up an additional table for each media item when browsing.

SubJunk:
Will review this ASAP

ik666:
It's ok. I see you're busy with all the other branches ...

SubJunk:
@ik666 yes and my SSD failed 2 days ago so I have been setting everything up again after buying a new one... Not fun haha

src/main/java/net/pms/network/mediaserver/nettyserver/RequestV2.java
```diff
@@ -287,11 +288,14 @@ public ChannelFuture answer(
 		}
 
 		if (uri.startsWith("api/")) {
-			ApiHandler api = new ApiHandler();
-			api.handleApiRequest(method, content, output, uri.substring(4), event);
-			ChannelFuture future = event.getChannel().write(output);
-			if (close) {
-				future.addListener(ChannelFutureListener.CLOSE);
```
SubJunk, line 294: Is it intentional that this close handler was removed? Just checking

src/main/java/net/pms/util/CoverArtArchiveUtil.java
```diff
@@ -493,7 +493,7 @@ private static void releaseCoverLatch(CoverArtArchiveCoverLatch coverLatch) {
 				connection = MediaDatabase.getConnectionIfAvailable();
 				// Check if it's cached first
 				if (connection != null) {
-					CoverArtArchiveResult result = MediaTableCoverArtArchive.findMBID(connection, mBID);
+					CoverArtArchiveResult result = MediaTableCoverArtArchive.findMBID(mBID);
```
SubJunk, line 496: Why is the connection sharing removed throughout this file?

src/main/resources/i18n/messages_de.properties
```diff
@@ -1,5 +1,6 @@
 # This file is automatically generated. Please do not edit this file. If you'd like to change the content please use crowdin.
 
+Audio.Like.MyAlbum=Meine Alben
```
SubJunk, line 3: You'll need to add the term to the English file if you want it to be translatable on Crowdin

SubJunk (changes requested):

src/main/java/net/pms/database/MediaTableAudiotracks.java
```diff
@@ -94,12 +95,10 @@ private static void upgradeTable(Connection connection, Integer currentVersion)
 					if (!isColumnExist(connection, TABLE_NAME, MBID_TRACK)) {
 						executeUpdate(connection, "ALTER TABLE " + TABLE_NAME + " ADD " + MBID_TRACK + " UUID");
 					}
-					break;
```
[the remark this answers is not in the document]
ik666: ok, I'll add the breaks again ...

src/main/java/net/pms/dlna/DLNAMediaAudio.java
```diff
@@ -876,4 +877,20 @@ public void setAudioProperties(AudioProperties audioProperties) {
 		}
 		this.audioProperties = audioProperties;
 	}
+
+	/**
+	 *
+	 * @return user rating (0 - 5 stars)
+	 */
+	public Integer getRating() {
```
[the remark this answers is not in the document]
ik666: I think @SurfaceS suggested to add rating attribute to files, because other types except audio can then also be rated. Basically a good idea. But you might be right, if a file (maybe a cue file) has multiple tracks, this logic fails. I think I'll move the rating column back to Audiotack table.

src/main/java/net/pms/network/mediaserver/handlers/ApiHandler.java
```diff
@@ -32,46 +33,46 @@ public ApiHandler() {
...
-		output.headers().set(HttpHeaders.Names.CONTENT_LENGTH, "0");
-		output.setStatus(HttpResponseStatus.NO_CONTENT);
-
+	public String handleApiRequest(HttpMethod method, String content, HttpResponse output, String uri, MessageEvent event) {
 		String serverApiKey = PMS.getConfiguration().getApiKey();
 		if (serverApiKey.length() < 12) {
 			LOGGER.warn("Weak server API key configured. UMS.conf api_key should have at least 12 digests.");
 			output.setStatus(HttpResponseStatus.SERVICE_UNAVAILABLE);
-			return;
+			return "Weak or no server API key configured. UMS.conf api_key should have at least 12 digests.";
```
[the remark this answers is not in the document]
ik666: Yes, characters is more precise.

src/main/java/net/pms/network/mediaserver/nettyserver/RequestV2.java
ik666: yes, the API call doesn't return right away, but later (and then it will be closed).

src/main/java/net/pms/util/CoverArtArchiveUtil.java
ik666: Because sharing wasn't used ... the connection pool is sufficient unless you'd like to batch process something ...

ik666:
I've added a backup/restore logic for user managed 'liked album' table, in case someone decides to clear the database. 

1. like albums
2. do a backup
3. clear database
4. do a restore

Liked albums appear again in "My Albums" folder. Thx @Nadahar for the hint ...

ik666:
Update: SearchRequest class handles now UPnP `SortCriteria` attribute.
- If you search for audio songs and include `-upnp:rating, +dc:title` as SortCriteria, you get high rated songs first, then sorted by song name. 
- If you search for audio albums and include `"-ums:liked, +dc:title"` as SortCriteria, you get liked albums first, then sorted by album title.

ik666:
@SubJunk I see you're very busy :) Anyhow, do you think, there's a chance to merge this?

SubJunk:
@ik666 yes sorry, I have been putting out fires in `master` :) I have one more high priority bug to fix (timeouts on browsing, I have a fix written in my head but I need my fingers to do it), then I will re-review this next. Sorry for the delay!

SubJunk:
Swap reviews? https://github.com/UniversalMediaServer/UniversalMediaServer/pull/2896 :)

ik666:
Sure, but your's was an easy one :)

src/main/java/net/pms/dlna/DLNAMediaAudio.java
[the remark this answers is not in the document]
SubJunk: I see. I didn't realize that was already discussed. It's ok either way for me

src/main/java/net/pms/network/mediaserver/handlers/ApiHandler.java
```diff
@@ -32,46 +33,47 @@ public ApiHandler() {
...
-		output.setStatus(HttpResponseStatus.NO_CONTENT);
-
+	public String handleApiRequest(HttpMethod method, String content, HttpResponse output, String uri, MessageEvent event) {
 		String serverApiKey = PMS.getConfiguration().getApiKey();
 		if (serverApiKey.length() < 12) {
-			LOGGER.warn("Weak server API key configured. UMS.conf api_key should have at least 12 digests.");
+			LOGGER.warn("Weak server API key configured. UMS.conf api_key should have at least 12 characters.");
 			output.setStatus(HttpResponseStatus.SERVICE_UNAVAILABLE);
-			return;
+			return "Weak or no server API key configured. UMS.conf api_key should have at least 12 characters.";
```
SubJunk, line 44: Now that the return string is being returned with the response, we should make it translatable

src/main/java/net/pms/dlna/DLNAMediaInfo.java
```diff
@@ -661,6 +662,8 @@ public void parse(InputFile inputFile, Format ext, int type, boolean thumbOnly,
 						Tag t = af.getTag();
 
 						if (t != null) {
+							addMusicBrainzIDs(af, file, audio);
```
SubJunk, line 665: Not sure if this applies here too, but did you see this commit I did? https://github.com/UniversalMediaServer/UniversalMediaServer/pull/2882/commits/a3844ee54e2d368e51d208a3355008f4e8927a45
I found that we were trying to parse tags for videos, which would always fail, so we should be careful to not do that here too

src/main/java/net/pms/dlna/DLNAMediaInfo.java
```diff
@@ -916,6 +920,31 @@ public void parse(InputFile inputFile, Format ext, int type, boolean thumbOnly,
...
+		try {
+			Tag t = af.getTag();
+			if (t != null) {
+				currentAudioTrack.setRating(StarRating.convertTagRatingToStar(t));
+			}
+		} catch (Exception e) {
+			LOGGER.trace("audio rating tag not parsed: " + e.getMessage());
+		}
+	}
+
```
SubJunk, line 947: These look very similar to the functions in libmediainfoparser, do these need to be here?

SubJunk (changes requested):

src/main/java/net/pms/dlna/DLNAMediaInfo.java
ik666: The MBID attributes in line 665 will only be added, if we have an audio tag. So this section should not be applicable to video files ... Therefore I think it's ok.

src/main/java/net/pms/dlna/DLNAMediaInfo.java
ik666: It's even doubled in the same method ... I'll remove it ... thx for this one. 

But: I actually don't know why LibmediaInfo is parsing all the metadata (into the database ?), and then again DLNAMediaInfo.parse() doesn't use it. Probably because database access (cache ?) can be turned off.

src/main/java/net/pms/dlna/DLNAMediaInfo.java
```diff
@@ -916,6 +921,48 @@ public void parse(InputFile inputFile, Format ext, int type, boolean thumbOnly,
 		}
 	}
 
+	/**
+	 * Extracts key value.
+	 *
+	 * @param key
+	 * @return If key is not available or blanc, NULL will be returned, otherwise string key value
+	 */
+	private String extractKeyValue(Tag t, FieldKey key) {
```
SubJunk, line 930: For this method and `extractKeyIntegerValue`, can they have more descriptive names since they are for Tags? Something like `extractAudioTagKeyValue` and `extractAudioTagKeyIntegerValue`?

src/main/java/net/pms/dlna/LibMediaInfoParser.java
```diff
@@ -573,7 +580,18 @@ private static void addMusicBrainzIDs(File file, DLNAMediaAudio currentAudioTrac
 				currentAudioTrack.setMbidTrack(val.equals("") ? null : val);
 			}
 		} catch (Exception e) {
-			LOGGER.trace("Audio Tag not parsed: " + e.getMessage());
+			LOGGER.trace("audio musicBrainz tag not parsed: " + e.getMessage());
+		}
+	}
+
+	private static void addRating(AudioFile af, File file, DLNAMediaAudio currentAudioTrack) {
```
SubJunk, line 587: Can this be `addAudioTrackRating`? Because we are in a big file that does much more than just audio files

SubJunk (changes requested):
Sorry to keep coming up with new things, I think this is the last review!
Have you tested it enough? I don't use this feature so I am trusting you on the testing part

src/main/java/net/pms/dlna/LibMediaInfoParser.java
ik666: Sure ... good idea.

src/main/java/net/pms/dlna/DLNAMediaInfo.java
ik666: Yes, I'll do that.

ik666:
@SubJunk Yeah, I am using this API now for over a month, coming up with new ideas as you see ... This really works great.

ik666:
@SubJunk I have made the naming changes ...

SubJunk (approved):

PR #2839, opened 2022-02-18
Samsung TV 2021 series (AU)

SurfaceS:
Based on the latest [notice](https://downloadcenter.samsung.com/content/UM/202109/20210910124253866/OSNDVBEUA-7.2.0_EM_OSCAR_EU_ENG_210910.0.pdf) from Samsung website.
Cover all the 2021 models for TV including special series.


src/main/external-resources/renderers/Samsung-TV-2021-0.conf
```diff
@@ -0,0 +1,115 @@
...
+# we will rely on modelName
+# modelName=QE75Q80AATXXC / QE50QN90AATXXC
+# ============================================================================
+#
+# try user agent even if most of time it's changed by user.
+UserAgentSearch = SEC_HHP_.*(Q[NEA]\d{2}Q[789]\d{1}A|Q[NEA]\d{2}QN[789]\d{1}A)
+UpnpDetailsSearch = Q[NEA]\d{2}Q[789]\d{1}A|Q[NEA]\d{2}QN[789]\d{1}A
+LoadingPriority = 5
+
+TranscodeVideo = MPEGTS-H265-AAC
```
SubJunk, line 27: For most people's hardware it's a better idea to do `MPEGTS-H264-AAC` or `MPEGTS-H264-AC3` because it is more likely to handle realtime encoding

SubJunk (changes requested):
This looks great. I left a note about the TranscodeVideo setting, and can you please add the relevant tests to https://github.com/UniversalMediaServer/UniversalMediaServer/blob/master/src/test/java/net/pms/configuration/RendererConfigurationTest.java

When the tests are added, you might find that some of the `LoadingPriority` numbers can be lower too

src/main/external-resources/renderers/Samsung-TV-2021-0.conf
SurfaceS: I was following the Samsung recommendations (H264 is not 8K capable).

But as TV handle all 8K codec it should not be a problem to lower to H264.

SubJunk (approved):
Very cool

PR #2849, opened 2022-02-26
Fixes from v11

### LINKED ISSUE #2343: Deinstallation does not remove autostart entry for wrapper.exe
The deinstallation routine does not remove the autorun entry for the wrapper.exe;

HKLM\System\CurrentControlSet\Services			

Digital Media Server	Universal Media Server: Java UPnP AV/DLNA Media Server		File not found: D:\ums\Universal Media Server\win32\service\wrapper.exe -s D:\ums\Universal Media Server\win32\service\wrapper.conf		

If I delete it by myself I do need administrator rights.

### LINKED ISSUE #2826: [UMS v10.15.0] 'Media Library' Subfolders no longer populated + Start-Up Media Scan now incredibly slow.
[ums_dbg_2022-02-13-09-20.zip](https://github.com/UniversalMediaServer/UniversalMediaServer/files/8054670/ums_dbg_2022-02-13-09-20.zip)

Since updating from v10.13.0 to v10.15.0, the Subfolders inside the "Media Library" Directory are no longer being populated. (ie, Recently Added/Played etc.)

Additionally, the Media Scan that occurs upon Application Start-Up now takes significantly longer. Prior to updating it took between 1m and 1m10s, and after updating it takes 12 to 14 minutes.

Unsure if issues are related to each other.
Debug logs are attached.
Issues are completely resolved by rolling back to v10.13.0.

Happy to help with troubleshooting following advice Re: possible fixes.

### LINKED ISSUE #2838: Attempt to remove medias.lock on install/startup
Sometimes the h2database lockfile is getting left behind during the upgrade to h2 v2, which breaks the upgrade and leaves UMS in a very broken state.

The Windows installer should be able to remove the file after it kills the process, which it already does. Then we should also try to remove it on UMS startup, for other operating systems. If it fails because of permissions it could leave a note in the logs or otherwise notify the user. It could be a good use for the SSE notification system that @SurfaceS built.

SurfaceS:
Delete windows service on uninstall.
fix #2343
Remove medias.lock.db file before db init.
fix #2838
fix #2826

### DIFF
--- a/src/main/external-resources/nsis/setup.nsi
+++ b/src/main/external-resources/nsis/setup.nsi
@@ -708,6 +708,13 @@ Section "Uninstall"
 	DeleteRegKey HKEY_LOCAL_MACHINE "${REG_KEY_UNINSTALL}"
 	DeleteRegKey HKCU "${REG_KEY_SOFTWARE}"
 
+	services::IsServiceInstalled 'Universal Media Server'
+	Pop $0
+	; $0 now contains either 'Yes', 'No' or an error description
+	${If} $0 != "Yes"
+		DeleteRegKey HKLM "SYSTEM\CurrentControlSet\Services\Universal Media Server"
+	${EndIf}
+
 	ExecWait 'netsh advfirewall firewall delete rule name=UMS'
 	ExecWait 'netsh advfirewall firewall delete rule name="UMS Service"'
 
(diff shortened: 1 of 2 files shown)


SubJunk:
Thanks!

PR #2858, opened 2022-03-02
Fixed failed database writes

### LINKED ISSUE #2856: Regression: Since 10.16.0, metadata tables are not populated for files
I have confirmed that since 10.16.0 the metadata tables (like `VIDEO_METADATA_ACTORS`) are not having values added for files, despite the logs saying they are. I have verified it with the h2database inspector tool.

It is still adding the values for TV series, just not files.

I'm looking into it now but wanted to post here in case you have any ideas @SurfaceS since that was the release with the h2database v2 upgrade

SubJunk:
There were some cases of a method establishing a shared db `Connection`, then calling another method which altered that connection using `setAutoCommit`, which caused unexpected results.
In this commit I have tried to establish a pattern where the method that establishes the `Connection`, is also the one that alters it, which gives better visibility over the state of the `Connection`.
Along the way I did minor optimizations that shave milliseconds.

This is a high priority bugfix so can I please get a review soon @UniversalMediaServer/developers 

Closes #2856

### DIFF
--- a/src/main/java/net/pms/database/MediaTableCoverArtArchive.java
+++ b/src/main/java/net/pms/database/MediaTableCoverArtArchive.java
@@ -154,39 +154,37 @@ public final class MediaTableCoverArtArchive extends MediaTable {
 				LOGGER.trace("Searching for Cover Art Archive cover with \"{}\" before update", query);
 			}
 
-			try (Statement statement = connection.createStatement(ResultSet.TYPE_SCROLL_SENSITIVE, ResultSet.CONCUR_UPDATABLE)) {
-				connection.setAutoCommit(false);
-				try (ResultSet result = statement.executeQuery(query)) {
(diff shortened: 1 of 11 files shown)


ik666:
From what I see, you just removed all manual commits ... I think your changes are just fine.

For the patter: As long as you don't want to spawn transaction over methods and/or want to have control over the commit point, best practice for atomic operations would probably be: the method who needs a connection acquires one itself by calling `MediaDatabase.getConnectionIfAvailable()` and release it at the end. This might be tricky of course for recursive calls :)

ik666 (dismissed):

SubJunk:
I removed the manual commits that were only doing one operation, since that is pointless, but I grouped more commits than before. For example here now there is one transaction instead of 19.
Thanks for approving so quickly! I will try to release this soon

PR #2867, opened 2022-03-06
fix BufferedOutputFileImpl return wrong len on read

### LINKED ISSUE #2866: Piped ffmpeg InputStream give rubbish bytes at end
When I tried to made HLS working, I was not understand why it's not working as I expected.
To debug, I created files directly from ffmpeg with the same command as the transcoder do (from log). And when I HLS directly from that files it worked as expected.
Then, I downloaded the files from the transcoder and compare hex in files.
The file are identical at start, but UMS transcoder's file have rubbish at end.

I then logged the number of byte transferred from the server in `WebInterfaceServerUtil.dump(in, os);`:
From ffmpeg to the pipe :
`[AVIOContext @ 00000213b04f0c80] Statistics: 1136836 bytes written, 0 seeks, 198 writeouts`
From the pipe to the renderer :
`[Thread-221] Sending stream finished after: 1146880 bytes.`

And looking at the difference on files, it is 10044 more rubbish bytes on the UMS transcoder's file.

It may be a bug only on windows platform.

SurfaceS:
fix #2866

### DIFF
--- a/src/main/java/net/pms/io/BufferedOutputFileImpl.java
+++ b/src/main/java/net/pms/io/BufferedOutputFileImpl.java
@@ -724,7 +724,7 @@ public class BufferedOutputFileImpl extends OutputStream implements BufferedOutp
 			return endOF - mb;
 		} else {
 			System.arraycopy(buffer, mb, buf, off, len - cut);
-			return len;
+			return len - cut;
 		}
 	}
 


SubJunk (approved):
This file could really do with some code comments and Javadocs, but this fix looks good

PR #2897, opened 2022-03-18
Bugfix: audio files get identified as video files -> Wrong 'protocolI…

ik666:
Audio files are being identified as video files at some points in the code, because the VideoCodec (VIDEOC attribute) is set to 'und' if no video is present, but UMS code checks against NULL later: 

DLNAMediaInfo line 1465 following 

```
} else if (codecV == null && codecA != null) {
	if ("ogg".equals(container) || "oga".equals(container)) {
		mimeType = HTTPResource.AUDIO_OGA_TYPEMIME;
	} else if ("3gp".equals(container)) {
		mimeType = HTTPResource.AUDIO_THREEGPPA_TYPEMIME;
	} else if ("3g2".equals(container)) {
		mimeType = HTTPResource.AUDIO_THREEGPP2A_TYPEMIME;
	} else if ("adts".equals(container)) {
[...]
```

For audio files the mime type will therefore not be checked delivering a wrong mime-type, leading to an incorrect protocol-info res-attribute (video/mpeg) instead of an audio/[...] in DIDL string. 

From my perspective NULL makes somehow more sense (changes in this PR).

### DIFF
--- a/src/main/java/net/pms/dlna/DLNAMediaInfo.java
+++ b/src/main/java/net/pms/dlna/DLNAMediaInfo.java
@@ -1434,7 +1434,7 @@ public class DLNAMediaInfo implements Cloneable {
 		}
 
 		if (mimeType == null) {
-			if (codecV != null) {
+			if (codecV != null && !codecV.equals(DLNAMediaLang.UND)) {
 				if ("matroska".equals(container) || "mkv".equals(container)) {
 					mimeType = HTTPResource.MATROSKA_TYPEMIME;
 				} else if ("ogg".equals(container)) {
@@ -1462,7 +1462,7 @@ public class DLNAMediaInfo implements Cloneable {
 				} else if (codecV.contains("mpeg") || codecV.contains("mpg")) {
 					mimeType = HTTPResource.MPEG_TYPEMIME;
 				}
-			} else if (codecV == null && codecA != null) {
+			} else if ((codecV == null || codecV.equals(DLNAMediaLang.UND)) && codecA != null) {
 				if ("ogg".equals(container) || "oga".equals(container)) {
 					mimeType = HTTPResource.AUDIO_OGA_TYPEMIME;
 				} else if ("3gp".equals(container)) {
--- a/src/main/java/net/pms/dlna/LibMediaInfoParser.java
+++ b/src/main/java/net/pms/dlna/LibMediaInfoParser.java
@@ -549,10 +549,6 @@ public class LibMediaInfoParser {
 				media.setContainer(DLNAMediaLang.UND);
 			}
 
-			if (media.getCodecV() == null) {
-				media.setCodecV(DLNAMediaLang.UND);
-			}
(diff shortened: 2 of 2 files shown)


Nadahar:
I don't know what has changed since I was involved, but I don't think this used to happen. I used to test quite a lot with audio files, and I'm pretty sure I would have picked up on that. I do remember there was a bug with some audio files, those with embedded thumbnails I think, where the thumbnail was parsed as "video" - but I'm pretty sure we fixed that back in the day.

I would recommend looking through changes to that code in the last years using Git blame (can't be that many), and you'd probably be able to find the culprit if this is a "common occurrence". If it's some very rare bug with some very particular audio files, it could of course have been there all the time.

You should also make sure that you don't use the "audio as video" hack that is in the code base. I think it applies to FLAC only, and it's an old hack that transcodes FLAC audio files into video files (with a black image) to allow the PS3 to play FLAC - since it can play FLAC as an audio track in video files but not as audio files. Go Sony!

ik666:
Hmmm, ok. Setting `codecv` to NULL was introduced in version 8.0.0.  Indeed I am using FLAC files for testing, but then, all places with `codecV == null` have to be changed to `codecV == DLNAMediaLang.UND` since `codecV == null` cannot happen. 

DLNAMediaLang.UND looks like a "null"-object then ... not very nice ...

Are you sure getting NPE's after this change ? With a clean DB I haven't seen any yet, but I don't have all the possible renderer ...

Nadahar:
@ik666 No, I'm not sure. It should have said it differently. It might work just fine, I'm just saying that from experience, lots of the code doesn't handle null, so introducing it where it hasn't used to be, runs the risk of NPEs all over the place.

If it can already be set to null in other places, it's probably fine. I still find it strange how this problem has "suddenly appeared" since I used to quite a lot of audio testing. But, watch out for the "audio as video" hack because it does some strange things and it could potentially be causing trouble.

ik666:
I'm asking myself the same question. I think as I introduced this PR, songs protocolInfo attribute was correctly set to audio for streaming when browsing into "My Album" folder (liked albums). And I was able to play songs from an album (extracting the correct streaming url). Then from time to time some songs just switched to video ... strange ...

SubJunk:
I guess this problem needs some more work to identify the bug before this fix is considered? Or have you found that this branch fixes the bug?

ik666:
@SubJunk Not really. This issue can be fixed in to ways ... @Nadahar suggested the other way around, which might be more the way the old code works. Therefore I'll submit an update.

ik666:
@SubJunk @Nadahar Now the codecV will be checked againt "UND". However this implementation blows up the database. For my test files from 8 MB (setting codecV to NULL) to 27 mb (setting codecV to "UND"). I'd recommend actually the "first fix", but due to the lack of many video files, I believe I cannot test it as needed.

SubJunk (approved):

PR #2902, opened 2022-03-25
editable server side playlist support

ik666:
This PR adds advanced playlist support to UMS. Playlist API is disabled by default. This service has the following capabilities:
- listing all available playlists
- adding songs to playlist
- removing songs from playlist
- creating empty playlist

Also see [usage documentation](https://support.universalmediaserver.com/books/configuration/page/external-api#bkmrk-playlist).

Edit:
- rate all songs (audio) by ums audiotrackId -> Songs do not necessarily have to be tagged with musicBrainzId


Nadahar:
I think a lot of the stuff you implement using the "API" could have been done using UPnP AV/DLNA. I know there is support for playlists (although I don't remember the details), and I also know there's support for uploading files from a client. One would think that by combining these, it might be possible to achieve the same that way. The difference would be that it would work for other combinations that just UMS and your software renderer implementation.

I realize that it would probably take some more work to figure out how to do it, studying standards etc - but I also think the reward would be much greater. This are just my thoughts and isn't in any way intended to be anything else.

ik666:
Thanks Nadahar for your thoughts! I've digged throw the UPnP AV CDS in Version 1 and Version 4 docs. Playlists are "supported" as container format and already implemented by UMS. But I believe, further support like "adding" and "removing"  items are not addressed, or at least not in a way that I understand.

Nadahar:
I think you need to look at the DLNA docs for a lot of this. The 2016 version (all the important stuff is there) is available for download for free here: https://spirespark.com/dlna/guidelines/

ik666:
Today, they want to be contacted :)

`Contact dca@spirespark.com to obtain a copy of the latest DLNA Guidelines` 

For the "edit / upload" idea: Even if an upload of an playlist would be possible, it's hard to do a valid "edit" without filesystem access ...

ik666:
Since it's now possible to identify the songs directly by its id, I've added a methods to song-rate songs identified by audiotrack_id.

Nadahar:
@ik666 It used to be downloadable from there for free. I assume that the same is still the case if you e-mail them - but they will probably try to sell you a newer version for money first. That said, there are other ways to get it. I can't post it here because it's copyrighted, but a lot of people have this specification.

I'd say it's pretty hard to develop both a renderer or a server without having the DLNA spec, so you really should get your hands on it. UPnP AV is just a "vague framework", DLNA is where most of the details are defined.

As to what is and isn't possible with playlists, you shouldn't conclude without actually checking. The spec is pretty long, my memory is limited, and I was never really interested in playlists when I looked in it. So, I don't claim to remember everything, I just know that there's some playlist stuff there, among other things.

ik666:
@Nadahar I'd love to see something in the specs, that I could use, but the real issue is that UMS doesn't identify any resources uniquely. A concrete resource might get an object id of "10" later "430". At that time "10" is not browsable any more! The UPnP spec strongly encourages ID's of a resource to be static (even after a server restart), but doesn't enforce this. Therefore, from a control point perspective, it's impossible to "add" or "remove" links to dynamic server resource ...

I've checked some control points. All have renderer based playlist support, none of them are able to create / edit server based playlists. All discussions end up with use this or that tool (mediamonkey, etc.) to edit the server based playlist directly on the filesystem where the music is stored. Intension of this API is to add/remove songs 
- while you're listening to songs
- while browsing throw the audio library
- that have been found by the search API
directly into the server based playlist. The best is UMS has direct access to file changes. There is no need to start another tool trying to find the songs from above use case.

Nadahar:
@ik666 I wasn't suggesting that UMS was able to handle it as it is. Adaptations would have to be done, but the same is true when you develop the "API".

It might not be possible, I just tried to suggest that you might want to check if it's actually possible or not using UPnP/DLNA. I'm not surprised that you haven't found another software renderer that can handle this. I must say that I don't know of one single proper software renderer implementation. I've tried "every" software renderer I've found, and they are all riddled with problems and non-compliance. I don't think anybody have had the motivation to make a proper implementation, and DLNA requires you to cash out a lot of money to get access to their "compliance tools", which means that all non-commercial projects are by definition excluded.

For hardware renderers, the situation is different. There is a lot of money involved there, so I assume that there could be quite a few implementations that are at least close to spec. Unfortunately, I can't buy them all to try, so I really don't know for sure though.

src/main/java/net/pms/database/MediaTableAudiotracks.java
```diff
@@ -132,6 +132,20 @@ private static void upgradeTable(Connection connection, Integer currentVersion)
 						LOGGER.trace("Indexing column RATING on table " + TABLE_NAME);
 					}
 					break;
+				case 7:
+					Connection con = DATABASE.getConnection();
```
SubJunk, line 136: There is already the `connection` variable to use here, we don't have to create a new one

src/main/java/net/pms/network/mediaserver/handlers/api/playlist/PlaylistManager.java
```diff
@@ -0,0 +1,243 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import net.pms.Messages;
+import net.pms.PMS;
+import net.pms.database.MediaDatabase;
+import net.pms.dlna.RootFolder;
+
+public class PlaylistManager {
+
+	private static final Logger LOG = LoggerFactory.getLogger(PlaylistManager.class.getName());
```
SubJunk, line 28: Most of our files call this `LOGGER`, can you please use that same convention?

src/main/java/net/pms/network/mediaserver/handlers/api/playlist/PlaylistService.java
```diff
@@ -0,0 +1,71 @@
...
+import org.jboss.netty.handler.codec.http.HttpResponseStatus;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import net.pms.Messages;
+import net.pms.network.mediaserver.handlers.ApiResponseHandler;
+
+public class PlaylistService implements ApiResponseHandler {
+
+	private static final Logger LOG = LoggerFactory.getLogger(PlaylistService.class.getName());
```
SubJunk, line 14: Same here

src/main/resources/i18n/messages.properties
```diff
@@ -1,4 +1,18 @@
 Api.Error.ApiKeyNotAvail=Weak or no server API key configured. UMS.conf api_key should have at least 12 characters.
+Api.Error.UnknownService=No service available for path
+Api.Playlist.AudiotrackIdUnknown=Unknown audiotrack id
+Api.Playlist.PlaylistAlreadyExists=Playlist already exists
+Api.Playlist.PlaylistCanNotBeCreated=Playlist can not be created
+Api.Playlist.PlaylistCreated=Playlist has been created
+Api.Playlist.PlaylistNameNoExt=Provided playlist name shall have no path and no extension
```
SubJunk, line 7:
```suggestion
Api.Playlist.PlaylistNameNoExt=Provided playlist name has no path or file extension
```

SubJunk (changes requested):

src/main/java/net/pms/database/MediaTableAudiotracks.java
sf666: I needed a connection without autocommit ...  OK, I could shuffle the autocommit flag. Not sure if this is more readable ...

src/main/resources/i18n/messages.properties
sf666: ok, the intension is to provide neither a path nor a file extension. I'll give it a new try ...

SubJunk:
@sf666 not sure if you noticed this failed the lint check?

ik666:
@SubJunk No I didn't. You're my hero. After I upgraded Eclipse in the past days my "checkstyle" plugin was gone ...

SubJunk (approved):

Mik-S-UMS:
I've had a question about the playlist API on the [forum](https://www.universalmediaserver.com/forum/viewtopic.php?t=16659). 

I don't know anything about it so can't help.

They asked what to put as `%{http_code}\n` for
```
curl -w "%{http_code}\n" -H "api-key: secret_password" http://localhost:5001/api/folderscanner/rescan
```

Looking over that [knowledge base](https://support.universalmediaserver.com/configuration/external-api) page I think some of the formatting html tags got converted to text when moving to the current system, like `<span class="s1">/api/like/</span><span class="s1">issongliked</span>`

Might be worth checking this over to see if it is correct as the documentation link in the OP no longer exists.
A guide on how to implement this would be helpful too.

ik666:
The `-w "%{http_code}\n"` is the http code output of the curl command on the command line. Nothing has to be put in here. In case of `200` appears, everything works ok. In case of something else like `401` (authentication failure) something went wrong. Actually I've discontinued work on the HTTP API and moved the API to an UPnP service, but this hasn't been merged.

PR #2905, opened 2022-03-26
Serialized version change recovery

SubJunk:
Currently if `ExifInfo` (a parent class involved in thumbnails) changes, or there's a significant JVM change causing the auto-generated UUID to change, UMS ends up in a broken state with regards to the cache/database. Serialization fails and then never recovers.

The strategy:
- [x] If a file thumbnail fails to deserialize, and there is a poster saved from the metadata API, we remove the old thumbnail and regenerate it from the poster
- [x] If a file thumbnail fails to deserialize, and there is **not** a poster saved from the metadata API, we remove the old thumbnail and reparse the file
- [x] If a TV series thumbnail fails to deserialize, and there is a poster saved from the metadata API, we remove the old thumbnail and regenerate it from the poster
- [x] If a TV series fails to deserialize, and there is **not** a poster saved from the metadata API (which will be very rare, this means there was an API metadata for the TV series but it did not include a poster, which I have never seen before), we remove the old thumbnail, and the IMDb ID for the TV series entry to cause it to retry the metadata API.


SubJunk:
@UniversalMediaServer/developers I think this is ready to review, is anyone able to do it?

SurfaceS (approved):

Issue #2912, opened 2022-03-31
HTTP console interface error: The element type "meta" must be terminated by the matching end-tag "</meta>".


esabol:
UMS 10.19 (but I've seen this with all the versions I've tried going back to 10.13, for the record) TRACE-level debug log:
```
INFO  2022-03-31 18:04:14.812 [HTTPv2 Request Worker 1] Media renderer was not recognized. Possible identifying HTTP headers:
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_#_#) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/99.0.4844.83 Safari/537.36
Upgrade-Insecure-Requests: 1
DNT: 1
Referer: http://192.168.1.###:500#/console/stop
TRACE 2022-03-31 18:04:14.812 [HTTPv2 Request Worker 1] Received a request from PlayStation 4 (192.168.1.###:#####):

GET /console/home HTTP/1.1

HEADER:
  Host: 192.168.1.###:500#
  Connection: keep-alive
  Upgrade-Insecure-Requests: 1
  DNT: 1
  User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_#_#) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/99.0.4844.83 Safari/537.36
  Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.9
  Referer: http://192.168.1.###:500#/console/stop
  Accept-Encoding: gzip, deflate
  Accept-Language: en-US,en;q=0.9

Renderer UUID=null
INFO  2022-03-31 18:04:14.815 [HTTPv2 Request Worker 1] [Fatal Error] :1:87: The element type "meta" must be terminated by the matching end-tag "</meta>".
TRACE 2022-03-31 18:04:14.815 [HTTPv2 Request Worker 1] Response sent to PlayStation 4:

HEADER:
  HTTP/1.1 200 OK
  Content-Type: text/html
  Server: Mac_OS_X-x86_64-10.#.#, UPnP/1.0 DLNADOC/1.50, UMS/10.19.0
  Content-Length: 308

CONTENT:
  Content isn't valid XML, using text formatting: The element type "meta" must be terminated by the matching end-tag "</meta>".
[...]

PR #2928, opened 2022-04-15
Declare HTML 5 doctype for console and reduce logspam

SubJunk:
This addresses #2912

### DIFF
--- a/src/main/java/net/pms/network/mediaserver/javahttpserver/RequestHandler.java
+++ b/src/main/java/net/pms/network/mediaserver/javahttpserver/RequestHandler.java
@@ -1376,7 +1376,7 @@ public class RequestHandler implements HttpHandler {
 			if (StringUtils.isNotBlank(name)) {
 				for (String value : headers.getValue()) {
 					header.append("  ").append(name).append(": ").append(value).append("\n");
-					if ("content-type".equalsIgnoreCase(name) && "/XML".contains(value.toUpperCase())) {
+					if ("content-type".equalsIgnoreCase(name) && value.toUpperCase().contains("/XML")) {
 						isXml = true;
 					}
 				}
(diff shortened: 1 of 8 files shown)


src/main/java/net/pms/network/mediaserver/javahttpserver/RequestHandler.java
```diff
@@ -1369,11 +1370,15 @@ private static boolean filterIp(InetAddress inetAddress) {
 
 	private static void logMessageSent(HttpExchange exchange, String response, InputStream iStream, RendererConfiguration renderer) {
 		StringBuilder header = new StringBuilder();
+		boolean isXml = false;
 		for (Map.Entry<String, List<String>> headers : exchange.getResponseHeaders().entrySet()) {
 			String name = headers.getKey();
 			if (StringUtils.isNotBlank(name)) {
 				for (String value : headers.getValue()) {
 					header.append("  ").append(name).append(": ").append(value).append("\n");
+					if ("content-type".equalsIgnoreCase(name) && "/XML".contains(value.toUpperCase())) {
```
esabol, line 1379: I think `"/XML".contains(value.toUpperCase())` is backwards. It should be `value.toUpperCase().contains("/XML")`, no?

SubJunk: Good point!

esabol (approved):

SurfaceS:
V11 log all the traffic and already take care of parsing only XML.
I think this will mess it.
All of that for the "console" webpage that no one see the goal.

SubJunk:
thanks for the investigation, this can be closed since it's already fixed in v11. i'll release the first v11 alpha version soon, there are just 2 bugs to fix first

PR #2961, opened 2022-05-04
Prevent users from enabling startup scanning while the cache is disabled

SubJunk:


SubJunk:
@UniversalMediaServer/developers is anyone able to review this please?

valib (approved):

Issue #2819, opened 2022-02-07
Clean up web interface player toolbar


SubJunk:
There's a separate toolbar below the web interface player that can probably be removed, along with its related flash player code, since Flash has been dead for a while. Users who still rely on it on older systems can stick with older UMS versions, but I doubt there is anyone still using that.

The other custom buttons on the row are the Download, Previous and Next buttons. We shouldn't lose that functionality, but it could be better designed to be more usable. For example the "previous" and "next" buttons might be bigger and more of a footer, the way that online learning websites tend to have it, with the name in them.

![image](https://user-images.githubusercontent.com/584228/152715890-92ed0507-faba-4140-a390-cbea8f194600.png)

PR #2971, opened 2022-05-14
Remove video.js second toolbar

SubJunk:
This also removes the ability to switch to the Flash player in the GUI, but I have left in the code for legacy users to use it via UMS.conf

Closes https://github.com/UniversalMediaServer/UniversalMediaServer/issues/2819


valib (approved):

SubJunk:
@valib thanks for reviewing :)

PR #3002, opened 2022-05-26
webinterface optimized

SurfaceS:
cache js and css
versioning js and css
move images under img
move js and css under util
move bump under util
set to reload js and css only after 7 days or newer UMS version used.


SubJunk:
Nice work

Issue #3004, opened 2022-05-26
Implement current Java UI in web interface


SubJunk:
It will be a big job but it just needs to be done. The Java UI has *always* been ugly, and only looks more dated over time, making us now look like a dinosaur.
People also have problems with it related to different pixel densities of their screens.
Also my day job involves working heavily with JavaScript, so it will make it much easier for me to develop on than the current UI.
Also there are a lot of very active UI libraries we can use that will save us a lot of work on things like accessibility, that either don't exist or are no longer maintained in Java.

I'm doing a POC first and will push it for opinions

Fixes:
- [x] #2993

SubJunk:
This has been mostly done, there can be individual issues for the remaining pieces and fixes

PR #3016, opened 2022-06-02
Web settings

SubJunk:
This is the feature branch for #3004 
It uses React with Mantine.
Existing Maven commands will compile and include it, then when UMS is running you can access it via `localhost:9001/settings/` or clicking "Setttings" via the system tray menu.
During development you can just use the `yarn start` command in the `/configuration-client` directory to develop with automatic serving, hot refreshing, etc.

TODO:
- [x] Auth
- [x] i18n
- [x] Populate the tabs
  - [X] General Settings
  - [X] Navigation Settings
  - [X] Shared Content
  - [X] Transcoding Settings
- [ ] Notifications about program updates via SSE
  - [x] Implement SSE
  - [x] Update memory usage
  - [x] Notify on server restart
  - [ ] Update server need restart - WIP @SurfaceS 
  - [x] Update accounts
  - [x] Update settings
- [x] Run the settings getter through the allowlist
- [x] Tooltips
- [x] #3046 
- [ ]  i18n all of the new interface
- [x] Define settings defaults in the server instead of client
- [x] Update translation keys after #3123
- [x] #3194
- [ ] #3195
- [x] #3208
- [x] #3363 Option to open web settings when server starts, enabled by default, and default the Java GUI to not do that anymore
- [ ] If the web settings are open when UMS is not open, display a helpful screen, instead of an infinite loader
- [x] Fix unable to add a web folder if it is no web folder already set #3351.
- [x] Fix unable to add web folder name on stream #3351.
- [x] #3350
- [x]  implement advanced settings #3359 
- [ ] implement renderers console
  - [x] Show renderers in real time
  - [ ] Control renderers
  - [ ] Edit renderers
- [ ] implement logs
  - [x] View server logs
  - [x] View files logs
  - [ ] Download full logs (needed ? we can already dl pack)
  - [ ] Move logs settings to the settings page and maybe add some option (trace only for part of app)
  - [x] Log's Filter options
- [x] implement player
  - [x] basic player
  - [x] add metadatas
  - [x] basic player on it's own server
  - [x] controlled player
  - [x] fix "Your player session was not received from the server." error on Player load
  - [x] fix web settings opening browser during Maven build
  - [ ] make player mobile browsable
  - [x] #3523 
  - [x] #3522 
  - [x] #3517 - WIP @SubJunk 

according to #3386
- [ ] implement all functionality of the old player in the web gui (either gui or player)
  - [x] new player should use the old player port
  - [x] add a link to the old player on the new player (people will then be able to browse like before)


configuration-client/src/components/Settings/Settings.tsx
```diff
@@ -164,38 +164,53 @@ export default function Settings() {
...
     if (!engine.isAvailable) {
       return (
         <Tooltip label={allowHtml(i18n.get['ThereIsProblemTranscodingEngineX']?.replace('%s', engine.name))} {...defaultTooltipSettings}>
           <ExclamationMark color={'orange'} strokeWidth={3} size={14}/>
         </Tooltip>
       )
+    } else if (items.includes(engine.id)) {
+      return (
+        <Tooltip label={allowHtml(i18n.get['TranscodingEngineXEnabled']?.replace('%s', engine.name))} {...defaultTooltipSettings}>
+          <ActionIcon size={10} style={{ cursor: 'copy' }} onClick={(e: any) => {setTranscodingEngineStatus(engine.id, false); e.stopPropagation();}}>
```
valib, line 201: I would recommend the ActionIcon size at least 20 because the size 10 is really small and it is not clear that clicking on it would change the state.

valib: The same is for the icons for moving the engine up or down.

SubJunk: Done

SubJunk:
@SurfaceS there's an item in the list `Update server need restart - WIP @SurfaceS` are you still working on it?

SurfaceS:
@SubJunk No, the logic is there.
It may need something to do with it, I'll look at that + the react web player after the db review.
A lot to do...

SubJunk:
Cool. No rush :) just keeping the list updated as we are getting close to beta I think

SubJunk:
Just wanted to note I'm making progress on Shared Content tab, hoping to finish it this week

SubJunk:
The Player section isn't working for me anymore on the latest code. I get an error that `Your player session was not received from the server` and then the logs show:

```
DEBUG 2022-09-22 11:47:15.603 [pool-6-thread-6] couldn't add 'My Music' folder because the media library is not initialized.
ERROR 2022-09-22 11:47:15.604 [pool-6-thread-6] RuntimeException in PlayerApiHandler: Cannot invoke "java.util.Collection.iterator()" because "this.val$c" is null
```

SurfaceS:
@SubJunk 

Browser-to-UMS was there to control renderers from player.
Is it still needed as we can do the same from renderers tabs ?

Subtitle translator is a tool. Is it needed ? And if yes, it should reside on the gui part, not the player one.

SubJunk:
@SurfaceS good questions. I guess it is useful for users to be able to send media to their renderer from the player, since that gives them the benefit of using our web player interface if they like it more than their renderer's one. Some renderers have very ugly DLNA/UPnP GUIs

The subtitles translator could either be removed (I'm not sure if it is maintained?), or split into its own project

SubJunk:
I'm working on the error on Player page load

SubJunk:
@UniversalMediaServer/developers  maybe we are ready to merge this? If users have problems with it they still have the Java GUI for now.

Also last time we did alpha and beta releases we didn't get much useful feedback, so I am thinking of doing the v12 release without betas. Any thoughts?

SurfaceS:
FYI, on the current code : 
GUI is on the old interface port (9001)
Player is on a new 9002 port by default (can be changed in GUI settings).
Player is not auth by default, and will allow download and remote controls (can be changed in GUI settings).
Player can be set to auth from the GUI accounts with perms.

SubJunk:
There are some minor design tweaks I want to make before merging this to main and releasing it as v12. I think we are very close to that now.

SurfaceS:
@SubJunk @js-kyle 
Why did we use BCrypt to cryt account password in db ?

It add at.favre BCrypt dependency, when we already have more robust crypting methods (like SHA512) in the apache commons.codec dependency.

js-kyle:
My understanding is that bcrypt is the most widely used for password hashes. SHA* hashes are designed for performance which is not great for passwords as they can be cracked much faster.

Would suggest we use one recommended by owasp https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html

SubJunk:
@SurfaceS I can't build this branch on macOS anymore, with the error `Project 'ums' is missing required source folder: 'src/main/external-resources/nsis'` any ideas?

SurfaceS:
> @SurfaceS I can't build this branch on macOS anymore in VS Code, with the error Project 'ums' is missing required source folder: 'src/main/external-resources/nsis' any ideas?

nsis is windows related, not macOS.
It's now located in src/main/external-resources/windows/nsis

Maybe your personal VS Code files (not included in PR) are looking for nsis ?

ik666:
@SurfaceS  @SubJunk I have an equal issue on MacOS. If you import the project into Eclipse, you can not start the application by running "UMS.launch" any more. I think the folder used to be named `UniversalMediaServer` now it's `ums`

therefore in the UMS.launch I could correct this with 

```
-    <stringAttribute key="org.eclipse.jdt.launching.PROJECT_ATTR" value="UniversalMediaServer"/>
+    <stringAttribute key="org.eclipse.jdt.launching.PROJECT_ATTR" value="ums"/>
```

or in the maven pom.xml the `artifactId` has to be `UniversalMediaServer` again ... I didn't recap who made the changes though ...

SurfaceS:
@SubJunk extra.nsh NSIS file was originally set as a global resource to filter its content.
On the last PR, I've moved it to the windows builder resource only.
It should fix your Mac error

SubJunk:
Nice one, thanks

Issue #2709, opened 2021-11-23
NullPointerException in browseSearchHandler


SubJunk:
I got this from some logs for version 10.10.1 from https://www.universalmediaserver.com/forum/viewtopic.php?f=4&t=14809
Not sure if it has been fixed since then but I wanted to make a record of it in case it hasn't been. Any ideas @ik666 ?

```
DEBUG 2021-09-05 12:22:50.636 [HTTPv2 Request Worker 2] net.pms.network.RequestHandlerV2 Caught exception: null
TRACE 2021-09-05 12:22:50.637 [HTTPv2 Request Worker 2] net.pms.network.RequestHandlerV2 
java.lang.NullPointerException: null
	at net.pms.network.RequestV2.browseSearchHandler(RequestV2.java:1176)
	at net.pms.network.RequestV2.browseHandler(RequestV2.java:1105)
	at net.pms.network.RequestV2.answer(RequestV2.java:687)
	at net.pms.network.RequestHandlerV2.writeResponse(RequestHandlerV2.java:414)
	at net.pms.network.RequestHandlerV2.messageReceived(RequestHandlerV2.java:270)
	at org.jboss.netty.channel.SimpleChannelUpstreamHandler.handleUpstream(SimpleChannelUpstreamHandler.java:70)
	at org.jboss.netty.channel.DefaultChannelPipeline.sendUpstream(DefaultChannelPipeline.java:564)
	at org.jboss.netty.channel.DefaultChannelPipeline$DefaultChannelHandlerContext.sendUpstream(DefaultChannelPipeline.java:791)
	at org.jboss.netty.handler.stream.ChunkedWriteHandler.handleUpstream(ChunkedWriteHandler.java:142)
	at org.jboss.netty.channel.DefaultChannelPipeline.sendUpstream(DefaultChannelPipeline.java:564)
	at org.jboss.netty.channel.DefaultChannelPipeline$DefaultChannelHandlerContext.sendUpstream(DefaultChannelPipeline.java:791)
	at org.jboss.netty.handler.codec.http.HttpChunkAggregator.messageReceived(HttpChunkAggregator.java:145)
	at org.jboss.netty.channel.SimpleChannelUpstreamHandler.handleUpstream(SimpleChannelUpstreamHandler.java:70)
	at org.jboss.netty.channel.DefaultChannelPipeline.sendUpstream(DefaultChannelPipeline.java:564)
[...]

PR #3035, opened 2022-06-07
Fixed lazy initialization error

ik666:
I think I got the (annoying) issue with uninitialized root folders fixed. Looked like a broken double check idiom ... 

If the root folder wasn't found resources like thumbnails couldn't be loaded ...


SubJunk (approved):

SubJunk:
I don't fully understand the problem but this change seems logical. Nice one

ik666:
I strongly believe it is this issue : #2709 which is fixed now.

PR #3036, opened 2022-06-07
web settings iam

SurfaceS:
create folder iam (Identity and Access Management)
moved logic into it
rebase functions related to database under their own table class
added groups
added group permissions
modify users to allow groups assignement
modify users to avoid brute force attack
escaping slash as json fail
removed text response as json expected


js-kyle (approved):

js-kyle:
This looks great. Are you planning to build an RBAC management UI on the web settings interface (just to check we don't double up work?)

SurfaceS:
No, it's just a clean/fix for part. I'm not touching the web settings interface (REACT) for now.

The brute force preventer is because some users use their interface server over internet, even if UMS is not (yet) intended to.
The IAM part is to allow RBAC for settings, gui, and may be use to have custom media folders by group of users, what is requested by some users.

The default user `ums`, id 0 in db, could be renamed, but should not be deleted as it is the admin user that have the wildcard permission. Same for the default group (id 0).

I may be wrong, but I think all permissions should sit under `net.pms.iam.Permissions`.
For exemple, you can add `public static final String REBOOT_SERVER = "reboot_server";`
Then you can ask if an account have the permission with `account.havePermission(Permissions.REBOOT_SERVER)`.

Permissions are group based.

For now, we have only the `ums` user under the `admin` group, and permission for it is `ANY`, so it's not needed to add the permission for it.

PR #3041, opened 2022-06-09
Websockets

js-kyle:
Adds a websockets implementation, to be used with the new UI


SurfaceS:
We need to know if we want to use `websockets` or `AJAX/SSE` for communication between the browser and UMS.
Both have pros and cons.
`websockets` have bidirectional support, but require its own service/ip port, authentication, etc, and is a bit more complicated to maintain/debug.
AJAX/SSE use the HTTP(S) service already in place (with auth done), but use AJAX to post/receive data and SSE to receive event from the server.

Exemple : Browser side, ajax post setting change. The server response let him know if it success.
On setting change, the server send a message to all SSE connections that a setting have changed.
Exemple 2 : New renderer founded -> the server send a message to all SSE connections that this happen with the renderer detail.

So the question is : do UMS will use the bidirectional communication from the server (the server ask the client for something).

I'm not telling we should use AJAX/SSE, I just say we need to think about it to take the good direction before implementing it.

js-kyle:
I don't have experience with SSE, but reading into it, it does look like it would be fit for purpose for what we are doing at this stage. I'll have a look at implementing this same Online Status feature using that method to compare

SubJunk:
It's a good point @SurfaceS thanks
I have a lot more experience with websockets than SSE but it's worth considering which option fits best for us. Keen to see that comparison @js-kyle

PR #3044, opened 2022-06-10
web-settings Server Sent Events test

SurfaceS:
added Server Sent Events impl.
added user id to jwt (as username can be changed during session).
centralized users under the interfaceserver.
fix typos.

You can then test SSE.

I added few func to allow you to broadcast a message thru `WebInterfaceServer.broadcastMessage(String message)` and `broadcastMessage(String message, String permission)` (that will allow UMS to not send data if the user doesn't have a permission).
message should be a JSON string as the test online use json.

Also, the user is invalid if the browser session keep it and UMS was down, as interfaceserver keep track of users.


js-kyle:
This is looking nice, but keen to clarify something

> Also, the user is invalid if the browser session keep it and UMS was down, as interfaceserver keep track of users.
Why is the session being invalidated? 

The intention of the websocket implementation was to allow the front end to have a realtime status updates if UMS is restarted for whatever reason, but this now doesn't work due to the 403's being returned

SurfaceS:
403 will appear only if UMS was killed (the application).
If a user use the restart server, it's not.

Message is received then `onMessage(JSON.parse(e.data))` parse the message as json. (maybe you doesn't want to).
On the `function onMessage(data: any) {` data should be the json object from your message.

I'm away for few days, so I'll not be able to help during this time.

js-kyle:
@SurfaceS I was attempting to use this feature to send memory use updates to the front end (just as an example to test), but no events were being sent to the browser. When I am attempting to broadcast messages (with an active SSE connection) this HashMap seems to be empty, so no events are sent https://github.com/UniversalMediaServer/UniversalMediaServer/pull/3044/files#diff-26fa09fb9ac981af3796ef477b386541e1c29f9c83f0cee5b8b48466ed37f8b7R112

So I was probably using it wrong. If possible are you able to add an example usage, of any server event sending a message to FE?

I do not follow the reason why a 403 is returned, when the front end attempts to connect when the program is restarted.The user has a valid JWT expiring in the future, and should be allowed access?

SubJunk:
@SurfaceS @js-kyle  this branch is falling behind, how should we resolve it?

js-kyle:
I am more inclined to use this approach opposed to websockets but my only hesitation is the 'issue' above where if the program is restarted, an attempt to make the sse will return a 403. I would not expect this behavior

It would be good to get a real example of this approach in use, such as a live memory use update

SurfaceS:
> my only hesitation is the 'issue' above where if the program is restarted, an attempt to make the sse will return a 403. I would not expect this behavior

It's no more the case, I modified the JWT session few days ago to handle a better compromise between security and comfort.
The IP included in it will allow us to confirm SSE request is well requested by the good IP.
With that, we will be able to recognize the user and confirm it's identity.

> It would be good to get a real example of this approach in use, such as a live memory use update

I've try react only for 1 week, so I'm not comfortable coding with it for now.
But I'll try to do something.

js-kyle:
Great, I think is looks like a good approach to move ahead with

Issue #3046, opened 2022-06-10
Web settings: Remove initial user


SubJunk:
This is a task to complete for https://github.com/UniversalMediaServer/UniversalMediaServer/issues/3004

At the moment there is a default user `ums` with a default password that is changed manually on first login. This allows anyone to create the initial user, but also requires hidden knowledge of the username and password. 

Instead, it would be just as secure and more user-friendly to have no default user. The first time someone goes to the configuration section they are greeted with a form to create the initial user, with any username or password.

SurfaceS:
I add some code #3063 to allow that but I'm not confident with react, so I cannot do a lot on this side.

When react app will handle the session data, we can delete the user creation on db creation.

I think that the "change password" should be replaced by "accounts", with 2 tabs on the page ("users", "groups").
Then on the "users" tab, a user can change his (or other if he have permission) password, username, display name, group, etc.
And permissions, name on the "groups" tab.

This also may enable in the future, the ability to assign custom shared content by group, as some users have requested.

PR #3063, opened 2022-06-17
web settings users

SurfaceS:
react :
added first login page
added session context
fix apple-touch-icon

java :
replace token username for user id
no admin handle/creation
keep logged users for server side check
action/renew token only if logged
few fix/log
auto-increment start with 1, not 0

I doesn't know react, so I was not able to correctly code it to handle the session state in App.tsx
So I commented on the file, if someone know how to handle this.
For now, I leave the old way (only check if token exists).

I follow the https://github.com/UniversalMediaServer/UniversalMediaServer/issues/3046.
If no admin found in db, then it'll reply session.isFirstLogin to true (so even if someone delete himself he will be able to recreate an admin account).


SubJunk (approved):

SubJunk:
@SurfaceS this looks great, thanks. I can't push to your fork so I'll merge it now so I can try to finish the react changes

PR #3078, opened 2022-06-24
web settings accounts

SurfaceS:
added Accounts Users/Groups implementation

removed ChangePassword
users db fix
added check on view/modify settings perms

next step will be to give the possibility to manage permissions for groups.

> Strange things is that when USM running, the React Router seems to not handle the url (we fall into webinterface home page as the url is not found). The links under `UserMenu` are not routed.
`/accounts`, `/settings`, and old `/changepassword` never worked on my side (I needed to cheat the `<Route index element={<Settings />} />` on `App` to have it on main page.
I think that routes are handled only under the `Router` element only, and as the user menu is outside, it doesn't work properly.

router fixed


js-kyle:
This looks good, I noticed that assigning a user's group does not appear to be working correctly - is that still to be built?

PR #3080, opened 2022-06-25
tooltips

js-kyle:


configuration-client/src/components/Settings/Settings.tsx
```diff
@@ -171,14 +172,14 @@ export default function Settings() {
                 sx={{ flex: 1 }}
                 {...form.getInputProps('server_name')}
               />
-
-              <Checkbox
-                mt="xl"
-                label={i18n['NetworkTab.72']}
-                {...form.getInputProps('append_profile_name', { type: 'checkbox' })}
-              />
+              <Tooltip label={getToolTipContent(i18n['NetworkTab.73'])} width={350} color="blue" wrapLines withArrow>
```
SubJunk, line 175: If the intention is to always use the options `width={350} color="blue" wrapLines withArrow` you could possibly use the spread operator to pass those in as a e.g. `defaultTooltipOptions` object, like we do with `form.getInputProps` in `Settings.tsx`

SubJunk (approved):
This looks good, I left a suggestion but not important

PR #3085, opened 2022-06-25
web settings DirectoryChooser fixes

SurfaceS:
> enforce security to /directories

Do not allow non logged users to browse your disk.

> moved parseQueryString to WebInterfaceServerUtil

Function may be use by all handlers.

> set getDirectoryResponse as static.

> getSubdirectories fix URISyntaxException thrown on windows

Get request with unescaped URI throw URISyntaxException.

> getSubdirectories use post instead of get.

Easier to create/handle the request.

> refactor parents to childrens, and parent to childrens to reflect the reality.

directory parents are directory container, not children.

> fix file separator for breadcrumbs on windows.

Use the correct file separator for all systems.

> added roots to enable select/browsing roots.

Added button at start to show the roots.
Before we cannot select a root folder (like "/" or "c:") and cannot browse other roots (like "e:")


SubJunk:
This looks good, I'm merging into branch on ums repo so I can work with it

PR #3087, opened 2022-06-25
web settings fix select

SurfaceS:
> fix select need string, not number

`Select` Mantine does not handle correctly the value.
Workaround by sending stringed numbers.

> back to default value if empty

If react receive empty string for setting, it will not back to default value as UMS does.

> real numberOfCpuCores handle

This was hard coded, and permit 0 value (UMS doesn't).


js-kyle:
This looks good

On this branch the `number_of_cpu_cores` form value is no longer showing the user value (just empty), though the values save if changed. I would expect the label to have a maximum of 8 and input populated with 3 for the below case

image

SurfaceS:
fixed

js-kyle (approved):

PR #3090, opened 2022-06-25
enforce security to /directories.

SubJunk:
moved parseQueryString to WebInterfaceServerUtil.
set getDirectoryResponse as static.
getSubdirectories fix URISyntaxException thrown on windows.
getSubdirectories use post instead of get.
refactor parents to childrens, and parent to childrens to reflect the reality.
fix file separator for breadcrumbs on windows.
added roots to enable select/browsing roots.

### DIFF
--- a/configuration-client/src/components/DirectoryChooser/DirectoryChooser.tsx
+++ b/configuration-client/src/components/DirectoryChooser/DirectoryChooser.tsx
@@ -71,19 +71,20 @@ export default function DirectoryChooser(props: {
           </Group>
         }
         overflow="inside"
+        size="lg"
       >
-        <Box sx={{ maxWidth: 700 }} mx="auto">
+        <Box mx="auto">
           <Paper shadow="md" p="xs" withBorder>
             <Group>
-              <Button
-                loading={isLoading}
-                onClick={() => getSubdirectories('roots')}
-                variant="default"
-                compact
-              >
-                <Devices2 />
-              </Button>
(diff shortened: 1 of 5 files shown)


SubJunk:
These changes look good. I need to change the root browsing part on macOS. Also I'm not sure about the change from GET to POST, since it seems like a GET to me. I know any request protocol can technically be used for anything, but think it's nice to be able to look at the protocol and know something about the route - a get gets, a post posts, a patch patches, a delete deletes, etc.

SubJunk:
@SurfaceS These changes look good. I need to change the root browsing part on macOS since it doesn't work the same as Windows, I'll also test on Windows to make sure they agree.
Also I'm not sure about the change from GET to POST, since it seems like a GET to me

SurfaceS:
I'm on windows, that's why I made these changes.
About the GET POST question.
I think you are confusing the goal of the request (Browser want data back, Browser just send data) and the method the request will used to send data.

Both POST and GET can send data.
With POST, you are not limited in what data you send, and the content itself is encrypted if over an HTTPS connection.
With GET, you are limited to ASCII, and the query is part of the URL, so not encrypted. GET can be cached, meaning you will have an outdated response if the browser take the cached one.
And so on.

[httpmethods](https://www.w3schools.com/tags/ref_httpmethods.asp)

Basically, I use GET only when their is no query, when I want to use the cache (js, css for ex).

PR #3092, opened 2022-06-26
Web settings languages

SurfaceS:
Change language on the fly.
Set default to user browser language if none defined.
Always use the one defined by the user, regardless the server language set.
Automatic RTL switch.
I18n on static values from server.


SurfaceS:
@UniversalMediaServer/developers  this is ready to review/try/merge

SubJunk:
It works well, there are some small problems that I'll fix after merge. Nice work this is really nice!

SubJunk (approved):

PR #3095, opened 2022-06-27
3004 web settings defaults

SubJunk:
Define defaults on the server so we have one list to maintain instead of two.


SubJunk:
@UniversalMediaServer/developers this is ready to review/merge

SubJunk:
Er, I mean now

SurfaceS:
I was searching why the `defaultConfiguration.number_of_cpu_cores` was always set to `settingsResponse.userSettings.number_of_cpu_cores`, despite that the json received from the server have good values on it from default and user.

I found that with `_.merge(settingsResponse.userSettingsDefaults, settingsResponse.userSettings);`, `settingsResponse.userSettingsDefaults` holds the destination object, so it is modified and it's not the intended behavior.

`_.merge({}, settingsResponse.userSettingsDefaults, settingsResponse.userSettings);` solve this by creating a new object to hold the destination object.

This fix is included in the #3102

PR #3102, opened 2022-06-28
3004 web settings defaults suggestion

SurfaceS:
just a suggestion for #3095

it use Map instead of JsonObject
it do not add valid empty key instead of adding then delete
if fix _.merge userSettingsDefaults as destination bug


SubJunk:
Nice fixes. I’m curious about why the change to Map?

SurfaceS:
> I’m curious about why the change to Map?

I've always preferred basic Java objects to heavyweight objects, but upon investigation, JsonObject isn't a big deal.

PR #3103, opened 2022-06-28
Feature/search caps

ik666:
This is a follow up to #2624. Since that merge, BubbleUPnP doesn't do any server side searches on UMS, except server side search is forced in BubbleUPnP configuration. Since some might like this feature, I'd recommend to enable SearchCaps again making it configurable.

I think there were some renderer that had issues with active search caps. Therefore, for compatibility purposes responding to upnp "GetSearchCapabilities" requests can be reverted to the old logic by setting PMS configuration key `upnp_search_caps_enabled` to `false`. Default is `true`.


ik666:
ups, again on wrong branch

PR #3113, opened 2022-06-30
3006 web settings Server Sent Events base

SurfaceS:
A base from #3044
Test things removed


js-kyle:
Not sure if this is ready for review yet, but on this branch I don't see any messages logged in ServerEventProvider when adding debugging/breakpoints. I have the SSE connection open however

SurfaceS:
If you add breakpoints on `if (event.event === "message") {`, it will stop each time a message is transmitted from the server.
As the memory usage updater is running, it should stop each 2 seconds.

js-kyle:
Hmm, ok I just realised that it works when I access the UI from `http://192.168.68.52:9001/`, but it does not work when running the ui in development mode using `yarn start`, which runs the UI at `localhost:3000`. Not sure why that is different. But I can try to work that out after merge

js-kyle (approved):

SurfaceS:
Should be ok to merge now.

Issue #3099, opened 2022-06-27
Translation keys


SubJunk:
Our source language file https://github.com/UniversalMediaServer/UniversalMediaServer/blob/b69a5f0d07984ff20ad7890120e608df3a09cef0/src/main/resources/i18n/messages.properties is a mess, as one would expect from 14+ years of translations without refactoring.
I propose it would be nicer to have simple keys that are an abbreviation of the value. I do that on another large project and it is more readable IMO.

What that means in reality is that this:
`FoldTab.UseInfoFromAPI=Use info from our API`
Would become:
`UseInfoFromAPI=Use info from our API`

Exceptions can be made when there is an explicit grouping or important context, like the `Language` section https://github.com/UniversalMediaServer/UniversalMediaServer/blob/b69a5f0d07984ff20ad7890120e608df3a09cef0/src/main/resources/i18n/messages.properties#L244-L283 but in general I think we make things more confusing instead of less, by adding unnecessary context.

This also gives us a good chance to get rid of the unreadable keys. So this:
`NetworkTab.14=Could not install the Windows service. It is probably already installed.`
would become:
`CouldNotInstallWindowsService=Could not install the Windows service. It is probably already installed.`

@UniversalMediaServer/developers any opinions on it?

Obviously there is the problem of all of the existing translations. I had a look in Crowdin to see what their support is for renaming keys, and I can't see any, so there will possibly be a manual task to update them. I can do that.

SurfaceS:
I agree that the source language file is a mess.
Especially because of the unreadable numbered keys.

Grouping have pros and cons.
This makes it easier for us to find if there is a word/phrase that would say the same thing.
On the other hand, sometimes we recreate a new entry because it is not based on the same group.

The thing to be careful is : 
For the same word/phrase in English, you can have several words/phrases in another language, depending on the context.

I don't know about Crowdin's renaming keys.

Nadahar:
@SubJunk There are more problems, but I don't know if you use all the features. If you use Crowdin "branches" (the crowdin plugin supports making Crowdin branches for git branches), Crowdin won't be able to correctly "merge" them when you merge the git branches. Also, you will lose comments, explanations, context information etc. provided for the strings you delete.

[...]

PR #3123, opened 2022-07-01
Renamed translation keys

SubJunk:
Implements https://github.com/UniversalMediaServer/UniversalMediaServer/issues/3099
This is incomplete, just making progress public

I have two sets of the translation files committed - there are the files we use, and then the files in the format/directory structure downloaded directly from Crowdin instead of via the Maven plugin. I'm converting them both. When it's all complete I will reupload the latter group to Crowdin


js-kyle (dismissed):

SubJunk:
The update seems to have gone well on Crowdin. I'll run the maven plugin on this branch today and if that goes well I'll merge this

PR #3126, opened 2022-07-02
3004 web settings about

SurfaceS:
Added about page


SubJunk (approved):

SubJunk:
Good idea to add that memory info to the About page

PR #3143, opened 2022-07-07
Fixed macOS version logic

SubJunk:
The version matching didn't account for versions of macOS higher than 10. This should fix sleep logic and false warnings


js-kyle (approved):
I don't get the incorrect log anymore

PR #3148, opened 2022-07-07
web settings impl opt out auth

### LINKED ISSUE #3142: Web settings: Option to opt-out of authenticated access
On the first page when a user is given the option to create their admin user, there should also be an option to opt-out. Think of a scenario like a person living alone on a private network, or just someone who doesn't care about such things.
We should encourage the use of auth, and default to it, but not enforce it IMO.

SurfaceS:
fix #3142

3 choices :
Disabled authentication = every one is admin.
Enabled authentication = every one need to login.
Enabled authentication + Auto localhost = every one need to login, except localhost that will be admin.


SubJunk:
Good idea on the third option, nice work

PR #3154, opened 2022-07-10
PmsConfiguration clean

### LINKED ISSUE #3155: User renderer's config
For now, the renderer configs are loaded from `"project.renderers.dir"` from `project.properties`.
This end in the `renderers` folder inside app directory.

Problem is that this folder, in most of OS, will be read only.

For the databases or logs, we use the profile directory to be able to write on it.

It may be a good idea to have the renderer configs provided by UMS in the current folder (under app folder), and the "custom" user config under the profile directory.

SurfaceS:
sort `NEED_RELOAD_FLAGS` and add goal for reload improvement (future PR).
fix multiple de referencing possible null pointer
set `final` to members called in constructor.
suppress use of obsolete `Vector` class from the `PmsConfiguration`
remove deprecated functions not used anymore
set `TempFolder` public as it is exposed

2nd commit
same than before (clean/fix) for `RendererConfiguration`.
Implement renderer config in profile directory.
fix #3155


valib:
I think that when renderers are discovered after reset than the each renderer network speed should be tested.

SubJunk:
Good idea, that's a safe time to do it

SubJunk (approved):

SubJunk:
Reviewed and tested. I can't find any bugs with it and it fixes a bug that was annoying me, which was a lingering reference to folders after removing them, even after clicking restart. Nice work @SurfaceS !

PR #3163, opened 2022-07-13
opt in sending date metadata for nextcp/2 renderer

ik666:
Opt in sending `dc:date` metadata.

@SubJunk Maybe you can help me to find out UMS's strategy for implementing new features. I see the aspiration to disable features by default, only because there are renderer out there in the world that might not get along with (even spec conform) data structures or response objects. Wouldn't it be more understandable to enable new features by default, and identify the bogus devices and opt that feature out for the specific renderer?

### DIFF
--- a/src/main/external-resources/renderers/Nextcp2.conf
+++ b/src/main/external-resources/renderers/Nextcp2.conf
@@ -7,6 +7,8 @@ RendererName = NextCP2
 
 UserAgentSearch = nextcp
 
+SendDateMetadataYearForAudioTags = true
+
 # Supported audio formats:
 Supported = f:ac3                       m:audio/vnd.dolby.dd-raw
 Supported = f:acdpm                     m:audio/x-adpcm


SubJunk:
@ik666 yes maybe you're right, maybe the approach is too cautious and limiting for features

I think we also need to implement better UPnP support. We have a BIG gap in how we interact with renderers via UPnP, since UPnP is supposed to be mostly configless but we don't treat it like that. We need to make our code more dynamic so we can react to the ORG_PN capabilities etc. that the renderers send us. Maybe if we get better at that two-way conversation between server and renderer, some of these problems will go away without need for custom config changes.

But yes your feedback is good and maybe my approach there wasn't the best one

SubJunk (approved):

ik666:
@SubJunk I agree in putting more effort in UPnP compatibility. Maybe one cause of this issue is that UMS doesn't talk between it's own (UPnP) components by UPnP protocol (MediaServer & ControlPoint & MediaRenderer = web interface). 

Probably the "supported file format" part can be removed from renderer config file, because the internal UPnP control point already retrieves the `protocolInfo` from the rendere ...

SubJunk:
Hi @ExSport great to see you. I agree, we still need what we have to customize support. It would be good to have the best of both worlds. We have code to parse getProtocolInfo responses but it's not hooked up to our filetype or transcoding logic. Ideally I think we would use the reported capabilities automatically, and that would greatly improve our support for unknown and future devices, and then our renderer configs can override and extend those capabilities.

One thing to remember is that there are lots of renderers out there that support an older style DLNA logic, and that is what PS3 Media Server was written to support, so they have no way to communicate their capabilities. We only started supporting UPnP protocol in v5, and not only did we never finish it, they have also changed it along the way so it can be hard to keep up 😄 

I think users these days are moving away from UPnP, they prefer to either watch things in a web browser on their laptops/notebooks/tablets, or via dedicated apps on their phones or TVs. That's a reason a lot of my efforts lately are put into the web interface, and I have an iPhone app I am tinkering with. But UPnP is still very important, especially while we don't offer apps for TVs. I hope we can continue to modernize to match what current users expect, and this year I think we are making great progress towards that!

ik666:
@SubJunk Your general idea to be universal is great, keep that please up. That's why I'd like to disagree on 

> I think users these days are moving away from UPnP, they prefer to either watch things in a web browser on their laptops/notebooks/tablets, or via dedicated apps on their phones or TVs.

I think it depends on the use case and the opposite is also true, because why should someone watch a movie on a laptop or tablet (or iPhone), if most people have a huge TV with DOLBY or ATMOS equipment installed in the living room? Same applies to music: There are user out there with 10k+ EUR audio equipment installed. Usually they have other needs regarding UI & feature set.

TV's usually have a build-in UPnP control point, that is used with works just fine (at least with renderer.conf files). Audio equipment manufacturer usually have build in UPnP support and ship their own control points as iOS/Android apps like Linn, Lumin, Denon, etc. able to access UMS talking UPnP. At least I am on this side, and my intrinsic motivation is to support UMS from that point of view :)

SubJunk:
@ik666 yes I agree, and I am closer to that too, I use UMS on my TV via the generic UPnP Media Server app. It is more younger people who often don't buy TVs, they seem happy to just have smaller screens and are less concerned with quality.

We should support all uses :) I guess I am focusing more on this area lately because our support for it isn't as good. I want us to do everything ;)

PR #3167, opened 2022-07-17
Main api json

SurfaceS:
use JsonObject instead of string to json to hashmap to json to string
fix bad stored values


SurfaceS:
sure

SubJunk (approved):

SubJunk:
Nice one, much nicer

PR #3173, opened 2022-07-18
added new database backend: postgresql

ik666:
Hi guys,

I've added a new database backend for Postgresql. As a side effect the connection pool is switched to "Hikari".

Purpose: tremendous speed improvement in database heavy actions like:
- Scanning folder (Library) (speed up initial scanning by magnitudes)
- Entering playlist files with many entries (>100) (from seconds to 'instant like access')
- UPnP search (from many seconds to milliseconds now)

People who used to wait for UMS to respond will get a smile on its face.

I think getting the database out of UMS makes it much more lighter. UMS feels more responsive now on my system. Probably this is still work in progress since I've only tested the audio part, because I don't have many video files.

By default, h2db is used. I've changed some h2db types for better compatibility. This has also to be double checked. 

Setup / configuration procedure for postgres:

1. Install postgres and login as a superuser
2. create database and user `ums` (and an optional test database)
```
create database ums;
create database ums_test;
create user ums with encrypted password 'ums';
grant all privileges on database ums to ums;
grant all privileges on database ums_test to ums;
```
3. grant access to user ums by editing postgres `pg_hba.conf` file
3.1 for remote access
```
host    ums             ums             192.168.1.0/24        md5
host    ums_dev         ums             192.168.1.0/24        md5
```
3.2 for unix sockets (this is recommended, UMS and postgres must be running on the same machine)
```
local   all             all                                     peer
```

4. configure database driver
4.1 for remote database server
```
database_backend = pg
database_user = ums
database_pass = ums
database_url = jdbc:postgresql://IP_ADDRESS/ums
```
4.2 for unix sockets connection

```
database_backend = pg
database_pass = ums
database_url = jdbc:postgresql://localhost/ums
database_user = ums
database_socket_factory = org.newsclub.net.unix.AFUNIXSocketFactory$FactoryArg
database_socket_factory_arg = /var/run/postgresql/.s.PGSQL.5432
```

Adapt unix socket file path in property `database_socket_factory_arg` to your installation. Ubuntu saves the socket file in the `/var/run/postgresql` folder. 

Comments welcome.

### DIFF
--- a/src/main/java/net/pms/database/Database.java
+++ b/src/main/java/net/pms/database/Database.java
@@ -83,35 +83,39 @@ public abstract class Database extends DatabaseHelper {
 	public Database(String name, String user, String password) {
 		logger = LoggerFactory.getLogger(Database.class);
 
-		PmsConfiguration c = PMS.getConfiguration();
-		HikariConfig config = new HikariConfig();
-		config.setUsername(c.getDatabaseUser());
-		config.setPassword(c.getDatabasePassword());
-		config.addDataSourceProperty("cachePrepStmts", "true");
-		config.addDataSourceProperty("prepStmtCacheSize", "250");
-		config.addDataSourceProperty("prepStmtCacheSqlLimit", "2048");
+		PmsConfiguration pmsConfig = PMS.getConfiguration();
+		HikariConfig hikariConfig = new HikariConfig();
+		hikariConfig.setUsername(pmsConfig.getDatabaseUser());
+		hikariConfig.setPassword(pmsConfig.getDatabasePassword());
+		hikariConfig.addDataSourceProperty("cachePrepStmts", "true");
+		hikariConfig.addDataSourceProperty("prepStmtCacheSize", "250");
+		hikariConfig.addDataSourceProperty("prepStmtCacheSqlLimit", "2048");
+
+		logger.info("using {} database backend ", PMS.getConfiguration().getDatabaseBackend());
 
 		if (isH2dbBackend()) {
 			h2dbInit(name, user, password);
-			config.setJdbcUrl(c.getDatabaseUrl() + ";CACHE_SIZE=131072");
+			String h2url = pmsConfig.getDatabaseUrl() + ";CACHE_SIZE=131072";
+			hikariConfig.setJdbcUrl(h2url);
+			logger.info("database url : {} ", h2url);
 			dbTypes = new H2dbTypes();
 		} else if (isPostgresBackend()) {
-			config.setJdbcUrl(c.getDatabaseUrl());
-			PMS.get();
-			PmsConfiguration pmsConfig = PMS.getConfiguration();
+			hikariConfig.setJdbcUrl(pmsConfig.getDatabaseUrl());
 			dbTypes = new PostgresTypes();
 			if (!StringUtils.isAllBlank(pmsConfig.getDatabaseSocketFactory())) {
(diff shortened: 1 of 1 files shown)


SurfaceS:
PostgreSQL is not known to be more efficient than H2.
On the paper, client/server model can't be more efficient than embedded, in the way that it use socket or so to communicate, and that add steps. Plus you need to maintain the db server, state, etc. that most people doesn't know/want.

You see a big gap just because PostgreSQL use it's own cache memory and settings, and use memory indexes.
In fact, used memory is JVM memory + PostgreSQL footprint + PostgreSQL db footprint.
If you allocate a good memory amount to H2, and customize the H2 settings, you should have about the same results or better.

By default, H2 doesn't use a lot of cache (16 MB if I remember), ending in many disk drive i/o (we have thumbnails blob in db that fill that in 1 request).

Also, in H2, default tables are Cached tables : datas and indexes are persistent on disk drive. With that, you don't need to have a system with a lot of memory, it run everywhere.

H2 also provide the Memory tables : datas are persistent on disk drive, but indexes live in memory. The cons is that it need to have some memory and required size will grow with the table size (like PostgreSQL).

So, to really compare both solutions, you should set a cache value to H2, and use CREATE MEMORY TABLE... instead of only CREATE TABLE.

In my opinion : 
PMS Database should not use custom engine but only `java.sql.*`
Your investigation show that SQL queries doesn't use standardized SQL / or standard Jdbc protocol => It should.
`org.h2` should be moved away from the common db files.
Media Database should be cleaned/split a lot.
H2 may use cache and/or Memory tables to be more efficient.

ik666:
@SurfaceS In theory for some artificial benchmarks you might be right. However, if you look at the h2db database file size, which is constantly growing over time to some GB, I believe h2db is not doing any maintenance tasks (recover disk space, update optimizer statics, rebuild indexes, ...). Besides that h2db full text searches doesn't use an index (current config) unless you use some h2db extensions (which are not configured). I can go along with adding some cache_size parameter for performance tuning. Not sure about memory table ... docs say, table should not get to large, whatever that means ... memory footprint seems not to be predictable which might be an issue over time.

Regarding moving to standards: As much as I'd agree here, I doubt a database independent solution can be made for all specific requirements.

SurfaceS:
@ik666 I don't say it's a bad idea to add engines possibility.
I just say that it will need implementation, support, update, help users on install, etc.

The current database need a full review / optimization. This is not done, so it's hard to begin a new implementation from it.

99% of users doesn't know what is a DB, and will not install a DB engine, start the DB service, maintain it, just for UMS.
The 1% left will not agree to be linked to a db engine (PostgreSQL in that case) and will ask support for MariaDB/My SQL, or any other, because they already have a DB server running.

That why I said we should move to standards SQL and JDBC usage, then the jdbcXXX.jar can be loaded dynamically without adding anything to the code.

ik666:
@SurfaceS Generally speaking, optimizing the h2db database is a good idea. I made your proposed changes and will do some testing. I also believe people don't care about the database (neither did I) unless it has an impact on the usability. 

Regarding the standard: As I mentioned some features like backup or restore a db-table have no ANSI SQL expressions. Optimized h2db full text search relies on non standard extensions and so on ... same "issue" OR mapper have: they all need individual "language"-files for each supported database ...

SurfaceS:
I agree.
That's why we don't use fulltext search or backup/restore.

ik666:
super seeded by #3177

PR #3174, opened 2022-07-19
UPnP search disable configuration item moved to MediaRenderer Configuration

ik666:
After the discussion in #3106 I think the configuration option for disabling UPnP search should be placed in MediaRenderer configuration, because it probably effects only a few devices.

UPnP search can be disabled for a specific renderer device by setting 
`upnp_search_caps_disabled = true`
default `false`.

### DIFF
--- a/src/main/external-resources/renderers/DefaultRenderer.conf
+++ b/src/main/external-resources/renderers/DefaultRenderer.conf
@@ -113,7 +113,6 @@ UpnpDetailsSearch =
 # UPnP search capabilities. In case a renderer has issues browsing or searching
 # content, set this property to false.
 # Default: true
-
 UpnpSearchCapsEnabled = 
 
 # Optional additional HTTP header for better detection. When defined, UMS also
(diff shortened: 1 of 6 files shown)


src/main/external-resources/renderers/DefaultRenderer.conf
```diff
@@ -109,6 +109,13 @@ UserAgentSearch =
 # Default: ""
 UpnpDetailsSearch = 
 
+# If UpnpSearchCapsEnabled is set to true, UMS will notify Media Renderer of available
+# UPnP search capabilities. In case a renderer has issues browsing or searching
+# content, set this property to false.
+# Default: true
+
+UpnpSearchCapsEnabled =
```
SubJunk, line 117:
```suggestion
# Default: true
UpnpSearchCapsEnabled = 
```

SubJunk (changes requested):
Just minor formatting

SubJunk (approved):

PR #3175, opened 2022-07-20
PR for speeding up h2db

ik666:
This is a follow up to get better h2db database performance. This branch contains the h2db changes from #3173.


SurfaceS:
I created a UMS branch for that, so we can collaborate on the same code

PR #3177, opened 2022-07-20
PR for speeding up h2db

SurfaceS:
#3175 from @ik666


SurfaceS:
As I can see, the table type (`MEMORY`/`CACHED`) is only set on table creation.
This means that the table type change will not be taken into account after creation.

I need to find how we can solve that.

ik666:
Maybe for such a a significant change it's ok to rebuild all tables by deleting the database ...

More complex : 
- Otherwise dump all tables just as `LikeMusic` does.
- delete all tables and create new ones (MediaDatabase has already methods for that), 
- and restore the content

SurfaceS:
@ik666  I think it was corrected on https://github.com/UniversalMediaServer/UniversalMediaServer/pull/3177/commits/7970fb3d7b528968fac6915ac1ba3a66c0a325cd

setDatabaseMediaCacheSize : MB
CACHE_SIZE : KB (MB * 1000)

SubJunk:
> Maybe for such a a significant change it's ok to rebuild all tables by deleting the database

It is important that users don't lose their fully played data. People will be relying on that and we can't recreate it when it is lost.

ik666:
OK, then the dump & restore as in `LikeMusic` for the "liked album" table should do it ...

SurfaceS:
Dump will save the table storage type.
I'm on it to find a proper way to do this.

ik666:
I wasn't precise :)

````
Script.process(connection, backupFilename, "", "TABLE MUSIC_BRAINZ_RELEASE_LIKE");
`````
will write just the table content to `backupFilename`.
````
RunScript.execute(connection, new FileReader(backupFilename));
````
will restore the content ... 

I think tables can be dropped and recreated in between ...

**Comment by ik666** [comment]
yes, you're right, postgres just dumps the content :|

Maybe the dump script can be rewritten after creation ... to
`SET DEFAULT_TABLE_TYPE 0;` ?

**Comment by ik666** [comment]
I've done some test on my system. Timings for Library Scanning thread: current main takes around 3h CPU time, postgres branch uses 40 CPU minutes, this branch finishes in 26 CPU minutes. All in all this PR feals good. Well done and thanks at @SurfaceS 

Is this PR stable now ?

**Comment by SurfaceS** [comment]
I've just added soft cache.
I also review the cache size to better fit H2 recommendation (64Mb each GB of JVM memory).

### I read a lot the H2 code and I understand that :

**Default TABLE_TYPE is CACHED.**
CACHE_SIZE can be 0 -> disabled.
CACHE_SIZE should be 64 Mb each Gb of Memory. 
Default CACHE_TYPE is LRU (Least Recently Used)
LRU clean the cache memory regularly to let the maximum JVM memory free.

Another option for CACHE_TYPE is the SOFT_LRU.
It internally add a second cache that keep data if it was requested multiple time.
It does not clean the cache memory except in case it reach the CACHE_SIZE limit.
It is an option between the CACHED table and the MEMORY table.

**With TABLE_TYPE set to MEMORY :**
It does not care about CACHE_SIZE / CACHE_TYPE.
It work like LRU but will never clean the cache memory, even if JVM have no more memory.

### **My personal test :** 
Tested after waiting everything was loaded (upnp founded, network speed done, etc).
I don't use scan on startup so it is not already parsed.
DB is on SSD.

CACHE SIZE : let the system cache calculate (510 MB)
CACHE TYPE : normal cache
TABLE TYPE : CACHED tables
```
INFO  16:15:47.095 [Library Scanner] Library scan completed in 12 seconds
INFO  16:16:21.948 [Library Scanner] Library scan completed in 9 seconds
INFO  16:17:00.456 [Library Scanner] Library scan completed in 9 seconds
INFO  16:17:34.441 [Library Scanner] Library scan completed in 9 seconds
```

CACHE SIZE : let the system cache calculate (510 MB)
CACHE TYPE : soft cache
TABLE TYPE : CACHED tables
```
INFO  16:10:57.569 [Library Scanner] Library scan completed in 12 seconds
INFO  16:11:28.068 [Library Scanner] Library scan completed in 9 seconds
INFO  16:11:50.308 [Library Scanner] Library scan completed in 8 seconds
INFO  16:08:30.235 [Library Scanner] Library scan completed in 8 seconds
```

CACHE SIZE : let the system cache calculate (memory tables don't care)
CACHE TYPE : normal cache (memory tables don't care)
TABLE TYPE : MEMORY tables (switched then restart UMS to test).
```
INFO  16:20:54.749 [Library Scanner] Library scan completed in 12 seconds
INFO  16:21:24.591 [Library Scanner] Library scan completed in 9 seconds
INFO  16:21:50.935 [Library Scanner] Library scan completed in 8 seconds
INFO  16:22:21.673 [Library Scanner] Library scan completed in 8 seconds
```

CACHE SIZE : No cache (0 Kb)
CACHE TYPE : normal cache (no cache don't care)
TABLE TYPE : CACHED tables
```
INFO  16:58:17.808 [Library Scanner] Library scan completed in 293 seconds
INFO  17:03:38.567 [Library Scanner] Library scan completed in 289 seconds
```
I stopped the tests here as it was long :)

CACHE SIZE : 16 Mb cache (16384 Kb)
CACHE TYPE : normal cache
TABLE TYPE : CACHED tables
```
INFO  17:05:27.860 [Library Scanner] Library scan completed in 13 seconds
INFO  17:05:58.297 [Library Scanner] Library scan completed in 9 seconds
INFO  17:06:25.360 [Library Scanner] Library scan completed in 9 seconds
INFO  17:06:47.727 [Library Scanner] Library scan completed in 9 seconds
```

**Comment by ik666** [comment]
Nice analytics! So the recommended combination is regular tables with SOFT_LRU enabled for best memory regards?

**Comment by SurfaceS** [comment]
It really depends your need.
MEMORY TABLE seems dangerous for low memory systems.
But the option is there. So people can try settings then adopt it.

I'll try to set the Connection Pool back to H2 to see if it end with the same results.

**Comment by ik666** [comment]
ok thx. Regarding the connection pool, I believe Hikari is a good choice.

**Comment by SurfaceS** [comment]
Maybe Hikari is a little bit better. I'm talking by about 1 or 2% (10 to 20 ms).
But my runs take about 10 sec when yours take minutes.

H2 have it own Connection Pool. That's why I tested to try to not add a dependency when it's not needed.

I'll leave it with Hikari until someone want to remove the deps.
I move the connection creation to the Database class to have Hikari's call only on the Database class.

**Comment by SurfaceS** [comment]
@ik666 are you OK with this PR ?

**Comment by ik666** [comment]
Yes. This PR is very nice.  Speed improvements are really noticeable and UMS is very responsive to browse requests. Code is more organized. 

I'm using this branch since one day ... no issues so far from my side. Let's give it a try.

**Comment by SubJunk** [comment]
This looks great, thanks for all the testing and benchmarking!
I have it running now and will test it. The startup scan went fast, but for me it always went fast, not sure why people get such different experiences with it.
If I find no bugs I'll merge.

**Comment by SurfaceS** [comment]
@SubJunk 
> The startup scan went fast, but for me it always went fast, not sure why people get such different experiences with it.

Neither do I.
In fact, there are several reasons.

**Media count :** 
When most of your media files are movies, you have little media taking up a lot of disk space.
But if most of your media files are audio files or images, you have a lot of media that takes up little disk space.

So you can end with 100.000 rows in the table when others have only 1000.

**Database size :** 
The larger the table grows, the more memory it needs to cache the data and the longer it will take to parse it.
If the cache is not optimized, the database engine will have to switch between/swap indexed data and therefore have a lot of disk access.

**Database location :** 
If the database is on a normal HDD, it will take much longer than if it is on a SATA SSD (good) or M2 NVME SSD (better).

**Media File type and Tables :** 
As I said in another pull request https://github.com/UniversalMediaServer/UniversalMediaServer/pull/3161#issuecomment-1186449240, the database schema and table definitions are messy.
Like, the `MediaTableFiles` must contain fields related to a file, not to a specific media type.
And I found BITRATE (for image ?), WIDTH (for audio ?), CODECV, IMBD datas, and so on.
This means that whenever I have an audio file, it will have a row containing fields not related to the file, but for a video file type, TV series episode, IMBD datas, image type, etc.

It should not.

In the state, this table definition slow down data parsing because it load everything for each media type.

**Conclusion :**
I'm reviewing the database schema, but it takes time to get a solution that won't break everything.
So it will be on another PR(s), and it is not expected to be released immediately.

File: src/main/java/net/pms/database/MediaTableTVSeries.java [hunk 1]
```diff
@@ -180,38 +179,38 @@ private static void createTable(final Connection connection) throws SQLException
...
+				"ORIGINCOUNTRY		VARCHAR, " +
+				"ORIGINALLANGUAGE	VARCHAR, " +
+				"ORIGINALTITLE		VARCHAR, " +
+				"PRODUCTIONCOMPANIES VARCHAR, " +
+				"PRODUCTIONCOUNTRIES VARCHAR, " +
+				"SEASONS			VARCHAR, " +
+				"SERIESTYPE			VARCHAR, " +
+				"SPOKENLANGUAGES	VARCHAR, " +
+				"STATUS				VARCHAR, " +
+				"TAGLINE			VARCHAR" +
```
SubJunk, line 213: Please use space-indentation when the indentation isn't at the start of the line, otherwise we end up with unpredictable indentation depending on IDE tab-size settings (as we can see this doesn't line up on GitHub)

SubJunk (commented):
Looks great to me, just a lint note

src/main/java/net/pms/database/MediaTableTVSeries.java
SurfaceS: I view that most all table use tabs.
I corrected

SubJunk:
Great, thanks! Nice work both of you!

SubJunk (approved):

SubJunk:
One note about this is that whenever my PC wakes from sleep, I see:
```
WARN  HikariPool-1 - Thread starvation or clock leap detected (housekeeper delta=1h26m34s388ms737µs800ns).
```
I don't think it's a big deal though

PR #3179, opened 2022-07-21
switched to memory tables

ik666:
The memory table were lost ?


ik666:
i see, it's configured.

PR #3181, opened 2022-07-21
read AudioFile tag only if coverart archive table is being updated.

ik666:
AudioFile tag is read even if it is not used any more later on. This should speed access to container with many audio playlists.


SubJunk (approved):

SubJunk:
Nice one!

PR #3191, opened 2022-07-25
3004 navigation settings

SubJunk:


configuration-client/src/components/Settings/Settings.tsx
```diff
@@ -141,6 +147,28 @@ export default function Settings() {
       });
   };
 
+  const resetCache = () => {
+    setLoading(true);
+    axios.post('/configuration-api/reset-cache')
```
js-kyle, line 152: The intention of the actions service was to utilise it for things like this, to prevent a large number of endpoints being created https://github.com/UniversalMediaServer/UniversalMediaServer/blob/3004-web-settings/configuration-client/src/services/actions-service.ts#L3 - so a new action would be added

SubJunk: Thanks, done

js-kyle (approved):

Issue #3206, opened 2022-07-31
Docker: use ENTRYPOINT instead of CMD to allow parameter passing


pataquets:
As a common practice, the main executable artifact/binary/script on a Docker container is expected to be on `ENTRYPOINT`, although you might find lots of cases being in the `CMD` as this is now. If you want to just run the container with the default options, there is no noticeable difference.
Using `ENTRYPOINT`, however has some differences:
- You can still customize the init command, but just by overriding `ENTRYPOINT` instead of `CMD` (there are the corresponding `command` and `entrypoint` statements in Docker Compose manifests, too).
- Now you can add command-line parameters/switches after the image name in the `docker run universalmediaserver/ums --options`.
- You can set some default `CMD` switches, which will be overridden if doing the above on `docker run`.

Happened to me while trying UMS, just dropping the improvement suggestion.

SubJunk:
This seems ok to me, without knowing a lot about it. Please submit a PR for it :)

pataquets:
@SubJunk: submitted #3212.

PR #3212, opened 2022-08-01
Dockerfile: Use ENTRYPOINT instead of CMD to launch UMS.

pataquets:
As discussed in #3206.

* Move init run in Dockerfile from `CMD` to `ENTRYPOINT` (see linked issue).
  To test, run `docker run IMAGE noconsole` (should fail to run, since there's no GUI).

* Wrap launch command and logic in a external init script and expose both JVM and UMS launch prameters via its respective environment variables.
  Test:
  ``` console
  $ docker run --rm -it -e UMS_PROFILE=/mydir -e JVM_JNA_NOSYS=false IMAGE console
  ```
  Check launch command switches using `ps` or `htop`.

* Also, removed [deprecated](https://docs.docker.com/engine/reference/builder/#maintainer-deprecated) `MAINTAINER` statement.


SubJunk:
@pataquets thanks for this. Does anything else need to change for existing Docker users or is this fully backwards-compatible?

SubJunk:
Sorry I let this become stale as we were prioritising other things. I honestly don't know whether this is a good idea, since our Docker releases have been developed and managed entirely by others - I just run the script to release.

@atamariya any opinion about this PR?

atamariya:
@SubJunk  It's ok approve - these are latest conventions.

SubJunk (approved):

PR #3217, opened 2022-08-06
Improved support for split episodes (e.g. S01E02a)

SubJunk:
Also speed improvement and better stacktrace on error in test, and fix java.lang.UnsupportedOperationException: JsonNull in main

### DIFF
--- a/src/main/java/net/pms/util/FileUtil.java
+++ b/src/main/java/net/pms/util/FileUtil.java
@@ -618,6 +618,7 @@ public class FileUtil {
 
 	private static final String SHOW_NAME_INDEX_MATCHER = "(?i) (S\\d{2}E\\d{2}\\w{1}|S\\d{2}E\\d{2}|S\\d{2}|S\\d{2}E\\d{2}-\\d{2}|\\d{4}/\\d{2}/\\d{2})";
 	private static final Pattern SHOW_NAME_INDEX_PATTERN = Pattern.compile(SHOW_NAME_INDEX_MATCHER + " - (.*)");
+	private static final Pattern SHOW_NAME_INDEX_FALLBACK_PATTERN = Pattern.compile(SHOW_NAME_INDEX_MATCHER);
 
 	/**
 	 * Same as above, but they are common words so we reduce the chances of a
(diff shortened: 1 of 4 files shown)


src/main/java/net/pms/util/FileUtil.java
```diff
@@ -1247,7 +1248,7 @@ public static String[] getFileNameMetadata(String filename, String absolutePath)
 						tvEpisodeName = convertFormattedNameToTitleCase(tvEpisodeName);
 					}
 				} else {
-					showNameIndex = indexOf(Pattern.compile(showNameIndexMatcher), formattedName);
+					showNameIndex = indexOf(Pattern.compile(SHOW_NAME_INDEX_MATCHER), formattedName);
```
SurfaceS, line 1251: `Pattern.compile(SHOW_NAME_INDEX_MATCHER)`

this one can also be statically compiled
eg:
`private static final Pattern SHOW_NAME_INDEX_FALLBACK_PATTERN = Pattern.compile(SHOW_NAME_INDEX_MATCHER);`

then

`showNameIndex = indexOf(SHOW_NAME_INDEX_FALLBACK_PATTERN, formattedName);`

src/main/java/net/pms/util/FileUtil.java
```diff
@@ -1224,14 +1227,12 @@ public static String[] getFileNameMetadata(String filename, String absolutePath)
 			if (tvSeason.length() > 1 && tvSeason.startsWith("0")) {
 				tvSeason = tvSeason.substring(1);
 			}
-			String showNameIndexMatcher = "(?i) (S\\d{2}E\\d{2}|S\\d{2}|S\\d{2}E\\d{2}-\\d{2}|\\d{4}/\\d{2}/\\d{2})";
-			pattern = Pattern.compile(showNameIndexMatcher + " - (.*)");
-			int showNameIndex = indexOf(pattern, formattedName);
+			int showNameIndex = indexOf(SHOW_NAME_INDEX_PATTERN, formattedName);
```
SurfaceS, line 1230: Maybe this line can be moved after the next line `if (isEmpty(movieOrShowName)) {`
as it is used only on this condition (don't process something we don't use).

SurfaceS (approved):

PR #3223, opened 2022-08-09
3004-web-settings upgrade

### LINKED ISSUE #3195: Web settings: Fix console errors
There are some console errors that appear on page load after login on the new settings page to be fixed

SurfaceS:
Affine gitignore.
update node.js from 16.15.1 to 16.16.0 (LTS).
update yarn from 3.2.1 to 3.2.2
set auto update yarn before build
update mantine from 5.0.3 to 5.1.1
update tabler-icons-react to 1.54.0
update dev deps

Fix define variables before they are used
Stop console logging error
Fix #3195
Use explicit string from numbers


configuration-client/package.json
```diff
@@ -52,16 +52,16 @@
...
     "@testing-library/react": "^13.3.0",
-    "@testing-library/user-event": "^13.5.0",
-    "@types/jest": "^27.5.1",
+    "@testing-library/user-event": "^14.4.2",
+    "@types/jest": "^28.1.6",
     "@types/lodash": "^4.14.182",
-    "@types/node": "^16.11.36",
-    "@types/react": "^18.0.14",
-    "@types/react-dom": "^18.0.5",
+    "@types/node": "^18.6.5",
```
SubJunk, line 62: The major version of this package should be in sync with the Node version or we will get potentially wrong types (not that it matters here for our frontend code, but it's a good practice)

pom.xml
```diff
@@ -1000,10 +1000,20 @@
...
 						</configuration>
 					</execution>
+					<!-- update yarn to latest version -->
+					<execution>
+						<id>yarn version</id>
+						<goals>
+							<goal>yarn</goal>
+						</goals>
+						<configuration>
+							<arguments>set version stable</arguments>
```
SubJunk, line 1014: I'm not sure about this step since it will slow down builds. When the web settings branch is merged to main, Renovate will keep yarn updated (that is actually a main reason I started using Renovate instead of Dependabot)

SubJunk (changes requested):

SurfaceS:
@SubJunk  done

PR #3224, opened 2022-08-10
Main swing gui separation

SurfaceS:
As we are looking to react gui, it's a good time to separate better `newgui` (swing) and the code.


SubJunk:
@SurfaceS I thought we would remove the Java GUI when the React one is stable, since I don't think we need the extra responsibility to develop two GUIs at once. Do you agree?

SurfaceS:
@SubJunk 
This PR is intended to separate GUI from code.
Then it will allow to change / remove / add / have multiple GUI's.
Before, the code was mixed with current swing GUI.
It's still the case for few things that I need to fix on another PR (like `RootFolder` with web folders).

> I thought we would remove the Java GUI when the React one is stable, since I don't think we need the extra responsibility to develop two GUIs at once. Do you agree?

It need to be done by steps.
Swing GUI is not only the `LooksFrame` class (main GUI), but also the `Splash`, `DbgPacker`, `LanguageSelection`, `ProfileChooser`, `Wizard`, the system try icon and so on.
All of these classes need to be taken in account.

Concerning the `LooksFrame` GUI (main), react is not handling everything for now, and will not in a week (more than setting I mean logs and renderers list/controls, ect).
Also, for now, if the server fail (like port already taken), the current GUI is visible, you can check logs and change settings.
Whit react, you end with a non accessible logs/settings.

I agree the current GUI should not be improved more (only fixed).

But first step may be to run both in parallel until react can handle every cases.

SurfaceS:
@SubJunk Concerning react gui
I think it will be good that it have its own service with a dedicated port, like 9002.
If it fail, it can then use a random port and be accessible thru the system tray icon (or port known thru the console output for fully headless).
This will prevent the system to be unreachable without GUI.
It's also good for people that doesn't want to share the GUI with others.

Also, I think I'll `SERVLET` the `configuration-client` to allow us to switch it to all server engine easily.

SubJunk:
@SurfaceS those are all very good points. I'll review this soon, it seems good

SubJunk (approved):

SubJunk:
I did some light testing and it works as before, nice work. The separation is much nicer!

PR #3226, opened 2022-08-12
3004 web settings separate

SurfaceS:
separate gui server
use port 9002, random if fail.
servlet


SubJunk (approved):

SubJunk:
I know almost nothing about servlets so there is a big element of trust involved here

PR #3272, opened 2022-08-22
Shared directories

SubJunk:
This is still in progress.

- [x] Add/remove/move shared directories
- [x] Toggle directory monitoring
- [x] Support removing web content items
- [x] Support adding web content items
- [x] Support editing web content items
- [x] Support reordering web content items
- [x] Manual scan of shared folders
- [x] Mark shared directory as fully played/unplayed (currently works on right-click in Java GUI but let's do buttons here)
- [x] Fix overlapping type dropdown when opened on web content section
I can't reproduce it with just Mantine, see https://codesandbox.io/s/elastic-hill-2e99fv so I think this is caused by the `react-movable` component
- [x] SSE for when scanning completes


SubJunk:
@UniversalMediaServer/developers any ideas about how to fix the `Fix overlapping` item above?

SubJunk:
@UniversalMediaServer/developers thanks for your help! I think this is ready for review

js-kyle (approved):

SurfaceS:
I think it need some work on web content.

If someone have no web content, it can't add one as the tab is empty.

SubJunk:
I'll fix that

PR #3289, opened 2022-08-25
3004 web settings player

SurfaceS:
A simple react player


SubJunk:
This is looking great and is very responsive. It's cool to see it all coming together with the settings, console and player

SurfaceS:
> It's cool to see it all coming together with the settings, console and player

I don't know if it's a good thing to mix the player (media service) with the gui.

I think most of users will want to keep the app gui away from the media service part (for security reasons) as some of them open the player port to the internet (even if I double check each time the risk when we add things to the gui).

SubJunk:
It's worth discussion and thought. In my opinion it's ok to have them together. If I look at similar programs like Plex they have everything in one web GUI, so I think users are used to it.

SurfaceS:
On this one : 
Merged the home / console
Player added to the web gui server
Player only on the web player server

SubJunk:
Is it a good time to merge this into the web settings branch? Then we can both work on it

SurfaceS:
Yes, it should be ok

PR #3293, opened 2022-08-26
Main guimanager

SurfaceS:
Swing separation part 2.


src/main/java/net/pms/PMS.java
```diff
@@ -547,19 +531,20 @@ private boolean init() throws Exception {
 		 * would be cleaner and could support other lifecycle
 		 * notifications (see above).
 		 */
-		FrameAppender.setFrame(frame);
+		//FrameAppender.setFrame(frame);
```
SubJunk, line 534: Why is this commented?

SurfaceS: fixed

SurfaceS:
removed `updateRenderer` from `PMS`, `IGui`/`LookFrame` and `StatusTab` classes.

In fact, `PMS` call the `IGui`/`LookFrame` `updateRenderer(renderer)` function.
Then `IGui`/`LookFrame` call the `StatusTab` `updateRenderer(renderer)` function.
Then `StatusTab` call the `renderer.updateRendererGui()` function.
As a `IRendererGuiListener`, `RendererItem.updateRenderer(renderer)` is called from the `renderer.updateRendererGui()`.

Directly use the `renderer.updateRendererGui()` instead as we have it from start.

SubJunk (approved):

SubJunk:
@SurfaceS After merging I noticed the Logs tab is broken

SurfaceS:
> @SurfaceS After merging I noticed the Logs tab is broken

I'll look at that now.

SurfaceS:
Oh, I maybe understand why.
The file [src/main/external-resources/logback.xml](https://github.com/UniversalMediaServer/UniversalMediaServer/pull/3293/files#diff-aa3c4e4874b50aa5264e6f2cdaaa94decb102f184d4776f7823b654009a4055c) was changed.
So if you run it without install, it keep the old logback.xml.
I had a blank log when until I install (or replace the logback file).

SubJunk:
@SurfaceS you're right, that was it. Thanks for looking into it

PR #3300, opened 2022-08-28
3004 web settings sharedcontent

SurfaceS:
For #3272

Fix array as string.
When you have only one folder in folders, the config give it as a string.
Javascript will then end in a pseudo array of char (instead of array of string).
You end with `SharedDirectory = ["c",":","\","video"]` instead of `["c:\video"]`.

Merge current web-setting (update mantine).

Fix swing gui has app code (move getFeedTitle to Feed class).


SubJunk (approved):

SubJunk:
Cool, thanks for the help

PR #3387, opened 2022-09-15
Access-Control-Allow-Origin set to "*"

ik666:
Hey guys. I'd like to add this policy for manipulating thumbnail images delivered by UMS in JavaScript. I need to crop images that are delivered as rectangle to squares. Without this policy the browser will not allow to do that.

### DIFF
--- a/src/main/external-resources/UMS.conf
+++ b/src/main/external-resources/UMS.conf
@@ -1808,6 +1808,14 @@ audio_likes_visible_root =
 # Default: ""
 managed_playlist_folder =
 
+# Access-Control-Allow-Origin response header
+# -------------------------------------------
+# Sets the Access-Control-Allow-Origin response header indicating whether the 
+# http server response (images, etc.) can be shared with requesting code from the 
+# given origin. By default, no Access-Control-Allow-Origin is set. 
+# Default: ""
+access-control-allow-origin =
+
 # ---< Other settings saved by UMS >------------------------------------------
 alternativeffmpegpath = 
 gui_log_search_case_sensitive = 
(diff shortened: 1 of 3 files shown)


SurfaceS:
Setting Access-Control-Allow-Origin to * is never a good choice.

SurfaceS:
Why don't u use your API to do that for you ?

ik666:
Access control can be set more restrictive, i.e. to GET requests. Actually, making it configurable (opt in) could also be a solution.

I'm uncertain, what you mean with using the/my API  ... thumbnail URL's in the DIDL response are generated by the DLNAResource class.

SubJunk:
@ik666 depending on the use case this can be done easily with CSS like this:

```
img.thumbnail {
  width: 175px;
  height: 175px;
  object-fit: cover;
}
```

ik666:
@SubJunk I think your renderer option is a good choise. CSS approach: I used to square images like you proposed with CSS in the past, but then I switched to a responsive design with flexbox div's resizing and wrapping images to cover most of the users screen leading to flexible image width and height values.

SubJunk:
@ik666 similar strategy should work with flexbox too, for example there is the `aspect-ratio` CSS property that will work with dynamic widths and heights. IMO CSS is a better choice for this than JavaScript, usually the CPU use is much lower for CSS, and less code

ik666:
@SubJunk You're absolutely right. I've searched for CSS solutions myself, but didn't find any until now. Thanks a lot, the `aspect-ratio` property did the trick. Must be a new one, haven't heard of it, yet :)

Regarding this PR ... maybe in the future someone can still find it useful to set the 'Access-Control-Allow-Origin' response header. Since it's now configurable one can narrow the domain now.

PR #3399, opened 2022-09-17
Don't try to extract metadata or do API lookups for phone backups

SubJunk:


src/main/java/net/pms/dlna/DLNAResource.java
```diff
@@ -5016,11 +5017,24 @@ public String getResolutionForKeepAR(int scaleWidth, int scaleHeight) {
 	 * @param file
 	 */
 	private void setMetadataFromFileName(File file) {
+		String absolutePath = file.getAbsolutePath();
+		if (
+			file != null &&
```
ik666, line 5022: if `file` is null, then the previous line `file.getAbsolutePath()` would throw a NPE ...

SubJunk: Done

Mik-S-UMS:
I know this is specifically for phone back ups, but I think there should be an option somewhere to prevent the same thing on personal or private files in certain folders.

SurfaceS:
More than that, code for specific renderers and specific OS versions should be avoided in main code.
It make it hard to maintain.

As [Mik-S-UMS](https://github.com/Mik-S-UMS) said, it should be an option somewhere to add files/folders user doesn't want to be api parsed, metadata parsed, or shared.

**A quick and dirty solution would be :** 
Add a boolean to `DLNAResource` class.
```
	private boolean skipMetadata = false;
	public boolean isSkipMetadata() {
		return skipMetadata;
	}
	protected void setSkipMetadata(boolean skipMetadata) {
		this.skipMetadata = skipMetadata;
	}
```
Modify `RootFolder.getiPhotoFolder()` after `iPhotoVirtualFolder` creation (line 759).
```
		iPhotoVirtualFolder.setSkipMetadata(true);
```
Modify `DLNAResource.addChild `
```
		if (skipMetadata) {
			child.setSkipMetadata(skipMetadata);
		}
```
Modify `DLNAResource.syncResolve()` 
```
			if (this instanceof RealFile && ((RealFile) this).getFile() != null && !skipMetadata) {
				setMetadataFromFileName(((RealFile) this).getFile());
			}
```

Mik-S-UMS:
I was thinking it might be a good idea to add a content type selector box for each shared folder. Something like
- Mixed
- Film
- Series
- Anime
- Music
- Pictures
- Other
- Private

This could also be helpful in directing UMS to the right service to get the best metadata and would help in those cases where dates are part of the title, or a film and series has the same name and they get misidentified. If set to mixed (the default setting) then it should work as it is now.

If set to private then those should not be logged too, just a message in the logs saying that "a folder has been set to private and will not be logged".

SubJunk:
These are all good points

SubJunk:
I don't know if the iPhoto hack would work if the user is just sharing their Pictures folder, which is our default

SurfaceS:
I don't know well mac os, but if I remember :
iPhoto Library should contain only images/personal movie records, so it do not need api/Metadata ?

SubJunk:
Yeah it can probably be applied to that too, but that's not what this PR is about - this PR is to ignore metadata for phone backups, which by default get stored in the `Photos Library.photoslibrary` directory, which is inside the default shared directories on macOS.
So that's different (but maybe overlap, I don't know) to iPhoto

SubJunk:
I started writing this as a configurable setting, but along the way I realised that's a different feature. This isn't the same as having folders that are optionally ignored, this is ignoring a folder on only one OS that should *always* be ignored, because Apple is storing photos as videos.
So I think this PR is conceptually good

SurfaceS (approved):

PR #3409, opened 2022-09-19
main fix

SurfaceS:
Fix the big hole.
Without that, everyone that have access to the port will be able to download your entire disk files, and use your computer as a proxy for all protocols faking you IP are doing it (mail, ftp, ssh, etc).


SubJunk (approved):

SubJunk:
Nice catch

SurfaceS:
This should be merged

PR #3443, opened 2022-09-26
New player page not loading

SubJunk:
WIP

### DIFF
--- a/src/main/java/net/pms/network/httpserverservletcontainer/HttpExchangeServletRequest.java
+++ b/src/main/java/net/pms/network/httpserverservletcontainer/HttpExchangeServletRequest.java
@@ -56,9 +56,11 @@ import javax.servlet.http.Part;
 import net.pms.PMS;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class HttpExchangeServletRequest implements HttpServletRequest {
-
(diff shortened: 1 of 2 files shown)


SubJunk:
@SurfaceS this fixes the player page not loading for me, but I'm not sure if it is the correct fix - maybe we want it to fail to load if there is no cookie. I'm not familiar enough with that new code to answer that

SurfaceS:
@SubJunk 
Concerning **`src/main/java/net/pms/network/httpserverservletcontainer/HttpExchangeServletRequest.java`**

`public Enumeration<String> getHeaders(String name)` should not return `null`, but an empty collection.

`Collections.enumeration(null)` will throw.
I checked the all Collections.enumeration, and that's the only one affected.

In fact, the `getHeaders(String name)` member should be changed, not the `parseCookies()` one (that should be reverted and not logged).

```
	@Override
	public Enumeration<String> getHeaders(String name) {
		List<String> headers = exchange.getRequestHeaders().get(name);
		if (headers != null) {
			return Collections.enumeration(headers);
		}
		return Collections.enumeration(new ArrayList<>());
	}
```

The `parseCookies` code will then not be affected by the "Cookie" header presence, and create the cookie container as requested.

Concerning **`src/main/java/net/pms/network/webguiserver/servlets/PlayerApiServlet.java`**
This is fine. 👍

SurfaceS:
merged in https://github.com/UniversalMediaServer/UniversalMediaServer/pull/3446

PR #3444, opened 2022-09-26
Main huge refactor

SurfaceS:
- upgrade all tests to full junit v5
**remove old junit dep v3 v4
**remove assertj-core dep

- refactor packages renderers -> devices -> players

- refactor encoders.players to encoders.engines
** as it was commented "The base class for all transcoding engines." and was named "Player"
** Player can confuse with renderer's player.

- moved platform related code to platform package

- added platform specific code to platform class

- fix some licenses headers

- remove all the old plugin (external) code.
** move StartStopListenerDelegate to service at it is used only by sleepmanager

- move Library scanner to service

- remove pre java 8 code that java now handle better than the impl from PS3 Media Server.
- create private constructor for class that should not be instantiated
- enum are always static, remove static modifier
- fix boolean literals should not be redundant
- remove superfluous throws
- array.size > 0 need to use array.isEmpty()
- use diamond operator <> when needed
- added private constructor when class is not meant to be instantiated
**(java create a public one in the background if not the case)
- use replace not replaceAll when regex not needed
- remove or annotate deprecated
- use StringBuilder, not StringBuffer
- add override when needed
- set protected modifier to abstract class
- replace string == to string.equals
- remove old Class.forName(dbdriver)
** not needed since java 6
- use expression switch
- fix no break in switch


src/main/java/net/pms/database/MediaTableTVSeries.java
```diff
@@ -325,7 +325,7 @@ public static Long set(final Connection connection, final JsonObject tvSeries, f
 						} else {
 							insertQuery = SQL_INSERT_ALL;
 						}
-						try (PreparedStatement insertStatement = connection.prepareStatement(insertQuery, PreparedStatement.RETURN_GENERATED_KEYS)) {
+						try (PreparedStatement insertStatement = connection.prepareStatement(insertQuery, Statement.RETURN_GENERATED_KEYS)) {
```
SubJunk, line 328: Why?

SurfaceS: In the interest of code clarity, static members of a base class should never be accessed using a derived type's name.

Doing so is confusing and could create the illusion that two different static members exist.

src/main/java/net/pms/util/SubtitleUtils.java
```diff
@@ -720,10 +728,11 @@ public static void deleteSubs() {
 	 * } and any ASS tags <code>
 	 * {\*}
 	 * </code> from subtitles file for renderers not capable of showing SubRip
-	 * tags correctly. * is used as a wildcard in the definition above.
+	 * tags correctly.* is used as a wildcard in the definition above.
```
SubJunk, line 731:
```suggestion
	 * tags correctly. * is used as a wildcard in the definition above.
```

SubJunk (commented):
Looks good and I tested it lightly, didn't notice anything broken

src/main/java/net/pms/dlna/Playlist.java
```diff
@@ -57,17 +55,8 @@ public void add(DLNAResource resource) {
 				}
 			}
 		} else {
-			String data = resource.write();
-			if (!StringUtils.isEmpty(data) && resource.getMasterParent() != null) {
```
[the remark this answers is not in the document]
SurfaceS: **From `DLNAResource`:**

```
	private ExternalListener masterParent;

	public void setMasterParent(ExternalListener r) {
		if (masterParent == null) {
			// If master is already set ignore this
			masterParent = r;
		}
	}

	public ExternalListener getMasterParent() {
		return masterParent;
	}
```
This code was removed as it is related to the old Plugins.

**From `Playlist`:**

As we are not using plugins, `resource.getMasterParent() != null` was always return `false`.

Then code after this was not needed.

src/main/java/net/pms/dlna/DLNAResource.java
```diff
@@ -743,16 +738,16 @@ public void addChild(DLNAResource child, boolean isNew, boolean isAddGlobally) {
 
 	/**
 	 * Determine whether we are a candidate for streaming or transcoding to the
-	 * given renderer, and return the relevant player or null as appropriate.
+	 * given renderer, and return the relevant engine or null as appropriate.
 	 *
 	 * @param renderer The target renderer
-	 * @return A player if transcoding or null if streaming
+	 * @return A engine if transcoding or null if streaming
```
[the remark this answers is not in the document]
SurfaceS: [224d714](https://github.com/UniversalMediaServer/UniversalMediaServer/pull/3444/commits/224d714db9fb6f1544ad048e79e317990ddd3a0a)

src/main/java/net/pms/util/SubtitleUtils.java
SurfaceS: [224d714](https://github.com/UniversalMediaServer/UniversalMediaServer/pull/3444/commits/224d714db9fb6f1544ad048e79e317990ddd3a0a)

SubJunk (approved):
Nice work, this is a good cleanup

PR #3464, opened 2022-10-02
Fixed failed parsing of files being moved/copied/modified

### LINKED ISSUE #3325: Files locked on filesystem if parsing fails
I think what happens is that if I am copying a group of videos to another hard drive that is monitored by UMS, UMS fails to parse the partial file, and then doesn't release it afterwards.
We have talked before about using a scheduling mechanism for file parsing too, where failed parses are stored temporarily in the database for retry. At the moment we retry after 3 seconds but often more time is needed before there is enough for MediaInfo to use.

SubJunk:
Fixes #3325

### DIFF
--- a/src/main/java/net/pms/dlna/RootFolder.java
+++ b/src/main/java/net/pms/dlna/RootFolder.java
@@ -1381,11 +1381,11 @@ public class RootFolder extends DLNAResource {
 	};
 
 	/**
-	 * Adds and removes files from the database when they are created or
+	 * Adds and removes files from the database when they are created, modified or
 	 * deleted on the hard drive.
 	 */
 	public static final FileWatcher.Listener LIBRARY_RESCANNER = (String filename, String event, FileWatcher.Watch watch, boolean isDir) -> {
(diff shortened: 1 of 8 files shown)


SurfaceS:
I would go for a Timer that parseFileForDatabase entries from a static locked files List<File> (with access lock) each 5 seconds ?

From `LIBRARY_RESCANNER`, if locked and not already exists in the list -> Log and add entry.
On the `parseFileForDatabase` -> Remove locked file entry if it exists in the List.

Then you will be sure that the file will be parsed at end and log only once.

src/main/java/net/pms/dlna/RootFolder.java
```diff
@@ -1445,6 +1445,22 @@ public static final void parseFileForDatabase(File file) {
...
+			return;
+		}
+
+		try (RandomAccessFile srcFile = new RandomAccessFile(file, "rw")) {
+			// no exception happened, so we can continue
+		} catch (Exception e) {
+			/**
+			 * This will happen a lot (every 500ms) for files that are being
+			 * copied/moved. Eventually the last one should succeed.
+			 */
```
SubJunk, line 1459:
```suggestion
```

SurfaceS (dismissed):

PR #3480, opened 2022-10-06
Add shutdown computer option to the Server Settings folder

### LINKED ISSUE #3475: (FEATURE) get back the plugins or turn off PC from ums dlna
I want a feature to turn off PC server from the ums device, like my PS4, it was possible with my ps3 when plugins were available, but now plugins don't exists, but some feature to turn off PC from my ps4 would be appreciated 👍

SubJunk:
Also adds custom icons for VirtualVideoAction

Closes #3475


ik666 (commented):
Wouldn't it be nicer to spend VirtualVideoAction a second constructor with 3 parameters, leaving the one with two params in place, instead of adding <code> null</code> to all existing constructor calls?

SubJunk:
@ik666 and @SurfaceS I used to do that until I saw files like [DLNAImage](https://github.com/UniversalMediaServer/UniversalMediaServer/blob/main/src/main/java/net/pms/dlna/DLNAImage.java) with many identically-named functions that give me a headache... that convinced me that it's better to just do the refactoring each time. That way at least IDE hinting makes it more readable

In other languages you would just do it like `myFunction(String oldParam, String newParam="Default value") { ...` it's one of my biggest complaints about Java

ExSport:
Someone wants to shutdown, other one sleep or hibernate to wake up the computer easily via WoL.
What about to offer all of them? 😎

SubJunk:
@ExSport yes I agree, those will be good features too

Issue #2618, opened 2021-08-22
Add OpenSubtitles hash to cache


SubJunk:
Some files may have their hashes computed more than once, for example after a failed API lookup is retried. That gives a very small delay (16ms in my testing) per file, but that would be more like <1ms with the cache, and would prevent IO

CodeLover254:
I will be glad to work on this one. :)

SubJunk:
@CodeLover254 great!

SubJunk:
@CodeLover254 I think this is waiting on you, will you return to it?

CodeLover254:
Hello @SubJunk. I have been swamped of late but I am happy to take on the task. Quick question,  the cache in reference here is the H2 database, right? My previous implementation had been relying on the in-memory cache.

SubJunk:
@CodeLover254 it's ok I understand being swamped :) yes, I think your in-memory cache was a good improvement, but the h2 database would be even better

CodeLover254:
Thanks for the clarification. I will use H2 for that.

SubJunk:
That never happened, but anyway we have removed OpenSubtitles support

PR #3485, opened 2022-10-08
OpenSubtitles Hash to Cache feature

CodeLover254:
In this pull request, I submit the caching feature for OpenSubtitles hashes in order to minimize latency and resolve issue #2618. The caching feature leverages the use of H2 database.


SubJunk:
@CodeLover254 great, thanks! You did a lot here and it looks good. I really like that you added tests for it! 🥇 

I wonder if it would be better to add it as a column to the existing FILES table, instead of making a new table. The hash is for the file so that makes sense I think. Then you can use the existing cache pattern where there is a getter/setter in `DLNAMediaInfo.java` and then the caching is handled by `MediaTableFiles:getData()` and `insertOrUpdateData()`

CodeLover254:
@SubJunk , didn't think of it that way. Guess am still learning. Thanks for the feedback. I can refactor it so that it uses the FILES table instead.

SubJunk:
I should have given those details in the issue to save you time. I'll try to make better issues

SubJunk:
Closing this because of the hopefully incoming new one :)

CodeLover254:
No problem. I will create a new one based on the discussions we had. :)

PR #3507, opened 2022-10-17
3004 react app and huge cleanup

### LINKED ISSUE #3501: Remove old engines
UMS use the `JUPnP (Netty)` engine as default since v11.0.0.
It's maybe time to remove old upnp engines :
`Sockets` that is not up to date.
`Netty` and `Java` that use the `SocketSSDPServer` (datagram) instead of `JUPnP` one.

I may add `ContentDirectory` service implementation to `JUPnP`, and put this engine in alpha for now to test it.

@SubJunk what do you think about this ?

SurfaceS:
Implements JUPnP ContentDirectory service.
Added JUPnP+ engines (netty and java) that are now serving ContentDirectory service directly from JUPnP.
Keeped JUPnP (netty) as default.
Removed old media server engines (sockets, netty and java with only register part from jupnp and use an really old way for SSDP part) (Fix #3501)
Uniformized Licence header over the whole project, react, js, css included.
Refactor PmsConfiguration to UmsConfiguration.
Move all web.conf file's code under its own WebSourcesConfiguration class (prepare #3508).
Move all ffmpeg.webfilters file's code under its own FFmpegWebFilters class (prepare #3508).
Clean on a lot of old way coding on many classes.
Set the right accessors on fields/members (public/private, etc).
Removed private class field used as local var only (declared on the local member only).
Set members static when appropriate.
Add getters/setters when needed (fields should not be directly acceded from others classes).
split range.time/byte to TimeRange and ByteRange with Range interface.
Remove UmsKeysDb never used.
Rename file pms.pid file to UMS.pid
Use interface base class for collections (eg. List for ArrayList).
Moved filesutils unix things under platform unix.
Use StandardCharsets directly instead of Charsets name.
Added some TODO for not understood nonsense code.


valib:
@SurfaceS nice job but there is a problem. The WEB subfolders are not under WEB folder but in the root folder like this
screen

SurfaceS:
I'll fix this tomorrow.
That's a inverted field. I've seen it on the WIP to uniform shared conf.

SurfaceS:
@valib fixed

SubJunk:
I'll test and review this weekend

src/main/java/net/pms/database/MediaTableMusicBrainzReleases.java
```diff
@@ -147,22 +145,36 @@ private static void createTable(final Connection connection) throws SQLException
...
 		}
 
-		@SuppressFBWarnings("EI_EXPOSE_REP2")
 		public MusicBrainzReleasesResult(final boolean found, final Timestamp modified, final String mBID) {
 			this.found = found;
 			this.modified = modified;
 			this.mBID = mBID;
 		}
+
+		public boolean isFounded() {
```
SubJunk, line 163: Same

src/main/java/net/pms/dlna/DLNAImageProfile.java
```diff
@@ -1256,22 +1258,10 @@ public DLNAComplianceResult checkCompliance(ImageInfo imageInfo) {
...
-			case DLNAImageProfile.JPEG_TN_INT:
-				checkJPEG(imageInfo, complianceResult);
-				break;
-			case DLNAImageProfile.PNG_LRG_INT:
-			case DLNAImageProfile.PNG_TN_INT:
-				checkPNG(imageInfo, complianceResult);
-				break;
-			default:
-				throw new IllegalStateException("Illegal DLNA media profile");
+			case DLNAImageProfile.GIF_LRG_INT -> checkGIF(imageInfo, complianceResult);
```
SubJunk, line 1261: I find switch statements easier to read the existing way, with one line per case

src/main/java/net/pms/dlna/DLNAImageProfile.java
```diff
@@ -1256,22 +1258,10 @@ public DLNAComplianceResult checkCompliance(ImageInfo imageInfo) {
...
+			case DLNAImageProfile.JPEG_LRG_INT, DLNAImageProfile.JPEG_MED_INT, DLNAImageProfile.JPEG_RES_H_V_INT, DLNAImageProfile.JPEG_SM_INT, DLNAImageProfile.JPEG_TN_INT -> checkJPEG(imageInfo, complianceResult);
```
SubJunk, line 1262: Some of these lines are getting long with this style of switch statement. I find the old style easier to read, with one case per line. Does anyone else have an opinion @UniversalMediaServer/developers ?

SubJunk:
I did some testing and can't see any problems

src/main/java/net/pms/dlna/DLNAImageProfile.java
```diff
@@ -1256,22 +1258,15 @@ public DLNAComplianceResult checkCompliance(ImageInfo imageInfo) {
...
-				throw new IllegalStateException("Illegal DLNA media profile");
+			case DLNAImageProfile.GIF_LRG_INT -> checkGIF(imageInfo, complianceResult);
+			case DLNAImageProfile.JPEG_LRG_INT,
+				DLNAImageProfile.JPEG_MED_INT,
+				DLNAImageProfile.JPEG_RES_H_V_INT,
+				DLNAImageProfile.JPEG_SM_INT,
+				DLNAImageProfile.JPEG_TN_INT -> checkJPEG(imageInfo, complianceResult);
+			case DLNAImageProfile.PNG_LRG_INT,
+				DLNAImageProfile.PNG_TN_INT -> checkPNG(imageInfo, complianceResult);
+			default -> throw new IllegalStateException("Illegal DLNA media profile");
```
SurfaceS, line 1269: It can be modified like that.

src/main/java/net/pms/dlna/DLNAImageProfile.java
SurfaceS: I modified it like that for readability :

```
		case DLNAImageProfile.GIF_LRG_INT -> checkGIF(imageInfo, complianceResult);
			case DLNAImageProfile.JPEG_LRG_INT,
				DLNAImageProfile.JPEG_MED_INT,
				DLNAImageProfile.JPEG_RES_H_V_INT,
				DLNAImageProfile.JPEG_SM_INT,
				DLNAImageProfile.JPEG_TN_INT -> checkJPEG(imageInfo, complianceResult);
			case DLNAImageProfile.PNG_LRG_INT,
				DLNAImageProfile.PNG_TN_INT -> checkPNG(imageInfo, complianceResult);
			default -> throw new IllegalStateException("Illegal DLNA media profile");
```

SubJunk (approved):
Nice work. This is a good set of improvements.

It took a long time to review. In the future if there are more big auto-formatting changes to make, it might be good to have one PR per change. For example one PR for the license headers, one for switch statement formatting, etc. and then there would be one for the non-formatting changes. Because it can be hard to find the "real" changes amongst the lint ones :)

PR #3521, opened 2022-10-21
fix audio id for TS video file used in m3u8 HLS

valib:


SurfaceS:
This seems not to be a problem related to HLS, but a parser one.
Parsers that fill data (mediainfo parser and ffmpeg one) mix the ID (Unique ID), media stream index, and type stream index (audio#0,audio#1,...).

This bug need to be addressed at source.

SubJunk:
should this be closed given the above information?

PR #3528, opened 2022-10-24
Bugfix/update renderer resource locator

ik666:
This PR fixes an issue (random renderer used) while generating the DIDL response by defining a generic renderer. This affects only the API.

### DIFF
--- a/src/main/java/net/pms/dlna/DbIdResourceLocator.java
+++ b/src/main/java/net/pms/dlna/DbIdResourceLocator.java
@@ -32,7 +32,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import net.pms.Messages;
 import net.pms.PMS;
-import net.pms.configuration.RendererConfiguration;
 import net.pms.database.MediaDatabase;
 import net.pms.database.MediaTableAudiotracks;
 import net.pms.database.MediaTableFiles;
@@ -52,11 +51,6 @@ public class DbIdResourceLocator {
 	 * This class is not meant to be instantiated.
 	 */
 	private DbIdResourceLocator() {
-		try {
-			API_RENDERER =  new ApiRenderer();
-		} catch (ConfigurationException | InterruptedException e) {
-			LOGGER.error("cannot instantiate API renderer", e);
-		}
 	}
 
 	public static DLNAResource locateResource(String id) {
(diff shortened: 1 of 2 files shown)


SurfaceS:
> issue (random renderer used)

In fact it take the default renderer from configuration if one is needed, and do not use any otherwise (to not filter).

> This affects only the API.

This affects all renderers as you hard implement a renderer, and set a conf filter on founded childs when it should not.

`DbIdResourceLocator` is not designed to have a renderer. It is a direct read on db for a media with its row id -> not UPnP -> no DIDL.

Renderer has it own `RootFolder` defined, and have all filtered data/childs on it.

You can't had a default renderer to `DbIdResourceLocator` without breaking it.

The only clean way to filter with a renderer on `DbIdResourceLocator` is to add a renderer param
```

	public static DLNAResource locateResource(String id) {
		return locateResource(id, null);
	}

	public static DLNAResource locateResource(String id, RendererConfiguration renderer) {
		return getDLNAResourceByDBID(DbIdMediaType.getTypeIdentByDbid(id), renderer);
	}

```

`public static DLNAResource getDLNAResourceByDBID(DbIdTypeAndIdent typeAndIdent, RendererConfiguration renderer) `
then :
```
		if (renderer != null) {
			res.setDefaultRenderer(renderer);
		}
```

ik666:
> > issue (random renderer used)
> 
> In fact it take the default renderer from configuration if one is needed, and do not use any otherwise (to not filter).

Actually, I was unclear. The first found and active transcoding engine is used if no renderer is given (is`NULL`). I think without any given renderer, transcoding makes no sense. Rethinking this, the PR3550 shows a smarter solution. Thanks for comment!

PS: The `DbIdResourceLocator` is written for identifying DLNAResources by database queries. It is used for delivering UPnP content doing browse and search requests and therefore is always generating DIDL objects (at the end).

ik666:
#3530 is a better solution. I investigate why the renderer is null later ...

SubJunk:
@ik666 I'm a bit lost with this bug, is it fixed? Can we close #3530 ?

ik666:
@SubJunk From my perspective, this (3528) bug is closed, however, I'd propose to merge #3530 into main before closing that one, because my comment there seems still to be valid (but I haven't debugged into the v12 branch yet) .

PR #3530, opened 2022-10-24
Do not use the first transcoding engine found, if no renderer is given.

ik666:
Alternative solution for #3528 . Handle empty RendererConfiguration. If no target renderer is specified, do no transcoding.


ik666:
@SurfaceS @SubJunk I think there is one more spot where - I think - unintentional transcoding might happen.

Down the method  `RootFolder.parseWebConf` some child Resources are added to the `parent` variable 
like these ... 
```
[...]
case "audiostream" -> parent.addChild(new WebAudioStream(uri, values[1], optionalStreamThumbnail));
case "videostream" -> parent.addChild(new WebVideoStream(uri, values[1], optionalStreamThumbnail));
```

In my environment `parent` has no defaultRenderer set. By the time `parent` is resolved, `DLNAResource.resolveEngine` is called without a renderer configuration (NULL). I'm asking myself, if `parent` should also set at least the defaultRenderer, or identify the correct one that is used?

SubJunk (changes requested):
I don't know if this makes sense on its own, because we can see further down that this method is designed to handle null renderers (for example on lines 750, and 779-780)

ik666:
@SubJunk No, it's not handled correctly. If NULL is given as renderer (just change the value in your IDE and trace the call further down), the first transcoding engine will still be used (also in 13.x branch), which makes no sense in my eyes.

SubJunk:
I'm still unsure about it given how much null handling exists in this method. I think at least let's clean up the unneeded null handling before merging it. So any part of that method that checks if `renderer` is `null` can be removed in this branch

SubJunk:
This kind of change makes me really wish we had more automated tests written to protect from unseen regressions.
But LGTM

SubJunk (approved):

ik666:
I know what you mean ...

PR #3536, opened 2022-10-25
update database statistics after scan

ik666:
generate database statistics after library scanner finishes.


SubJunk:
What does it do?

ik666:
This command updates the selectivity statistics of tables. The selectivity is used by the cost based optimizer to select the best index for a given query. From time to time index might degenerate. This helps to "clean up" statistics after a restart (or manual rescan).

SubJunk (approved):

SubJunk:
Ok, thanks for explaining

Issue #3523, opened 2022-10-21
Web settings: Fix shared content not updating on web


SubJunk:
Reproduction steps:
1) Add a new shared folder in the new web settings area and click Save
2) Go to the Player area, observe the new folder isn't there, even if you refresh the page

I can see the update if I go to port 9002, but not 9001

SubJunk:
@SurfaceS I think this should be fixed before we merge the branch for v12, because it will affect every new user who uses the web player, which is a lot of people

SurfaceS:
@SubJunk : Have a try on this one https://github.com/UniversalMediaServer/UniversalMediaServer/pull/3561.
It is a confs cleanup that I was working on, and that should handle shared changes differently.

ik666:
@SurfaceS I've merged the new branch. I do not see my configured shared resources in the root folder any more. I'm using a headless environment. The files are available over the media library though.

SurfaceS:
@ik666 
can you post your SHARED.conf and what is missing on your root ?

SubJunk:
My problem is fixed, but I'll leave it for this new one and delay the release until it's resolved

ik666:
@SurfaceS Maybe I've missed something, but I don't see a SHARED.conf ... where should it be and what is the content? If this is a new config element, maybe the folder settings (UMS.conf) from 11.x can be merged?

And I am missing the "My Album" folder displayed in the RootFolder.

ik666:
@SubJunk @SurfaceS I think my issue is fixed now. I had a NPE during scan (see #3604). Maybe this is why nothing showed up. Also the SARED.conf is now being created and merged as expected ... Thx all.

Issue #3508, opened 2022-10-17
Uniformizing settings files format


SurfaceS:
UMS use a lot different file format to store settings.

`/UMS.conf` use java properties file format.
`/renderers/*.conf` files use java properties file format.

`/ffmpeg.webfilters` use custom file format
`/UMS.cred` use custom file format
`/data/UMS.tmpmgr` use custom file format
`/data/dynpls/*.ups` use custom file format
`/data/resume/*.resume` use custom file format

We may look if it is possible to avoid custom's file format.

~~`/VirtualFolders.conf` use JSON format~~
~~`/WEB.conf` use custom file format~~

- [x] `WEB.conf`, `VirtualFolders.conf` and `folders`'s setting part from `UMS.conf` should be stored on the same file, like `shared.conf`. #3561
- [ ] `/UMS.conf` and `/renderers/*.conf` files could be updated from `commons-configuration` to `commons-configuration2`

SubJunk:
Good idea

SubJunk:
This has been done

PR #3561, opened 2022-11-01
3004 shared confs

SurfaceS:
**About Shared content :**
- regroups all shared content configs in one file (`SHARED.conf`).
- added ability to add virtual folders from react GUI.
- added ability to virtual folders to seat in a path (not only on root as before).
- added ability to deactivate a share without delete it (then it keep DB data for it).
- adapt react GUI and swing GUI (swing GUI need more work on it, but I'm not expert on this).
- read old user's files (`WEB.conf` and `VirtualFolders.conf`) and build a brand new format `SHARED.conf` if `SHARED.conf` is not present.
- keep old user's files (`WEB.conf` and `VirtualFolders.conf`) for retro compatibility, but maybe it can be deleted ?
- keep old `UMS.conf` value for shared for retro compatibility, but maybe it can be deleted ?
- resources `WEB.conf` and `VirtualFolders.conf` deleted from package as they are not used anymore for new installs.

**About App conf :**
- cleanup `UmsConfiguration`.
- cleanup `UMS.conf` file.

**About Device and Renderer confs :**
- cleanup `DeviceConfiguration`.
- cleanup `RendererConfiguration`.
- cleanup `renderers.conf` files affected.
- split `DeviceConfiguration` and `DeviceConfigurations` that handle all the loaded ones (and some code that will end in the `ConnectedRenderers` at end).
- split `RendererConfiguration` and `RendererConfigurations` that handle all the enabled ones.
- added `ConnectedRenderers` to should handle at end all founded devices.


SurfaceS:
partially fix #3508

SubJunk:
I have verified it fixes #3523

src/main/java/net/pms/configuration/DeviceConfiguration.java
```diff
@@ -39,15 +32,12 @@
 public class DeviceConfiguration extends UmsConfiguration {
 	private static final Logger LOGGER = LoggerFactory.getLogger(DeviceConfiguration.class);
 
-	public static final int DEVICE = 0;
-	public static final int RENDERER = 1;
-	public static final int PMSCONF = 2;
+	private static final int DEVICE = 0;
+	private static final int RENDERER = 1;
+	//private static final int UMSCONF = 2; // not used
```
SubJunk, line 37:
```suggestion
```

SubJunk (approved):

PR #3566, opened 2022-11-03
2d to 3d conversion for VR

threedguru:
**2D to 3D Conversion for VR**

This pull request provides support for watching 2D Videos in 3D within a VR Headset such as the Meta Quest 2 via a media player such as the SkyBox VR Media player. The linked video provides a demonstration and screen shots of UMS with the additional functionality explained in the commentary:

https://streamable.com/ijojmx

The code and configuration changes include; An AviSynth(+) 2d to 3d conversion script (including corresponding UMS GUI updates - algorithm selection, output format, resizing options), addition of AviSynth+ multithreading support (optionally generates AviSynth+ compatible multithreading script rather than AviSynth - they changed the syntax), changes to the implementation of the resume video function for AviSynth/FFMpeg transcoding (it did not seem to function previously), extension of video format support without the need for external codecs via the use of FFMS2 for AviSynth(+) (the libraries for which are added), explicit VR media player/renderer support (for SkyBox VR Player), extended support for NVIDIA and AMD GPU hardware via new encoding options for FFMpeg.

**Hardware and Software Requirements:**

**Hardware:**

A VR Headset (an [Oculus/Meta Quest 2 VR Headset ](https://www.meta.com/gb/quest/products/quest-2/) has been used for proving)
A capable Windows PC with high-end NVIDIA or AMD GPU and CPU (an [MSI GF65 Thin Laptop ](https://www.msi.com/Laptop/GF65-Thin-10UX/Specification)has been used for proving). This laptop has an NVIDIA GeForce RTX 3060 GPU. 
A high-speed home Wireless Network (the [AX1800 Whole Home Mesh Wi-Fi 6 System ](https://www.tp-link.com/uk/home-networking/deco/deco-x20/)has been used for proving)

**Software:**

[Sky Box VR Player](https://skybox.xyz/): A media player that supports 3D playback and DLNA for a range of headsets from Meta, Valve and Steam VR. 
[Universal Media Server ](https://www.universalmediaserver.com/): A free DLNA, UPnP and HTTP/S Media Server, (obviously with the additions contained in this pull request).
[AviSynth](https://avs-plus.net/)[+ ](https://avs-plus.net/)(64 bit): “Frameserver” video processing tool with scripting, a modern fork of [Avisynth](http://avisynth.nl/index.php/Main_Page) with multithreading and 64-bit support. It is recommended to install the 64 bit version that includes the vc redistributables from here: https://github.com/AviSynth/AviSynthPlus/releases/download/v3.7.2/AviSynthPlus_3.7.2_20220317_vcredist.exe
[K-Lite Codec Pack](https://codecguide.com/download_kl.htm): A collection of DirectShow filters, VFW/ACM codecs, and tools. Codecs and DirectShow filters needed for decoding audio and video formats (with GPU hardware acceleration).

I have tested the changes against the main branch. It is advisable to use AviSynth+ 64 bit rather than the AviSynth 32 bit version (that currently ships with UMS). It might work well enough with the 32 bit version of AviSynth, but it needs more proving. AviSynth+ should be installed as a pre-req. 

I have not changed the existing default configuration settings for UMS. In order to turn on the 2D to 3D conversion functionality, it needs to be enabled in the AviSynth/FFMpeg transcoder settings tab. I also recommend enabling the NVIDIA h264 hardware encoding in the FFMpeg transcoder settings page (a new setting). I needed to implement hardware encoding to achieve the necessary speed and quality trade off. I have set the SkyBoxVRPlayer renderer file to request MPEGTS-H264-AC3. Changing H265 settings will have no impact on SkyBoxVRPlayer therefore. There is scope for more tuning and proving with AMD graphics hardware too. I have attached a couple of screen shots of the settings that I used for proving. 

AviSynthFFMpeg - Snapshot

FFMPegVideo - Snapshot

I would also highlight that, in the absence of access to a VR Headset, it is possible to at least run/test/debug the AviSynth(+) generated 3D scripts with VirtualDub2, available from here: http://virtualdub2.com/ (I use the 64 bit version) . However, doing this requires a bit more explanation than I can give here. Attached is an example screen shot:

VirtualDub2

For completeness and to aid understanding, I have provided a brief overview of the theoretical background and links to earlier work on this topic. I have incorporated and modified a couple of the earlier AviSynth scripts referenced in this pull request. 

**The Pulfrich Effect/Phenomenon:**

The Pulfrich Effect is a psychophysical percept wherein lateral motion of an object in the field of view is interpreted by the visual cortex as having a depth component, due to a relative difference in signal timings between the two eyes.

The Pulfrich effect has been utilised to enable a type of stereoscopy, or 3-D visual effect, in visual media such as film and TV. As in other kinds of stereoscopy, glasses are typically used to create the illusion of a three-dimensional image. By placing a neutral filter (e.g., the darkened lens from a pair of sunglasses) over one eye, an image, as it moves right to left (or left to right, but not up and down) will appear to move in depth, either toward or away from the viewer.

The usage applied in the work presented here, with a VR headset, does not require glasses. The mechanism to induce the effect is somewhat different. The technique involves displaying a time delayed video frame to one eye compared to the other. This induces the 3D pulfrich effect.

**Previous Work and Acknowledgement:**

This on-the-fly real-time 2D to 3D work is built on the foundation of several articles dating back to 2009. These are listed below:

- [Impressive Realtime 2D to 3D Video Conversion for 3D Vision - 2009](https://3dvision-blog.com/957-impressive-realtime-2d-to-3d-video-conversion-for-3d-vision/) 

- [2D to 3D Realtime Video Conversion with ](https://3dvision-blog.com/1039-2d-to-3d-realtime-video-conversion-with-avisynth-proof-of-concept/)[Avisynth](https://3dvision-blog.com/1039-2d-to-3d-realtime-video-conversion-with-avisynth-proof-of-concept/)[ (Proof of Concept) - 2009](https://3dvision-blog.com/1039-2d-to-3d-realtime-video-conversion-with-avisynth-proof-of-concept/)

- [2D to 3D Realtime Video Conversion Script for ](https://3dvision-blog.com/1220-2d-to-3d-realtime-video-conversion-script-for-avisynth-v0-3/)[Avisynth](https://3dvision-blog.com/1220-2d-to-3d-realtime-video-conversion-script-for-avisynth-v0-3/)[ v0. 3 - 2010](https://3dvision-blog.com/1220-2d-to-3d-realtime-video-conversion-script-for-avisynth-v0-3/)

- [A Collection of Scripts for 2D to 3D conversion with ](https://sourceforge.net/projects/voix/files/2d3/)[AviSynth](https://sourceforge.net/projects/voix/files/2d3/)

- [Convert 2D Movies to 3D for Virtual Reality – 2018](https://f4cepa1m.com/blog/f/convert-2d-movies-to-3d-for-virtual-reality)

- [Simple Script for 2D to 3D Video Conversion – 2022](https://www.reddit.com/r/OculusQuest/comments/sq5lua/simple_script_for_2d_to_3d_video_conversion/)

A range of 2D to 3D conversion algorithms have been implemented in the AviSynth scripting language above. This work leverages a couple of these scripts as a base.


threedguru:
OK, I've recreated the pull request. I've looked at the conflicts. I manually merged the conflicted files yesterday on the web-settings branch so they are good to go (with the versions in the pull request) except for messsages.properties and FileWatcher (I don't know why that's popped up as I had no interest in changing it).

src/main/external-resources/UMS.conf
```diff
@@ -911,6 +911,22 @@ ffmpeg_sox =
 # Default: "auto"
 ffmpeg_gpu_decoding_acceleration_method =
 
+# H264 GPU encoding acceleration method
```
SubJunk, line 914:
```suggestion
# H.264 GPU encoding acceleration method
```

src/main/external-resources/UMS.conf
```diff
@@ -911,6 +911,22 @@ ffmpeg_sox =
...
+# -----------------------
+# Available list provided is:
+# libx264, libx264rgb, h264_amf, h264_nvenc, h264_qsv, nvenc, nvenc_h264
+# Select nvenc methods for NVIDIA cards, amf for AMD.
+# Default: "libx264"
+ffmpeg_gpu_H264_encoding_acceleration_method =
+
+# H265 GPU encoding acceleration method
```
SubJunk, line 922:
```suggestion
# H.265 GPU encoding acceleration method
```

src/main/java/net/pms/configuration/UmsConfiguration.java
```diff
@@ -2647,6 +2767,63 @@ public boolean getAvisynthMultiThreading() {
...
+	}
+
+	/**
+	 * Returns true if we should use FFMS2 instead of DirectShowSource in
+	 * AviSynth.
+	 *
+	 * @return True if we should use FFMS2 instead of DirectShowSource in
+	 *         AviSynth.
+	 */
+	public boolean getAvisynthUseFFMS2() {
```
SubJunk, line 2787: Please change `get` to `is`

src/main/java/net/pms/configuration/UmsConfiguration.java
```diff
@@ -2647,6 +2767,63 @@ public boolean getAvisynthMultiThreading() {
...
+	public void setAvisynth2Dto3D(boolean value) {
+		configuration.setProperty(KEY_AVISYNTH_2D_TO_3D, value);
+	}
+
+	/**
+	 * Returns true if we should convert 2D video to 3D in AviSynth.
+	 *
+	 * @return True if we should convert 2D video to 3D in AviSynth.
+	 */
+	public boolean getAvisynth2Dto3D() {
```
SubJunk, line 2823: Same

src/main/java/net/pms/configuration/UmsConfiguration.java
```diff
@@ -2804,6 +3005,129 @@ public boolean getFfmpegAvisynthInterFrameGPU() {
...
+	}
+
+	/**
+	 * Returns true if we should use FFMS2 instead of DirectShowSource in
+	 * AviSynth.
+	 *
+	 * @return True if we should use FFMS2 instead of DirectShowSource in
+	 *         AviSynth.
+	 */
+	public boolean getFfmpegAvisynthUseFFMS2() {
```
SubJunk, line 3025: Same

src/main/java/net/pms/configuration/UmsConfiguration.java
```diff
@@ -2804,6 +3005,129 @@ public boolean getFfmpegAvisynthInterFrameGPU() {
...
+	public void setFfmpegAvisynth2Dto3D(boolean value) {
+		configuration.setProperty(KEY_FFMPEG_AVISYNTH_2D_TO_3D, value);
+	}
+
+	/**
+	 * Returns true if we should convert 2D video to 3D in AviSynth.
+	 *
+	 * @return True if we should convert 2D video to 3D in AviSynth.
+	 */
+	public boolean getFfmpegAvisynth2Dto3D() {
```
SubJunk, line 3061: Same

src/main/java/net/pms/configuration/UmsConfiguration.java
```diff
@@ -2804,6 +3005,129 @@ public boolean getFfmpegAvisynthInterFrameGPU() {
...
+	public void setFfmpegAvisynthHorizontalResize(boolean value) {
+		configuration.setProperty(KEY_FFMPEG_AVISYNTH_HORIZONTAL_RESIZE, value);
+	}
+
+	/**
+	 * Returns true if we should resize the input 2D video for 3D conversion in AviSynth.
+	 *
+	 * @return True if we should resize.
+	 */
+	public boolean getFfmpegAvisynthHorizontalResize() {
```
SubJunk, line 3119: Same

src/main/java/net/pms/newgui/engines/AviSynthFFmpeg.java
```diff
@@ -99,7 +126,128 @@ protected static JComponent config(String languageLabel) {
...
+		horizontalResize.setContentAreaFilled(false);
+		horizontalResize.addItemListener(new ItemListener() {
+			@Override
+			public void itemStateChanged(ItemEvent e) {
+				CONFIGURATION.setFfmpegAvisynthHorizontalResize((e.getStateChange() == ItemEvent.SELECTED));
+			}
+		});
+		builder.add(GuiUtil.getPreferredSizeComponent(horizontalResize), FormLayoutUtil.flip(cc.xy(1, 8), colSpec, orientation));
+
+		// builder.addLabel(Messages.getString("HorizontalResizeResolution"), cc.xy(1, 8));
```
SubJunk, line 218:
```suggestion
```

src/main/resources/i18n/messages.properties
```diff
@@ -248,6 +248,22 @@ EditUmsConfigurationFileManually=Edit UMS configuration file manually
...
+AlgorithmForConverting2Dto3D=Conversion Algorithm:
+PulfrichBase2Dto3DAlgorithm=Pulfrich Base
+PulfrichLighting2Dto3DAlgorithm=Pulfrich and Lighting Depth Detection
+OutputFormat3D=3D Output Format:
+FullSBS3dFormat=SBS (Full Side by Side) - Left Eye First
+FullTB3dFormat=TB/OU (Full Top/Bottom a.k.a Over/Under) - Left Eye First
+HalfSBS3dFormat=HSBS (Half Side by Side) - Left Eye First
+HalfTB3dFormat=HTB/HOU (Half Top/Bottom a.k.a Half Over/Under) - Left Eye First
+HalfUpSBS3dFormat=HSBS Upscaled (Half Side by Side) - Left Eye First
+HalfUpTB3dFormat=HTB/HOU  Upscaled (Half Top/Bottom a.k.a Half Over/Under) - Left Eye First
```
SubJunk, line 261:
```suggestion
HalfUpTB3dFormat=HTB/HOU Upscaled (Half Top/Bottom a.k.a Half Over/Under) - Left Eye First
```

src/main/resources/i18n/messages.properties
```diff
@@ -248,6 +248,22 @@ EditUmsConfigurationFileManually=Edit UMS configuration file manually
 Empty=Empty
 Enable=Enable
 EnableAvisynthVariableFramerate=Enable AviSynth variable framerate change into a constant framerate (convertfps=true)
+Avisynth2Dto3DSettings=2D to 3D Conversion Settings
+AlgorithmForConverting2Dto3D=Conversion Algorithm:
```
SubJunk, line 252: A lot of these keys don't match the convention of the file, which is generally to match the first few words of the string. For example this one would be `ConversionAlgorithm`

SubJunk (changes requested):
This looks good to me, just some minor convention changes

src/main/java/net/pms/configuration/UmsConfiguration.java
threedguru: Changed to is

src/main/java/net/pms/configuration/UmsConfiguration.java
threedguru: Changed to is

threedguru:
Should be good to go now.

src/main/resources/i18n/messages.properties
```diff
@@ -248,6 +248,26 @@ EditUmsConfigurationFileManually=Edit UMS configuration file manually
 Empty=Empty
 Enable=Enable
 EnableAvisynthVariableFramerate=Enable AviSynth variable framerate change into a constant framerate (convertfps=true)
+2Dto3DConversionSettings=2D to 3D Conversion Settings
```
SubJunk, line 251: The keys look good now, can you please sort it alphabetically?

SubJunk (changes requested):
Thanks for your changes, just one more formatting change I can see

SubJunk:
I'm testing it now. One thing to note is that the new settings have been added to the Java GUI but not the React one.
I need to add some instructions about how to develop it to the repository, but for now these are the instructions that were in the PR for the web settings:

Existing Maven commands will compile and include it, then when UMS is running you can access it via `localhost:9001/settings/` or clicking "Settings" via the system tray menu.
During development you can just use the `yarn start` command in the /configuration-client directory to develop with automatic serving, hot refreshing, etc.

src/main/resources/i18n/messages.properties
threedguru: Done. I have sorted all of the keys in the file and run UMS to test.

threedguru:
I saw the react UI. It didn't enable advanced settings without manual intervention (it was enabled for the thick client) so I did not realise the full extent of the current react implementation. I'll implement the react UI when I can. Do you have a VR headset for testing?

SurfaceS:
I just had a quick look on this, not in details.

As I understand, this PR is for VR headset (concern only few devices).
Be careful about configurations settings.
Some of them are renderer settings only, some others are general app settings.
For example, FFmpegGPUH265EncodingAccelerationMethod is a general app setting as it concern local UMS server hardware.
But maybe isFfmpegAvisynthHorizontalResize is a renderer concern ?

Also, the windows nsis setup need to be arranged.

@SubJunk AviSynth+ 3.7.2 may become the default AviSynth version ?
For now, it is installed only on windows.

https://avs-plus.net/ argue it is binaries for Arch Linux (Community, [x86-64](https://www.archlinux.org/packages/community/x86_64/avisynthplus/)), Arch Linux ARM (Community, [arm](https://archlinuxarm.org/packages/arm/avisynthplus), [armv6h](https://archlinuxarm.org/packages/armv6h/avisynthplus), [armv7h](https://archlinuxarm.org/packages/armv7h/avisynthplus), [aarch64](https://archlinuxarm.org/packages/aarch64/avisynthplus)) and have windows/macOS installers.

threedguru:
> I just had a quick look on this, not in details.
> 
> As I understand, this PR is for VR headset (concern only few devices). 

Well, in truth this supports any device that can display 3D content with DLNA media server support. It converts 2D video to 3D. I cannot overstate the range of content that this brings into the 3D world. With the demise of 3D TVs from 2017 onwards, this set of devices includes ALL VR headsets (e.g. 10 Million+ Oculus Meta Quest 2 devices sold - just one type), DLP Projectors as well as legacy 3DT TVs and all manner of 3D display devices such as the LumePad (displays in 3D without glasses or a headset). More specifically it supports the the media players that run on VR headsets and other devices including; Sky Box VR Player, Big Screen, Virtual Desktop, Pigasus etc. etc). I will add support for all of those. I hope any concern you have about "few devices" is alleviated - that is simply an incorrect statement. 
 
Be careful about configurations settings. Some of them are renderer settings only, some others are general app settings. For example, FFmpegGPUH265EncodingAccelerationMethod is a general app setting as it concern local UMS server hardware. But maybe isFfmpegAvisynthHorizontalResize is a renderer concern ?
> 

That is an interesting observation, but that setting is not just to match the resolution of the renderer. Given the CPU/GPU intensity of the conversion task, the user may choose to deliver the video at a resolution that is lower than the full capacity of the renderer/device in order to support the available bandwidth and to support transcoding of the video in real-time to avoid buffering. Might be worth some further discussion.

> Also, the windows nsis setup need to be arranged.
> 

I have already added to the pom the necessary settings to install the conversion script, supporting DLLs and the FFMS2 ffmpeg based codec alternative. Did this concern relate to any other elements in addition?

> @SubJunk AviSynth+ 3.7.2 may become the default AviSynth version ? For now, it is used by UMS only on windows.
> 

AviSynth+ is definitely a better alternative than AviSynth. As you may have seen the multi-threading syntax is different in AviSynth+ than AviSynth (I added an option to enable AviSynth+ multithreading support). The 64 version of AviSynth+ is also a good bit faster. I was planning to write, ideally generic, installation instructions for UMS to support 2D to 3D conversion. In the interest of exttracting maximum performance I have proven the work with the 64 bit version of AviSynth+, which I pre-install (UMS then does not install the 32-but AviSynth over the top of it). It may well work well with the 32 bit AviSynth version currently shipped with UMS, but I would need to go back and re-prove that.

For now, this is probably Windows only as I don't think the GPU as well as AviSynth+ support is there for Linux and Mac for Nvidia and AMD cards with ffmpeg in quite the same way. I am prepared to be proven wrong though.

> https://avs-plus.net/ argue it is binaries for Arch Linux (Community, [x86-64](https://www.archlinux.org/packages/community/x86_64/avisynthplus/)), Arch Linux ARM (Community, [arm](https://archlinuxarm.org/packages/arm/avisynthplus), [armv6h](https://archlinuxarm.org/packages/armv6h/avisynthplus), [armv7h](https://archlinuxarm.org/packages/armv7h/avisynthplus), [aarch64](https://archlinuxarm.org/packages/aarch64/avisynthplus)) and have windows/macOS installers.

threedguru:
I have not added a specific renderer for the BigScreen media player, but the 2D->3D support works really well with this UMS version. BigScreen supports the 3D SBS format. I cannot tell you how amazingly well this works. See brief example video:

https://streamable.com/b4viu6

SurfaceS:
First thanks for you contribution.
When I said few devices, I mean compare to billions of TV renderers.
I not mean this should not be implemented.

After a quick look on this (and I may have wrong) :

It seems that you add an option for ffmpeg to use hardware encoding for x264/265, what is nice.
It will impact all renderers on all OS, so it should be reviewed twice.
Future : Maybe it is a way to find what lib is actually available on the system (hardware+OS) with a ffmpeg cmd.


Then, you use AviSynth+.
AviSynth+ have many builds. (x86, x64, x86-xp, and so on) and have no compatibility with windows < XP (what is another things).
Plugins need to be on the same arch than avisynth+ build.
You provide mvtools2.dll, ffms2.dll on a 64bit version.
If we not provide the avisynth build, we cannot know the one that was installed by the user.
And it can change after install.
So either we provide the complete package on the installer (avisynth+ and plugins) and someone have to maintain this, or UMS can just check the presence of those libraries in the avisynth plugin folder, as we already do for `vsfilter`.

At end, `convert2dto3d.avsi` is the only needed file, the other ones depends if we want to package everything out of the box.
The good things is that `convert2dto3d.avsi` doesn't depends on arch or OS, and should be able to run on every servers (linux as well) with the correct libs (avisynthplus, ffms2, avisynth-plugin-mvtools2).

**I'll let @SubJunk choose what it preferable for UMS.**


`KEY_FFMPEG_AVISYNTH_ENABLE_PLUS_MODE` -> I don't think this settings is needed as we can check the `avisynth.dll` version when checking the `avisynth` presence to know that, and it is more accurate.
I'll work on this to add a founded version.

Also, we need to check each libraries presence that is needed for 2dto3d (or ffms2) before putting it on the script.

**About code :**
On this PR https://github.com/SurfaceS/UniversalMediaServer/tree/2d3d.
I moved the `AviSynth3DFormat` away from UMS conf to add it to Renderer conf as it depends to renderer.
Also, it allow AviSynth engine to not convert to 3D for non 3D renderers (nonsense).
Now, the AviSynth script add 2D to 3D only if the renderer conf have a AviSynth3DFormat.

threedguru:
> First thanks for you contribution. When I said few devices, I mean compare to billions of TV renderers. I not mean this should not be implemented.
> 

Ok, sorry if I misunderstood. Billions is more than millions. See my responses below. 

As a broad point, some of my design decisions were as a result of the voyage of discovery that I was engaged in when getting into UMS and making the solution work. I did not know at the start whether this 2d to 3d stuff would work effectively at all in the context of UMS. I am an architect, but I needed to make it work first and performance was key to that. I think there is scope for revisiting some of these decisions. I'll explain why I did it this way and then the way forward may become clearer.

I was learning about UMS as I went, as I had never looked in detail at it before, although I do remember using the PS3 Media Server when my son had a PS3. That was a while ago!

> After a quick look on this (and I may have wrong) :
> 
> It seems that you add an option for ffmpeg to use hardware encoding for x264/265, what is nice. It will impact all renderers on all OS, so it should be reviewed twice. Future : Maybe it is a way to find what lib is actually available on the system (hardware+OS) with a ffmpeg cmd.

Well, yes there is always scope for improvement. I used the list from the current version of ffmpeg shipped with UMS. However, this command could be used to extract the available codecs:


> 
> Then, you use AviSynth+. AviSynth+ have many builds. (x86, x64, x86-xp, and so on) and have no compatibility with windows < XP (what is another things). Plugins need to be on the same arch than avisynth+ build. You provide mvtools2.dll, ffms2.dll on a 64bit version. If we not provide the avisynth build, we cannot know the one that was installed by the user. And it can change after install. So either we provide the complete package on the installer (avisynth+ and plugins) and someone have to maintain this, or UMS can just check the presence of those libraries in the avisynth plugin folder, as we already do for `vsfilter`.
> 
> At end, `convert2dto3d.avsi` is the only needed file, the other ones depends if we want to package everything out of the box. The good things is that `convert2dto3d.avsi` doesn't depends on arch or OS, and should be able to run on every servers (linux as well) with the correct libs (avisynthplus, ffms2, avisynth-plugin-mvtools2).
> 
> **I'll let @SubJunk choose what it preferable for UMS.**
> 
> `KEY_FFMPEG_AVISYNTH_ENABLE_PLUS_MODE` -> I don't think this settings is needed as we can check the `avisynth.dll` version when checking the `avisynth` presence to know that, and it is more accurate. I'll work on this to add a founded version.
> 
> Also, we need to check each libraries presence that is needed for 2dto3d (or ffms2) before putting it on the script.
> 
> **About code :** On this PR https://github.com/SurfaceS/UniversalMediaServer/tree/2d3d. I moved the `AviSynth3DFormat` away from UMS conf to add it to Renderer conf as it depends to renderer. Also, it allow AviSynth engine to not convert to 3D for non 3D renderers (nonsense). Now, the AviSynth script add 2D to 3D only if the renderer conf have a AviSynth3DFormat.

SubJunk:
Good discussion here, thanks everyone. Yes I think it sounds good to detect the AviSynth version. I had always meant to release InterFrame for AviSynth+ and had users requesting it, but I don't use it anymore so I have put my time into other things. My TV does a decent job of it on its own.

threedguru:
> Good discussion here, thanks everyone. Yes I think it sounds good to detect the AviSynth version. I had always meant to release InterFrame for AviSynth+ and had users requesting it, but I don't use it anymore so I have put my time into other things. My TV does a decent job of it on its own.

There are a number of related things. Do you mean detecting both AviSynth/AviSynth+ and separately also detecting the x86/x74 (32 bit/64 bit). It would be great to do both and then make sure that the appropriate DLLs shipped with the product 32/64 bit are installed. The latter would mean looking again at interframe and mvtools. Are you also amenable to shipping AviSythn+ with UMS instead of AyiSynth? AviSynth is long in the tooth.

threedguru:
@SubJunk and @SurfaceS, I have implemented the react UI for the AviSynthFFMpeg transcoder but it does not seem to set the defaults and save/retrieve the settings consistently. I have tried to follow other similar examples in the code. What am I missing?

SurfaceS:
@threedguru 
you merged too early.

Now it's fixed.

threedguru (commented):
**Please Re-consider Moving 3D Output Format to Renderer File:**

I feel that moving the 3D output format to the renderer might be a mistake. I suspect that the norm with UMS is that the best device format is not directly related to the IT capabilities of the home network and server, but very largely dependent on the output device itself. However, for the resource intensive 3D work, the best format IS heavily dependent on the home network, and GPU/CPU capabilities of the server as well as the device/renderer itself as well as the specific video. The best format would be full resolution (formats 1,2,5 or 6), but I have made the default 4 - half resolution as a good starting point. This is not the optimum resolution, but the lowest common denominator in terms of the output file size from which the user can then choose to go higher, if they have a powerful enough server. The 3D media players, that I have looked at (SkyBoxVRPlayer,BigScreen and Pigasus), all support multiple 3D formats and the user might want to experiment on a per video level which one works best. Options 5 and 6 actually deliver double the original resolution in order to cater for devices that halve the size of the resolution received (either in width or height) - which means a larger file size and heavy processing requirement - which their server PC might not support. I'm just saying that the best 3D output format is not just (or even mainly) dependent on the renderer and so it should not be in the renderer file.
In addition, the ability to change the format while watching is well enabled through UMS's new react UI as you can do that without removing the headset via the device's web browser. This is virtually impossible if it is in the renderer file.
The next item is related and perhaps more of an issue. Whilst, I have successfully created renderer files for "Sky Box VR Player" and "BigScreen", I do not seem to be able to do that for the Pigasus media player (which I have purchased  just to set up the renderer). Its User Agent has no distinguishing features at all:

User-Agent: Android/10 UPnP/1.0 Cling/2.0

A Google search indicates that this User Agent is also matched by some versions of Kodi. How would one go about enabling 3D conversion for this device only? At the moment it just displays the Chromecast renderer? One would have to change the default renderer. This would also mean, as I have mentioned before, that there would need to be a custom renderer file for every 3D supporting device. It would also additionally mean that all AviSynth content would be served in 3D for that device, which may not be what the user wants.
I suggest retaining the existing UI selection rather than moving the setting to the renderer file. If there are other ways of addressing the items above then fine, let me know. UMS has many features that I am not aware of.

threedguru:
It now builds and runs, but it still does not work properly. The 2D to 3D AviSynth stiuff does not work any more. The settings are messed up. Last time I debugged it, it was because the tempFolder was null so it could not write the .avs file. I deliberately did not go into the react UI this time (that is a did not create a user) and it still does not work.

SurfaceS:
Reverted the setting on UMS, added a boolean on the renderer conf.
Fixed the tmp error.

The avisynth is now provided by UMS itself (UMS/bin on install), and it will look if the libraries are presents (avisynth, ffms2, mvtools2, convert2dTo3d script).

You should see something like this on the log.
```
INFO  20:02:35.347 [main] Founded AviSynth version 3.7.2.0
INFO  20:02:35.347 [main] Founded ffms2
INFO  20:02:35.347 [main] Founded mvtools2
INFO  20:02:35.347 [main] Founded convert2dTo3d script
INFO  20:02:35.348 [main] Transcoding engine "AviSynth/FFmpeg" is available
```

Checks are important if someday you want to use 2dto3d on linux/mac (just checking if the libraries are installed, as they exists for it).

SurfaceS:
about `User-Agent: Android/10 UPnP/1.0 Cling/2.0`
You need to use `UpnpDetailsSearch` as the User-Agent is too common.

threedguru:
> Reverted the setting on UMS, added a boolean on the renderer conf. Fixed the tmp error.

Are you saying that you reverted the change so that you can select the 3D Output format in the UI again and the temp folder error is also fixed? If so, thanks. How would I see those changes? A pull did not update the relevant files for me.
> 
> The avisynth is now provided by UMS itself (UMS/bin on install), and it will look if the libraries are presents (avisynth, ffms2, mvtools2, convert2dTo3d script).
> 
> You should see something like this on the log.
> 
> ```
> INFO  20:02:35.347 [main] Founded AviSynth version 3.7.2.0
> INFO  20:02:35.347 [main] Founded ffms2
> INFO  20:02:35.347 [main] Founded mvtools2
> INFO  20:02:35.347 [main] Founded convert2dTo3d script
> INFO  20:02:35.348 [main] Transcoding engine "AviSynth/FFmpeg" is available
> ```
> 
> Checks are important if someday you want to use 2dto3d on linux/mac (just checking if the libraries are installed, as they exists for it).

Again, how do I see those changes?

My thoughts on 3D output format:

Make the default renderer enable sending of 3D content instead of disabling it and define a default format (e.g. 4 - HTB/HOU). Then explicitly disable 3D output in the existing 2D renderer files. This would mean we don't need specific renderer files for any new 3D renderers. I suggest that users with a VR headset are more likely to have a TV that is already supported by UMS with a renderer file than the other way round. In this case both devices would work as intended. If they hit a problem, because there is no renderer for their TV, then they can define a renderer for their TV with 3D disabled. This would be a rarer occurrence than needing to provide a file for every software media player that can play 3D content - we are not detecting the VR hardware after all, but the media player - I guess there is no way to detect the hardware. What happens now with UMS and Smart TVs that presumably can have multiple different media players - do you detect those? 

It would also be good to leave 3D Output format in the renderer as a preference but also add UI selection back so that it can be overridden in the UI (add an additional default option that uses whatever the renderer default setting is), but let drop down selection override it. 

The boolean flag that you mention sounds similar but I haven't seen the code to follow exactly what you mean.

threedguru:
@SurfaceS I briefly tested the updates in your pull request. It's good. I like the changes and the 2D to 3D stuff works again. The react UI additions that I made still do not work properly yet though. I will take a look when I can.

SurfaceS:
PR https://github.com/threedguru/UniversalMediaServer/pull/3
fix react UI settings

threedguru:
> about `User-Agent: Android/10 UPnP/1.0 Cling/2.0` You need to use `UpnpDetailsSearch` as the User-Agent is too common.

The react UI changes are good. It seems to almost fully work now. I had to change String to Integer in the Thick Client for some of the drop down boxes - it went back and forward whilst I was trying to get it work, I see your last change made them ints again. There is potentially an issue with the "convert to constant fps", it did not show the default correctly in react when I first went in.

How do I get UMS to show me the UpnpDetails that it has gathered from the renderer (PigaSus) to create a pattern. It just picks up Chromecast and I cannot find a way of getting it to show me any other details. I even had to use the debugger to find the User Agent string - which was no help.

We are getting there I think.

threedguru:
@SubJunk and @SurfaceS, it is working well now. I wanted to give you guys an idea of the user experience in a VR headset with the VR players that we now have renderers for and with the React UI in place. This video shows the Avatar 2 trailer being played in each of the players; Big Screen, Pigasus and Sky Box together with changing the UMS settings via the React UI from the Oculus Web browser. It is very cool! BigScreen seems to not be as reliable as discovering UMS as the other two - not our problem I guess:

https://streamable.com/fijffp

threedguru:
Past tense of "to find" is "found" not "founded". There is another verb "to found" with past tense "founded", but that was not required here.

threedguru:
@SurfaceS / @SubJunk The seek/resume functionality for AviSynthFFmpeg does not seem to work anymore. It did the last time I tested it. It should not use the "-ss" argument to ffmpeg to implement this (I changed this behaviour as it would have been massively inefficient for a 2d->3d video as the seek would happen after the conversion rather than before), it should be handled within the script, before conversion. I can see that the detection of AviSynthFFmpeg in FFMPegVideo changed recently to use this form:

		boolean avisynth = this instanceof AviSynthFFmpeg;

There used to be an avisynth() method that returned true or false. I do not know if this change, and related ones, is the culprit. This code and the corresponding line in the debug log shows that, right now, it is incorrectly picking up the avisynth flag (for example when requested to advance 30 seconds):

		// Set seeks
		if (params.getTimeSeek() > 0 && **!avisynth**) {
			cmdList.add("-ss");
			cmdList.add(String.valueOf(params.getTimeSeek()));
		}

Because it ouputs this in the debug log:

Starting "C:\Program Files (x86)\Universal Media Server\bin\ffmpeg.exe" **-ss** 774 -i "Z:\SCL8F1~L\DS9...

This is obviously incorrect. It should be triggering the avs script to hande this with these lines:

				lines.add("seekFrame=int(video2d.FrameRate*" + **timeSeek** + "+0.5)");
				lines.add("video2dFromSeekPoint=Trim(video2d,seekFrame,0)");

Any idea why this is happening now?

SubJunk:
@threedguru this is so cool, and really making me want to get the VR gear! Thanks for the videos, they have been really helpful

As for the regression, is that just between v11 and v12, or did it happen before? There has been a lot of refactoring recently so it would be good to narrow it down

threedguru:
> @threedguru this is so cool, and really making me want to get the VR gear! Thanks for the videos, they have been really helpful

Thanks. I think it is cool too. I am keen to make it stable with resume etc., there are a lot of variables, see below. I do not think any of this is specific to the 2D to 3D conversion processing:

> 
> As for the regression, is that just between v11 and v12, or did it happen before? There has been a lot of refactoring recently so it would be good to narrow it down

The regression does not seem to be between v11 and v12. It is difficult to pin down. I made a commit that at least seems to make UMS again invoke the AviSynth script generation, on a resume, instead of passing the -ss parameter to ffmpeg (which fails). This is effectively a reversion of the line (method name changed though):

		boolean avisynth = this instanceof AviSynthFFmpeg;

To:

		boolean avisynth = isAviSynthEngine();

Following that change, what seems to happen now is the resume process thread is killed off. Probably within DLNAResource. When I made resume work the first time round, I noticed that Sky Box Media player (at least) was issuing multiple requests to seek to the same resume point. With each request DLNAResource killed the previous process before it completed - I guess the 3D generation process takes longer than a typical transcode would. I therefore extended DLNAResource so it ignores duplicate seek requests ( this has been in the code for some time now ) so that it does not kill resume processes before they have had time to complete:

			if (lastTimeSeek == params.getTimeSeek()) {
				LOGGER.debug("Duplicate time seek request: " + params.getTimeSeek() + " seconds, ignoring");
			} else {

However, that no longer seems to be enough. See the attached log extract. This log line shows that it is killing a correct resume process when it should not be:

DEBUG 2022-11-17 13:05:26.922 [External Process Stopper] Stopping process: ffmpeg.exe-62

The code to avoid killing processes on a duplicate seek request is still there though, so I do not understand what is happening, see the last line of the log extract, for example:

DEBUG 2022-11-17 13:05:41.817 [jupnp-netty-worker-22] Duplicate time seek request: 94.0 seconds, ignoring

I do not understand why this earlier sequence is not also ignored rather than killing the in-flight process:

DEBUG 2022-11-17 13:05:26.916 [jupnp-netty-worker-20] AviSynth will seek to time index: 94.0 seconds
DEBUG 2022-11-17 13:05:26.918 [jupnp-netty-worker-20] Creating pipe \\.\pipe\ffmpegvideo_116_1668690326918
DEBUG 2022-11-17 13:05:26.922 [External Process Stopper] Stopping process: ffmpeg.exe-62

This process should have been allowed to continue as the new request is a duplicate of the earlier request here, albeit 15 seconds later. 15 seconds is not an unreasonable length of time to allow AviSynth to seek to the position in the video (in this case 94 seconds in) and begin generation of the 3D version of the 1080p video from that point:

DEBUG 2022-11-17 13:05:11.478 [jupnp-netty-worker-2] AviSynth will seek to time index: 94.0 seconds

In addition (and I am not sure if this is a one off - it probably is not), UMS seems to get confused about the audio stream on a resume. The Avatar 2 trailer video plays fine from the start, but the resume processing throws an error in ffmpeg related to not being able to find the audio stream - why should that be? See second log file extract and the lines below. It starts playing from the start fine. I tried this with both DirectShowSource and FFMS2 with the same result:

This is playing from the start - which works:

DEBUG 2022-11-17 12:53:13.118 [jupnp-netty-worker-8] AviSynth will seek to time index: 0.0 seconds

No error beyond this point until the resume request here:

DEBUG 2022-11-17 12:53:36.538 [jupnp-netty-worker-9] AviSynth will seek to time index: 66.0 seconds

Where ffmpeg generates this error:

DEBUG 2022-11-17 12:53:38.485 [ffmpeg.exe-47-2] Stream map '0:a:-1' matches no streams.

It is the same video and script so this should not happen.

This is all quite confusing.

[resume_error_log.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/10032341/resume_error_log.txt)

[resume_audio_error_log.txt](https://github.com/UniversalMediaServer/UniversalMediaServer/files/10032476/resume_audio_error_log.txt)

threedguru:
Virtual Desktop is a very popular VR application. This is how we can support via VLC Media Player. I had to modify the VLC renderer file to enable the 2D to 3D AviSynth stuff. I am not entirely convinced that having it off by default is the right move:

Virtual Desktop:

The approach with Virtual Desktop is slightly different than to the other renderer clients in that Virtual Desktop does not support DLNA/UPNP and so cannot talk to UMS directly. However, VLC does support DLNA/UPNP. The key is the View/Playlist menu in VLC Media Player for Desktop.

The approach is to connect VLC Media Player to UMS and then play the 2D to 3D converted video in VLC viewed through Virtual Desktop. 

There is an option to enable 3D SBS viewing mode in Virtual Desktop which does a really great job of displaying VLC's 3D SBS output (generated by UMS) in glorious 3D.

See video for a demo (excuse placement of explanatory text - it has moved during rendering for some reason):

https://streamable.com/gzk9xk

SubJunk:
@threedguru nice resume fix, thanks! The VLC playback looks great. What is remaining on this branch, or is it ready for review/merge now?

threedguru:
It can be merged now.

A general issue, not related to the mod, is being able to fast forward and
rewind on AviSynth stuff in general. I don't know if this is a UMS or
renderer issue. Any thoughts? It should work like resume, but the renderers
or UMS don't treat it the same.

threedguru:
@SubJunk, @SurfaceS so when is this going in? I've been telling people that this will be in release 12, but I can see that 12 has already been released. Some have paid patreon to get it and the feature isn't there. I've had to make a development build now to get more feedback on testing and proving. Avatar 2 is out on 16th December. This will re-energise interest in 3D. I don't understand why this is not in Release 12 yet.

SubJunk:
@threedguru I wanted to separate this from the first v12 release to make it easier to track down potential bugs. I think it can be in the next release, 12.1

SubJunk (approved):

SubJunk:
@threedguru are there any details in particular you want me to add to the GitHub, forum, Facebook and Twitter release announcements? For now I just have

```
Added 2D to 3D conversion for virtual reality, for full details see https://iwantaholodeck.com/on-the-fly-2d-to-3d-video-conversion-with-universal-media-server-ums/
```

In the changelog, but it might be good to have images/videos to share to help people understand what the feature is and why they might want it

threedguru:
Ok, that makes sense. Thanks for the update.

threedguru:
> @threedguru are there any details in particular you want me to add to the GitHub, forum, Facebook and Twitter release announcements? For now I just have
> 
> ```
> Added 2D to 3D conversion for virtual reality, for full details see https://iwantaholodeck.com/on-the-fly-2d-to-3d-video-conversion-with-universal-media-server-ums/
> ```
> 
> In the changelog, but it might be good to have images/videos to share to help people understand what the feature is and why they might want it

Thanks for asking. The installation and configuration guide link that you have is fundamental. These videos also provide good illustrations of the various aspects:

UMS Web UI, 2D to 3D Function and Pigasus VR Player:
https://streamable.com/2x3h9b

BigScreen VR Player with UMS Dynamically Converting Avatar 2D Trailer to 3D:
https://streamable.com/isj75w

How it works explainer with SkyBox VR Player example:
https://streamable.com/ijojmx

Algorithm Example - Avatar Trailer 2D to Full 3D SBS:
https://streamable.com/4od2ns

SubJunk:
Ok great. I will prioritise the release for Avatar timing

threedguru:
> Ok great. I will prioritise the release for Avatar timing

That is definitely a good move. I'll update the text and download link to the official release in the guide as soon as it is released. I've made a minor update for now to say it will be out around the time of the Avatar film release. So, as you had before this is the install and configuration guide url link to use:

https://iwantaholodeck.com/on-the-fly-2d-to-3d-video-conversion-with-universal-media-server-ums/

SubJunk:
It looks like your website is down so I can't get that image
image

threedguru:
Hi,

Yes, having network issues. Engineer called. I have attached the image
file. I'll fix the hosting.

SubJunk:
I don't think the image came through. It's ok, we have time. The release is ready to go when the site is back up

threedguru:
The website is back up. It should all be working now.

SubJunk:
Sure I can add you to the thanks section in the readme if that's what you mean. The release is done now on Patreon so you can do the edits on the blog post.

threedguru:
Cool, thanks, I've edited the blog accordingly.

Mik-S-UMS:
@threedguru could you help a few people on the forum that are having difficulties with 3D conversion as I don't know anything about how it works. 

One wants to know if [MPC-HC can be used instead of VLC](https://www.universalmediaserver.com/forum/viewtopic.php?t=15571) and another has the [conversion stopping after a few min](https://www.universalmediaserver.com/forum/viewtopic.php?t=15574)

Is there a place where you prefer people can contact you for help with this when you are not on here or the forum?

threedguru:
I have a blog site with all of the details here:

https://iwantaholodeck.com/on-the-fly-2d-to-3d-video-conversion-with-universal-media-server-ums/

People can sign up for help there. I suggest you have a skim through that
page just for your own understanding on how I have been telling people to
set it up.

It's picking up quite a bit right now. I don't think MPC-HC supports DLNA.

Mik-S-UMS:
The `UserAgentSearch` for the Pigasus.conf is too generic as it only matches "Android". There is another [user](https://www.universalmediaserver.com/forum/viewtopic.php?p=47328#p47328) where their Cambridge Audio player is being identified as Pigasus as that also uses "Android" in the User-agent string.

I'll do a conf for the CA player but the pigasus.conf could be improved with a more definite match.

PR #3576, opened 2022-11-05
Fix TV series never being found locally by IMDb ID

SubJunk:


SurfaceS:
`/Program Files (x86)/` stay the installation directory only if it is the destination folder on the registry.

SubJunk:
@SurfaceS I added a new one instead now

js-kyle (approved):

PR #3577, opened 2022-11-05
Confs

SurfaceS:
**UMS :**
-  refactor `configuration` into `umsConfiguration` for clarity

**BaseConfiguration:**
- created class that is base class for `UmsConfiguration` and `DeviceConfiguration`.

**UmsConfiguration :**
- extends `BaseConfiguration`
- remove extends `RendererConfiguration` as it has nothing to do with it.

**ConnectedRenderers**
- created class that handle all the connected renderers arrays.
- it will need to move found upnp renderers arrays on it one day (not urgent).

**Renderer :**
- renderer is now a fully found renderer device.
- it's not anymore the base class for `RendererConfiguration` what is nonsense.
- `RendererConfiguration` was confusing between a renderer device or only a renderer conf.
- now have non settings things from `RendererConfiguration` / `DeviceConfiguration` on itself.

**RendererConfiguration :**
- extends `BaseConfiguration`
-  moved `getConfigurationAsJsonString()` to the `SettingsApiServlet`

**DeviceConfiguration : -> RendererDeviceConfiguration**
-  refactored to `RendererDeviceConfiguration`.
-  now use the provided `umsConfiguration` instead of extending `UmsConfiguration`.
-  extends `RendererConfiguration` as it should.
-  moved statics `RendererConfiguration` specific to device (old `deviceConfs`) to `RendererConfigurations` (devices conf are RendererConfiguration with id).
-  moved non setting things to `Renderer` class.
-  moved renderers inet<->uuid (old `XREF`) to `ConnectedRenderers`.


src/main/java/net/pms/network/webguiserver/servlets/SettingsApiServlet.java
```diff
@@ -359,4 +361,37 @@ public static String getConfigurationUpdate(String key) {
...
+				if (SettingsApiServlet.haveKey(strKey)) {
+					String strValue = Objects.toString(value);
+					//do not add non acceptable empty key then it back to default
+					if (StringUtils.isNotEmpty(strValue) || SettingsApiServlet.acceptEmptyValueForKey(strKey)) {
+						//escape "\" char with "\\" otherwise json will fail
+						propsAsStringMap.put(strKey, strValue.replace("\\", "\\\\"));
+					}
+				}
+			}
+		);
```
SubJunk, line 392:
```suggestion
		configurationAsProperties.forEach((key, value) -> {
			String strKey = Objects.toString(key);
			if (SettingsApiServlet.haveKey(strKey)) {
				String strValue = Objects.toString(value);
				//do not add non acceptable empty key then it back to default
				if (StringUtils.isNotEmpty(strValue) || SettingsApiServlet.acceptEmptyValueForKey(strKey)) {
					//escape "\" char with "\\" otherwise json will fail
					propsAsStringMap.put(strKey, strValue.replace("\\", "\\\\"));
				}
			}
		});
```

SubJunk (approved):

PR #3590, opened 2022-11-07
react web player renderer review

SurfaceS:


SurfaceS:
- create `WebGuiRenderer` from old `WebRender` with only needed code, and remove most of the non `Renderer` base code.
- create `WebGuiPlayer` from old `WebPlayer` with only needed code.
- added connected username to the renderer display name (like Me@Firefox).
- refactor react gui player `token` to more comprehensive `uuid`.
- fix many bugs in react player.
- `PlayerApiServlet` now use `WebGuiRenderer` (`Renderer`) instead of `RootFolder` (uniform code).
- moved old react gui player `RootFolder` array to a `Renderer` array on `ConnectedRenderers` (uniform code).
- better handle player sse.

SubJunk (approved):

PR #3605, opened 2022-11-13
fix #3603 fully played

### LINKED ISSUE #3603: Mark directory as fully played feature gone from Java GUI
@SurfaceS this probably happened in the refactoring of the shared content tab. You used to be able to right-click a directory and mark it as played or unplayed, but now nothing happens on right-click. Can you look?

SurfaceS:
Doesn't find how to add popup only for specific row (seems not implemented in java).

BTW, it fix #3603


SubJunk (approved):

SubJunk:
@SurfaceS thanks

PR #3637, opened 2022-11-21
fix fileless folders

### LINKED ISSUE #3631: Multiple permissions prompts on macOS since v12
Using the default shared directories, the OS prompts me repeatedly for access to Photos. It pops up about 30 times until it's done, and it starts again each time I open the app. It also does it for the Downloads folder if I share that, screenshots are in https://github.com/UniversalMediaServer/UniversalMediaServer/issues/3622#issuecomment-1320442520

In v11 it only prompts one time.

image

I think the reason is that for some reason, v12 is not saving the security/privacy preference, while v11 did that. I can see in my list of apps with permissions, that v12 is not in the list:

image

but in v11 those things get added to that list:

image

SurfaceS:
Do not add fileless FolderContent

May fix #3631


SubJunk:
Tested and it fixes that problem, thanks!

SubJunk (approved):

PR #3648, opened 2022-11-23
fix metadata + added fields on react player

SurfaceS:
DLNAMediaVideoMetadata -> refactored to MediaVideoMetadata (not dlna related)

**MediaTableAudiotracks, MediaTableChapters, MediaTableSubtracks**
- give data themself (not from the MediaTableFiles).

**MediaVideoMetadata**
- now contain everything under it (api data as well).
- direct update MediaTableVideoMetadata
- add plot, tagline and votes

**TvSerieMetadata**
- created class with api data.
- direct update MediaTableTVSeries

**APIUtils**
- fix country is string with comma not array
- set api url static final
- fully set MediaVideoMetadata
- fully set TvSerieMetadata

**React player** 
- fix logo and poster
- added Rated to reflect old interface
- added Year Started to reflect old interface
- added Total Seasons to reflect old interface
- added IMDB link to reflect old interface

about #3575, after some requests, it's appear the API server does not give these data.

**API :**
- fix missing data db save.
- fix thumbnail creation that take the parser one, when the API poster is created/saved before the parser finish.
- fix API request frequency when a result does not contain imbdid.

PR #3657, opened 2022-11-28
added languages for API

SurfaceS:
Add all languages for the API.
Then the UMS player have translated strings according to selected language.


SurfaceS:
Sorry, it is based on the #3648 so it is hard to see the changes.

SurfaceS:
@SubJunk  : To see the real changes, check here :
https://github.com/SurfaceS/UniversalMediaServer/pull/122

SurfaceS:
merged main and added some quick code for the api side to get ride of the local info.movito.themoviedbapi use.

SurfaceS:
modified according to https://github.com/UniversalMediaServer/api/pull/1145

added tmdbId on tables
added reference more readable on tables
refactored some table columns

SurfaceS:
Tested after the api merge.

Works great.
Go on a movie (or series, episode) on player, change language, enjoy.

SurfaceS:
added modified column to tables (future use, to fix [#831](https://github.com/UniversalMediaServer/api/issues/831)).

On react interface :
update mantine -> 5.9.5.
added tmdb attribution on react about page (as asked from rules).
fix react player media page title.
added tmdb link to react player media page.
added tagline to react player media page.

FYI : 
Most movies/tv episodes stored in the **api database** does not have the tmdb id or tmdb tv id stored.
The api store it only from the last PR.
So even now, UMS requests will not have it if movie/tv episode was found before the [PR](https://github.com/UniversalMediaServer/api/pull/1145).

Most tv shows stored in the **api database** have tmdb id stored.
But UMS database do not store it until this PR, so all current stored tv shows do not have the tmdb id.

Basically, that mean that tmdb link to react player media page will be visible on few media for now.

SurfaceS:
should be done.

src/main/java/net/pms/database/MediaTableTVSeries.java
```diff
@@ -737,6 +782,18 @@ public static JsonObject getTvSeriesMetadataAsJsonObject(final Connection connec
 						result.addProperty("rating", MediaTableVideoMetadataIMDbRating.getValueForTvSerie(connection, id));
 						result.add("ratings", MediaTableVideoMetadataRatings.getJsonArrayForTvSerie(connection, id));
 						result.addProperty("released", MediaTableVideoMetadataReleased.getValueForTvSerie(connection, id));
+						if (lang != null && !"en-us".equalsIgnoreCase(lang)) {
+							VideoMetadataLocalized loc = MediaTableVideoMetadataLocalized.getVideoMetadataLocalized(connection, id, true, lang, imdbID, "tv", tmdbId, null, null);
+							if (loc != null) {
+								loc.localizeJsonObject(result);
+								//temp fix to store tmdbID if it was not before
```
SubJunk, line 789: I'm hesitant about temporary fixes since they often stick around as we forget to remove them, and it can be hard to know when it's ok to remove them. If this is really the best strategy, can we leave a comment in the code that details exactly what the plan is, including when this can be removed?

SurfaceS: In fact, it is not a temp fix as the situation can be found even with the current code.

It test the presence of the tmdb ID, and store it if it was not present.
So it will happen only once.

SubJunk:
I'm having a hard time testing it because when I change the language on the web interface, it gets stuck with a loading overlay that doesn't end

SurfaceS:
> I'm having a hard time testing it because when I change the language on the web interface, it gets stuck with a loading overlay that doesn't end.

It is a bug with rtl not related to this PR.
I never found why it do that.

SurfaceS:
> the language is changed, but I don't see any different metadata.

sorry... bug fixed

SurfaceS:
> I'm having a hard time testing it because when I change the language on the web interface, it gets stuck with a loading overlay that doesn't end.

@SubJunk Found it : the lazy load break the theme things.

Fixed.

SubJunk:
@SurfaceS nice work, I'll test more today

SubJunk:
@SurfaceS this works great now, it's amazing. It's so smooth. I think there will be many happy users with this

SubJunk (approved):

PR #3692, opened 2022-12-15
x265 changes:

SubJunk:
- Use ultrafast preset for higher fps
- Fix buffer underflow errors
- Stop passing in MPEG-2 video options


SubJunk:
@UniversalMediaServer/developers I have been running this and it has fixed the problems I had with x265 transcoding, it would be great to get it reviewed

Issue #3698, opened 2022-12-16
Shared Contents Location Not Saved


OldMan100:
No matter how any times I've tried, version 12.0.1 will not save the addresses I've placed in shared contents location, whether I use a Windows Letter or Network address to specific the location of my files. After the program is shutdown and re-started, only the three user location are still there. This is the same through the web interface. Through both the program and interface, everything looks and appears to be saved, once the program is shutdown and restarted, all the locations a user enters are gone. Once I re--installed 11.6, everything was good again.

[ums_dbg_2022-12-15-22-44.zip](https://github.com/UniversalMediaServer/UniversalMediaServer/files/10243377/ums_dbg_2022-12-15-22-44.zip)

SubJunk:
Can you please try this version and let me know if that works? It works for me https://www.universalmediaserver.com/uploads/UMS-boblatimer.exe

OldMan100:
After installing 13, whatever is causing 12.0.1 from saving any shared locations add by the user, is present in version 13. Right now, version 11.6.1 is the last good version.

SurfaceS:
@SubJunk v13 does not have the fix PR

SubJunk:
I’m away from my build computer but will upload a fixed version for you in the next few days

SubJunk:
@OldMan100 this should be fixed in this version https://www.universalmediaserver.com/uploads/UMS-OldMan100.exe
Please let us know if it works

infectormp:
@SubJunk I installed (clean install) version you provided and it doesn't save newly added Shared Content entry. UMS saves only entry added through wizard at first run.

SubJunk:
@infectormp is that when adding it via the old settings or the browser? Can you make a new issue for it?

Issue #3697, opened 2022-12-15
Unknown video feeds & podcasts listed as shared content, cannot be removed


aemondis:
Upon first load of the GUI and creation of an admin user, the shared folders list numerous podcasts, video feeds, etc. that cannot be removed. These are not channels I created, or have even accessed before. When I try to remove them via the GUI, then click save, they reappear again. Refer to screenshots and attached trace:
![image](https://user-images.githubusercontent.com/22867555/207986662-3e92a50d-d092-48f4-b2e2-689f78732f30.png)
![image](https://user-images.githubusercontent.com/22867555/207986791-1c4f29e2-bfb6-42dc-8913-86503b2ba457.png)

[ums_dbg_2022-12-16-10-18.zip](https://github.com/UniversalMediaServer/UniversalMediaServer/files/10241376/ums_dbg_2022-12-16-10-18.zip)

aemondis:
I can also confirm that this version does not resolve the issue: https://www.universalmediaserver.com/uploads/UMS-boblatimer.exe

It seems however that I can mark any of those feeds as "disabled" with no issue, and this saves successfully. When I click "Remove" via the GUI the item disappears from the web page, but once I click save it says "successful", only for the items to reappear immediately at the bottom of the list of shared content.

SubJunk:
Thanks for reporting this. That's the correct set of default sample channels but yeah that's a bug that they can't be removed. I'll have a look

aemondis:
Glad to hear I'm not going crazy!

I do have something new to add to this. After closing the UI in the browser (without closing the browser), then re-loading the UI and logging in, the shared folder changes now appear to have taken effect, and all the defaults have gone...

The joys of working with physical files! FWIW - this is Win11 on an insider build, so always possible there's some weird interaction going on behind the scenes with the way files are accessed on disk.

SurfaceS:
this should be fixed in #3699 and #3703

SubJunk:
I’m away from my build computer but will upload a fixed version for you in the next few days

SubJunk:
@aemondis this should be fixed in this version https://www.universalmediaserver.com/uploads/UMS-OldMan100.exe
Please let us know if it works

PR #3699, opened 2022-12-16
Shared Configurations

SurfaceS:
don't reload when writing file.
`readAllConfigurations` better handle when it need to save.

### DIFF
--- a/react-client/src/components/SharedContent/SharedContent.tsx
+++ b/react-client/src/components/SharedContent/SharedContent.tsx
@@ -44,6 +44,9 @@ export default function SharedContent() {
     if (sse.userConfiguration === null) {
       return;
     }
+    if (sse.userConfiguration['shared_content']) {
+      configuration['shared_content'] = undefined;
+    }
     const userConfig = _.merge({}, configuration, sse.userConfiguration);
     sse.setUserConfiguration(null);
     setConfiguration(userConfig);
--- a/src/main/java/net/pms/configuration/sharedcontent/SharedContentConfiguration.java
+++ b/src/main/java/net/pms/configuration/sharedcontent/SharedContentConfiguration.java
@@ -123,9 +123,11 @@ public class SharedContentConfiguration {
 	}
 
 	public static void updateSharedContent(SharedContentArray values, boolean save) {
+		LOGGER.debug("New shared content configuration sent.");
 		boolean updated = false;
 		synchronized (SHARED_CONTENT_ARRAY) {
 			if (!values.equals(SHARED_CONTENT_ARRAY)) {
+				LOGGER.debug("Updating shared content configuration");
 				SHARED_CONTENT_ARRAY.clear();
 				//check viability
 				for (SharedContent sharedContent : values) {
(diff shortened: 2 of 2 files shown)


SubJunk:
Not sure if this was ready for testing but I tried it and I have the same bugs as before, and I can reproduce this too where closing and starting UMS undoes new folder additions https://github.com/UniversalMediaServer/UniversalMediaServer/issues/3698

SurfaceS:
I can't reproduce this error from this PR.
if I add or remove with both gui, I see the java one instantly update.

The only bug is a visual bug an react side, due to :
`const userConfig = _.merge({}, configuration, sse.userConfiguration);`

In fact, when configuration.shared_content contain thing on it, merge will keep it.

ExSport:
Hi guys
It seems I spotted similar bug when cleaning up some default shares which stayed in the list, only reordered until I force refreshed the browser. Hopefully the fix will resolve this behavior as well :-) Thank you for great work, btw!
More info here: https://github.com/UniversalMediaServer/UniversalMediaServer/discussions/3687#discussioncomment-4422359

SubJunk:
Using this latest version I reproduced the following bug:
I removed two shared items via web interface (one folder and one image feed) and then hit save. The items are correctly updated in the Java GUI, but the web interface added the folder back, and duplicated the last entry:
image
Refreshing the page fixes that, and the content is still saved correctly after restarting the program. So the bug where the content comes back after closing UMS is not consistent. I guess there are a few different bugs in this area to fix. I'm going to prioritize this and hopefully do a release very soon. Today if I can fit in the timing

SurfaceS:
@SubJunk did you try the latest ?? 

> The only bug is a visual bug an react side, due to :
> const userConfig = _.merge({}, configuration, sse.userConfiguration);
> 
> In fact, when configuration.shared_content contain thing on it, merge will keep it.

https://github.com/UniversalMediaServer/UniversalMediaServer/pull/3699/commits/bc2f73a763d7dc2c4ec090001436bdf18710ad0a

SurfaceS:
> So the bug where the content comes back after closing UMS is not consistent. I guess there are a few different bugs in this area to fix.

I think the first commit https://github.com/UniversalMediaServer/UniversalMediaServer/pull/3699/commits/d40a6bf606dec3e9254cea3060cd0e9f0fa20a40 fix that.

SubJunk:
You're right, https://github.com/UniversalMediaServer/UniversalMediaServer/commit/bc2f73a763d7dc2c4ec090001436bdf18710ad0a seems to fix the removal bug. The quickrun maven command didn't update the react app when I tested the first time.
I will do some more playing around to see if there are any other bugs but this looks good

SubJunk:
It seems a similar bug is still there when I remove an entry via Java GUI. Here I removed two entries so I get two duplicates at the end:
image

SubJunk:
@SurfaceS I think that is fixed here https://github.com/UniversalMediaServer/UniversalMediaServer/pull/3703
so between our fixes I think we have a good state now

SubJunk (approved):

SurfaceS:
fix #3697
fix #3698

PR #3727, opened 2022-12-24
Enhancement to 2D to 3D Conversion Algorithm

threedguru:
The base [2D to 3D conversion algorithm](https://iwantaholodeck.com/background-and-further-reading/), currently implemented in UMS, takes two time delayed video frames and maps them to the left and right eye views.

This is both an extremely simple and effective approach. It relies on the [Pulfrich effect](https://iwantaholodeck.com/background-and-further-reading/). It does, however, require that there is relatively good correlation between the video frames selected for the left and right eye views. If this is not the case then the approach temporarily breaks down. This typically happens when the source video contains fast moving action that generates excessive change from video frame to video frame or where there is a wholesale change of scene.

These large scale frame to frame changes, translating to large‌‌ discrepancies in the left and right eye views, both destroy the 3D Pulfrich effect and cause a brief feeling of discomfort or disorientation for the viewer.

This algorithmic enhancement completely eliminates any sense of discomfort or disorientation due to rapid motion or scene changes. It does this by first detecting rapid motion and scene changes between frames. It then ensures that, for those instances, the same view for the left eye and right eye is kept, thereby eliminating the risk of disorientation as the view temporarily reverts to 2D. The video samples below provide a demonstration of the algorithm in action.

The first video shows a 2D to 3D converted version of the "Black Panther: Wakanda Forever Trailer". It uses the Pulfrich base algorithm, but in addition a mask has been generated where rapid motion or scene changes are detected in the video. The mask shows up as a pink frame allocated to the right eye (in bottom half of the video). The format of this video and the videos below is 3D Half Top-Bottom (HTB):

https://streamable.com/0bvkkg

The next video shows the result of replacing the pink mask, generated for the right eye, with the same frame content as used for the left eye - effectively returning the video to 2D for those frames:

https://streamable.com/92g1oe

The next video shows the result using the original base Pulfrich algorithm with no motion or scene detection for comparison purposes:

https://streamable.com/w1bkac

The main code enhancement is to the "convert2dTo3d.avsi" AviSynth script, plus the addition of the supporting library "masktools2", the other library that it uses "mvtools2" is already included in UMS.

The revised script uses the motion vector analysis (MAnalyse) and scene change detection routines (MSCDetection) from the 
mvtools2 library, see documentation and github repository below:

http://avisynth.org.ru/mvtools/mvtools2.html#version2
https://github.com/pinterf/mvtools
 
In addition, it uses the mt_merge routine from the masktools2 library,
see documentation and github repository below:
 
http://avisynth.nl/index.php/MaskTools2/Mt_merge
https://github.com/pinterf/masktools
 
Although only a few lines of script are required to invoke these routines, they are very effective in addressing the excessive motion and scene change issues outlined above.

See article here for further details: https://iwantaholodeck.com/enhanced-2d-to-3d-algorithm/

This pull request also updates the "SkyBoxVRPlayer.conf" renderer file, as the most recent version of Sky Box VR Player has changed the User Agent so that the previous version of the configuration no longer works.


UMS.launch
```diff
@@ -11,7 +11,7 @@
     </listAttribute>
     <stringAttribute key="org.eclipse.jdt.launching.CLASSPATH_PROVIDER" value="org.eclipse.m2e.launchconfig.classpathProvider"/>
     <stringAttribute key="org.eclipse.jdt.launching.MAIN_TYPE" value="net.pms.PMS"/>
-    <stringAttribute key="org.eclipse.jdt.launching.PROJECT_ATTR" value="ums"/>
+    <stringAttribute key="org.eclipse.jdt.launching.PROJECT_ATTR" value="UniversalMediaServer"/>
```
SubJunk, line 14: This line seems to go back and forth, like here https://github.com/UniversalMediaServer/UniversalMediaServer/commit/4adae84f2300e24d3a6820543e9a169bc9cef21c
@ik666 @threedguru can you two figure it out? I don't use Eclipse

src/main/external-resources/renderers/SkyBoxVRPlayer.conf
```diff
@@ -10,12 +10,13 @@ RendererIcon = SkyBoxVRPlayer.png
 # This renderer has sent the following strings:
 #
 # User-Agent: Linux/4.19.81+, UPnP/1.0, SKYBOX STUDIO for UPnP devices/1.14.12
+# User-Agent: Linux/4.19.81+, UPnP/1.0, Portable SDK for UPnP devices/17.1.5
 # ============================================================================
 
-UserAgentSearch = SKYBOX\s{1}STUDIO
+UserAgentSearch = Linux/4.19.81+
```
SubJunk, line 16: It's a shame this is so generic. Can you add a test for it to this file to help us detect conflicts? https://github.com/UniversalMediaServer/UniversalMediaServer/blob/main/src/test/java/net/pms/configuration/RendererConfigurationTest.java

src/main/java/net/pms/configuration/ConfigurableProgramPaths.java
```diff
@@ -16,27 +16,31 @@
...
+
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.ThreadSafe;
-import net.pms.util.ExternalProgramInfo;
-import net.pms.util.ExecutableInfo;
-import net.pms.util.FFmpegProgramInfo;
-import net.pms.platform.PlatformProgramPaths;
-import net.pms.util.FileUtil;
+
```
SubJunk, line 30: There are a few files here with these blank lines added in the imports, can that be reverted?

SubJunk (changes requested):
It all looks good, just the blank lines in imports, and the eclipse config attribute

UMS.launch
threedguru: That attribute needs to have a value that is the same as the name of the eclipse project (which is the same as the folder name of the project). My eclipse project is called UniversalMediaServer, I guess someone else has named theirs ums.  The attribute was set to the value UniversalMediaServer when I first started working on UMS earlier this year - someone must have changed it to match their project - and I find that I need to keep changing it back or I cannot run UMS in Eclipse as it complains about the project being missing. Obviously, the eclipse project name needs to be standardised - to UniversalMediaServer (I would suggest!).

src/main/java/net/pms/configuration/ConfigurableProgramPaths.java
threedguru: Done

src/main/external-resources/renderers/SkyBoxVRPlayer.conf
threedguru: I've added tests for SkyBox, Pigasus and Bigscreen VR Players.

UMS.launch
SubJunk: I see, that makes sense to me. Can you rename yours to UniversalMediaServer @ik666 ?

SubJunk (approved):

PR #3731, opened 2022-12-26
Stop treating LG video thumbnail requests as play/stop events

SubJunk:
LG TVs send play requests in their Media Player app to get dynamic thumbnails on hover, which causes all kinds of problems, like marking the video as fully played if the thumbnail requested is near the end of the video. Here we detect the unique user-agent of those requests and treat them differently.

Note: I didn't make it an option in the renderer configs because I want it to be future-proof too, even when we don't have a config yet

### DIFF
--- a/src/main/java/net/pms/dlna/DLNAResource.java
+++ b/src/main/java/net/pms/dlna/DLNAResource.java
@@ -3076,7 +3076,7 @@ public abstract class DLNAResource extends HTTPResource implements Cloneable, Ru
 								} else {
 									res = res.clone();
 								}
-					
+
 								root.stopPlaying(res);
 							}
 						}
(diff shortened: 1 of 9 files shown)


SurfaceS:
I don't think you want to mess up with `StartStopListenerDelegate`, `startPlaying` or `stopPlaying`.
Using it say _I'm playing this media_, with all the thing under it (gui, db, analyze, etc).

I think you would prefer something more simpler like that (java http server modified as well as the jetty one): 
https://github.com/SurfaceS/UniversalMediaServer/commit/399932f44fdcd94ed960b94aaac4668d2fcc2ffc

SubJunk:
@SurfaceS you're right, that's nicer code. I'll test it

SubJunk:
It works too so I'll replace this

PR #3735, opened 2022-12-27
Remove Content-Length header for chunked response

kabel:
Including both Content-Length and Tranfer-Encoding: chunked is an HTTP protocol violation and breaks proxies.


SurfaceS:
Thx for your contribution.
We do not need chunked if we know the content length, fixed in #3742.

PR #3743, opened 2022-12-30
Bundle of fixes

SubJunk:
The changes are described in the commit messages, please review @UniversalMediaServer/developers 
I have more changes coming to improve muxing of HDR content but this is a good start I think. I have been running it personally so it is well tested


SurfaceS (approved):

SurfaceS:
FYI only : Careful with HDR and hardware ffmpeg encoder.
Some cards will fail with HDR10.

PR #3758, opened 2023-01-04
2D to 3D Enhancement: Added Motion Compensation to Right Eye View

threedguru:
Enhancement to 2D to 3D Algorithm:

Further reduces risk of discomfort:

* Reduced the threshold at which the algorithm rejects a frame for 2D to 3D processing, as the result of too much frame to frame motion, from about 20% to 10% of the frame.
* Introduced a new element that performs full frame motion compensation on the right eye view so that it always aligns precisely with the left eye view.

See details and examples of output here:

https://iwantaholodeck.com/algorithmic-tuning-motion-and-alignment/


SubJunk:
@threedguru nice work!

SubJunk (changes requested):
This looks great, just the new lines in the imports

threedguru:
Thanks. I'll get rid of the lines again, but I think that there may be Eclipse formatting rule preferences, perhaps custom defined for UMS at some point in the past, that keep putting these blank lines back in. I could be wrong - I will take a closer look in the morning. Did you buy a headset yet? :)

threedguru:
@SubJunk as I suspected, there is an xml file in the root of UniversalMediaServer called Eclipse.Formatter.xml. Its from 2017 for eclipse preferences for code formatting. I kept getting lint errors (and still do). I imported this xml files on the assumption that it would apply UMS coding standards for Eclipse. That does not seem to be the case, as you have flagged blank lines in imports a couple of times now... I can see it makes a lot of references to lines after imports etc. See these for example:

<setting id="org.eclipse.jdt.core.formatter.blank_lines_after_imports" value="1"/>
...
<setting id="org.eclipse.jdt.core.formatter.blank_lines_before_imports" value="1"/>
<setting id="org.eclipse.jdt.core.formatter.blank_lines_after_package" value="1"/>

Every time I open/save a file in Eclipse it applies the rules defined in this file. 

I have no idea of what the current rules should be. Someone in the know should review these preferences and update them or delete this file. Should I be using it at all?

threedguru:
@SubJunk this is ready to go now.

SubJunk:
Oh nice find! I didn't get a headset yet, I don't have a lot of spare money but would love one! I just upgraded my 10 year old TV though which I'm really happy about

SubJunk (approved):

PR #3765, opened 2023-01-09
Fixed unshared content not being cleaned up after scan

SubJunk:


src/main/java/net/pms/configuration/sharedcontent/SharedContentConfiguration.java
```diff
@@ -147,6 +150,13 @@ public static void updateSharedContent(SharedContentArray values, boolean save)
 					PMS.get().resetRenderersRoot();
 				}
 				updated = true;
+
+				if (wasFolderUpdate) {
+					// Rescan to add/remove Media Library content
+					if (CONFIGURATION.getUseCache() && !LibraryScanner.isScanLibraryRunning()) {
+						LibraryScanner.scanLibrary();
```
ik666, line 157: Shouldn't the current running thread be stopped, and a new library scan started ?

SubJunk: Good point, that's changed now

ik666 (approved):

PR #3770, opened 2023-01-11
Support for Renderer Setting of KeepAspectRatio and KeepAspectRatioTranscoding in 2D to 3D Videos

threedguru:
BigScreen, to take one example, is a popular and free VR media player, supporting headsets from a number of vendors. It has a number of different virtual environments in which you can watch videos, from grand cinemas, to private appartments. In each of these environments, the screen varies in size - but it is always fixed at an aspect ratio of 16:9. 

UMS does support the renderer tags of KeepAspectRatio and KeetAspectRatioTranscoding, which force an aspect ratio of 16:9. However, UMS currently ignores these tags for videos that are in 3D format. UMS does not currently treat 2D to 3D converted videos as "3D" as it references the source media info, but also does not behave correctly in the context of the range of potential 3D output formats in terms of forcing the 16:9 aspect ratio.

This pull request detects the KeepAspectRatio and KeepAspectRatioTranscoding renderer tags in the context of a 2D to 3D transformation, checks whether the original media is in 16:9 format and if not uses an AviSynth script (CropResize) to pre-process the video into 16:9 format BEFORE the 2D to 3D conversion. This therefore means that the resultant video is in 16:9 format whatever the 3D output format in use. See link for CropResize below:

https://forum.videohelp.com/threads/393752-CropResize-Cropping-resizing-script

You can see it in action in this video - along with UMS chapter support (which I just discovered - Eureka moment):

https://streamable.com/qrqcr4


threedguru:
@SubJunk are you about to do a release? Is there any chance that we could get this in?

SubJunk (approved):

SubJunk:
@threedguru good timing :)

PR #3776, opened 2023-01-13
Fix default 2D to 3D frame stretch factor for Web UI for revised 2D to 3D algorithm

threedguru:
The Web UI defaults are maintained in a different place than for the thick client. I missed this when changing the default for the frame stretch factor. The new algorithm for 2D to 3D conversion works better with a default of 8 rather than 5. Please include this small change in the upcoming release.


SubJunk (approved):

SubJunk:
It would be cool to combine the defaults

PR #3780, opened 2023-01-14
Removed maven-idea-plugin and maven-site-plugin

SubJunk:
I don't use these plugins, does anyone else @UniversalMediaServer/developers ?


ik666:
@SubJunk I don't use those plugins.

PR #3801, opened 2023-01-29
Fixed exception when folders are populated on startup

SubJunk:


SurfaceS:
I would change the `LibraryScanner.scanLibrary()` directly instead of `SharedContentConfiguration`.

```
			if (PMS.get().getRootFolder(null) != null) {
				long start = System.currentTimeMillis();
				PMS.get().getRootFolder(null).startScan();
				LOGGER.info("Library scan completed in {} seconds", ((System.currentTimeMillis() - start) / 1000));
				analyzeDb();
			}
```

SurfaceS (approved):

PR #3822, opened 2023-02-09
UMS-Fixes

LMSDS:
2 small fixes


SubJunk:
Thanks for these changes, they look good. Is there a specific bug you encountered that `LC_ALL` fixed?

SubJunk:
@LMSDS ?

LMSDS:
Sorry for the late Reply :-)
The LC_ALL fix the VLC Version parsing on Systems, which use other Language settings (in my case de_AT.UTF-8):
VLC media player 3.0.18 Vetinari (revision 3.0.13-8-g41878ff4f2)
VLC-Version 3.0.18 Vetinari (3.0.13-8-g41878ff4f2)

With LC_ALL:
VLC media player 3.0.18 Vetinari (revision 3.0.13-8-g41878ff4f2)
VLC version 3.0.18 Vetinari (3.0.13-8-g41878ff4f2)

Because the parsing awaits English responses, but UMS also works with other local language settings, its better to declare for external processes, what we want. In feature it's possible, that other external encoders also start to i18n responses.

On the other side, it's also possible to fix the VLC version parsing alone here:
`https://github.com/UniversalMediaServer/UniversalMediaServer/blob/b00b8a4983d3db55a941fa0ae2dd818a8fd8455d/src/main/java/net/pms/encoders/VLCVideo.java#L676`
from "VLC version\\s+[^\\(]*\\(([^\\)]*)" to "VLC media player\\s+[^\\(]*\\(([^\\)]*)"

PR #3886, opened 2023-04-11
feat: delay freeing up sleep mode in Windows 11

### LINKED ISSUE #3883: Sleep prevention is too soft / OS too aggressive
So the UMS sleep prevention is set to "During playback", but I'd like a "During playback and reasonable time after" option.

Why?

- Right now, sleep is really only prevented during playback, so for instance, when your SO comes in, you hit pause, resume afterwards -> computer asleep, day ruined.
- Or when your current play ends, you browse other titles, you hit play when you found something -> computer asleep, day ruined.

Is it possible to not only prevent sleep during playback, but rather reset the OS' sleep counter?

SubJunk:
Fixes #3883

### DIFF
--- a/src/main/java/net/pms/platform/windows/WindowsUtils.java
+++ b/src/main/java/net/pms/platform/windows/WindowsUtils.java
@@ -666,4 +666,12 @@ public class WindowsUtils extends PlatformUtils {
 		return true;
 	}
 
+	/**
+	 * A readable way to see if this is Windows 11 or greater,
+	 * becuase Windows 11 identifies itself as 10.0.0, FFS...
+	 *
+	 */
+	public static boolean isWindows11OrGreater() {
+		return WindowsUtils.getOSVersion().isGreaterThanOrEqualTo("10.0.0");
+	}
 }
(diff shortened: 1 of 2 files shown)


js-kyle (approved):

src/main/java/net/pms/platform/windows/WindowsUtils.java
```diff
@@ -666,4 +666,12 @@ public static boolean uninstallWin32Service() {
 		return true;
 	}
 
+	/**
+	 * A readable way to see if this is Windows 11 or greater,
+	 * becuase Windows 11 identifies itself as 10.0.0, FFS...
```
Nadahar, line 671: @SubJunk I really don't understand what this JavaDoc means - I mean "readable"? It seems to me like it still applies to Windows 10 as well, so shouldn't it at least say something about needing "improvement" to actually tell if it's Win11? I mean, there must be some way to tell using some Windows API.

SubJunk: Good point, I'll use a name-based method instead

SubJunk: There is another problem with this code - it seems that Windows is now never allowed to sleep in Windows 11, so the `doAllowSleep()` with the 5 minute delay isn't working. I can see that the code does run after 5 minutes in the logs.
Maybe it's because of the `run()` loop in `AbstractSleepWorker` conflicting with the timing, any ideas?

PR #3943, opened 2023-06-14
Update Free-Freebox.conf

ouaklafoud:
without this, no x265 works , will solve so many problems for french people using freebox


SubJunk:
Thanks for this. It would be good to specify support more specifically, for example like this file https://github.com/UniversalMediaServer/UniversalMediaServer/blob/main/src/main/external-resources/renderers/LG-OLED-2020%2B.conf#L70-L77
But I will merge this because it is an improvement for now

ouaklafoud:
thank you very very much and i will.
Just to let you know it took me very very long time to figure this out, and there was no workaround known on the web ( browsed around 50 pags for this problem with no solution), this makes it all works :)

SubJunk:
@ouaklafoud thank you for your effort :) I will put your name in the changelog for credit

PR #3955, opened 2023-06-24
Main fix threads and minor platform utils

SurfaceS:
ThreadGroup is not demonized by default, and daemon option will be removed.
Created SimpleThreadFactory that allow uniformed use of ThreadFactory

ProcessUtil :
getProcessId from process itself (since java 9)
move specific platform code under it's own platform utils.
move process utility code under ProcessUtil.


SubJunk:
@SurfaceS welcome back!
the code looks cleaner this way so I am happy to merge. Are there any bugs that are fixed by it? I am wondering for the changelog

SubJunk (approved):

SurfaceS:
@SubJunk Not a lot of free time ;)

> Are there any bugs that are fixed by it? I am wondering for the changelog

Normally no, but it may work with the process id for windows / macos now.
The background thread for metadata is now properly named (was "opensubtitle ....")

SubJunk:
@SurfaceS I understand about free time :) cool, I will merge

PR #3960, opened 2023-06-27
fix player media library

### LINKED ISSUE #3948: "Show Media library" not working correctly in web settings
Just saw a [post](https://www.universalmediaserver.com/forum/viewtopic.php?t=15773) on the forum that if they turn off the media library then it will also stop metadata.

I have checked on the 13.4.1 snapshot build and I can't see the option to show the media library. Instead the Resume folder was marked as being on even though I have that off. Hovering over the Resume option brings up the Media Library tooltip instead (the other options don't show tooltips too)

Clipboard Image

Toggling the Resume option changes `show_media_library_folder` in UMS.conf

I think this is related to #3774

SurfaceS:
fix #3948

### DIFF
--- a/react-client/src/components/Player/Player.tsx
+++ b/react-client/src/components/Player/Player.tsx
@@ -223,7 +223,7 @@ export const Player = () => {
                   theme.colorScheme === 'dark' ? theme.colors.dark[9] : theme.colors.gray[0],
                 },
               })}
-              onClick={() => sse.askBrowseId(media.id)}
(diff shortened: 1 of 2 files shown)


SubJunk:
I already fixed that here https://github.com/UniversalMediaServer/UniversalMediaServer/pull/3952

SurfaceS:
@SubJunk not seen your PR #3952

Quite the same, but allow to see the actors and others metadata.

SubJunk (approved):

PR #3968, opened 2023-07-01
WIP Access control

SubJunk:


SurfaceS:
Since Update changelog

[f4ee979](https://github.com/UniversalMediaServer/UniversalMediaServer/pull/3968/commits/f4ee97930f6877630714fee660d1d37e57a3ee23)
- getDefaultFolders() is now saved on platform utils

[c49f98d](https://github.com/UniversalMediaServer/UniversalMediaServer/pull/3968/commits/c49f98dd59f8c22aee8c7d52c5e54ad9cb2de705)
- Uniformisation of tables and SQL code
- Use of PreparedStatement (quicker)
- Creating new record from ResultSet.moveToInsertRow
- Changed SQL type for COL_NRAUDIOCHANNELS to INTEGER (was BigDecimal ??)
- Uniformisation of indexes names TABLE_NAME + "_" + COLUMN (+ "_" + COLUMN) + "_IDX"
- Moved the media status (playback count, ...) under it's own class.

[4b655d2](https://github.com/UniversalMediaServer/UniversalMediaServer/pull/3968/commits/4b655d22c6f1cff30b6704979fbeeb8d6ce05430)
- Move media info away from dlna package.
DLNAMediaInfo -> net.pms.media.MediaInfo
DLNAMediaLang -> net.pms.media.MediaLang
MediaType ->  net.pms.media.MediaType
DLNAMediaAudio -> net.pms.media.audio.MediaAudio
DLNAMediaChapter -> net.pms.media.chapter.MediaChapter
DLNAMediaSubtitle -> net.pms.media.subtitle.MediaSubtitle

[439a847](https://github.com/UniversalMediaServer/UniversalMediaServer/pull/3968/commits/439a8470a7e37d3cc3b212f7d3ee61d1bca3d112)
- created FailSafeProcessWrapper to avoid use of multiple thread
- created FFmpegParser
- fix videoWithinH264LevelLimits result is set only for the first renderer asking

SurfaceS:
@SubJunk 
I think this should be merged to main before starting the users changes (shares, media status, renderer link, etc).

SubJunk:
I can't get it to compile on Windows, I get this error:

```
[INFO] [MAKENSIS] Error: invalid VIProductVersion format, should be X.X.X.X
[INFO] [MAKENSIS] Error - aborting creation process
```

src/main/java/net/pms/database/MediaTableFiles.java
```diff
@@ -1184,4 +1097,31 @@ public static DLNAThumbnail getThumbnailByTitle(final Connection connection, fin
 
 		return null;
 	}
+
+	//TODO : review this
```
SubJunk, line 1101: Does this still need to be done?

SurfaceS: yes, it will need a complete review on it's usage on a next PR.

SurfaceS:
> I can't get it to compile on Windows, I get this error:
> 
> ```
> [INFO] [MAKENSIS] Error: invalid VIProductVersion format, should be X.X.X.X
> [INFO] [MAKENSIS] Error - aborting creation process
> ```

you made a mistake on versioning in POM :
`<version>13.6.0</version>` -> `<version>13.6.0-SNAPSHOT</version>`
`<project.version.short>13.6.0-SNAPSHOT</project.version.short>` -> `<project.version.short>13.6.0</project.version.short>`

SubJunk:
Oh you're right, my mistake! I'll test this branch now

SubJunk:
It's so cool. I like your UI implementation

SubJunk:
@SurfaceS it looks like this has been causing problems for users. pipin reported problems and had to revert to 13.6.0 here https://www.universalmediaserver.com/forum/viewtopic.php?p=47982#p47982 and icenlord gave some possible hints here https://www.universalmediaserver.com/forum/viewtopic.php?p=48086#p48086

pipin has also reported that v14 beta does not resolve the problems https://www.universalmediaserver.com/forum/viewtopic.php?p=48181#p48181

SubJunk:
Ah I see. When I reverted the access control changes from `main`, I only reverted one of the PRs:

In https://github.com/UniversalMediaServer/UniversalMediaServer/pull/4036 I rolled back https://github.com/UniversalMediaServer/UniversalMediaServer/pull/4018 but not this PR. So the changes snuck into main.

SurfaceS:
> I did but v14 shows a bit incosistent behaviour.
> UMS is shown as a media server on my devices, but it takes a few restarts to show the shared content

I think that the V14 cache was not yet ready (V14 rescan everything).

> Edit: In that last post, he post be describing the problem I have been seeing, where folders show up empty the first time I browse. Maybe we should revert these features from main to keep v13 clean, and save them for v14, what do you think?

I don't see where the empty folder can be linked to this feature.
If the renderer was blocked, you will receive nothing from the server (kill connection).
And a refresh will do the same result.

> pipin has also reported that v14 beta does not resolve the problems https://www.universalmediaserver.com/forum/viewtopic.php?p=48181#p48181

In V14, I've just added a menu item that reset all filter settings.
It clean everything and set allowing by default for renderer filter and network filter.

image

I also set a guard that the option "Block Devices by Default" for network device is only accessible from localhost and an information (as the client will not be able to reach the server anymore if not localhost).

Also, now the Network Devices tab is refreshed after a setting change.
(that may be the cause of some trouble as the user was not having the real time info).

SurfaceS:
> I thought it might be related to the changes with IDs. For me, when I get the bug, the folder is empty at first, then if I go back into it 2 seconds later, it works. That is with v14 that I have been running for a long time. It seems to happen when I first put the TV on, I thought it might be the TV cache having a different ID for those folders.

Yes, only V14 is affected, that's why I said "I don't see where the empty folder can be linked to this feature." (filters).

> I know you made the IDs persistent for files, are they persistent for containers too?
Yes, containers ids are persistent as well.

SubJunk:
> Yes, only V14 is affected, that's why I said "I don't see where the empty folder can be linked to this feature." (filters).

Yeah I agree, not related to this feature. It came up on the forum because the user tried v14-b1 to see if that fixed their problems.

PR #3971, opened 2023-07-06
Access control

SurfaceS:
@SubJunk  I was thinking something like that.


SubJunk:
@SurfaceS great! I will review and test it within a few hours, I am doing some Dolby Vision things first

PR #3978, opened 2023-07-07
Prevent sample files showing up in the TV shows area of Media Library

SubJunk:


SurfaceS:
I think this one should not be merged now.

My next PR will remove video track info from `FILES` table, and move it to `VIDEOTRACKS` table.

Also, filtering on a "greater than double value" is maybe not the best way to do.

Sample can be a flag on `VIDEO_METADATA` table, like the `ISTVEPISODE` one (`ISSAMPLE`, or `ISTRAILER`).

SubJunk:
@SurfaceS cool, I agree. It is potentially still a good index to add here https://github.com/UniversalMediaServer/UniversalMediaServer/pull/3978/files#diff-552ef7c3f32721c1801d78dd004a1f15a637b21e22215c79192b3b8ed552bffeR467 because we already filter by duration for movies. Or we could add an `isMovie` boolean and start using that

SubJunk:
Superceded by https://github.com/UniversalMediaServer/UniversalMediaServer/pull/5377

PR #3990, opened 2023-07-11
Extract subs for ffmpeg 2

SubJunk:


SubJunk:
thanks @valib for the help
I am resurrecting this because I need it for HDR functionality, because we can't hardcode subtitles to some HDR streams

PR #4018, opened 2023-08-03
Access control

SurfaceS:
`MediaInfo` is only a container spec info like it should.
Video spec was moved to `MediaVideo`, allowing multiple video track on a container.
`MediaInfo` have now the `MediaParser` member that is stored on the db.
It add the ability to reparse a media by clearing this columns.
Modified the Table `FILES` according to this.
Created the Table `VIDEOTRACKS` for `MediaVideo`.
Stop storing video spec numbers into string.
Refactor `net.pms.media.metadata.*` to `net.pms.media.metadata.video.metadata.*`
Renamed some functions and members to reflect what they really are.

`MediaAudio` is now only an audio track spec info like it should.
`MediaAudioMetadata` now have all the audio Metadata like `MediaVideoMetadata` does (nothing to do with the audio track specs).
Modified the Table `AUDIOTRACKS` according to this.
Modified the Table `AUDIOTRACKS` colums names to reflect more what's on it.
Created the Table `AUDIO_METADATA` for `MediaAudioMetadata`.
Removed `LIKESONG` columns and functions as it was just set and never read.
I think that was superseed by staring system and never deleted.

`Engines` -> Stop storing the temp renderer config to the engine class static as it could interfer on multiple renderer transcoding.

`Parsers` -> Moved most of the parser under the `Parsers` package.
Merge the (same/duplicate) logic to there.

`MediaInfoParser` -> created full utility streams class that contain every possible request with comment.

Test was moved on their own package.
Removed the need to create PMS instance when checking Parsers
Added test for audio with tags.
Added test for images file.
Added direct parser tests (that show a lot of false/bugged parse data).

And maybe some other things I forgot.

Also, `firstAudioTrack` (and video as well) was renamed to `defaultAudioTrack`, as I found some default audio track not been the first audio stream on some files. Now MediaInfo and ffmpeg parser look for default and forced.


SurfaceS:
**About the Show and hide folders based on renderers**

This PR was intended to split things and have a better view on what we need to change (and fix some issues as well).

Showing/allowing access to resources based on user group is the easy part.

The hard part, as it come from my head :
`MediaInfo` (with tracks) should be common to all (immutable) -> a change on it will be on every DLNAResource (object by ref).
`MediaAudioMetadata` and `MediaVideoMetadata` as well. It's not user or renderer specific.

Things like "Staring" should be separated : metadata staring, and user staring (that why I was not agree with staring directly on/modify audio files, that should contain the metadata staring).

`MediaStatus` should be user aware, then a user can track his own fully played, played count, etc.
That mean, `Thumbnails` should not be common to all, but only to a user (reflecting fully played, position, etc).
`MediaInfo` thumbnail should be the unaltered thumbnail or maybe deleted.

I also see that if multiple renderer using the same user, the Root may not be the same according to renderer settings (split folders, abilities, etc).
That's why I think each renderer should have it's own DLNAResource object, sharing some common objects (like `MediaInfo` for all, and `MediaStatus` based on the user).

What if setting is delete/move the file on fullyplayed if many users ?
Maybe on renderer other than WebRenderer a user choice on root folder ?? with a 4 digit code as an option ?

src/main/java/net/pms/encoders/FFMpegVideo.java
```diff
@@ -1048,7 +1040,7 @@ public synchronized ProcessWrapper launchTranscode(
 			cmdList.add("0:V");
 
 			cmdList.add("-map");
-			cmdList.add("0:a:" + (media.getAudioTracksList().indexOf(params.getAid())));
+			cmdList.add("0:a:" + (media.getAudioTracks().indexOf(params.getAid())));
```
SurfaceS, line 1043: this may use params.getAid().getId()

SurfaceS:
just fixed a long time bug that sometime prevent webplayer to handle correctly the stream.
removed the 1s delay before transcode for it.

for tracks, UMS was never delete tracks that are not more in the file as it just add or update.

added optional id for things like OGG id, etc.

SurfaceS:
From what I understand :

Each renderer have it's own RootFolder (a DLNAResource).
The RootFolder is filled with childs (DLNAResource).

DLNAResource is an object that can be shared with all renderer.
The first renderer that access the DLNAResource will scan for child, with it's conf.
Then any renderer can update some field, and use the childs created for the previous renderer.
Also, if no thumb, it fill the MediaInfo thumb according to it's conf, even if other renderer doesn't have the same conf.
At end, a renderer is using DLNAResource for another one.

My solution :
I really think that DLNAResource should be linked to a renderer MediaStore, and never be shared between renderers.

GlobalIdRepo should be deleted.

Replaced by MediaInfoRepo -> immutable (common to all) MediaInfos, Tracks, Metadata.
Replaced by MediaStatusRepo -> common to user MediaStatus.
And things like Thumbs directly on the DLNAResource.

Childs may be automatically updated with Listeners.

@SubJunk  What do you think of this ?

SubJunk:
@SurfaceS that seems very logical to me. I think it's a good idea. I'll try to provide some context for historical decisions below, because they might help.

In the past we have run into memory problems, so that is important to consider when implementing solutions that increase the amount of things in memory like that. Some renderers crawl all folders in the background (LG TVs for example) so any solution should imagine that kind of situation. Those memory problems were likely caused by bugs too. Lately things have been very stable IMO, I haven't noticed any memory or stability problems for at least a year.

At the moment the Global ID Repo has a good feature: it lets us discover and reuse objects in memory instead of always creating new ones, but it also lets Java tell us to forget them if it is running out of memory. That made a huge improvement to stability for power users with tens of terabytes of files.

I think what you described is worth trying though because only a subset of the file data will be duplicated for different renderers. Hopefully that will be efficient enough to not fill up the memory.

SurfaceS:
@SubJunk Ok, i'll start that on Monday.

I think what is filling a lot memory is thumbs.

For now, this PR is ready to review, and fix a lot of background things (missed info, wrong track, webplayer hang).
The "bad" side is that it will reparse all files once.

SubJunk:
@SurfaceS do you want to rebase this onto `main`?

SurfaceS:
Merged main / fix conflicts

Java :
update dependency h2 to v2.2.220
migrateDatabase to version 3

React client :
update dependency @ mantine to v6.0.18
update dependency @types/node to v18.17.3
update dependency @types/react to v18.2.19
update dependency @typescript-eslint/eslint-plugin to v6.3.0
update dependency @typescript-eslint/parser to v6.3.0
update dependency eslint-plugin-react to v7.33.1

src/main/java/net/pms/dlna/RealFile.java
```diff
@@ -244,18 +250,23 @@ public synchronized void resolve() {
...
 						try {
-							MediaInfo media = MediaTableFiles.getData(connection, fileName, file.lastModified());
+							MediaInfo media = MediaTableFiles.getMediaInfo(connection, fileName, file.lastModified());
 
 							setExternalSubtitlesParsed();
 							if (media != null) {
 								setMedia(media);
 								if (configuration.isDisableSubtitles() && getMedia().isVideo()) {
 									// clean subtitles obtained from the database when they are disabled but keep them in the database for the future use
+									// FIXME this hack should not be done
```
SubJunk, line 260: Is there a plan to fix this?

SurfaceS: Yes.
This should not be done as a renderer can override the "disable subs", and then all renderer will be without subs.

SubJunk (dismissed):

SubJunk:
The code looks good to me, I will do some testing before merging

SubJunk:
I installed it on Windows 11 and the program won't launch. It's not updating the debug.log either. Very weird but maybe the compilation went wrong (NSIS bug? we are on an old version and I noticed they did Windows 11 fixes in v3...) so I'll try recompiling

SubJunk:
I don't know if this is an existing bug, but the web player is not loading at all during the initial scan. We should check if that's the same for the latest release. The last time I tested it, it was working, because that's how I tested the REALTIME_LOCK functionality

SubJunk:
It might depend on the port being used.
If I click "Web player" in the system tray right-click menu, it takes me to http://192.168.1.128:9002/ which does not load. But if I click "Settings" in that menu and then go to Player in the burger menu, it takes me to http://localhost:9001/ which works

react-client/yarn.lock
```diff
@@ -3467,9 +3442,9 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@types/react@npm:*":
-  version: 18.2.17
-  resolution: "@types/react@npm:18.2.17"
+"@types/react@npm:*, @types/react@npm:18.2.19":
```
SubJunk, line 3445:
```suggestion
"@types/react@npm:*":
```

SubJunk (approved):

PR #4028, opened 2023-08-12
Access Control - part 2

SurfaceS:
**Done for now :**
fix scan button state
fix FileWatcher error logged on shutdown
fix sse with less blocker sync as log can block a lot

moved all virtual to virtual dlna.virtual
moved some non dlna to correct package
rebuild the library parser
moved chapter hls to hlshelper
renderers now have their own rootfolder
renderers now have their own globalids

implement database cache memory usage show
implement MediaResource clear children
implement GlobalIdRepo clear
implement Renderer resetRootFolder -> clear all children
delete rootfolder, etc. on renderer deletion

implement group authorization for shares
implement user pin code
implement user avatar

implement user filestatus
now save the default mimetype
refactored library things from net.pms.dlna to net.pms.library
refactored DLNAResource -> LibraryResource
added DidlHelper from old DLNAResource
added DlnaHelper from old DLNAResource

organize imports
code cleaning

added copy files status on each users for filling user status
some user want to keep it, others can reset it on the shared folder

fix AudioTrackCount
webgui : split home renderer/device as device ip hostname take long time

implement RendererUser
implement ContentDirectoryError

**This PR is awaiting testing and review**

Users can now choose to link a renderer to a user on the renderer gui.
That should enjoy users that want to hide some things for other (like for kids).
Also, as file status is now based on user, other viewers will not disturb user file status.

I forgot the user choice on upnp for now as it require too much to do.
Sad as it was a good idea on my head.


SubJunk:
I forgot to say I am reviewing and testing this. There is a lot here and it looks great so far

SubJunk:
On the GUI I can authorize Groups for Shared Content:
image

but then in the Renderer menu I can authorize Accounts:
image

So they don't match up. Maybe we should be able to do both Accounts and Groups on both areas?

SubJunk:
I also somehow got the server into a state where every folder shows up empty on the TV, and the server is responding to browse requests with:

```
================================== HTTPSERVER RESPONSE BEGIN ====================================
HTTP/1.1 701 The specified ObjectID is invalid

HEADER:
  Content-Length: 0
```

but there are no user or group restrictions set

SurfaceS:
> On the GUI I can authorize Groups for Shared Content:
but then in the Renderer menu I can authorize Accounts:

It is the intended use.
Shared content is restricted on group based.
Renderer is assigned to an account (user)

Then, for example, my wife and me are on the same group (Parents), so we have the same shared folders.
But my renderer is set to my account.
I then have my own file status.

SurfaceS:
> I also somehow got the server into a state where every folder shows up empty on the TV, and the server is responding to browse requests with:

That is the normal behavior when an object is not founded (added on this PR).
Before, the renderer was waiting forever if you reply with a 200 code and nothing on the content.

That can happen if you where browsing home->films->somefolder, and you did something that reset the renderer root folder.
(change user, rights, etc)
Going back to the root is the option.

I'll work on a better "root folder rebuild on change" later.

SubJunk:
@SurfaceS it seems that folders are not displaying new content on my TV in this branch. I haven't looked into it but it feels like the SystemUpdateId isn't as sensitive as it was before

SubJunk:
@SurfaceS these code conflicts should go away if you rebase it on the `v14` branch

SurfaceS:
I'll have a look today

SurfaceS:
moved to V14

PR #4038, opened 2023-08-21
StarRating service fix / SQL exception

ik666:
This code update was missed while refactoring the audio table ...

- Moved SQL code vom rating service to MediaTable class.
- fixed SQL to reflect new table structure


SubJunk:
Will leave this for @SurfaceS to review

SubJunk:
It might be better to base work on `SurfaceS:ac-root-renderer` instead of `v14` since that is the most recent code that will merge into v14 soon

SubJunk:
This was added to v14 branch

PR #4107, opened 2023-09-30
V14 - Localize video metadata

SurfaceS:
Save thumbnail source.
Localize video metadata for server language (non web gui).
Localize default video thumbnails for server language (non web gui). Store thumbnail source.
fix duplicate background api request.

It will check for thumbnail loc on first mediainfo request (may delay get).
As v14 reset all mediaInfo, it should not be a problem.


SubJunk:
Is this waiting for review or testing?

SurfaceS:
I was testing it before merge.

PR #4124, opened 2023-10-09
Bug Fixed - 3669

Priyanka-GD:
Live Subtitles folder shows up in "Recently Added/Played" lists on web 

- Removed Live Subtitle folder from the display on left panel

Screenshot 2023-10-09 at 4 56 02 PM


SubJunk:
@Priyanka-GD thank you for your contribution! I don't think this is the right fix though.
To reproduce the bug, enable the `Show the "Live Subtitles" folder` setting, and observe that the folder appears on the _first page` of the web player

Priyanka-GD:
@SubJunk I tried to reproduce the bug. I checked "Show Live subtitle" folder 
image

But the folder didn't appear on the first page of the web player. Could you please elaborate on how to reproduce this bug? I want to fix it.

SubJunk:
@Priyanka-GD when I enable that setting, here is my web player:
image

PR #4129, opened 2023-10-13
fix: an incorrect property is used in migrating `ip_filter`

dytlzl:
## Description
Fix the incorrect property that is used in the process of migrating the old configuration `ip_filter`.

## Motivation
While the original code sets a property called `KEY_BLOCK_RENDERERS_BY_DEFAULT` to `true`, considering that `KEY_NETWORK_DEVICES_FILTER` corresponds with `KEY_BLOCK_NETWORK_DEVICES_BY_DEFAULT`, the configuration that should be enabled is `KEY_BLOCK_NETWORK_DEVICES_BY_DEFAULT`, not `KEY_BLOCK_RENDERERS_BY_DEFAULT`.


SubJunk (approved):

SubJunk:
@dytlzl nice find, thanks

PR #4133, opened 2023-10-13
Added more MP4 ORG_PN values

SubJunk:
This should produce values according to the June 2016 DLNA spec as detailed in `DLNA Guidelines June 2016 - Part 2 Media Format Profiles.pdf`

image


SubJunk:
This is in v14 now which is better than doing another db migration in main

PR #4140, opened 2023-10-16
Fix for - Checking whether correct java version is installed in the system that is compatible with maven version #4150 + #3007

### LINKED ISSUE #3007: Java 18
How do we know when we should use Java 18? Should we make a list of high risk dependencies to check compatibility with?
In Node.js we have package.json that lists version compatibility, is there a similar thing for Java?

Priyanka-GD:
Check whether the correct Java version is installed in the system and compatible with pom.xml version.
Have also coded JUnit test cases for the method

image

image


src/main/java/net/pms/PMS.java
```diff
@@ -880,60 +875,52 @@ public static void main(String[] args) {
 		boolean displayProfileChooser = false;
 		boolean denyHeadless = false;
 		File profilePath = null;
-
+		checkCompatibilityWithJavaVersion();
```
SubJunk, line 878: I think this line and the method it calls are the only changes in this file, can you revert the other changes?

src/test/java/net/pms/util/FileUtilTest.java
```diff
@@ -589,7 +589,12 @@ public void testGetFilePermissions() throws FileNotFoundException {
 		assertTrue(FileUtil.getFilePermissions(new File("")).isFolder(), "CurrentFolderIsFolder");
 		assertTrue(FileUtil.getFilePermissions(new File("")).isReadable(), "CurrentFolderIsReadable");
 		assertTrue(FileUtil.getFilePermissions(new File("")).isBrowsable(), "CurrentFolderIsBrowsable");
-		assertTrue(FileUtil.getFilePermissions(new File(System.getProperty("user.dir"))).isFolder(), "user.dirFolderIsFolder");
+		try {
+			assertTrue(FileUtil.getFilePermissions(new File(System.getProperty("user.dir"))).isFolder(), "user.dirFolderIsFolder");
+		}
+		catch(NullPointerException ex){
+			//Exception
```
SubJunk, line 596: Why is this change here?

SubJunk (changes requested):

src/test/java/net/pms/util/FileUtilTest.java
Priyanka-GD: Reverted other changes.
For FileUtilTest.java - in the build pipeline, it was throwing NullPointerException, so I kept it inside the try/catch block 

IMG_2863

SubJunk:
@Priyanka-GD it looks like an automatic formatter ran on this branch, can you remove those changes please?

SurfaceS:
I don't think that's a good way to check the java version from maven.

1 : End-user doesn't have the pom file, so it will fail on regular use.
2 : UMS have a `project.properties` that is filled by Maven, so it can have the compiler java version from here.
3 : It add dependencies on the final end-user jar. Tests and test dependencies should sit and stay on the test package (`<scope>test</scope>`).
4 : `System.clearProperty("user.dir")` break others test code that rely on it.

The #3007 is not about checking the current version. It is about having a tool that check if we can use previous/next java version without dependency/code problem.

Priyanka-GD:
> I don't think that's a good way to check the java version from maven.
> 
> 1 : End-user doesn't have the pom file, so it will fail on regular use. 2 : UMS have a `project.properties` that is filled by Maven, so it can have the compiler java version from here. 3 : It add dependencies on the final end-user jar. Tests and test dependencies should sit and stay on the test package (`<scope>test</scope>`). 4 : `System.clearProperty("user.dir")` break others test code that rely on it.
> 
> The #3007 is not about checking the current version. It is about having a tool that check if we can use previous/next java version without dependency/code problem.

Okay! Got it. Will try to look into it.

Issue #3052, opened 2022-06-13
Media metadata does not show in logs if exists in DB


Mik-S-UMS:
Something I miss from the logs that is helpful to track down problem files for users is seeing the media info for that file in the logs.
When a new file is scanned it's mediainfo is shown in the logs but if a file has already been scanned and it's metadata is stored in the database it does not log them so I have no idea why some videos is not playing on the renderer.

Could it be possible to show that in trace logs when the metadata already exists?

SubJunk:
Yeah that would be possible to do it when it's read from the database, as long as it's in TRACE mode so that we don't bloat the logs for the average user

SubJunk:
Done thanks to @Priyanka-GD

PR #4145, opened 2023-10-17
Enhancement #3052 Media metadata does not show in logs if exists in DB

Priyanka-GD:


src/main/java/net/pms/dlna/DLNAResource.java
```diff
@@ -727,19 +727,20 @@ public void addChild(DLNAResource child, boolean isNew, boolean isAddGlobally) {
...
 					addChildInternal(child, isAddGlobally);
+				} else {
+					LOGGER.trace("Details on media being imported :" + child);
 				}
-
 				if (resumeRes != null) {
 					resumeRes.setDefaultRenderer(child.getDefaultRenderer());
 					addChildInternal(resumeRes);
+				} else {
+					LOGGER.trace("DLNAResource is null :" + resumeRes);
```
SubJunk, line 739:
```suggestion
```

SubJunk (changes requested):
We don't need to log `resumeRes` which will always be `null` here

SubJunk (approved):

PR #4171, opened 2023-10-26
add limit to item path to avoid OOM

SurfaceS:


ik666:
Just in case of relevance: I was live profiling UMS, and I as far as I remember, the `UPDATE_IDS` map was filled with many mediaStoreId's instances having exact the same value. I think (not sure) the parentId was -1. Maybe the while loop should check for `>0` instead of `!= 0`.

PR #4185, opened 2023-11-02
fixed RequestHandlerV2 to pipe api requests.

ik666:
This PR handles API requests to URI `/api` again.


SurfaceS:
Your code bypass everything (renderer, log, permissions, etc).

If it doesn't serve the api, it's because the request doesn't find the renderer on this line :
`renderer = ConnectedRenderers.getRenderer(ia, headers.get(HttpHeaders.Names.USER_AGENT), headers.entries());`

Is this the normal behavior  ?

ik666:
@SurfaceS 
This API is meant to be used by other programs and not by renderer ... This is the intended behaviour.

PR #4188, opened 2023-11-03
locate correct resource for given DBID id

ik666:
This PR fixes discovery of resources in case objectID is DBID format encoded.


SurfaceS:
This should not be fixed in http server. (or if you fix something on one, you should replicate it on the others).

I fixed this on the last PR, under the mediaStore itself.

PR #4209, opened 2023-11-04
V14 check java version

SurfaceS:
example code for #4140


SubJunk:
The author of #4140 seems to have gone away

PR #4037, opened 2023-08-21
v14

SubJunk:
Feature branch for v14

https://github.com/UniversalMediaServer/UniversalMediaServer/issues/3969

**Done for now :**
fix scan button state
fix FileWatcher error logged on shutdown
fix sse with less blocker sync as log can block a lot
fix AudioTrackCount
fix multi renderer creation on webguiplayer
fix RELEASEDATE as timestamp
fix typo in column name

rebuild the store parser
moved chapter hls to hlshelper
renderers now have their own MediaStore
renderers now have their own globalids

implement database cache memory usage show
implement StoreResource clear children
implement GlobalIdRepo clear
implement Renderer resetRootFolder -> clear all children
delete rootfolder, etc. on renderer deletion

implement group authorization for shares
added iTunes, iPhoto and Aperture to SharedContent
implement user pin code
implement user avatar

implement user filestatus
now save the default mimetype
refactored library things from net.pms.dlna to net.pms.store
refactored DLNAResource -> StoreResource
added DidlHelper from old DLNAResource
added DlnaHelper from old DLNAResource
split store container / store item
merge LibraryScanner and MediaScanner -> StoreScanner

organize imports
code cleaning

added copy files status on each users for filling user status
some user want to keep it, others can reset it on the shared folder

implement RendererUser
removed cache option.
created thumbnail store

update h2 database to v2.224
remove apache httpcore dep
add thread name for debug

**UPnP :**
implement ContentDirectoryError
added systemUpdateId event (was never implemented)
partially added jupnp mediaserver didl results

**React app**
update mantine 6.0.21 -> 7.1.0
update web-vitals 3.4.0 -> 3.5.0
removed unnecessary deps
added source-map-explorer 2.5.3 -> add analyzing the bundle size
split home renderer/device as device ip hostname take long time
add loading on renderer screen
add loading on device screen
add loading on player renderer creation
change shared settings table to box (nicer)
fix breadcrumbs size
fix logs page
fix can't save settings from webgui
added status line on webgui
added flag for movie/tv countries
added original movie/tv series title and country if not same as title
added Rating bar
displayed logged user on the top bar
main screen default to player if not allowed to view settings
always show scroll bar on the user list

remove Apache HttpClient5 deps
fix registerExternalSubtitles for VirtualVideoAction
remove apache httpcore dep
add thread name on speedtest
add thread name for debug
fix multi renderer creation on webguiplayer
fix coverart archive tests
added jupnp DIDL-Lite with classes
[...]

PR #4217, opened 2023-11-06
MediaStoreIds using long

ik666:
@SurfaceS I've again an issue with DBID resources. At the moment the MediaStore concept is not able to process ID's other than long numbers. This makes all storageContainer relying on DBID id's fail to resolve child resources properly. Child resources do not have an "id",  resulting in broken items (thumbnails, etc).

Some examples are:
- browse to the music album "99 red ballons" identifies by musicBrainzID : `$DBID$MUSICBRAINZALBUM$189002e7-3285-4e2e-92a3-7f6c30d407a2`
- browse to a playlist identified by database Table `FILES` with  `ID = 307` : `$DBID$PLAYLIST$307`

I'm not very happy with this solution (PR), because it fixes only the "$DBID$PLAYLIST$..." uri path. Therefore: 

I'm asking myself, if you could just remove the method "getLongId()" for identifying a parent folder and use the current objectId (as String)? This should probably fix the above mentioned issues for all cases.

Or you find a better (general) solution, how to handle this issue?

### DIFF
--- a/src/main/java/net/pms/store/container/PlaylistFolder.java
+++ b/src/main/java/net/pms/store/container/PlaylistFolder.java
@@ -31,6 +31,8 @@ import net.pms.dlna.DLNAThumbnailInputStream;
 import net.pms.formats.Format;
 import net.pms.formats.FormatFactory;
 import net.pms.renderers.Renderer;
+import net.pms.store.DbIdMediaType;
+import net.pms.store.MediaStoreIds;
 import net.pms.store.StoreContainer;
 import net.pms.store.StoreResource;
 import net.pms.store.item.FeedItem;
@@ -52,6 +54,8 @@ public final class PlaylistFolder extends StoreContainer {
 	private final boolean isweb;
 	private final int defaultContent;
 	private boolean valid = true;
+	private static final String DBID_PLAYLIST = DbIdMediaType.GENERAL_PREFIX + DbIdMediaType.TYPE_PLAYLIST.dbidPrefix;
+	private static final int DBID_SPLIT = DBID_PLAYLIST.lastIndexOf("$") + 1;
 
 	public PlaylistFolder(Renderer renderer, String name, String uri, int type) {
 		super(renderer, name, null);
@@ -315,4 +319,30 @@ public final class PlaylistFolder extends StoreContainer {
 		return null;
 	}
 
+	@Override
+	public Long getLongId() {
+		if (getId().startsWith(DBID_PLAYLIST)) {
+			String databaseId = getId().substring(DBID_SPLIT);
+			return Long.valueOf(databaseId);
+		}
+		return super.getLongId();
+	}
+
+	@Override
+	public boolean isRefreshNeeded() {
+		return true;
+	}
+
+	public void doRefreshChildren() {
+		for (StoreResource playlistItem : getChildren()) {
+			try {
+				Long id = MediaStoreIds.getMediaStoreResourceId(playlistItem);
+				if (id != null) {
+					playlistItem.setId(id.toString());
+				}
+			} catch (Exception e) {
+				LOGGER.trace("doRefreshChildren", e);
+			}
+		}
+	}
 }


SurfaceS:
In fact, you are trying to mix direct acceded items (DBID), non UPnP compliant, and MediaStoredId'ed items that follow the UPnP rules.

In UPnP, every item/container must have a parent, and a unique ID.
In your usage, $DBID$MUSICBRAINZALBUM$189002e7-3285-4e2e-92a3-7f6c30d407a2 may have (for example) a parent id of 0 (root) and on the same time another parent ID.
That is not permitted by UPnP.
In UPnP point of view, that is 2 different objects that must have different object ID.
UPnP have a field (`item@refID`), that allow unique ID'ed object to have the id property of the item being referred to (that could be referred by many objects).

That's why for now this does not work.

On your PR, you give the database file id (non UPnP related) as parent ID, so giving 0 will have the same result.

> I'm asking myself, if you could just remove the method "getLongId()" for identifying a parent folder and use the current objectId (as String)? This should probably fix the above mentioned issues for all cases.

It will only hide the broken UPnP things.
UPnP allow a string like $DBID$MUSICBRAINZALBUM$189002e7-3285-4e2e-92a3-7f6c30d407a2 for ID, but it need to be unique.

> find a better (general) solution, how to handle this issue?

yes, that's a better solution.

On witch case do you use something like : $DBID$MUSICBRAINZALBUM$189002e7-3285-4e2e-92a3-7f6c30d407a2 or $DBID$PLAYLIST$307 ?

Because I'm sure you can use the non-UPnP breaking way to achieve the same thing.

ik666:
@SurfaceS Actually, the two above mentioned container are the parent container and I think they don't break the UPnP rules. The only problematic thing is that `MediaTableStoreIds` is storing objects ID's as long value, even tough string values are permitted by UPnP.

The DBIDResourceLocator creates a PlaylistFolder ( = parent) with the object ID `$DBID$PLAYLIST$307`. RealFile (audio) items are being added as child items. The child items cannot use `$DBID$PLAYLIST$307` as a parent ID,  therefore I made this workaround and "invent" a long-id, which is needed because of storing long values in the database:

Please have a look at `MediaTableStoreIds` around line 145 : `long parentId = resource.getParent().getLongId();`

If you'd change the line to `String parentId = resource.getParent().getId();` everything would just work fine I believe. I just wanted to check back, if used the long datatype on purpose for something needed I didn't think of.

The idea behind DBID is the ability to identify UPnP object over time (also after a restart of UMS). They can be used for permanent links (like favorites) , because the target DLNA object doesn't change over time. 

In my use case (`$DBID$PLAYLIST$307`): I can directly "browse" to an audio playlist (.m3u8) that has the ID 307 in the FILES table, without the need to navigate through the "filesystem".

SurfaceS:
> The idea behind DBID is the ability to identify UPnP object over time (also after a restart of UMS). They can be used for permanent links (like favorites) , because the target DLNA object doesn't change over time.

New store keep it.
Ids are stored in db, and the path is recreated if you browse it directly after UMS restart.
It is also reused if you browse the parent.

Old way (changing between restarts) was not UPnP compliant.

> In my use case ($DBID$PLAYLIST$307): I can directly "browse" to an audio playlist (.m3u8) that has the ID 307 in the FILES table, without the need to navigate through the "filesystem".

With storeIds, you will be able to do the same thing with the storeId (that will never change).

I better understand your need.

It need work on it.

ik666:
> In your usage, $DBID$MUSICBRAINZALBUM$189002e7-3285-4e2e-92a3-7f6c30d407a2 may have (for example) a parent id of 0 (root) and on the same time another parent ID.
That is not permitted by UPnP.

`$DBID$MUSICBRAINZALBUM$189002e7-3285-4e2e-92a3-7f6c30d407a2` doesn't have different parentID's (it's always the string "0") and I'd like to stick to the "DBID" logic. Actually all "$DBID$xxx" 1st-level container have a parentID of "0". The only issue is, that the child items (usually a RealFile DLNAResource) inside the DBID container fail with a cast exception inside the MediaStore since it's parentID is obviously not a long value (in line 145).

But I think I see where you come from and think to understand your point. I'm not sure which section you're referencing to: In `UPnP-av-ContentDirectory-v1-Service.pdf`

> 2.5.3. A_ARG_TYPE_ObjectID
This variable is used in conjunction with those actions that include an ObjectID parameter. ObjectID parameters uniquely identify individual objects within the Content Directory Service.

No word about parentId. I don't think the tuple (objectID, parentID) has to be unique, since all UPnP actions (browse, play, etc.) are being done with an unique objectID. 

Just to line up thoughts: I think an ObjectID may have different parents, and I'm not sure , where you see the issue (UPnP breach) ... May I construct a simple example:

Starting with an empty media directory:

1) place one audio file in the root directory. Let's say "audio.mp3".
2) place an "playlist.m3u8" file in the same directory. The playlist should have one entry (line), referencing the audio file: "../audio.mp3"

The audio.mp3 file can now be navigated / referenced by 2 ways:

- Browse to "0" -> identified audio.mp3 item id -> do UPnP actions like play using this id.
- Browse to "0" -> browse into the playlist -> identify audio.mp3 item id -> do UPnP actions like play using this id.

In my scenario: audio.mp3 has 2 different parent-id's but from my understanding may have the same item id lets say "$DBID$FID$2". This can also be used over time (after restart) to identify an object and do UPnP actions on that. I don't see the benefit in generating new ID's?

SurfaceS:
I think you are confusing the ContentDirectory object and the ending media / metadata.

This example from `UPnP-av-ContentDirectory-v1-Service.pdf` page 46, show you what they want.
It end to the same object with the same metadata/content/url, but they have different Id as they are unique object in the ContentDirectory point of view (not in the same parent).
Here, they add refID to item ID 20 (optional) to let know the renderer they are the same objects and others deeper Tracking Changes Option things.

```
<item id="15" parentID="12" restricted="false">
	<dc:title>Playing in the pool</dc:title>
	<dc:date>2001-10-25</dc:date>
	<upnp:class>object.item.imageItem.photo</upnp:class>
	<res protocolInfo="http-get:*:image/jpeg:*" size ="25000">
		http://10.0.0.1/getcontent.asp?id=15
    </res>
</item>
<item id="20" refID="15" parentID="13" restricted="false">
	<dc:title>Playing in the pool</dc:title>
	<dc:date>2001-10-25</dc:date>
	<upnp:class>object.item.imageItem.photo</upnp:class>
	<res protocolInfo="http-get:*:image/jpeg:*" size="25000">
		http://10.0.0.1/getcontent.asp?id=15
    </res>
</item>
```


From `UPnP-av-ContentDirectory-v4-Service`

page 36 :
```
Each object that is managed by a ContentDirectory service is assigned a unique ID by the
implementation. This object ID, which is exposed via the object’s @id property, provides
a unique identity for the object. The @id property is essential for reliably identifying a specific
object among those hosted by a particular ContentDirectory service because no other object
metadata (or combinations of metadata) provides a distinct identity for the object.
For example, two distinct objects, each with a unique @id property value, can have identical
metadata (except for the @id property) such as two objects representing the same content.
Without the @id property, there is no way to definitively distinguish between the two objects.
Even when there are differences in the metadata of two objects, those differences can be
temporary and do not provide a reliable way to distinguish the two objects.
```

page 41 :
```
Within the ContentDirectory service tree hierarchy, an object resides directly below a
container if that object’s @parentID property value equals the @id property value of that
container. That container is called the parent (container) of the object (one level up) and the
container is said to have a parent relationship with the object . The object is called a child
(object) of the container (one level down) and the object is said to have a child relationship
with that container.

A child object can be either an item object or a container object.

Any object can only have one parent container, except for the root container, which has none
(indicated by setting its @parentID property value to “-1”).
```

page 46 :
```
For example, if there are three playlist containers that all contain child items representing the same
song, the ContentDirectory service implementation can store one item that contains all of the song’s
metadata and store two (smaller) reference items that simply point to the one referenced item.
When a reference item is browsed (via Browse() or Search() actions), it shall be returned as a valid
DIDL-Lite object (for example, @id, dc:title, etc. properties are required).
The metadata of the returned object shall be an exact copy of the metadata from the referenced item
except for any of the following:
The reference item shall not inherit the following property values from the referenced item:
 @id
 @parentID (but can have the same value as the @parentID property of the referenced item if they reside in the same container.)
```

When you change the parent Id of an object, it is like if the object was moved (in ContentDirectory service point of view).
When you set parentId to 0, the renderer think this object reside in the root folder (what is not the case).

At end, it "may" work on most of renderer, or not if the renderer is aware, but fact is that is break ContentDirectory service implementation.

SurfaceS:
```
An object is called a descendant object of a container if it is a child of that container or if it is
connected to that container through one or more intermediate child relationships (any level
down). A descendant object can be either an item object or a container object. All objects
(except the root container) are descendants of the root container.
```

That is why my fix [add limit to item path to avoid OOM](https://github.com/UniversalMediaServer/UniversalMediaServer/pull/4037/commits/20d3c0d5c218f06d7c1a1c87e6bc5a0725c7f7df) should not exists if UMS follow ContentDirectory service implementation.

ik666:
> That is why my fix [add limit to item path to avoid OOM](https://github.com/UniversalMediaServer/UniversalMediaServer/pull/4037/commits/20d3c0d5c218f06d7c1a1c87e6bc5a0725c7f7df) should not exists if UMS follow ContentDirectory service implementation.

I think I disagree here. 

Probably, this fix is only needed, because your MediaStore accepts only LONG as ObjectID's, behaving in my opinion non UPnP compliant. Items with a parentID of, let's say "123ABC" will eventually get a parentID set to "NULL" because of the conversion exception from String to Long in the class `StoreResource` line 162 : `return Long.valueOf(getId());`. Finally this is the non UPnP compliant behaviour..

Why not just do it right an accept Strings like "123ABC" as an objectID / parentID in the MediaStore implementation (which is no big deal I believe).

ik666:
> This example from `UPnP-av-ContentDirectory-v1-Service.pdf` page 46, show you what they want.
> It end to the same object with the same metadata/content/url, but they have different Id as they are unique object in the ContentDirectory point of view (not in the same parent).
> Here, they add refID to item ID 20 (optional) to let know the renderer they are the same objects and others deeper Tracking Changes Option things.

No no no ... You're only quoting section 2.8.5.2 of the CDS v1 manual.

The step before is 2.8.5.1. : 
```
request: CreateReference( “13”, “15” )
response: CreateReference( “20” )
```

This example is telling the CDS to create a reference which will be then available later within a `SEARCH` request. This is something different. And I don't think UMS has implemented the `CreateReference` service.

ik666:
> At end, it "may" work on most of renderer, or not if the renderer is aware, but fact is that is break ContentDirectory service implementation.

I don't see any break in the "ContentDirectory service implementation". Also quoting "page 46" which is an explanation of reference items will not explain, why the tuple (objectID, parentID) has to be unique. I don't see any evidence for that in the docs.

But again, I might be wrong, and regarding this PR, making MediaStore aware of objectID/parentID of type String will just solve everything.

SurfaceS:
> No no no ... You're only quoting section 2.8.5.2 of the CDS v1 manual.

I can quote all part of the spec, they all give the same result.

> why the tuple (objectID, parentID) has to be unique. I don't see any evidence for that in the docs.

Just have a look on the UPnP AV basis.
http://upnp.org/resources/documents/UPnP_AV_tutorial_July2014.pdf

Page 19 : Objects: Hierarchy
CDS exposes tree of meta data objects.

Hierarchy maintained by @id and parentID (here you have part of the answer)
Hierarchy Starts with Object “0”

Page 20 :
Have required properties:
• @id Unique object identification in the hierarchy (hierarchy is the all tree of meta data objects).


> Why not just do it right an accept Strings like "123ABC" as an objectID / parentID in the MediaStore implementation (which is no big deal I believe).

It's about efficiency concern.
Id need to be stored and indexed in the database.
ParentId as well.
Using long is faster and take less memory than varchar in every aspect.
Next Id assignation is simpler.

As you see (I expect) that Id should be unique in the CDS, a media can create a lot of CDS objects.
For each media, you will have a new object in "by author", "by album", "by date", and more if "by date/by author" etc, etc... mean maybe 50 or more objects.
1000 songs = > 50000 objects.

Then you will be happy that the database look for a long instead of a string.

The string is just a human reading friendly thing.

I made change on the last PR that assign a storeId to VirtualFolderDbId (and implement translation on it at the same time).

Let me know if it solve your problem.

ik666:
@SurfaceS Yeah, I don't want to argue to much, I know where you come from and agree a leaf in a tree implementation can only have one parent ... you might be right.

I think your approach storing DBID's or container/folder in the STORE_IDS "folder structure" is a good approach storing the missing folder information. I'll check it and report ...

ik666:
@SurfaceS Nice work, thanks. With current V14 I get a `StackOverflowError`, because of a recursion of the VirtualFolderDBID with i18nName "MyAlbums". I believe `StoreContainer` line 129 make a recursive call. 

StackTrace is:

````
INFO  07:29:57.397 [Media Scanner] java.lang.Exception: Stack trace
INFO  07:29:57.402 [Media Scanner] 	at java.base/java.lang.Thread.dumpStack(Thread.java:1380)
INFO  07:29:57.403 [Media Scanner] 	at net.pms.store.StoreContainer.addGlobalRepo(StoreContainer.java:464)
INFO  07:29:57.403 [Media Scanner] 	at net.pms.store.StoreContainer.addChildInternal(StoreContainer.java:456)
INFO  07:29:57.403 [Media Scanner] 	at net.pms.store.StoreContainer.addChildContainer(StoreContainer.java:325)
INFO  07:29:57.403 [Media Scanner] 	at net.pms.store.StoreContainer.addChild(StoreContainer.java:129)
INFO  07:29:57.403 [Media Scanner] 	at net.pms.store.MediaStore.addVirtualMyMusicFolder(MediaStore.java:793)
INFO  07:29:57.403 [Media Scanner] 	at net.pms.store.MediaStore.<init>(MediaStore.java:103)
INFO  07:29:57.403 [Media Scanner] 	at net.pms.renderers.Renderer.getMediaStore(Renderer.java:259)
INFO  07:29:57.403 [Media Scanner] 	at net.pms.store.StoreContainer.addGlobalRepo(StoreContainer.java:464)
INFO  07:29:57.403 [Media Scanner] 	at net.pms.store.StoreContainer.addChildInternal(StoreContainer.java:456)
INFO  07:29:57.403 [Media Scanner] 	at net.pms.store.StoreContainer.addChildContainer(StoreContainer.java:325)
INFO  07:29:57.403 [Media Scanner] 	at net.pms.store.StoreContainer.addChild(StoreContainer.java:129)
INFO  07:29:57.403 [Media Scanner] 	at net.pms.store.MediaStore.addVirtualMyMusicFolder(MediaStore.java:793)
INFO  07:29:57.403 [Media Scanner] 	at net.pms.store.MediaStore.<init>(MediaStore.java:103)
INFO  07:29:57.403 [Media Scanner] 	at net.pms.renderers.Renderer.getMediaStore(Renderer.java:259)
INFO  07:29:57.403 [Media Scanner] 	at net.pms.store.StoreContainer.addGlobalRepo(StoreContainer.java:464)
INFO  07:29:57.403 [Media Scanner] 	at net.pms.store.StoreContainer.addChildInternal(StoreContainer.java:456)
INFO  07:29:57.403 [Media Scanner] 	at net.pms.store.StoreContainer.addChildContainer(StoreContainer.java:325)
INFO  07:29:57.403 [Media Scanner] 	at net.pms.store.StoreContainer.addChild(StoreContainer.java:129)
INFO  07:29:57.403 [Media Scanner] 	at net.pms.store.MediaStore.addVirtualMyMusicFolder(MediaStore.java:793)
INFO  07:29:57.403 [Media Scanner] 	at net.pms.store.MediaStore.<init>(MediaStore.java:103)
INFO  07:29:57.403 [Media Scanner] 	at net.pms.renderers.Renderer.getMediaStore(Renderer.java:259)
INFO  07:29:57.403 [Media Scanner] 	at net.pms.store.StoreContainer.addGlobalRepo(StoreContainer.java:464)
INFO  07:29:57.403 [Media Scanner] 	at net.pms.store.StoreContainer.addChildInternal(StoreContainer.java:456)
INFO  07:29:57.403 [Media Scanner] 	at net.pms.store.StoreContainer.addChildContainer(StoreContainer.java:325)
INFO  07:29:57.403 [Media Scanner] 	at net.pms.store.StoreContainer.addChild(StoreContainer.java:129)
INFO  07:29:57.403 [Media Scanner] 	at net.pms.store.MediaStore.addVirtualMyMusicFolder(MediaStore.java:793)
INFO  07:29:57.403 [Media Scanner] 	at net.pms.store.MediaStore.<init>(MediaStore.java:103)
````

**Comment by SurfaceS** [comment]
No, addVirtualMyMusicFolder should not sit in initializer. As the store in not yet created, it'll create a new one indefinitely on getMediaStore call.

**Comment by SurfaceS** [comment]
Should be fixed.

**Comment by ik666** [comment]
> Should be fixed.

Yes it is working. Thanks.

**Comment by ik666** [comment]
> Should be fixed.
@SurfaceS One exception: The `MyMusic` folder is not visible in the WebUI below the Audio-Folder any more as in v13.

**Comment by ik666** [comment]
@SurfaceS I've still one issue with browsing storeID's in regard to the storeID concept. If I navigate to a playlist identified by DBID, i.e. `$DBID$PLAYLIST$307?` i'll get a storeID of let's say 453. 

If I BROWSE to this resource, I get no answer (HTTP_500). `MediaStore` seems to fail (re-)creating the resource around line 470. The local variable libraryIds contains actually the Playlist Resource, but it doesn't get into account ... I remember you said something about it the resource has to be browsed. You mentioned the storeID can be used as a permanent link over time, or am I wrong here?

**Comment by SurfaceS** [comment]
> One exception: The MyMusic folder is not visible in the WebUI below the Audio-Folder any more as in v13.

Are you talking about "My Albums", or this is another thing ?


> You mentioned the storeID can be used as a permanent link over time, or am I wrong here?
It should, yes.

But in your example : 
You browse `$DBID$PLAYLIST$307` -> direct access to something -> this object will have the unknown parent of root, and an ID 453.
A UMS restart will not be able to find any more this resource as it is, in fact, not present in the root.


Another example : You browse normally the tree Root (0) -> My Albums (153 for example) -> An album (567).
After a restart, a call on the id 567 : 
If 567 is already on the memory store, give it.
If not, get the parent tree from db (here 0, 153).
Then update the tree : 0 Root, if 153 is here refresh it, and it should have the 567 on it.
If not, it is that tree has changed (folder renamed, moved, or so), or the file is more in the container, and UMS should respond by a ContentDirectory error code 701 (NO_SUCH_OBJECT).

**Comment by ik666** [comment]
> Are you talking about "My Albums", or this is another thing ?

No it's different. It's the permalink use case I mentioned above. From a logical point of view the CDS tree and the filesystem tree have the same structure. Therefore I thought if I find the physical playlist file, the physical folder the file lies in, should also be present in the CDS tree ... 

I looked into the database. The library scanner is adding files to the database, but not the corresponding folder's.

**Comment by SurfaceS** [comment]
In fact yes and no.
The parent can be the container from the real folder (what you suggest), but also a container "by date", and so on.
And each parent have it's own id. That's the funny (not really) things.

Also, the search `ContentDirectory` function should not return it's own id on founded objects, but the founded object parentId on the starting container tree (that's why it has a `containerId` parameter).
But that is too much to do for a fix on this PR.

`MediaLibrary` do not use direct access nor search feature, so everything if fine there.

On the last PR, I add a constraint that accessing an object like `$DBID$PLAYLIST$307` will work only if this object was browsed before (and then known by upnp store ids).
The store can then find a working parent on the tree.

**Comment by ik666** [comment]
>In fact yes and no.
The parent can be the container from the real folder (what you suggest), but also a container "by date", and so on.
And each parent have it's own id. That's the funny (not really) things.

Yeah, but I still think your argumentation you gave in the above chat regarding the CDS tree is correct. If you identify an object by objectID there can only be exact one parentID. Finally, I was convinced after I found the CDS action `MoveObject` in the UPnP spec (v4) :

The `5.5.13 MoveObject()` service action moves a resource from one container to another.  IN parameter are `objectID` and `NewParentID`. This means a resource identifies by an objectID will get a new parent objectID.

If the same resource should be visible in another location i.e. "by date" a new objectID with a refID (to the original resource) has to be used. 

I'd like to suggest an extension of the MediaStore concept.

1) Each discovered folder (by LibraryScanner) must be stored in the database, given an objectID and parentID (FOLDER table). Here we construct the CDS tree.
2) Each discovered file (by LibraryScanner), must be stored in the database, given an objectID and parentID (FILES table). ParentID is the folder that was discovered in step 1. We "physically" link a resource in the CDS tree.

Now objects can be browsed or even moved by `MoveObject`, operating on ID's generated in step 1 & 2

Virtual resources:

3) Virtual container resources (playlists-folders, "by date", youtube, etc.), should be added to the FOLDER table.
4) item-objects which are located in the "physical" CDS tree must not expose their original objectID, but a new one. The original objectID must be referenced (REFID). For that, we link in the REFID table a FILES.ID objectID with a FOLDER.ID. For this link we calculate a unique objectID as "refid", which can then be used while browsing virtual folder constructing the correct refID="..." entry.

Bildschirmfoto 2023-11-09 um 13 52 49

I think it's almost what STORE_IDS wanted to do, but may a little cleaner to understand. What do you think?

**Comment by ik666** [comment]
In the above scenario the "deeplink" or "permanent link" `$DBID$PLAYLIST$307` would point to the objectID in the CDS tree, knowing the object and it's parent. 

Maybe we could prefix the LONG output "objectID" with a table ident like "FILES_123", "FOLDER_234", "REFID_456" just to know where to lookup the resource identified by the long value.

**Comment by SurfaceS** [comment]
I was thinking the exactly same idea at start.
But after, I found that depending on the renderer, the file can have different parent Id (for example, if the renderer need to split a container for large item number) or for some iso/zip/etc that, if the renderer accept it, it will be given as item or container with items, and surely other cases.

So, the best end was to have an id that is linked to another id (container).
If we ask for a media from the container, the objectId returned will have the same id as before if it exists.
If we ask for an object Id, it will be the same object in the same container as before or not be given if : 
- the container is not anymore valid, like if I change settings to not share anymore this folder, it's container will not be accessible anymore until you share it again, or the container was deleted, etc.
- the container do not anymore contain this id (file/folder removed).

I also see that refId is optional and will complicate a bit the thing, but may be used.
In the background, that tell the renderer to load first the metadata from the refId object, then adjust with the id object (object may have different metadata).
Then the service need to maintain updateId for each as now without ref.

>The original objectID must be referenced (@refId).

remember that the refId need to reference an existing and accessible objectId in the tree, as it will be load first from the renderer on request.

> In the above scenario the "deeplink" or "permanent link" $DBID$PLAYLIST$307 would point to the objectID in the CDS tree, knowing the object and it's parent.

Here, you will have the "called" main id. But it don't care witch one you have as you are calling a orphaned object, and maybe you want another container as parent.
To end in a specific container, you need to call the real CDS ID.

The last PR will end on the first found valid object when you are calling a orphaned object.

> Maybe we could prefix the LONG output "objectID" with a table ident like "FILES_123", "FOLDER_234" just to know where to lookup the resource identified by the long value.

With what we have now, browsing "123" will give an item and "234" a container.

**Comment by SurfaceS** [comment]
> The 5.5.13 MoveObject() service action moves a resource from one container to another. IN parameter are objectID and NewParentID. This means a resource identifies by an objectID will get a new parent objectID.

UPnP allow a lot of things if well implemented.

That's why I said few time ago that the nextcp api should not exists.
When you want to do something, most of time you can have the solution under UPnP CDS.

LikedAlbums, Starring etc - could be done by an UpdateObject call changing metadata.
Playlist -> CreateObject for the playlist, CreateObject for the items.

**Comment by ik666** [comment]
> That's why I said few time ago that the nextcp api should not exists. When you want to do something, most of time you can have the solution under UPnP CDS.
> 
> LikedAlbums, Starring etc - could be done by an UpdateObject call changing metadata. Playlist -> CreateObject for the playlist, CreateObject for the items.

Probably true. We could implement playlist management and other things into the CDS service. But then we have to identify the CDS items uniquely.

**Comment by ik666** [comment]
> I also see that refId is optional and will complicate a bit the thing, but may be used.
In the background, that tell the renderer to load first the metadata from the refId object, then adjust with the id object (object may have different metadata).
Then the service need to maintain updateId for each as now without ref.

I don't think so: see section `5.2.21 reference, reference item, referenced item`:
The reference item shall contain a @refID property, whose value shall be equal to the value of the @id property of the referenced item. Note: Control points can use the existence of the @refID property to distinguish between a referenced item and all of the reference items that point to it.

`Control points can use the existence of the @refID` ->  The renderer or control points doesn't load the metadata from the referenced object. All metadata must already be delivered by the browse request to a VIRTUAL_OBJECT_ID (above diagram). The VIRTUAL_OBJECT_ID must be used to render all metadata from the original object following the FILE.ID path.

**Comment by ik666** [comment]
> I was thinking the exactly same idea at start.
> But after, I found that depending on the renderer, the file can have different parent Id (for example, if the renderer need to split a container for large item number) or for some iso/zip/etc that, if the renderer accept it, it will be given as item or container with items, and surely other cases.

But let's please follow the idea to have a static 1:1 representation between CDS tree and the system directory tree with a data structure as pictured above. Maybe some tweaks must be made, but it simplifies many thing I believe.

From my point of view we have an issue in logic because we're mixing object types in the FILES table. We should store CDS item objects into FILES table, CDS container objects in the FOLDER table. For example:

`C.2.1.4 playlistItem:item` 
> A playlistItem is required to have a res property for playback of the whole sequence. This res property is a reference to a playlist file authored outside of the ContentDirectory service (for example, an external M3U file).

->  `PlaylistItem` is the real `.m3u8` file.

`C.2.2.2 playlistContainer:container`
> [...] a control point needs to separately initiate rendering for each child object, typically in the order the children are received from a Browse() action.

The playlist can be represented by a file or by a container. Both can/should/must(?) be delivered to a control point / renderer. 

Same concept can be applies to all other item/container types you mentioned (MKV, ZIP, etc.) There is an "item" representation of the file, which can i.e. be used to "move" with "MoveObject", or be browsed as a container representation. I've extended the for this concept.

You could even decide by renderer to present either the item or container representation instead of both.

The datastructure would look like this:

Bildschirmfoto 2023-11-10 um 09 55 58

**Comment by ik666** [comment]
Even though I thinks it's not neccesary, storing a renderer specific tree should also be possible (see STORE_IDS). 

>Then the service need to maintain updateId for each as now without ref.

Probably it's enough to maintain the updateId on the container resource.

It would look like this:

Bildschirmfoto 2023-11-10 um 10 28 56

**Comment by SurfaceS** [comment]
> Probably true. We could implement playlist management and other things into the CDS service.

Not on this PR (we are in bug fixing stage).

>  But then we have to identify the CDS items uniquely.

It is the case for now. An id will always end to the same object.

> VIRTUAL_OBJECT_ID things

CDS Items are not only file.id based. It can be virtual, part of file, actions, track on a DVD, etc.
Same for CDS folders.
Then, what append when, like movies, you present the same file with multiple options (transcoding, language, subtitles, and so on) ? as each one need a unique item id.

It will need a huge work without any benefit (maybe more bugs :)).
And that will end to the same state as now : You need to know something from the UMS db to access an object.
Object Id on this PR.

On v13, file ID, track ID or playlist ID are just db fields UMS gave you.
Like $DBID$PLAYLIST$307 (given by UMS for itself) mean it is a playlist that should load the file ID 307.
This info was useful by UMS because as it does not keep track on what it serve, it need to have back something it understand.

> playlistItem vs playlistContainer

That is a good example.
One renderer can read playlist files, another one can't.
The first renderer will have playlist parent (assign id 150), that contain playlist item (assign id 151).
The second renderer will have playlist parent 150 (because first renderer already assign an id to this container), that contain a container (assign id 152), that contain all medias on the playlist (assign id 153 and 154).
If the second renderer set that it can now read playlist itself (config) and refresh, it will end to 150-151 like the first one.
A third renderer appear 1 month later and is set to read playlist files, but also to have the playlist container shown.
It will have the playlist parent 150, the playlist item 151, the playlist container 152 that contain the medias on the playlist 153 and 154.

This is already handle by the actual db.

> It would look like this:

What is the final goal ?

**Comment by ik666** [comment]
Don't worry, no intention to implement playlist management right now :)

> CDS Items are not only file.id based. It can be virtual, part of file, actions, track on a DVD, etc.

Agree, and yes, all those "virtual" items should be parsed and added to the Files table (maybe we should rename the table to StoreItem). All (virtual) folders should be added to the FOLDER table (maybe it should be named StoreContainer). All get a valid ID and can be directly identified (renderer independent).

> What is the final goal ?

To implement the spec as described in CDS V4, Anned D, Section2.

>D.2 Generating Object ID Values
As discussed in subclause 5.2.3 and 5.2.4, control points can benefit when objects preserve their identify (i.e. retain the value of their @id property even when going off-line)

I believe this should apply to CDS items and CDS container objects.

I want to get rid of "$DBID$PLAYLIST$307" and use the objectId generated by the UMS CDS as you proposed.

Staying with your example above I'd like to just browse to ID 152. The DIDL result must present a valid parentID objectID (150). Both resources should be browsable.

I also want to be able "remember" / deep link / permanent link objectID 150 (the folder). I question myself how to keep track of the folder if you don't store it in the database (at least I don't see it yet).

You said:
> With storeIds, you will be able to do the same thing with the storeId (that will never change).

But I had an issue:
>@SurfaceS I've still one issue with browsing storeID's in regard to the storeID concept. If I navigate to a playlist identified by DBID, i.e. $DBID$PLAYLIST$307? i'll get a storeID of let's say 453.

Actually, it's not important how I get the (storeID) objectID 453. Obviously, UMS has no idea where the item with objectID 453 is stored. Doesn't sound UPnP compliant. 

Let's say a second control point / renderer ist accessing objectID 453, it should work. The second renderer should be able to walk the CDS tree up towards root ("0"). 

Soft facts
- Documentation of the solution for other developers
- If possible to "keep it simple"

**Comment by ik666** [comment]
> This is already handle by the actual db.

OK, I'll do some more investigation, what's going on.

**Comment by SurfaceS** [comment]
> I also want to be able to "remember" / deep link / permanent link objectID 150 (the folder). I question myself how to keep track of the folder if you don't store it in the database (at least I don't see it yet).

It is stored in db.

> Actually, it's not important how I get the (storeID) objectID 453. Obviously, UMS has no idea where the item with objectID 453 is located in the CDS tree. Doesn't sound UPnP compliant.

As an id is linked with a parentId, and it's parent as well, it can be located on the CDS.
And that why I had prefer to not set the fake ID of 0 that would not, and instead find the first working one if you ask an orphaned id "$DBID..." (or null if none found).

> Let's say a second control point / renderer ist accessing objectID 453, it should work. The second renderer should be able to walk the CDS tree up towards root ("0").

It is already the case now.
If the second control point / renderer is not accessing it, it mean it does not have access to it (file/folder/permissions/settings/not registered).
If you browse the webPlayer, login, UMS create a renderer.
Browse to a folder or a file.
Shutdown UMS, restart.
Refresh the webpage. You will end in the same folder/item. (viewable with the breadcrumb).
You can copy the url, logout, login (you end in a new renderer).
Paste url, it will end the same.

**Comment by ik666** [comment]
>Browse to a folder or a file.

I don't want to browse to a file or folder, because it's not my use case. I have over 60.000 audio files in over 3.000 directories and your solution is that I should manually click through every folder, before UMS starts working? 

Giving objects identity lazy by the time of "browsing" has some - let's say - drawbacks. UPnP search seems therefore to be broken now. 

**case 1) do not browse any folder below root**

1) start with a clean database and wait until shared folder scan is finished.
2) Do a UPnP search `Search("0", "dc:title contains "Material Girl")`
3) I have one hit in the database. The audio metadata is linked to an entry in the FILES table.

How should the ID's (id & parentID) of the DIDL result be constructed?

```
<DIDL-Lite 
[...]
    <item id="18" parentID="13" restricted="0">
</DIDL-Lite>
``` 

**case 2) browse MediaLibrary **

1) start with a clean database and wait until shared folder scan is finished.
2) browse through the MediaLibrary -> Audio 
  -> "By date"
  -> "All files"
  -> By author -> "Madonna"
  -> and so on
  [...]
  Let's assume the audio file with the title containing "Material Girl" has been "browsed" multiple times.

3) Do a UPnP search `Search("0", "dc:title contains "Material Girl")`
4) I have one hit in the database. The audio metadata is linked to an entry in the FILES table.

How should the ID's (id & parentID) of the DIDL result be constructed?

From a user perspective would you expect only one hit or many ? I'd personally expect only one hit, the item which is located in the "physical" file system. 

I believe everything would be ok, if the LibraryScanner would "browse" to every physical folder it is scanning, filling the STORE_IDS table with the physical CDS tree. As a side effect, DBID's will probably work again.

ik666:
> I believe everything would be ok, if the LibraryScanner would "browse" to every physical folder it is scanning, filling the STORE_IDS table with the physical CDS tree. As a side effect, DBID's will probably work again.

Implemented with PR #4240, therefore this one is obsolete.

PR #4219, opened 2023-11-06
fix: RequestV2 missing TextContent (body)

ik666:
Some API request need the POST body, that wasn't set in RequestV2.


SurfaceS:
added under the `uri.startsWith("api/")` condition as it break code after it if not.

PR #4222, opened 2023-11-07
assing id to VirtualFolderDbId

SurfaceS:
translate VirtualFolderDbId


ik666:
Thank you for your support. I'll check that ...

PR #4236, opened 2023-11-11
Bump Java to 17.0.9+11

SubJunk:


SurfaceS:
@SubJunk As I see you updated the svn, there is also that binaries that should be updated #4237.

PR #4240, opened 2023-11-13
UPnP search / eager generate storeIDs while MediaLibrary scan

ik666:
@SurfaceS With this PR I believe I can implement UPnP search again, overcoming the issue described in #4217.

### DIFF
--- a/src/main/java/net/pms/store/MediaScanner.java
+++ b/src/main/java/net/pms/store/MediaScanner.java
@@ -33,6 +33,7 @@ import net.pms.configuration.sharedcontent.SharedContentConfiguration;
 import net.pms.configuration.sharedcontent.SharedContentListener;
 import net.pms.database.MediaDatabase;
 import net.pms.database.MediaTableFiles;
+import net.pms.database.MediaTableStoreIds;
 import net.pms.gui.GuiManager;
 import net.pms.platform.PlatformUtils;
 import net.pms.renderers.ConnectedRenderers;
@@ -133,6 +134,14 @@ public class MediaScanner extends StoreContainer implements SharedContentListene
 
 					// Display and log which folder is being scanned
 					if (storeContainer instanceof RealFolder) {
+						if ("scanner".equals(resource.getName())) {
+							// "scanner" is the root folder where shared resources are located
+							resource.setId("0");
+						}
+						// eager scan physical file system
+						MediaStoreId childId = MediaTableStoreIds.getResourceMediaStoreId(child);
+						child.setId(childId.getId() + "");
+
 						String childName = child.getName();
 						LOGGER.debug("Scanning folder: " + childName);
 						GuiManager.setStatusLine(Messages.getString("ScanningFolder") + " " + childName);
@@ -160,6 +169,16 @@ public class MediaScanner extends StoreContainer implements SharedContentListene
 					}
 				} else if (!running) {
 					break;
+				}
+
+				if (running && child instanceof RealFile && child.getParent() instanceof RealFolder) {
+					// We're building the "physical" CDS tree for real items and objects
+					MediaStoreId childId = MediaTableStoreIds.getResourceMediaStoreId(child);
+					if (childId != null) {
+						child.setId(childId.getId() + "");
+					} else {
+						LOGGER.trace("couldn't determine storeID for child : {} | parent : {} ", child, child.getParent() != null ? child.getParent() : "NULL");
+					}
 				}
 			}
 		} else {


ik666:
@SurfaceS  Thanks for you help past days. I took me some time to figure out exactly how the MediaStore concept works.  UPnP search for items seems to work now. I'll have a look at container types later.

ik666:
@SurfaceS Look's finished so far. UPnP seems to work as again. What do you think?

SurfaceS:
> not to store any DBID objects below root or "0". They reside in a separate folder.

They should end to valid objects and container.
For a Browse, they should sit under the asked container (and relation stored to the `MediaTableStoreIds`).
For a Search, objects added to the results should have their own real parent id for parent id (as stored in `MediaTableStoreIds`), and you can add them to a fake container. Just not store the search relation on the database.

`VirtualFolderDbId nonVisibleDbIdFolder` is not a solution.

> The UPnP-search looks up only physical resources.

In fact, the UPnP-search should find what is under the ContainerID passed as argument (I think this is not implemented anywhere).
It can be others containers or items that match the search.
It allow parent ids to be the searched ContainerID, but also a child container included in the ContainerID tree.

For UPnP-Browse with BrowseDirectChildren for BrowseFlags, all objects returned should have the browsed object as parent id.

For UPnP-Browse with BrowseMetadata for BrowseFlags, only one object is returned and should have the browsed object as id.

> All DBID resources are reconstructed from the CDS tree (Store_IDS) or being only once created.
> No need to find playlists over API, because UPnP-StorageFolder search is working and can be used to search for them

Be careful that an id from the `MediaTableStoreIds` could refer to an object not for your renderer/user.
It could be one for another renderer (settings), or another user (shares permissions).

That why you should always test if this object is available on the renderer `MediaStore`.
So you should get all the ids matching what you want, then ask your `MediaStore` to give you the object.

**In short :**
If you query the db for files/folders (`MediaTableFiles`), you should test each file is allowed for the user before creating the resource.
This can be done with `renderer.hasShareAccess(file)`.

If you query the db for ids (`MediaTableStoreIds`), you should test each id is available on the current renderer CDS.
`renderer.getMediaStore().getResource(id.toString());` -> if `StoreResource`, you can add it, if `null`, you should not use it.
_(if `StoreResource` is returned, it mean the renderer CDS was able to find the object in memory or recreate the object if not anymore on the memory.)_

ik666:
@SurfaceS I've started to change the lookup logic as suggested:

>If you query the db for files/folders (MediaTableFiles), you should test each file is allowed for the user before creating the resource. This can be done with renderer.hasShareAccess(file).

This seems to work

>If you query the db for ids (MediaTableStoreIds), you should test each id is available on the current renderer CDS.
renderer.getMediaStore().getResource(id.toString()); -> if StoreResource, you can add it, if null, you should not use it.
(if StoreResource is returned, it mean the renderer CDS was able to find the object in memory or recreate the object if not anymore on the memory.)

This seems not to work, if the resource hasn't been visited / browsed / discovered yet. `renderer.getMediaStore().getResource(id.toString());`gives NULL (without reconstructing/recreating the path to the resource) for the current renderer. If I manually browse to the resource before initiating the search request it works.

Do I have to manually walk the tree up adding all the folders on the path until the first "discovered resource"?

ik666:
@Surface. Got it. The reconstruction of the CDS tree for a resource not in memory had an issue.

ik666:
>VirtualFolderDbId nonVisibleDbIdFolder is not a solution.

@SurfaceS With last commit MusicBrainz album folders and persons are in their own MediaLibrary audio folder. If an album has no MusicBrainzID, the folder is located from below the MediaLibrary "By Album" folder. 

> In fact, the UPnP-search should find what is under the ContainerID passed as argument (I think this is not implemented anywhere).

Yes, you're right. Can be done and should be on the todo list. I haven't found a control point that uses a subtree search yet. But I must admit, I'm testing only with BubbleUPnP (Android) , LINN App und Lumin App (iOS)

I think your points are resolved. We're done with this PR ?

ik666:
@SurfaceS Just FYI (can be fixed later). The MusicBrainz recordID has only a "title", no "album. The date can be almost anything, YYYY,  YYYY-MM-DD or even other way around. Should be finally a string in the MBID table. Only the "year" part should be parsed for MediaInfo.

SurfaceS:
> @SurfaceS Just FYI (can be fixed later). The MusicBrainz recordID has only a "title", no "album". The date can be almost anything, YYYY, YYYY-MM-DD or even the other way around. Should be finally a string in the MBID table. Only the "year" part should be parsed for MediaInfo.

Yes, we had the same on movie/tv series api.
We end with an INTEGER for year and DATE for date (if needed).
But the api / parser / ect need to well parse the infos.

Issue #4148, opened 2023-10-17
Add ability to shut down computer from web settings


Vbarbarian:
Currently UMS supports shutting down the computer via DLNA, by the `#--Server Settings--#` folder:
![image](https://github.com/UniversalMediaServer/UniversalMediaServer/assets/584228/b1c91837-19b1-49e5-8835-9d14aebc470e)
but this is not a user-friendly thing to do, to play a video to trigger an action, it is just a workaround because that type of interface does not support buttons.

On our web interface we can be more user friendly (while still supporting the existing DLNA way on renderers) by adding the setting to the `Settings` area in the web interface.

- [ ] should be a button to shut down the computer that UMS is running from via the web settings
- [ ] should _still_ be a fake video on the `#--Server Settings--#` virtual folder that can do it via renderers
- [ ] should hide the `#--Server Settings--#` virtual folder on the web player

SubJunk:
Good point, that functionality should be added to the web settings

SurfaceS:
Do you mean COMPLETELY SHUTDOWN THE SERVER ? (the computer)

wolf-haus:
@SubJunk is this still up for grabs? I'll pick it up if so.

SubJunk:
@wolf-haus that would be great thank you, I have added details into the issue to describe the outcome better

wolf-haus:
@SubJunk - to clarify, is the shutdown button to be added to the `Settings` (`/settings`) area? The `Tools` (`/actions`) area has a number of buttons that perform actions on the server / application, so would it make sense to group the shutdown button with those?

SubJunk:
@wolf-haus yeah the actions area would be good, you're right

wolf-haus:
@SubJunk created PR: https://github.com/UniversalMediaServer/UniversalMediaServer/pull/4250

SubJunk:
This will be in the next release. Thank you and happy new year @wolf-haus

PR #4250, opened 2023-11-17
Add ability to shut down computer from web settings #4148

wolf-haus:
Added a button to the Actions screen of the web UI, which will trigger an API call to shut down the host machine, on which that instance of UMS is running.

As far as I can tell, a Mac or Linux machine won't shut down unless UMS is running with `root` privileges; the application will stop, but nothing further will happen - I have tried to make this clear in the confirmation popup.


SubJunk:
@wolf-haus this looks great, I will test it soon

SubJunk:
This works well, thank you. It will be in the next release

PR #4252, opened 2023-11-17
V14 - ik666

SurfaceS:
@ik666 

I think you should base your work on this.

It better implement a real discover on VirtualFolderDbId, then it will be updated on browse (as sql based).
Also, the MusicBrainz things are handled by MusicBrainzAlbumFolder.

for the search, I removed the fake updateID, replaced by the real CDS updateID.

On the search `getLibraryResourceFromSQL`, the result objects should not be created there.

It should be handled by `DbIdResourceLocator.getLibraryResourceByDbTypeIdent`
Something like : if no result, you know with your dbid type where the result should end.
Find this container, then `doRefreshChildren()`.
If still no result, that mean you do not have result.


ik666:
OK, thx for comment. I'll check in which branch to integrate.

ik666:
@SurfaceS I've integrated your changes in the PR #4240. It was easier for me since it was already ahead of this branch. i think it became a good solution.

ik666:
@SurfaceS In the other branch, MusicBrainzID's are not store in the CDS any more. They took way to much resources, if you have many albums. 

The person items are needed to restore navigation back to parent person views.

ik666:
Maybe you could merge the other branch and we have a new touchpoint to start with.

PR #4282, opened 2023-12-06
read correct tags for artists and album-artists

ik666:
Reads correct ARTISTS and ALBUM_ARTISTS tags. Fallback to old logic in case tags are missing.

### DIFF
--- a/src/main/java/net/pms/parsers/MediaInfoParser.java
+++ b/src/main/java/net/pms/parsers/MediaInfoParser.java
@@ -1187,12 +1187,12 @@ public class MediaInfoParser {
 		MediaAudioMetadata audioMetadata = new MediaAudioMetadata();
 		audioMetadata.setSongname(StreamContainer.getTrack(MI, 0));
 		audioMetadata.setAlbum(StreamContainer.getAlbum(MI, 0));
-		String albumArtists = StreamContainer.getAlbumArtists(MI, 0);
+		String albumArtists = MI.get(StreamKind.GENERAL, 0, "ALBUM_ARTISTS");
 		if (StringUtils.isAllBlank(albumArtists)) {
 			albumArtists = StreamContainer.getAlbumPerformer(MI, 0);
 		}
 		audioMetadata.setAlbumArtist(albumArtists);
-		String artists = StreamContainer.getArtists(MI, 0);
+		String artists = MI.get(StreamKind.GENERAL, 0, "ARTISTS");
 		if (StringUtils.isAllBlank(artists)) {
 			artists = StreamContainer.getPerformer(MI, 0);
 		}
--- a/src/main/java/net/pms/parsers/MediaInfoParser.java
+++ b/src/main/java/net/pms/parsers/MediaInfoParser.java
@@ -1187,8 +1187,16 @@ public class MediaInfoParser {
 		MediaAudioMetadata audioMetadata = new MediaAudioMetadata();
 		audioMetadata.setSongname(StreamContainer.getTrack(MI, 0));
 		audioMetadata.setAlbum(StreamContainer.getAlbum(MI, 0));
-		audioMetadata.setAlbumArtist(StreamContainer.getAlbumPerformer(MI, 0));
-		audioMetadata.setArtist(StreamContainer.getPerformer(MI, 0));
+		String albumArtists = StreamContainer.getAlbumArtists(MI, 0);
+		if (StringUtils.isAllBlank(albumArtists)) {
+			albumArtists = StreamContainer.getAlbumPerformer(MI, 0);
+		}
+		audioMetadata.setAlbumArtist(albumArtists);
+		String artists = StreamContainer.getArtists(MI, 0);
+		if (StringUtils.isAllBlank(artists)) {
+			artists = StreamContainer.getPerformer(MI, 0);
+		}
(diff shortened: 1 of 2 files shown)


SurfaceS:
@ik666 
I had recreated this binding library few months ago to be accurate with the mediainfo sources. And this tags (`ARTISTS` `ALBUM_ARTISTS`) are not valid ones.

Looking at the mediainfo c++ code :
`PERFORMER` -> Performer
`ARTIST` -> Performer (if not set), Composer
`ENSEMBLE` -> Accompaniment (if not the Performer or Composer)
`ALBUM ARTIST`, `ALBUMARTIST` -> Album/Composer, Album/Performer (if Performer not set)

!! The files under `src/main/java/net/pms/parsers/mediainfo/` should only be changed according to the mediainfo sources.

SurfaceS:
https://github.com/UniversalMediaServer/UniversalMediaServer/pull/4037/commits/03f1b890d3031856cfb84830614fd1fa5bb65906

I've just updated the mediainfo parameters from sources.
As you can see, their is not so much changes (most of them are descriptions) and neither `ARTISTS` or `ALBUM_ARTISTS` are on the mediainfo source code.

ik666:
I've checked myself `MediaInfo/Tag/File_VorbisCom.cpp`. Looks like you haven't taken the other file formats into account ... 

There are some default TAGs read ... 

```
             if (Key==__T("ADDED_TIMESTAMP"))        Fill(StreamKind_Common,   0, "Added_Date", Ztring().Date_From_Milliseconds_1601(Value.To_int64u()/1000));
        else if (Key==__T("ALBUM ARTIST"))           AlbumArtists.push_back(Value);
        else if (Key==__T("ALBUM"))                  Fill(StreamKind_Common,   0, "Album", Value);
        else if (Key==__T("ALBUM_COMMENT"))          Fill(StreamKind_Common,   0, "Comment", Value);
        else if (Key==__T("ALBUMARTIST"))            AlbumArtists.push_back(Value);
        else if (Key==__T("ARTIST"))                 Artists.push_back(Value);
        else if (Key==__T("AUTHOR"))                 Fill(StreamKind_Common,   0, "WrittenBy", Value);
        else if (Key==__T("BUYCDURL"))               {}
        else if (Key==__T("BWFVERSION"))             // bext
```
But at the end, it just reads all given VORBIS comments ... and they are all accessible ...

SurfaceS:
> But at the end, it just reads all given VORBIS comments ... and they are all accessible ...

I don't know what version you use, but on my computer, latest version do not show any of this tags (from mediainfo cmd with a flac that have tags in it).

From UMS, (MediaInfo v23.10 and 23.11) direct flac tags return nothing (just checked now).
Maybe have a try on your computer.
```
		String[] testTags = {"ARTIST", "ARTISTS", "ALBUMARTIST", "ALBUM_ARTISTS", "MUSICBRAINZ_ORIGINALALBUMID"};
		for (String testTag : testTags) {
			String testRes = MI.get(StreamKind.GENERAL, 0, testTag);
			if (!testRes.isEmpty()) {
				LOGGER.info(testTag + " tag is " + testRes);
			}
		}
```

BTW, the files under `src/main/java/net/pms/parsers/mediainfo/` should only be changed according to the mediainfo sources (supported by the author).
So if any non supported `get` has to be used, it should not pollute the official one but use its own `get` as `MI.get(StreamKind.GENERAL, 0, "ARTISTS");` on UMS `MediaInfoParser.java` (like on the previous test).

ik666:
You don't see the tags, most likely because you haven't set them. Just use a wide spread tagging tool like MusicBrainz Piccard, Songkong, etc. for tagging.

Actually, I'm not sure what you're up to. As I mentioned, the official source code reads all VORBIS comments at the end ...
```
else                                Fill(Stream_General, 0, comment.SubString(__T(""), __T("=")).To_UTF8().c_str(), Value);
    FILLING_END();
```

It is done, because VORBIS comments are by design only a list of KEY=VALUE pairs. VORBIS does not enforce a specific scheme like ID3 ... That's why you have to "try out" what's set by the majority of the tagging tools. For example "album artist" can be ALBUM_ARTIST, ALBUMARTIST, ALBUM ARTIST ... 

I don't want to pollute anything. Please feel free to put the code where you like it, I don't mind. Just want UMS to work right.

ik666:
> I don't know what version you use, but on my computer, latest version do not show any of this tags (from mediainfo cmd with a flac that have tags in it).

```
% mediainfo -h
MediaInfo Command line, 
MediaInfoLib - v23.07
```

It produces the output I've given above.

ik666:
> When I said "pollute" is that these files need to stay out of parameters not filled, advised and commented by the author. For special case, we have a way to directly access those parameters by calling them directly.

OK. I see what you mean.

> Note : Did MediaInfo not give the right values calling getPerformer and getAlbumPerformer ?

No it gave me the performer value 
`Alex Grube / Claus Fischer / Emma Kelly / Gabriel Kelly / Hanno ...`

> Bad thing for you is that :
 else if (Key==__T("MUSICBRAINZ_ALBUMID"))    {}
is just throwing away this tag (and some others MUSICBRAINZ ones).

Yeah, that sucks. Maybe, that's why Piccard also sets `MUSICBRAINZ_ORIGINALALBUMID`, it's a copy and still accessible in the tags :)

PR #4323, opened 2023-12-31
Fix various documentation typos

luzpaz:
Found via `codespell -q 3 -S "./CHANGELOG.md,./src/main/resources/i18n,./react-client,./src/main/external-resources,./target/classes/resources/i18n,./src/main/java/net/pms/util/Iso639.java" -L abitrate,als,bu,caf,childrens,childs,datas,medias,reencode,serie,setts,vie`


SubJunk (approved):
Cool, thanks

PR #4325, opened 2024-01-03
Remove buggy access control from v13

SubJunk:
This was supposed to be reverted before the 13.7.0 release but that did not fully happen. This also adds a migration back to the old config for now


SubJunk:
This isn't ready yet because I will need to update the database files and related files because the versions got bumped. I am doing that now

SurfaceS:
@SubJunk what is the goal of this PR ?

SubJunk:
@SurfaceS to remove the access control feature from the v13 releases. It was supposed to be reverted and introduced in v14, but I made a mistake and only reverted half of it months ago

SubJunk:
@SurfaceS I think this is ready now. I will upload it to the forum for testing

Issue #4315, opened 2023-12-22
'Check for updates' button locks up UMS (13.8.0)


IntelliMoo:
The 'Check for updates' button displays a blank white "Universal Media Server Auto Update" dialog and locks up UMS (13.8.0).  The dialog and main window cannot be closed by normal means including Alt+F4.

SubJunk:
I have confirmed this, thanks for reporting! Fixing now

SubJunk:
This is fixed since 13.9.0

PR #4109, opened 2023-10-03
System tray icon on update

tylovejoy:
Issue #2352

The system tray icon will have a small red circle in the top right corner if there is an update available.

If there is anything I am missing or done incorrectly, please let me know!

PR #4332, opened 2024-01-04
fix update dialog hanging

tylovejoy:
Fix update dialog box hanging. 

Discussed here: https://github.com/UniversalMediaServer/UniversalMediaServer/pull/4109#issuecomment-1876264311

Solves: #4315

### DIFF
--- a/src/main/java/net/pms/newgui/LooksFrame.java
+++ b/src/main/java/net/pms/newgui/LooksFrame.java
@@ -684,7 +684,9 @@ public class LooksFrame extends JFrame implements IGui, Observer {
 					autoUpdater.pollServer();
 				}
 				AutoUpdateDialog.showIfNecessary(this, autoUpdater, isStartup);
-				PlatformUtils.INSTANCE.addSystemTray(this, autoUpdater.isUpdateAvailable());
+				if (autoUpdater.isUpdateAvailable()) {
+					PlatformUtils.INSTANCE.addSystemTray(this, true);
+				}
 			} catch (NoClassDefFoundError ncdfe) {
 				LOGGER.error("Error displaying AutoUpdateDialog", ncdfe);
 			}


SubJunk (changes requested):
Thanks for the fast response. It still crashes when an update is available. You can see it by setting the version to an older one and running this

Unfortunately I think this bug will cause a lot of problems because everyone with automatic updating enabled will see UMS crash on startup. I should have tested it more.

I will make a change to disable automatic updating for the broken versions

tylovejoy:
Strange- I tested that and it worked fine on my machine. My apologies for not catching this before submitting.

SubJunk:
@tylovejoy interesting that it worked for you, maybe the code hadn't updated when you tested it? Here's a screenshot after I changed the version string to 13.7.1. Can you recheck?
image

Thanks for apologizing, I should have tested it more, plus the way the auto updating code is written it is very easy to introduce bugs. It is using a code pattern where it is heavily bound together so any small change often introduces a bug.

SubJunk:
I think this will prevent those versions from ever finding an update while enabling future versions to find them https://github.com/UniversalMediaServer/UniversalMediaServer/pull/4333

SubJunk:
I think the problem is that the change event is fired twice, because of this observable method https://github.com/UniversalMediaServer/UniversalMediaServer/blob/a6586f671eaccde874bc6c446f0de1c200689976/src/main/java/net/pms/newgui/LooksFrame.java#L669
I noticed that observables are deprecated in Java now, in part because of thread safety problems.

SubJunk:
I have narrowed down the problem to the line that removes the previous TrayIcon now, so I think I am close to a fix for that

SubJunk:
This problem is fixed in https://github.com/UniversalMediaServer/UniversalMediaServer/pull/4333
Unfortunately it involves using a magic number (5000ms) but I had problems otherwise, and didn't want to refactor the entire updating code today 😆 
In testing I verified that it behaves on startup and when the button is clicked

Another thing I noticed is that the update icon is the Windows one, which looks strange on macOS because that has a different tray icon. Are you able to make a variation for macOS too @tylovejoy ?

PR #4374, opened 2024-01-10
added status to splash

SurfaceS:
added dot timer status to see something is running


SubJunk (commented):
@SurfaceS let's keep the old splash screen, the new text style doesn't look right to me. Also are we not able to translate the new status texts?

PR #4456, opened 2024-02-08
Bump MediaInfo to 24.01

SubJunk:


SurfaceS:
@SubJunk 
I think you should use this one for mac :
Headers was added to it.
https://mediaarea.net/download/binary/libmediainfo0/24.01.1/MediaInfo_DLL_24.01.1_Mac_x86_64+arm64.tar.bz2

PR #4487, opened 2024-02-18
add advice how to register a TMDB API key.

SurfaceS:
add Java gui option


SubJunk:
It could be good to add an article that explains that functionality on support.universalmediaserver.com too
I would love for us to have a lot more content on there

Mik-S-UMS:
Yes this definitely needs more documentation on what it does and why it is needed.

I tried to figure this out and made an account on TMDB but the only options to create an API key was to be a developer or professional. Both options need a lot of personal information to generate one.

From how I understood it the API key should belong to the developer of the program and not a user of that program.

SurfaceS:
General Settings
External outgoing traffic

UMS can use an external network for several things:
- get UMS updates.
- find missing information on your media (title, genre, actors, etc.).

If you want to prohibit UMS from doing this, you can uncheck “Enable external network”.

You have two choices to discover missing information on video media.

The easiest way (default) is to use our API to get it for you.

The other one is to use the TMDB API directly with your own TMDB account.
This solution also gives you more control over selecting/refreshing TMDB information directly from the GUI.
This will require a valid TMDB API key, which can be obtained from your TMDB account settings page.
For those who do not yet have a TMDB API key, they can register for it by clicking on the API link under the account settings.
Select the developer's API key and fill in the information that suits you (the key will be yours).

SurfaceS:
> From how I understood it the API key should belong to the developer of the program and not a user of that program.

The problem is that we cannot share the UMS TMDB api key.
And, the implemented way TMDB give is not secure :
https://www.themoviedb.org/talk/63d51446955c65008478dfab

SurfaceS:
You can see the reply from Travis Bell (the founder of TMDB) :
https://www.themoviedb.org/talk/58ff62be925141080b0094f9



> Ya, this comes down to how you want to publish your project. I'll add that since we don't charge for the API, it's not usually a big deal to get people to register their own but if you make the choice to include yours, the main issue you might run into is that someone could take it an abuse it. In that case we block it and you would need to get a new one, but everyone who is using your project would also need to do that (whether it's an update you publish or otherwise).
> 
> Your call on the matter but that kind of explains both ways to think about it.

PR #4507, opened 2024-02-29
Ensure resolution is divisible by 2 after scaling

future-figs:
I had an issue where some video files would not play with the renderer's max resolution set to 640x480. It turned out that the `scale` filter had calculated an output resolution with an odd number of pixels for the height, which caused libx264 to crash.

This PR adds a black 1-pixel pad in either axis if needed, with the output centered. This makes the resolution divisible by 2 along both axes. It only applies to one usage of the `scale` filter because the others either get their output resolutions from the renderer's max resolution (instead of calculating some oddball res), or already have even-number padding implemented (using `convertToModX`). So, now all cases should be accounted for.

One bit of inconsistency is that multiples of 2 are enforced, instead of multiples of 4 as in the existing code with `convertToModX`. If multiples of 4 are required for some reason, then this should be edited to make it so.


SubJunk:
@future-figs thanks for this, yes please make it mod4 because some players can't play some mod2 resolutions. Panasonic TVs being one of them

future-figs:
Thanks for explaining, changed to multiples of 4.

SubJunk:
Thanks!

PR #4521, opened 2024-03-05
Fix TV series with the same title but different startYear being treated the same

SubJunk:


SubJunk:
This seems almost ready but I discovered in testing it makes some media library folders empty. Though only on the TV, not the web player which is weird..

SubJunk:
@SurfaceS it seems to be working now, can you review?

SurfaceS (approved):

PR #4533, opened 2024-03-16
Allow FFmpeg to hardcode subtitles to Dolby Vision video track to be muxed with tsMuxeR

SubJunk:
I have not verified that FFmpeg supports this yet so this is pending testing


SubJunk:
This is difficult because FFmpeg does not automatically output the same format metadata as the input file, so we need to detect and add a lot of flags like:

```
-pix_fmt yuv420p10le 
-x265-params colorprim=bt2020:transfer=smpte2084:colormatrix=bt2020nc:range=limited:master-display="G(13250,34500)B(7500,3000)R(34000,16000)WP(15635,16450)L(10000000,50)"
-bsf:v hevc_metadata=colour_primaries=9:transfer_characteristics=16:matrix_coefficients=9
```

and that is just for HDR, not even Dolby Vision yet.
I hope FFmpeg make it more simple but there is a big arguing thread that makes it seem like they won't do it https://trac.ffmpeg.org/ticket/7037 :(

SubJunk:
This could be resurrected if FFmpeg supports it in the future

Mik-S-UMS:
Just had a look at that Q&A and there was a reply today saying it is supported since FFmpeg 7.1

> Since FFmpeg 7.1, it's as easy as adding parameter -dolbyvision 1

I don't know if this applies to subtitles or preserves the right metadata though.

SubJunk:
Yeah that's amazing timing haha, maybe someone was following this issue. I'll reopen it

SubJunk:
Unfortunately there is a bug in FFmpeg that causes it not to add the correct metadata in MPEG-TS files. I confirmed it using FFmpeg and MediaInfo so it's not something to do with UMS. But tsMuxeR can take that stream and produce the correct metadata still.

Mik-S-UMS:
Someone on the [forum](https://www.universalmediaserver.com/forum/viewtopic.php?p=50329#p50329) mentioned that WebOS 2025 now supports DV in MKV so in theory these will no longer need to be remuxed, at least for LG TVs. 

@SubJunk do you have this update yet? 
Apparently their media player has a memory leak that causes it to quit but don't have any definitive info on this.

SubJunk:
Wow, that will be amazing if it's true. I submitted a request to LG about 3 years ago asking for it and they said maybe one day 😆 I will try it

SubJunk:
They disappeared :(

SubJunk:
My TV is only on WebOS 2023 so I can't check

Mik-S-UMS:
I think you can manually update but at the moment it is probably better to not update as there are other problems with it.

That user did post a link and I did a bit of further research in this and updated the post on the forum with links to what I found.

SubJunk:
I'm very interested to follow this, it will be a game changer if they fix the problems now

SubJunk:
I don't think I will work on this anymore because webOS added that DV support. It could technically still be useful but I think probably any TV that supports DV also supports almost all common subtitles

PR #4571, opened 2024-04-09
FFmpeg transcoder streams softsubs in MP4

SubJunk:
This means we don't have to hardcode the subtitles which saves a lot of initial processing by FFmpeg

Also muxes Dolby Vision without tsMuxeR


SubJunk:
@valib thanks for the help! I was programming without autocomplete or anything 😆

SubJunk:
This is working for me on my TV. I need to do some more manual testing to make sure the subtitles are synchronized properly with the video and audio

PR #4602, opened 2024-04-22
do range check to prevent exception

ik666:


SubJunk:
@ik666 it failed on the linter

PR #4638, opened 2024-05-08
allow concurrent MediaInfoParser parsing

SurfaceS:
Allow concurrent `MediaInfo` parsing.
Keep `MediaInfoParser` static, properly close `MediaInfo` handles, don't mess tests.

for @ik666


SurfaceS:
included in v14.1

PR #4640, opened 2024-05-08
allow concurrent MediaStore retrieval

SurfaceS:
allow concurrent MediaStore retrieval.

Removed synchronized on members.
Added synchronized on ids.


SurfaceS:
included in v14.1

PR #4652, opened 2024-05-14
fix #4650

### LINKED ISSUE #4650: V14 - Cannot open DLNA on TV it crashes.
Hi, 

i updated the UMS from V13 to V14 and after the update i found two issues. (Tried it with clean install its the same)
First when the TV finds the UMS after selecting it in the TV the TV responds with "that the page cannot be displayed" and the server generates an error i included the logs.
Second problem is when i launch the server as service all folder names are wrong and the half of the names are missing. Maybe its a language problem see the shared.conf in the zip. When the server is launched without the service then the problem is not there. 
Its a Samsung TV UE32H6200.
Thanks.

SurfaceS:
fix #4650


SurfaceS:
@SubJunk 

do you need more info about this ?

SubJunk:
No this looks good

PR #4656, opened 2024-05-15
CDS UpdateObject

ik666:
- Implemented CDS UpdateObject base functionality
- Added CRUD operations for `upnp:rating` attribute
- Moved StarRating from internal API to UpdateObject implementation


SurfaceS:
@ik666 nice to move StarRating to CDS.

I reviewed only `UmsContentDirectoryService` part, as it can be simplified a lot quickly.

For the other part, I may have a look later.
I don't know if you can use the `Property` class from `net.pms.network.mediaserver.jupnp.support.contentdirectory.result` to avoid using a `NodeList` as result.

I think this can be improved later.

ik666:
@SurfaceS Actually, I was quite surprised to see how complex `UpdateObject` is defined in the spec. It has single attribute updates and multiple attribute updates in combination with empty/NULL strings. Input values are XML fragments, but not always well formed. Really annoying ... maybe the code got more complex as needed, but it survived some test cases I've written.

Regarding `NodeList`: It's not used as a "result" which is delivered back to the caller, but used internal by the Handler. Since `currentTagValue` and `newTagValue` supplied by the control point, can have more then one node the `list` implementation is used instead of `Node`. Definitely, the key value pairs could be copies to another list structure, yet I'm unsure if there is a benefit for a developer? Agree, to check back later ...

ik666:
As a side note : This PR only updates `AudioItem` objects. This could also be extended to video and image.

SurfaceS:
@ik666 I made a PR on your git.
It fix few things, typo, move to correct directories.

I added more tests to handle more "strange" input.
You'll see that now `getFragments` (old `splitCsvList`) return the expected fragments in all case (empty or existing metadata, never null or missing).

ik666:
@SurfaceS Thanks for review. Have you committed anything here? I see no update. I've pushed this PR branch to the UMS git in the path `feature/CdsUpdateObject`. We can also work there and create a new PR.

SurfaceS:
https://github.com/ik666/UniversalMediaServer/pulls

@ik666  here

ik666:
@SurfaceS OMG, got it ... the pull request is in my repo :)

PR #4666, opened 2024-05-16
main fix #4653

### LINKED ISSUE #4653: v14.0 release.  Issue with root directory drive scan
I am getting these errors in the logfile when UMS scans the disks for content.
 ERROR A plugin has attempted to add a null child to "root"

Hence, the directory scan only includes films under a directory "to be seen" and not the other three disks full of films.
This has never happened before with any other version.  It has only now occurred after upgrading to the v14.0 sent from the Patreon site.
On the three disks in question, all film directories are in the root directory and not in a subdirectory.

I hope someone can help me with this.  Thanks!!

[ums_dbg_2024-05-14-09-23.zip](https://github.com/UniversalMediaServer/UniversalMediaServer/files/15308987/ums_dbg_2024-05-14-09-23.zip)

SurfaceS:
@SubJunk 
I added more log.

I'll need a browse to understand what is going on.
fix #4653


SurfaceS:
@SubJunk
[allow hidden SharedContent root.](https://github.com/UniversalMediaServer/UniversalMediaServer/pull/4666/commits/7811a7c3a6a0b8639dad92f9b8bba369672285cd)

should do the trick.

PR #4700, opened 2024-05-26
main fix #4694

### LINKED ISSUE #4694: The wrong numeration in 14's series.
An insignificant but quite important for browsing bug found in new version.
изображение_2024-05-26_074155344

SurfaceS:
`TVEPISODENUMBER` is a string that can contain multiple episode num
eg "10-16"
Take the first one and convert it to integer for sorting
fix #4694 

For future, I think that `TVEPISODENUMBER` should be renamed `TVEPISODE`, stored as integer, then we can add `TVEPISODETO` or something like that, if multi episodes are still needed.


SubJunk:
@dependabot recreate

SurfaceS:
@SubJunk 
it also fix the lint on main and merge changelog

SurfaceS:
superseded by #4712

PR #4740, opened 2024-06-08
 A better Video Coding Level implementation (#1242)

SurfaceS:
fix #1242


src/main/java/net/pms/encoders/MEncoderVideo.java
```diff
@@ -500,7 +500,7 @@ public ProcessWrapper launchTranscode(
 		} else if (isAviSynthEngine()) {
 			deferToTsmuxer = false;
 			LOGGER.trace(prependTraceReason + "we are using AviSynth.");
-		} else if (defaultVideoTrack.isH264() && renderer.isH264Level41Limited() && !isVideoWithinH264LevelLimits(defaultVideoTrack, renderer)) {
+		} else if (defaultVideoTrack.isH264() && renderer.getH264LevelLimit() < 4.2 && !isVideoWithinH264LevelLimits(defaultVideoTrack, renderer)) {
```
SubJunk, line 503: I think now we can make this check more specific by modifying `isVideoWithinH264LevelLimits` to work with the new logic, then change this line to:
```suggestion
		} else if (defaultVideoTrack.isH264() && !isVideoWithinH264LevelLimits(defaultVideoTrack, renderer)) {
```

src/main/java/net/pms/encoders/TsMuxeRVideo.java
```diff
@@ -254,7 +254,7 @@ public ProcessWrapper launchTranscode(
 			 * In reality this won't cause problems since renderers typically don't support above 4.1 anyway - nor are many
 			 * videos encoded higher than that either - but it's worth acknowledging the logic discrepancy.
 			 */
-			if (defaultVideoTrack.isH264() && renderer.isH264Level41Limited() && !isVideoWithinH264LevelLimits(defaultVideoTrack, renderer)) {
+			if (defaultVideoTrack.isH264() && renderer.getH264LevelLimit() < 4.2 && !isVideoWithinH264LevelLimits(defaultVideoTrack, renderer)) {
```
SubJunk, line 257: Same here

src/main/external-resources/renderers/Sony-PlayStation4.conf
```diff
@@ -23,7 +23,6 @@ RemoveTagsFromSRTSubtitles = false
 # PS4 generally does not support above level 4.2 and we don't have a
 # setting for that so we use the 4.1 setting to ensure compatibility
 # See https://manuals.playstation.net/document/gb/ps4/music/mp_format_m.html
-H264Level41Limited = true
```
SubJunk, line 26: I think we should set this to `H264LevelLimit = 4.2`

SubJunk (changes requested):
This is great, thanks for working on it. I commented a couple of related enhancements I think would make sense at the same time

SubJunk (approved):

PR #4752, opened 2024-06-10
Migrate eslint to flat config format

SurfaceS:
Migrate `eslint` configuration to flat config format (compatible since ESLint v8 and v9 required).
https://eslint.org/docs/latest/use/configure/migration-guide

This PR also run `eslint` before build (was not the case before).

https://github.com/UniversalMediaServer/UniversalMediaServer/pull/4749 should be able to merge without error (if we want to).


SubJunk:
This could be ok, but I thought we might wait until `typescript-eslint` is using v9 too. Otherwise we have two versions of ESLint https://github.com/typescript-eslint/typescript-eslint/blob/main/package.json#L88
What do you think?

SurfaceS:
On the yarn file, I see that `eslint@npm:8.57.0` was replaced by `eslint@npm:9.4.0`.
No more ref to `eslint@npm:8.57.0`.

This PR also `eslint` before build (was not he case before).
I checked, and `eslint` warn/error as before.

SurfaceS:
I had split the PR without updating to eslint v9.

https://github.com/UniversalMediaServer/UniversalMediaServer/pull/4749 should be able to merge without error (if we want to)

SubJunk (approved):

SubJunk:
Ok cool. I don't mind if you want to bump ESLint to v9, I don't feel strongly about it

PR #4755, opened 2024-06-11
Migrate eslint configuration format

SurfaceS:
Migrate `eslint` configuration to flat config format (compatible since ESLint v8).

https://eslint.org/docs/latest/use/configure/migration-guide


SurfaceS:
superseed

PR #4768, opened 2024-06-17
lookup existing storeID for DBID folders

ik666:
Should fix StoreContainer message `Node (xxx) already has an ID (xxx), which is overridden now. The previous parent node was: xxx`.

Looks up StoreID from database and loads resource from mediaStore ... 

@SurfaceS I have an issue with this PR. The MediaStore dosen't deliver the resource which was found before by searching the MediaStore `MediaStoreIds.getMediaStoreIdsForName` in line 105. Even tough the ID was found in line 119: 

`return (VirtualFolderDbId) renderer.getMediaStore().getResource(Long.toString(ids.get(0)));` delivers NULL.

I think the MediaLibrary->Audio folder doesn't discover any childs (from DB) in the method `recreateResource` (line 522 and following). Any idea?

### DIFF
--- a/src/main/java/net/pms/store/DbIdLibrary.java
+++ b/src/main/java/net/pms/store/DbIdLibrary.java
@@ -101,45 +101,64 @@ public class DbIdLibrary {
 		}
 	}
 
-	private void setAudioLikesFolder(List<StoreResource> backupChildren) {
-		if (audioLikesFolder == null) {
-			audioLikesFolder = new VirtualFolderDbId(renderer, "MyAlbums", new DbIdTypeAndIdent(DbIdMediaType.TYPE_MYMUSIC_ALBUM, null));
-		}
-		if (PMS.getConfiguration().displayAudioLikesInRootFolder()) {
-			if (backupChildren.contains(audioLikesFolder)) {
-				renderer.getMediaStore().addChildInternal(audioLikesFolder, false);
-			} else {
-				renderer.getMediaStore().addChild(audioLikesFolder);
+	private VirtualFolderDbId lookupDbIdFolder(DbIdMediaType typeRootFolder) {
+		List<Long> ids = MediaStoreIds.getMediaStoreIdsForName(typeRootFolder.toString());
+		if (ids.size() > 0) {
+			LOGGER.debug("found audio folder with ID " + ids.get(0));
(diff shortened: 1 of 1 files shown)


SurfaceS:
The new audio PR will not use DBID folders.

ik666:
@SurfaceS  It's actually not related to DBID folder ... the MediaLibrary -> Audio Folder stays empty. The issue seem to be missing/no code to fill up children for the Audio folder (in the reverse lookup from an item up to the root).

SubJunk:
I see this has the `not ready` label, is it still active or should we close it?

ik666:
@SubJunk It's actually not ready because of an other related issue mentioned above: MediaStore's `getResource` or to be precise `recreateResource` doesn't build the tree up to the "Audio" MediaLibraryFolder, ending one stop to early, running in a NPE ... This PR would fix the error `Node (xxx) already has an ID (xxx), which is overridden now`, but I think it has no priority ... listening or watching videos is not effected ... we can close it now, or wait until the mentioned PR arrives?

SubJunk:
@ik666 I see, you can decide whether to keep it open, I was just checking :)

SubJunk:
Is this waiting for me to review?

ik666:
yes, it has been working for a while on my personal branch.

SubJunk:
I don't fully understand it but that's ok, it has been waiting for ages and you have tested it

SubJunk (approved):

src/main/java/net/pms/store/DbIdLibrary.java
```diff
@@ -101,45 +101,64 @@ private void addChildToMediaLibraryAudioFolder(VirtualFolderDbId dbIdFolder) {
 		}
 	}
 
+	private VirtualFolderDbId lookupDbIdFolder(DbIdMediaType typeRootFolder) {
+		List<Long> ids = MediaStoreIds.getMediaStoreIdsForName(typeRootFolder.toString());
+		if (ids.size() > 0) {
+			LOGGER.debug("found audio folder with ID " + ids.get(0));
```
SubJunk, line 107:
```suggestion
			LOGGER.debug("found audio folder with ID {}", ids.get(0));
```

SubJunk (approved):

PR #4779, opened 2024-06-23
Remove OpenSubtitles

### LINKED ISSUE #4778: Stop calculating OpenSubtitles hashes
These are no longer used by the API so we do not need them. Some operations should get a bit faster when it is removed

SurfaceS:
fix #4778


SubJunk:
@SurfaceS is this ready or is there more to do?

SurfaceS:
Yes, this is ready

PR #4789, opened 2024-06-30
Bump MediaInfo to 24.06

SubJunk:
# Description of code changes

...

# Checklist


SurfaceS:
From the MediaInfo issues, a reply to a user is :
> In your file, the SMPTE ST 2094 App 4 profile is compatible with HDR10+ Profile B but colour primaries and transfer characteristics are not (they are both BT.709 instead of BT.2020 & PQ) + no mastering display color primaries, [which are mandatory for HDR10+](https://en.wikipedia.org/wiki/HDR10%2B#HDR10+_content_profile) so we decided to remove this tip because feedback from users was that it was misleading, they were thinking that the file is compatible with HDR10+.


From MediaInfo 24.06 :
`
"extra":{"ConformanceErrors":[{"HEVC":[{"GeneralCompliance":"Bitstream parsing ran out of data to read before the end of the syntax was reached, most probably the bitstream is malformed (frame 0+79, time 00:00:00.071+00:00:01.506, offset 0x1FE9+0x3D84DB)"}]}`

seems that the hdr10 test file is not well formed, that is why not does not advert hdr10 things.

SurfaceS:
fix HDR10 test file
fix MPEGTS duration
(I2087, MPEG-TS: general duration includes before and after PCR offsets)

PR #4804, opened 2024-07-10
main fix #4803

### LINKED ISSUE #4803: Web GUI on 14.1.0 with Linux not working
A few users on the [forum](https://www.universalmediaserver.com/forum/viewtopic.php?t=16043) have found that the web GUI is not working in 14.1.0 on Linux.

I had a look at their logs and did not see any errors but my best guess is something to do with the switch to Jetty.

**Boss** has done a bit of experimenting and found that it could be the react index.html

> The problem is: ums-14.1.0/web/react-client/index.html.
If I copy ums-14.0.2/web/react-client/index.html
to
ums-14.1.0/web/react-client/index.html
It all works fine. 

I am guessing this is just restoring Netty so something about Jetty is not working right on Linux, assuming it was this version that made the switch.

SurfaceS:
fix #4803 
Web GUI on 14.1.0 with Linux not working

### DIFF
--- a/react-client/vite.config.ts
+++ b/react-client/vite.config.ts
@@ -12,6 +12,9 @@ export default defineConfig({
     sourcemap: true,
     chunkSizeWarningLimit: 1600,
     rollupOptions: {
+      output: {
+          hashCharacters: 'hex'
+      },
       onwarn(warning, defaultHandler) {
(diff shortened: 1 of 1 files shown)


SubJunk:
Thanks, I will do a new release now

PR #4886, opened 2024-08-08
3box setup / authorization / resource access error

ik666:
I've an issue requesting thumbnails in a UPnP 3 box setup. 

If the accessing control point is a web application, physical IP-access to the resources like thumbnails is done by the users web browser. If the web application and the web browser do not run on the same machine, no renderer can be associated and downloading thumbnails (or the underlying resource) fails with a 413 error.

This PR: In case authentication is disabled, an auth check is not really needed.

### DIFF
--- a/src/main/java/net/pms/network/mediaserver/servlets/MediaServerServlet.java
+++ b/src/main/java/net/pms/network/mediaserver/servlets/MediaServerServlet.java
@@ -117,7 +117,9 @@ public class MediaServerServlet extends MediaServerHttpServlet {
 					 * fixme : non-renderer should advise a special uuid.
 					 * renderer = null;
 					 */
+				}
 
+				if (renderer == null) {
 					/**
 					 * In in a 3-box setup: If the accessing control point is a web application, physical IP-access to the resources
 					 * like thumbnails is done by the users web browser. If the web application and the web browser do not run on
--- a/src/main/java/net/pms/network/mediaserver/servlets/MediaServerServlet.java
+++ b/src/main/java/net/pms/network/mediaserver/servlets/MediaServerServlet.java
@@ -31,6 +31,8 @@ import java.util.ArrayList;
 import java.util.Date;
 import java.util.Locale;
 import java.util.TimeZone;
+import net.pms.PMS;
+import net.pms.configuration.RendererConfigurations;
 import net.pms.dlna.DLNAImageInputStream;
 import net.pms.dlna.DLNAImageProfile;
 import net.pms.dlna.DLNAThumbnailInputStream;
@@ -115,6 +117,16 @@ public class MediaServerServlet extends MediaServerHttpServlet {
 					 * fixme : non-renderer should advise a special uuid.
 					 * renderer = null;
 					 */
+
+					/**
+					 * In in a 3-box setup: If the accessing control point is a web application, physical IP-access to the resources
+					 * like thumbnails is done by the users web browser. If the web application and the web browser do not run on
+					 * the same machine, no renderer can be associated. If authentication is not enabled, just use the
+					 * default renderer.
+					 */
+					if (!PMS.getConfiguration().isAuthenticationEnabled()) {
+						renderer = RendererConfigurations.getDefaultRenderer();
+					}
 				}
 			}
 


SurfaceS:
We already discuss about this.

When UMS discover a renderer, it advertise content adapted to it (renderer capability, rights, etc).

If the control point is not the renderer itself (3th party), this control point is responsible of presenting/adapting (or not) the right media uri to the renderer.

So UMS should keep the advertised renderer when the final renderer is not the control point, and not try to adapt content to it.

If I remember well, you had a problem with a control point that send the url to a renderer, then leave the network (UMS will destroy this renderer, so you are not able to access the advertised uri).

On this PR, you "only" set to default renderer when the renderer is unkown.
But the advertised id may not be available for default renderer, and will not be the advertised DLNA profile.

I don't think this is the right way to handle this.
We should find a way to create an uri with some kind of unique id that contain the final asked media, format, user, stream, etc.
Then when you access this uri, you will always get what is advertised.

SubJunk:
I agree with the comment, it seems like this would introduce a different set of problems

ik666:
Yeah, to bad we have always the same issue. UMS is not really aware of a 3 box setup at the moment, thinking the control point is a renderer. Actually, in this particular case, the control point is not able to download/display the thumbnail URL given by a CDS browse request. No renderer are involved  ... 

However, If we add the control point uuid to the renderer map, the issue also seems to be gone ... maybe this suits better ...

sf666:
@SurfaceS This PR is changed: We should at least add the device uuid to the list of known renderer, until a better solution is found.

ik666:
> I agree with the comment, it seems like this would introduce a different set of problems

Ah, is it so? I think dropping/breaking UPnP AV client support with v14 is the bigger problem.

SubJunk:
@ik666 can you explain how v14 broke UPnP AV client support?

ik666:
@SubJunk I thought I made it clear. Sorry for not being able to explain the issue so everyone can understand it. 

Let's say we have the following setup:

A control point (BubbleUPnP) identifies a MediaServer (UMS) and a Renderer (LG TV).

1) The control point is browsing the MediaServer and gets a resource URL's for a movie. If you look at the resource URL, you'll see, that the UUID of the control point it part of the path in case it could be determined, else NULL will be in place instead. 

2) The control points sends the resource URL of the movie to the Renderer for playing. (This is the AV client interaction). The movie will not play.

Why is it so? The MediaServer (UMS) checks if the UUID of the given URL (from the control point) matches with the Renderer UUID. The control point and the renderer don't share the same UUID and therefore the movie will not play. This is of course complete nonsense and violates the UPnP Device Architecture.

Most likely [this forum topic is](https://www.universalmediaserver.com/forum/viewtopic.php?t=16197) related to it.

Besides that :
- UMS identifies the wrong renderer in Step 2, because the renderer will be chosen by the given UUID (with is the control point and not the renderer).
- Actually, UMS is not able to read the UUID of the control point in step 1, because if you look at the `UmsContentDirectoryService.browse` implementation, only the IP and the http headers are available during browsing. That's why UMS tries to map the given IP address to a renderer it has discovers before. If you have multiple clients on the same IP address, this is of course a complete mess.  => That's why the current implementation works only in a 2-box setup (control point and renderer share the same IP address -> share the same UUID over the IP mapping), but fails with multiple [clients on the same IP (forum)](https://www.universalmediaserver.com/forum/viewtopic.php?t=16545).

SubJunk:
@ik666 sorry I missed this reply somehow. Thanks for explaining, I see the problem. Will you fix it?

ik666:
@SubJunk The main blocking point is the IP - UUID association. I think this association was introduced to support the "authorization" / "account" logic for specific devices (TV's in a 2 box setup), but the solution wasn't thought to the end, lacks of teamwork and UPnP architectural understanding, leading us to this situation we have now. 

Therefore, making UMS UPnP compliant (3-box setup ready) again would also mean to remove all UUID logic - including "new features introduced in V14"  -> authorization for devices identified by UUID (authorization by IP address is not effected). Reading this thread - where you've obviously difficulties merging 2 lines of workaround code - I'm not sure if you're willing to go this way, yet.

It's not that hard to "fix" it on the server side. I've already done the basic steps (removed all authorization, because I don't need it -> IP based access control has to be put back in business) on the forked UMS repo in my account (`main_home` branch), but some alignments need to be done in the UI, which must be done by someone else (you).

SubJunk:
@ik666 I think the authorization and account features are important to keep because our users were requesting it for a long time.

Maybe we can make a renderer config setting to define devices as control points, and then they get a different account/auth logic?

I think UMS and its predecessor PMS were not designed for this situation from the start, hence the configs being called "renderers". Maybe we need another folder called "control-points" that has a different kind of config?

> The MediaServer (UMS) checks if the UUID of the given URL (from the control point) matches with the Renderer UUID. The control point and the renderer don't share the same UUID and therefore the movie will not play. This is of course complete nonsense and violates the UPnP Device Architecture.

To make sure I understand, in this case the renderer is able to play content but not browse it? So it needs BubbleUPnP to push content to it?

ik666:
@SubJunk The authorization and account features can stay in place, but only for IP based "identification" (like i.e. serviio does it). The UUID binding makes the 3-box setup to break up.

Therefore, my suggestion is to remove UUID-based authorization / enforcement logic. In a 3-box setup, accounts are assigned separately to control points (by IP) and renderers (also by IP). This means a control point may be able to browse all content, but playback could still be restricted if the target renderer has a different account.

For example, a control point with full access might not be able to play a movie on a kids’ TV, since the TV is tied to a limited account.

> Maybe we can make a renderer config setting to define devices as control points, 

Yes, that makes sense. We only need a single ControlPoint “renderer” that covers all control points (TVs, BubbleUPnP, VLC, etc.). This special entry would accept all media types without any transcoding, since control points is the only device that interacts with the UMS content directory.

When the actual resource is requested by a renderer, UMS should then resolve the correct renderer.conf.

> and then they get a different account/auth logic?
We can keep using the existing IP-based auth logic. The only issue right now is that control point devices don’t show up, since they’re being ignored or associated with renderer by UUID - IP mapping (and are therefore hidden). 

> I think UMS and its predecessor PMS were not designed for this situation from the start, hence the configs being called "renderers". Maybe we need another folder called "control-points" that has a different kind of config?

Control points don’t actually need any configuration from UMS’s side. As mentioned above they only browse the content directory and don’t require special handling.

ik666:
> To make sure I understand, in this case the renderer is able to play content but not browse it? So it needs BubbleUPnP to push content to it?

Basically yes. 

In regards to UPnP device architecture:
We need to distinguish between renderer and control points. The **renderer device** has not the ability to browse content. The renderer has only the ability to play resources.
The **control point device** is only browsing content and has not the ability to render content.

If a hardware includes a **control point device** and a **renderer device** like TV's do, we have a 2-box setup. And only in that setup the current authorization works without any issues.

ik666:
@SubJunk  > sorry I missed this reply somehow. Thanks for explaining, I see the problem. Will you fix it?

I had a look at it, but with the way authorization currently works, this issue can’t be fixed in a general way. I'll think about a workaround, but then authorization has to be disabled by the user.

PR #4922, opened 2024-08-24
Fix video quality

SubJunk:
# Description of code changes

Users have reported a big regression in video quality recently on the forum
https://www.universalmediaserver.com/forum/viewtopic.php?p=48784
https://www.universalmediaserver.com/forum/viewtopic.php?p=48788

The problem is mostly that the PS3 config says it supports H.264 inside MPEG-PS, but when I tested it this was not true. There were other bugs that intersected to cause this problem, which is that when there is a video stream that is compatible with PS3, UMS was incorrectly transcoding the stream, but also not applying the video quality parameters, so the transcoding happened with the default FFmpeg quality for MPEG-2 video which isn't very good

I fixed other bugs along the way:
- [x] Changed PS3 config to support AAC-LC inside MPEG-PS and MPEG-TS, I have tested it with 2 and 5.1 channel AAC-LC
- [x] Fixed the checks in MEncoder and FFmpeg engines to defer to tsMuxeR - they were checking compatibility in MPEG-PS but it should be MPEG-TS
- [x] Fixed support for remuxing supported video streams when the goal is MPEG-2 video
- [x] Logging and formatting improvements

# Checklist
- [x] Any relevant issues are [referenced](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#referencing-issues-and-pull-requests)
- [x] Any relevant [Knowledge Base](https://github.com/UniversalMediaServer/knowledge-base) articles are written/modified

### DIFF
--- a/src/main/external-resources/renderers/Sony-PlayStation3.conf
+++ b/src/main/external-resources/renderers/Sony-PlayStation3.conf
@@ -49,7 +49,7 @@ Supported = f:wma                    n:2             m:audio/x-ms-wma
 
 # ============================================================================
 #
-# Taken from: http://manuals.playstation.net/document/en/ps3/current/video/filetypes.html
+# Taken from: https://manuals.playstation.net/document/en/ps3/current/video/filetypes.html
 # Note: Some things on the following list have been experimentally disproven,
 # hence why our settings are sometimes different.
 #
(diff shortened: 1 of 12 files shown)


SubJunk:
This has been confirmed as working on the forum

PR #4947, opened 2024-09-02
fix #4884 and #4937

### LINKED ISSUE #4884: Sony XBR-49X850B  Content could not be acquired
# How to contribute

In order to help diagnose and fix your issue we often need your "debug information". It's easy to save this information and attach it to a GitHub issue:

Hi
Files are mp4 on a MacBook pro.  Using Safari.  I have a log zipped which is attached.  please help.  Does the number of files have anything to do with this problem?
[ums_dbg_2024-07-31-15-07.zip](https://github.com/user-attachments/files/16533225/ums_dbg_2024-07-31-15-07.zip)

* Start UMS.
  * Go to the `Logs` tab.
  * Press the `Create TRACE logs` button which restarts UMS in the "TRACE" mode. If that doesn't work for some reason, either select `Trace` from the `Log Level` dropdown at the bottom or set `log_level = TRACE` in the configuration file and restart UMS manually.
  * Reproduce the problem.
  * Click `Pack debug files` on the lower left.
  * Click `Zip selected files`.
  * Save the zip file to a location you will remember.
  * Attach the zip file to the GitHub issue.

Additionally, if the problem happened in the browser, it can be useful to know about any browser console errors, so please paste them into the issue too.

### LINKED ISSUE #4937: A to Z limit support broken
This is a comment from Patreon:
Guys, the server is not detecting my shared drived. This problem began with 14.4.0, and I just simply went back to 14.3.0 and everything worked fine again. Now the same issue is with 14.5.0, and it will not see my shared drives, so the available items is 0. This was one of the issues I reported which resulted in corrections before.
The server somes up, however when it attempts to scan the shared drives, it sees nothing, cleans up the database and then just sits there. I am running Win10 22H2 and 3 of the 4 drives are shared at root level. One drive has a folder that is shared. I hope we can get this thing back to normal. I have again gone back to 14.3.0 and everything is fine. Thanks!
Any ideas what could have broken this @SurfaceS @ik666 ?

SurfaceS:
fix #4884
fix #4937

`MediaTableMusicBrainzReleases` (`MUSIC_BRAINZ_RELEASES`) update was crashing.
1st, it prevent the media db to update/create other tables.
2nd, the db status is then `DatabaseStatus.CLOSED`.
This bug is their from an update on V14 that add ARTIST_ID, GENRE, and so on.

So, the user do not have db update/creation that sit after `MediaTableMusicBrainzReleases` update since then.
Now that we use the db to assign UPnP id, UMS was parsing well, but was unable to give an Id for UPnP.

In short, it has nothing to do with Shared drive support (#4937), neither with Sony XBR-49X850B Content (#4884).
That's only a side effect.

On this PR :
Fix `MediaTableMusicBrainzReleases` update.
Rename indexes to add table name like we do everywere. (eg: `ARTIST_IDX` may be movie artist, and so on, so it should be renamed `MUSIC_BRAINZ_RELEASES_ARTIST_IDX`)
Clean `MediaTableMusicBrainzReleases`.

### DIFF
--- a/src/main/java/net/pms/database/MediaTableMusicBrainzReleases.java
+++ b/src/main/java/net/pms/database/MediaTableMusicBrainzReleases.java
@@ -31,14 +31,16 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * This class is responsible for managing the MusicBrainz releases table. It
- * does everything from creating, checking and upgrading the table to
+ * This class is responsible for managing the MusicBrainz releases table.
+ *
+ * It does everything from creating, checking and upgrading the table to
  * performing lookups, updates and inserts. All operations involving this table
  * shall be done with this class.
  *
  * @author Nadahar
  */
 public final class MediaTableMusicBrainzReleases extends MediaTable {
+
 	private static final Logger LOGGER = LoggerFactory.getLogger(MediaTableMusicBrainzReleases.class);
 	public static final String TABLE_NAME = "MUSIC_BRAINZ_RELEASES";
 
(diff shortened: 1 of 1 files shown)


SurfaceS:
### Developer advice :

- When we update the db, always use the `IF_EXISTS` or `IF_NOT_EXISTS` on `CREATE`, `ALTER`, `DROP` statement.
- When naming indexes, name should look like `TABLE_NAME + CONSTRAINT_SEPARATOR + COL_1 + CONSTRAINT_SEPARATOR + COL_2 + IDX_MARKER` that for example, end to `FILE_ID_MODIFIED_IDX`. It is then easier to drop it or change it. And it avoid collision between tables.

SubJunk (approved):

PR #4951, opened 2024-09-03
mvn frontend error with configured maven proxy settings

ik666:
There is an issue with the maven-frontend-plugin in combination with configured proxy settings in the maven `settings.xml` file. The plugin passes an argument that appearently is not supported by the used yarn 4 installation for the react client. 

Disabling forwarding the proxy setting to the yarn client, solves this issue.


SubJunk (approved):

SubJunk:
Nice one

PR #4964, opened 2024-09-10
Added missing StoreIds indicies

ik666:
Speeding up search requests. MediaTable methods used by search request were not using any indices.


ik666:
mixed PR ... closing

PR #4966, opened 2024-09-10
Speed up Resource creation

ik666:
Speed up Resource creation. Added missing indices for reading video metadata. This PR makes a huge difference on my audio library, even though there aren't any video files (factor 3-5).


SubJunk:
Looks like a great idea, but tests failed

ik666:
> Looks like a great idea, but tests failed

Maybe, we need tests for the tests ...  :)

SubJunk (approved):

PR #4968, opened 2024-09-11
fix HTTP 400 bad uri

ik666:
UMS generates some URI's which are not processed by Jetty like `/ums/thumbnail/419b4fa5-c752-f1f0-0000-000005b0528a/77463/JPEG_SM_Rhye%2FLewisOfMan.jpg` because they are ambiguous. This PR activates Jetty LEGACY support, so thumbs are transferred ... 

Alternatively, the URI has to be unambiguous (not realized in this PR).

### DIFF
--- a/src/main/java/net/pms/network/mediaserver/jupnp/transport/impl/jetty/ee10/JettyServletContainer.java
+++ b/src/main/java/net/pms/network/mediaserver/jupnp/transport/impl/jetty/ee10/JettyServletContainer.java
@@ -24,6 +24,7 @@ import net.pms.network.mediaserver.servlets.MediaServerServlet;
 import net.pms.network.mediaserver.servlets.NextcpApiServlet;
 import org.eclipse.jetty.ee10.servlet.ServletContextHandler;
 import org.eclipse.jetty.ee10.servlet.ServletHolder;
+import org.eclipse.jetty.http.UriCompliance;
 import org.eclipse.jetty.http2.server.HTTP2CServerConnectionFactory;
 import org.eclipse.jetty.server.HttpConnectionFactory;
 import org.eclipse.jetty.server.Server;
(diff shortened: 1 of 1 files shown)


SubJunk:
What does ambiguous mean in this context?

ik666:
> What does ambiguous mean in this context?

From Jetty's point of view it's documented here: https://jetty.org/docs/jetty/12/programming-guide/server/compliance.html#uri 

I believe the encoding of the path separator "/" as %2F stops Jetty from delivering the resource.

valib:
@ik666 what is the original name of the file  stored in your PC? To have the "/" inside the file name is strange and we need to know if it is the UMS case or something else.

ik666:
It's not a file. I was searching for an Artist. Probably, same applies for search results of type album ...

And the Artist has a "/" ... 

search

valib:
@ik666 seems like wrong path or name for the virtual folder.

ik666:
@valib Yeah. The thumbnail actually gets delivered even when you cut the part after the objectID like this `/ums/thumbnail/419b4fa5-c752-f1f0-0000-000005b0528a/77463/`

Generating thumb url could therefore be done without encoding the "title" attribute and we wouldn't run into the HTTP 400 error. I think we have two places where the DIDL result is generated ...

SubJunk:
Is there a reason we need that encoded in the name? Could it just be another character like -?

ik666:
I think the special character "/" or all characters that will result in an "%" encoding can be just filtered out.

SurfaceS:
> Alternatively, the URI has to be unambiguous (not realized in this PR).

> Is there a reason we need that encoded in the name? Could it just be another character like -?

I think we should not break UriCompliance rules that follow the HTTP / URI / Servlet specs for illegal & suspicious characters.
Alternative (unambiguous URI) is better.

see #4993.

PR #4969, opened 2024-09-13
fixed recursive delete of files -> containerFiles -> files and so on

ik666:
There is a delete loop, ending in a StackOverflowError ... 

`deleteEntry` is only called from `MediaTableFiles `.


SubJunk (approved):

SubJunk:
Cool, I saw this error recently

PR #4032, opened 2023-08-19
Fix seeking during transcoding on LG TVs

SubJunk:
I'm trying to be able to seek in transcoded videos on my LG TV, and these changes almost work, but the audio gets far behind the video after seeking.

This does already have the advantage of letting the TV resume playback on transcoded files, because the TV remembers the right request to make in bytes

Does anyone have an idea about why that would happen? @UniversalMediaServer/developers

PR #4970, opened 2024-09-13
Fix seeking in transcoded videos

SubJunk:
# Description of code changes

V2 of https://github.com/UniversalMediaServer/UniversalMediaServer/pull/4032
Some media players, including LG TVs, send BOTH byte-ranges and timeseek ranges, and this was breaking seeking for transcoding because we had logic to prefer the byteseek in all cases. Now we prefer the timeseek for transcoded videos, only if we have defined the renderer as seek by time exclusive

It is working as it is for most files, but for some reason when Dolby Vision files are seeked, there is a black screen and the audio plays. The TV does not recognize that as a problem, it just seems to think it is receiving a black video. Maybe a keyframe thing? I am experimenting with different settings to try to resolve that. If any @UniversalMediaServer/developers have ideas let me know

Edit: The problem is more broad than Dolby Vision files, the problem is with tsMuxeR output with this config. I'm doing a lot of testing to figure out why and will put results here.

Edit 2: tsMuxeR output is played properly by the TV with `-ss 0` but when seeking is used, the TV only accepts the audio. So I think it is a keyframe problem and I'm testing that

## Testing results for tsMuxeR

Audio works but video is out of sync after taking a LONG time to display anything with the following:

```
Video track:
[ffmpeg.exe-81] Starting ./ffmpeg -ss 877.0 -i video.mkv -c copy -copyts -start_at_zero -f hevc -y -y -y \\.\pipe\1726368429812ffmpegvideo
OR
[ffmpeg.exe-81] Starting ./ffmpeg -ss 877.0 -i video.mkv -c copy -copyts -f hevc -y -y -y \\.\pipe\1726368429812ffmpegvideo

Audio track:
[ffmpeg.exe-82] Starting ./ffmpeg -ss 877.0 -i video.mkv -ac 6 -f ac3 -c:a ac3 -ab 448k -y \\.\pipe\1726368429812ffmpegaudio01

tsMuxeR options:
[tsMuxeR.exe-83] Starting ./tsmuxer ums-tsmuxer.meta \\.\pipe\1726368429813tsmuxerout.ts
with contents:
MUXOPT --no-pcr-on-video-pid --new-audio-pes --no-asyncio --vbr --vbv-len=500
V_MPEGH/ISO/HEVC, "\\.\pipe\1726368460900videoout", track=1
A_AC3, "\\.\pipe\1726368460901audioout", track=2
```


SubJunk:
I made a support ticket for FFmpeg to add support for adding DV metadata to MPEG-TS, because then we won't need tsMuxeR for this anymore
https://trac.ffmpeg.org/ticket/11193

SubJunk:
@dependabot recreate

SubJunk:
@SurfaceS this is working for me now, but only with this hack https://github.com/UniversalMediaServer/UniversalMediaServer/pull/4970/files#diff-ba7a7eef129f4ceba66a325e08707372dc35544a5db48afc9277c28fe16c1dbbR1069 to stop tsMuxeR being used during seeking. Do you have any ideas about why it would not work? The TV accepts output from FFmpeg while seeking, but not from tsMuxeR.

SubJunk:
I will try to go back in time and test whether tsMuxeR seeking has ever worked for devices I have. I think I remember it working at some point, but right now with LG and VLC it does not work

SubJunk:
@UniversalMediaServer/developers ok I think tsMuxeR engine does not support seeking. I have reproduced on the VLC app on macOS now, and my phone, and my LG TV using LG media player. So I will call that a general bug that has always existed, and make a new issue for it. In the meantime this PR is working great on my TV, please review it

SurfaceS:
@SubJunk 
I should yes. I receive a lot of [@dependabot](https://github.com/dependabot) under your name, maybe it was on this batch.

I can't test this on LG, I don't have any, but the other part is working fine on my setup.

SurfaceS (dismissed):

src/main/java/net/pms/encoders/FFMpegVideo.java
```diff
@@ -1066,6 +1066,9 @@ public synchronized ProcessWrapper launchTranscode(
 			} else if (!EngineFactory.isEngineAvailable(StandardEngineId.TSMUXER_VIDEO)) {
 				deferToTsmuxer = false;
 				LOGGER.warn(prependTraceReason + "the configured executable isn't available.");
+			} else if (params.getTimeSeek() > 0) {
+				deferToTsmuxer = false;
+				LOGGER.debug(prependTraceReason + "the renderer will display a blank screen, no good explanation for this yet.");
```
SubJunk, line 1071:
```suggestion
				// this condition can be removed when https://github.com/UniversalMediaServer/UniversalMediaServer/issues/5113 is fixed
				LOGGER.debug(prependTraceReason + "the renderer will display a blank screen, no good explanation for this yet.");
```

PR #4998, opened 2024-09-25
main fix #2664

### LINKED ISSUE #2664: Update folder contents when thumbnail is updated
I realised that my comment on #2599 might have been forgotten being on a closed issue

I found that on PS4 that it will not reload a folder until the contents of that folder has changed, that means if a file is marked as played the PS4 does not see any change and does not display the updated thumbnail

A solution would be to say that the contents has changed when the thumbnail is updated

SurfaceS:
main fix #2664

added query on updateid to thumbs url.
That should force the control point to download the refreshed thumb.


SubJunk:
It was a good attempt but didn't work

SurfaceS:
Was missing old didl lite engine changes.
Should be good now

SurfaceS:
After a long review, that should be OK now.

- added `query` on updateid to thumbs url -> That should force the control point to download the refreshed thumb.
That fix the react gui as well.

- log `query` string on trace log (more info on url now).

- implements `etag` and `If-None-Match` -> that allow server to reply `NOT_MODIFIED` if the object was not updated.

- implements `Cache-Control` header with `max-age=86400` (1 day) for `http/1.1+` instead of `Expire` with 116 days (still use for HTTP/1.0).

- implements option for **AlbumArtURI UPnP hack**.
All TV conf use it for now, but it is an option now so it can be not used.

- implements option for **versioned UPnP ObjectId hack**.
Fix for Samsung TV that does not refresh UPnP Objects but store them permanent with the first seen data.

- implements option to **force AlbumArtURI UPnP Profile**.
This allow a renderer to only present one AlbumArtURI UPnP Profile.

- Samsung TV 2019+ confs updated.

SurfaceS:
@SubJunk this one is good to go

SubJunk (changes requested):
Wow that's a lot of investigation done. I will test it on my LG TV since its thumbnails get stuck too. Can you please also add the new renderer config options into DefaultRenderer.conf?

SubJunk:
I'm still testing on LG but this was confirmed to work by the user so I'm merging it now.

There is still a potential gotcha for the future, which is that people who make new Samsung TV configs might forget or not realize they should add these settings, and encounter the problem again, so I wonder if a more hardcoded solution could be better

SubJunk:
The config flags don't seem to make a difference to my TV, which is ok, it's good it is fixed on Samsung

PR #5076, opened 2024-10-28
Implements UPnP log levels

SurfaceS:
- Implements UPnP log levels (avoid spamming / big log file) for readability.
It is quite similar that what we get before upgrading to JUPnP (old cling).
It is a real time setting, so we can still log debug in real time the datagrams/io UPnP.

- Log contentdirectory DOMException, fix UPnP genre.

### DIFF
--- a/src/main/java/net/pms/network/mediaserver/jupnp/support/contentdirectory/UmsContentDirectoryService.java
+++ b/src/main/java/net/pms/network/mediaserver/jupnp/support/contentdirectory/UmsContentDirectoryService.java
@@ -392,10 +392,10 @@ public class UmsContentDirectoryService {
 				checkInput(modelObjectToAdd);
 
 				StoreResource resource = null;
-				if (modelObjectToAdd.getItems().size() > 0) {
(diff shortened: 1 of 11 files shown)


src/main/java/net/pms/network/mediaserver/jupnp/support/contentdirectory/UmsContentDirectoryService.java
```diff
@@ -423,7 +423,7 @@ private void checkInput(Result modelObjectToAdd) {
 		if (modelObjectToAdd.getContainers().size() > 1) {
 			LOGGER.trace("more than 1 container ... using first found.");
 		}
-		if (modelObjectToAdd.getContainers().size() > 0 && modelObjectToAdd.getItems().size() > 0) {
+		if (modelObjectToAdd.getContainers().isEmpty() && modelObjectToAdd.getItems().isEmpty()) {
```
valib, line 426: @SurfaceS is it correct? What about to use 

    if (!modelObjectToAdd.getContainers().isEmpty() && !modelObjectToAdd.getItems().isEmpty()) {

PR #5099, opened 2024-11-07
main partial fix #5095

### LINKED ISSUE #5095: default location for media server is the users home folder
So not really a functionality issue I guess but from a security standpoint I don't think the server should default to the users home folder without explicitly creating a new user just for the media server program to run on. I mean you download the software and run it on your linux computer and it just scanned everything in your home folder and sent the metadata to the api.universalmediaserver.com website. I mean they said this is a media server that doesn't spy on you. It sure seems like it does.

SurfaceS:
better handle Linux video, audio and picture Dir as default share

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -700,6 +700,11 @@
 			<artifactId>jsoup</artifactId>
 			<version>1.18.1</version>
 		</dependency>
+		<dependency>
+			<groupId>dev.dirs</groupId>
+			<artifactId>directories</artifactId>
+			<version>26</version>
+		</dependency>
 	</dependencies>
 	<build>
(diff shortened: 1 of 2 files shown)


SubJunk:
Is this ready to review/merge? Also can you please describe the difference here for the changelog?

SurfaceS:
yes, it is ready.

It fix #5095 (partially as their is more on the issue, but maybe nor relevant).

It use the xdg-user-dir to set default shared folders (if user do give any share) on Linux.
It share user's video/music/pictures folders instead of the all user's home folder.

SubJunk (approved):

PR #5121, opened 2024-11-19
Main - new windows installer

SurfaceS:
### current installer (old) :
- become the old (maybe last) x86 installer, but still support x86-64
- fixed all users installation registry
- fixed all users installation folder on x86-64
- build to UMS-setup-x86.exe
- warn that a x64 version exists if it run under x64 system

### new windows installer (x86-64 only) :
- fixed all users installation folder
- fixed all users installation registry
- added portable installation mode
- added current user only installation mode
- added update mode
- build to UMS-setup.exe
- can't run under x86 system

### UMS :
- fix for windows all users install default shares is now all users files
- typo on conf file
- implements portable


SurfaceS:
should be ready for review

SubJunk (dismissed):
The code looks good and I tested both installers, they seem to work well

Issue #4954, opened 2024-09-07
Disabling subtitles in UMS plays audio in different language than the default. Unchecking this box works fine. 


magnetite2:
[ums_dbg_2024-09-06-19-20.zip](https://github.com/user-attachments/files/16913916/ums_dbg_2024-09-06-19-20.zip)
# How to contribute

In order to help diagnose and fix your issue we often need your "debug information". It's easy to save this information and attach it to a GitHub issue:

  * Start UMS.
  * Go to the `Logs` tab.
  * Press the `Create TRACE logs` button which restarts UMS in the "TRACE" mode. If that doesn't work for some reason, either select `Trace` from the `Log Level` dropdown at the bottom or set `log_level = TRACE` in the configuration file and restart UMS manually.
  * Reproduce the problem.
  * Click `Pack debug files` on the lower left.
  * Click `Zip selected files`.
  * Save the zip file to a location you will remember.
  * Attach the zip file to the GitHub issue.

Additionally, if the problem happened in the browser, it can be useful to know about any browser console errors, so please paste them into the issue too.

magnetite2:
19:14:24 INFO  Buffered logging turned OFF

19:14:24 INFO  Verbose file logging pattern enforced

19:14:24 INFO  Getting gui configuration file : C:\ProgramData\UMS\GUI.conf

19:14:25 INFO  Starting Universal Media Server 14.4.0

19:14:25 INFO  Based on PS3 Media Server by shagrath, copyright 2008-2014

19:14:25 INFO  https://www.universalmediaserver.com

19:14:25 INFO  

19:14:25 INFO  Build: db5479e91 (2024-08-10)

19:14:25 INFO  Working directory: C:\Program Files (x86)\Universal Media Server

19:14:25 INFO  Temporary directory: C:\Users\Jeff\AppData\Local\Temp\UMS

19:14:25 INFO  Logging configuration file: C:\Program Files (x86)\Universal Media Server\logback.xml

19:14:25 INFO  Logfile: C:\ProgramData\UMS\debug.log

19:14:25 INFO  

19:14:25 INFO  Profile directory: C:\ProgramData\UMS

19:14:25 INFO  Profile directory permissions: drwx

19:14:25 INFO  Profile configuration file: C:\ProgramData\UMS\UMS.conf

19:14:25 INFO  Profile configuration file permissions: -rwx

19:14:25 INFO  Profile name: MAGNETITE

19:14:25 INFO  

[...]

PR #5138, opened 2024-12-01
Transcode if any audio is unsupported

SubJunk:
# Description of code changes

If at least one audio track is not compatible, the file must be transcoded.

The reason it needs to be like this is because some files have the audio tracks matching the Blu-rays, which is often a DTS variant main track, with other tracks like commentary in Dolby Digital (AC-3), and for a media player that does not support DTS that could result in the file being streamed with only the commentary track playable, which is really annoying.

# Checklist

### DIFF
--- a/src/main/java/net/pms/configuration/FormatConfiguration.java
+++ b/src/main/java/net/pms/configuration/FormatConfiguration.java
@@ -721,8 +721,21 @@ public class FormatConfiguration {
 				renderer
 			);
 			finalMimeType = mimeType;
-			if (mimeType != null) { // if at least one audio track is compatible, the file can be streamed.
-				return finalMimeType;
+
+			/**
+			 * If at least one audio track is not compatible, the file must be transcoded.
+			 *
+			 * Explanation: This was different in the past (see link below).
+			 * The reason it needs to be like this is because some files have the audio tracks
+			 * matching the Blu-rays, which is often a DTS variant main track, with other tracks
+			 * like commentary in Dolby Digital (AC-3), and for a media player that does not
+			 * support DTS that could result in the file being streamed with only the commentary
+			 * track playable, which is really annoying.
+			 *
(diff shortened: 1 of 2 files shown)


SubJunk:
This was awaiting review for 2 weeks so I will merge it, I have tested it

ExSport:
Hello @SubJunk 
A lot of users have some 4K videos with many high quality audio tracks, where e.g. one of 4 tracks is unsupported 7.1 whatever. 
Now user can stream videos easily, with correct audio. 
After the change all such files will be transcoded what will be unusable (realtime 4k transcoding, problematic fwd/rwd).
There should be another options like transcode audio track only, detect documentary track and ignore transcoding when another supported track is available, give configuration option to override such behavior. 
I am sure there are more occasions when user wants to stream file instead of transcoding and when, in minor cases, user needs some specific audio track, there is TRANSCODE folder for it? 
I suppose there will be more complains that video worked but now doesn't work vs. I hear documentary mono audio only. Or am I wrong?😏
Btw. Happy Christmas time and thank all of you for your work on this project👍

SubJunk:
@ExSport this is good feedback, thanks. Yes another way to do it would be try to determine whether a track is commentary. That was what initially caused this PR - I had a file that had DTS main audio, then AC-3 commentary track, and since my TV does not support DTS I couldn't play the main audio.

We do mux the 4k video track usually too, depending on the media player.

I don't know if we can reliably detect which track is a commentary track though?

Mik-S-UMS:
This reminded me of #4954 where disabling subtitles (as they were PGS and needed transcoding) would play a different audio track. 

I guess different languages could also use different audio codecs so it might not just be commentary tracks that need to be detected. I'm not sure on the best solution though, just that language settings will need to be taken into account too.

SubJunk:
For now I will revert my change, thanks for the comments. I will try to detect the commentary track based on the title. In my case it was called `commentary` so we can check for that. It is far from perfect but maybe is a better change than this one

SubJunk:
Oh and happy Christmas time too @ExSport thank you!

SubJunk:
I still get commentary tracks... I will need to fix this

PR #5151, opened 2024-12-05
main fix yarn dev update

### LINKED ISSUE #5013: Web player development workflow not working
What I expect is to run `yarn dev` within the `react-client` directory and that launches an instance of vite that connects to the UMS server. But what happens is that it launches a version of the React client that does not update with code changes.

I can run `npx vite build` manually and then access the changes via http://localhost:9001/ but I can't figure out how to get a live development environment working.

SurfaceS:
Adjust chunk size limit
yarn dev now update with code changes.

fix #5013

### DIFF
--- a/react-client/package.json
+++ b/react-client/package.json
@@ -5,7 +5,7 @@
   "proxy": "http://localhost:9001",
   "type": "module",
   "scripts": {
-    "dev": "vite",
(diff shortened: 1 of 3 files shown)


SubJunk (approved):

SubJunk:
Cool, thanks!

PR #5167, opened 2024-12-17
transcoding issue WebStream / check if renderer support native stream

ik666:
WebStreams should only be transcoded, if the renderer doesn't support the streams mime-type. 

This PR checks if the renderer configuration has a "supports line" for the streams mime-type. In that case no transcoding is being done.

### DIFF
--- a/src/main/java/net/pms/store/StoreItem.java
+++ b/src/main/java/net/pms/store/StoreItem.java
@@ -63,6 +63,7 @@ import net.pms.store.container.ChapterFileTranscodeVirtualFolder;
 import net.pms.store.item.DVDISOTitle;
 import net.pms.store.item.RealFile;
 import net.pms.store.item.VirtualVideoAction;
+import net.pms.store.item.WebStream;
 import net.pms.util.ByteRange;
 import net.pms.util.FileUtil;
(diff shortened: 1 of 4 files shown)


src/main/java/net/pms/store/StoreItem.java
```diff
@@ -442,6 +443,20 @@ public TranscodingSettings resolveTranscodingSettings() {
 			return null;
 		}
 
+		// WebStreams should be checked against the mime-type of the stream
+		if (this instanceof WebStream ws) {
+			if (getMediaInfo() != null && renderer.getFormatConfiguration() != null) {
+				if (StringUtils.isAllBlank(getMediaInfo().getMimeType())) {
+					LOGGER.warn("cannot read MIME-TYPE of stream {} ...", ws.getUrl());
```
SubJunk, line 450: Should this line also return early?

ik666: If we cannot read the mime-type of the web-stream, we may need transcoding. Therefore, this line shouldn't return early. I haven't seen any web-streams without mine-type set, yet ...

SubJunk (approved):

PR #5233, opened 2025-01-07
CDS BrowseDirectChildren / unnecessary discovery of children of children

ik666:
While browsing the CDS, children of children are being looked up resulting in a slow response in some cases, where children have themself many entries (`StoreItems`). 

As an example: I have a playlist folder. In this folder are 10 Playlists having 100 entries. For browsing the folder, its just enough to discover the 10 `PlaylistsFolder` container. At the moment also the children of the `PlaylistFolder` are being looked up, resulting in additional 1000 lookups, that are not needed to fulfill the actual request.

There is a comment about DVD structure. Maybe there exists some objects that need to be deep inspected. If so, it probably should be done on the StoreResource implementation itself.


SubJunk:
This change looks good to me, but I want a second opinion from @SurfaceS since he has been more across the refactoring of these files

SurfaceS:
The thing is that, as you need it for your implementations, you add a lot of `resolve()`, `discover()`, `parse()`, `discoverChildren()`, `doRefreshChildren()`, etc here and their, ending in discovering the all things most of time before this code is reach.

This code was their to avoid overload on `getResources` with childs.
Also, it was their as resources was shared between renderers. So it need to be refreshed with the config for the renderer asking it.
Now it is no more the case.

So, it should have no side effect to remove it.

SubJunk (approved):

PR #5272, opened 2025-02-03
main fix 5203

### LINKED ISSUE #5203: Regression: support for Windows short names
In the past we would support very long file names (at least on Windows) by falling back to short names, but now the files just are not added. I had to rename them to be shorter so UMS would see them

SurfaceS:
fix #5203

# Description of code changes
use of Win32 file prefixing and conventions
refactoring `getShortFileNameIfWideChars` to `getSystemPathName`
refactoring `getShortPathNameW` to `getSystemPathName`

### DIFF
--- a/src/main/java/net/pms/platform/windows/WindowsUtils.java
+++ b/src/main/java/net/pms/platform/windows/WindowsUtils.java
@@ -109,6 +109,10 @@ public class WindowsUtils extends PlatformUtils {
 		if (path != null) {
 			File file = new File(path);
 			String resolved = file.getAbsolutePath();
+			/*
+			 * On Windows the maximum short path is 260 minus 1 (NUL) -> 259 char.
+			 * But for directories it is 260 minus 12 (8.3 file) minus 1 (NUL) to allow for the creation of a file in the directory -> 247.
+			 */
 			if (resolved.length() > 247) {
 				if (resolved.length() > 32000) {
 					LOGGER.warn("Cannot access file with path exceeding 32000 characters");
--- a/src/main/java/net/pms/platform/windows/WindowsUtils.java
+++ b/src/main/java/net/pms/platform/windows/WindowsUtils.java
@@ -17,7 +17,6 @@
 package net.pms.platform.windows;
 
 import com.sun.jna.Native;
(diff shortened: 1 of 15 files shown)


src/main/java/net/pms/platform/windows/WindowsUtils.java
```diff
@@ -108,37 +105,21 @@ public File getKLiteFiltersDir() {
...
-				int r = Kernel32.INSTANCE.GetShortPathNameW(pathname, test, test.length);
-				if (r > 0) {
-					String result = Native.toString(test);
-					LOGGER.trace("Using short path name of \"{}\": \"{}\"", pathname, result);
-					return result;
+	public String getSystemPathName(String path) {
+		if (path != null) {
+			File file = new File(path);
+			String resolved = file.getAbsolutePath();
+			if (resolved.length() > 247) {
```
SubJunk, line 112: Why 247?

SurfaceS: On Windows the maximum short path is 260 minus 1 (`NUL`) -> 259 char.
But for directories it is 260 minus 12 (8.3 file) minus 1 (`NUL`) to allow for the creation of a file in the directory -> 247.

The long path prefix (`\\?\`) is working with short path as well.
For example, a file like `C:\ProgramData\UMS\UMS.conf` can have `\\?\C:\ProgramData\UMS\UMS.conf` as working path even if it is not more than 259 char (the max for short file path).

So it is no need to check if it is a directory or a file, just prefix as long path if it is more than 247 char.

SubJunk: Interesting, thanks for explaining. I'll add that explanation as a code comment for future people

SubJunk, line 112:
```suggestion
			/*
			 * On Windows the maximum short path is 260 minus 1 (NUL) -> 259 char.
			 * But for directories it is 260 minus 12 (8.3 file) minus 1 (NUL) to allow for the creation of a file in the directory -> 247.
			  */
			if (resolved.length() > 247) {
```

Issue #4923, opened 2024-08-25
Crash on media scan on Docker


kieferlam:
Hello. I have UMS running in docker using universalmediaserver/ums image. The container crashes and restarts continously when I specify my media folder and start the scan. Because the container is continously crashing and restarting, I cannot pack the debug files per the instructions. I have set debug log level to TRACE. These are the relevant logs from the docker container logs:

> INFO  22:11:07.277 [main] Web player is available at: http://192.168.1.253:9002
INFO  22:11:07.323 [SpeedStats background worker 1] Renderer 50EX700_Series found on address: 192.168.1.54
INFO  22:11:07.517 [main] Universal Media Server is now available for renderers to find
INFO  22:11:07.518 [Media Scanner] Media scan started
/usr/include/c++/13.2.1/bits/stl_vector.h:1125: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](size_type) [with _Tp = ZenLib::Ztring; _Alloc = std::allocator<ZenLib::Ztring>; reference = ZenLib::Ztring&; size_type = long unsigned int]: Assertion '__n < this->size()' failed.

After this, it restarts and cycles through this routine of startup, media scan started, crash.

SubJunk:
Looks related to MediaInfo, what's the MediaInfo version?

kieferlam:
How can I get this information?

SubJunk:
Can you see if it still happens if you update MediaInfo to 24.06?

kieferlam:
How do I go about doing this?

ga30r7:
Hello, i have the same exception in the log with the latest `universalmediaserver/ums` docker image.

The server version is:
Universal Media Server 14.5.0
database engine version 2.3.232
MediaInfoLib - v24.04

Im not an expert to update the `mediainfo` lib in linux, so would you like to give me an example command to do that?

Thanks for the help!

SubJunk:
It looks like 24.04 is the latest MediaInfo in that environment for now :( https://pkgs.org/search/?q=mediainfo

You can compile and install MediaInfo from source but I don't have instructions for that. This is where to get the source code: https://mediaarea.net/en/MediaInfo/Download/Source

SubJunk:
We did a new release recently, it might be worth checking if that fixed this, please let us know

kieferlam:
Still crashing with the same error at the same point.
[...]

PR #5273, opened 2025-02-04
new Dockerfile using Ubuntu 22.04 LTS and JDK 21 (Temurin)

### LINKED ISSUE #5264: Crashes when trying to scan a specific file.
UMS version: 14.10.0 (Docker)

If there is a file in the scan range of UMS, UMS will crash immediately.
When this file is deleted or moved out of the scan range, it will return to normal after the container is restarted.

The attachment is a test file, the file type is ProTracker module.
This file has been confirmed to play normally on Windows.
[{E0ADFAA7-D4E1-6978-451D-4A585DF0BC7A}.zip](https://github.com/user-attachments/files/18565479/E0ADFAA7-D4E1-6978-451D-4A585DF0BC7A.zip)

**Not all files of the above types will cause UMS to crash. Files of the same type and in the same path will not cause UMS to crash.**

Information before the crash:
```
/usr/include/c++/14.2.0/bits/basic_string.h:1352: std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::reference std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::back() [with _CharT = char; _Traits = std::char_traits<char>; _Alloc = std::allocator<char>; reference = char&]: Assertion '!empty()' failed.
```

ik666:
This Dockerfile uses Ubuntu 22.04 LTS and Eclipse JDK 21 (Temurin). This may fix #5264 and #4923.

@SubJunk Please check if this build is running fine with the `ant` build.


SubJunk:
Ok cool I will check when I have time, thanks

SubJunk:
I tried to test it but I am having trouble getting Docker to run on my PC. Does anyone else have it running already? @SurfaceS ?

SubJunk:
@SurfaceS I publish our releases to Docker Hub here https://hub.docker.com/r/universalmediaserver/ums
It could be a Maven profile to make it easier. I publish it from my old x64 MacBook since that is the only thing I have that Docker works on

SurfaceS:
@SubJunk 
I think that this PR workaround the issues by `-Xms2048m -Xmx4096m` instead of `-Xmx512M`.
The other part is using `ubuntu:22.04` instead of `alpine`,  `Java 21 Temurin` instead of `Java 17 official`, and build the jar inside the docker (strange way).
That's why it is so big (`JRE` + `JDK` + `sources` + `build`).

I don't think `ubuntu:22.04` is better than `alpine` (for `docker`) nor `Java 21 Temurin` vs `Java 17 official`.

@SubJunk 
try #5338
I removed completely the heapmem things, as `java 11+` is aware of the container memory.
Having a fixed heapmem can end in short in memory, or fail to load (in case the system do not have the asked memory)

SubJunk:
Good point, that heapmem was really low before. Thanks

PR #4918, opened 2024-08-23
Extended UPnP CreateObject / ImportUri support

ik666:
This PR adds

extended CDS createObject support 
- `object.container.storageFolder` => creates and adds folders to the CDS
- `object.item` => creates and adds empty objects (for audio or video file type)

import/update CDS objects
- implements UPnP `importUri` attribute for importing / updating CDS content identified by objectID

PR #5293, opened 2025-02-18
new UPnP ums service

ik666:
I've seen some other media server with UPnP services for server interaction. I found that idea quite useful and that's why I have made this PR, which adds an UMS service to the media server device. It's designed for extended configuration and action capabilities.

It's main goal is to give media devices the ability to interact with UMS in an UPnP intended way (a standard way). Since the UPnP services are self describing, they are easier to use as a custom made API, which is removed. The existing services `rating` and `like` were remapped to the UPnP services.


SubJunk:
@dependabot recreate

SurfaceS:
@ik666 
I don't have time to review this for now but a quick look end to -> it cannot be merge as this :
```
	 * Restricting service access violates the UPnP-av-ContentDirectory contract and should default to true.
	 * See UPnP-av-ContentDirectory-v4-Service.pdf section 5.2.25 Device Protection Option. Correct way is to
	 * implement a DeviceProtection service to restrict access.
```
UPnP ContentDirectory from V4 :
5.2.25 Device Protection Option

> The ContentDirectory service does not include any mechanism to restrict access to service actions/content based on Roles, Control Point Identities or User Identities. Due to this openness, content and services on MediaServers can be accessed by all control points/users in the UPnP network. **This leads to violation of content privacy in MediaServers**. Implementation of the ContentDirectory service CONTENT_PROTECTION feature addresses these limitations.

The violation is to let the media server content be accessed by anyone, not to restrict adding, modifying or deleting ContentDirectory object or media file.

> UPnP DeviceProtection service defines three Roles: Public, Basic and Admin.
> For control points that do not implement DeviceProtection or remain unregistered, a generic Role of (Public) is automatically assigned.

When you said `should default to true` (no restriction), UPnP say it default to generic Role of (Public) (the more restricted access).
When you said `Correct way is to implement a DeviceProtection service to restrict access.` UPnP say `Correct way is to implement a DeviceProtection service to allow access.`.


**Restricting writes on the media server files from control points is a must, not an option.**

> B.1.4 @restricted
> Namespace: DIDL-Lite Property Data Type: xsd:boolean Multi-Valued: NO
> Description: The read-only required @restricted property indicates whether the object is modifiable. If set to “1”, the ability to modify or delete a given object is confined to the ContentDirectory service implementation. Therefore, a control point cannot add, modify or delete metadata from a restricted object. Additionally, control points are not able to add, modify or delete any children of a restricted container. However, the @restricted property does not propagate to descendant objects. Note however, that metadata of a restricted object can still change due to internal ContentDirectory service implementation manipulations.
> If set to “0”, a control point can modify the object’s metadata and add, delete, or modify the object’s children.

We advertise all CDS objects to `@restricted=1`.
Change on CDS is from internal ContentDirectory service implementation manipulations (UMS itself).
All normal control point understand that they cannot add, modify or delete CDS objects.
Also, the ContentDirectory service (UMS) should check if the control point is allowed to add, modify or delete CDS objects before processing the request.

That is a security concern.
I already fix #4918 with #5199.
Without this security patch, anyone in the network can delete all your media server files.
I don't think that end users will be happy with that.

@SubJunk 
Did you end with the same review ?

More globally, in my opinion : 
- UMS should not change media files by default (as now).
- All things related to a file (metadata) can reside on the database, even more now as UMS is multi-user, with their own play count, rating, and so on.
- In case UMS may modify media files, end user should agree with that when changing the setting (advertise him before that the file can be broke, etc). I think that @UniversalMediaServer/developers already fight on this when many years ago a PR allowed audio rating to be written directly into the media file (still present).
- If we add the ability to add/remove/modify media file from external request, authentication should be on, and user should be notified of that when changing the setting.
- If we add the ability to add/remove/modify media files from external request without authentication, user should be warn of that, and the setting should be hidden.

ik666:
@SurfaceS 
For the sake of happiness, we can just delete the comment. No "default behaviour" code was changed. The previous API is just linked to this service.

> Correct way is to implement a DeviceProtection service to allow access.

As far as I can follow your point: If you think this to the end and regarding the fact, that UMS doesn't implement a DeviceProtection service (http://www.upnp.org/specs/gw/UPnP-gw-DeviceProtection-v1-Service-20110224.pdf), in conclusion no one would have any access.

> Also, the ContentDirectory service (UMS) should check if the control point is allowed to add, modify or delete CDS objects before processing the request.

As I have pointed out in the past, UMS doesn't identify control points, instead, control points are treated as renderer. And due to missing DeviceProtection service, control points cannot identify themself at the moment.

And yes, adding more cool features to UMS (adding content media, adding/modifying metadata, etc.) makes no sense if UMS always sends `@restricted=1`. UMS should respect it's configuration. I personally would rather like to keep UMS universal and interactive.

ik666:
@SurfaceS 
>Without this security patch, anyone in the network can delete all your media server files.

Just to be precise: At the moment, only CreateObject is implemented, updating existing objects is not implemented. So no: no one can delete all files. IMHO I doubt, that anyone in your home would implement a UPnP Service just for doing that. Since people in your home usually have physical access to hardware, there are easier way to delete everything. 

Many UPnP device implementations lack authentication mechanisms, and by default assume local systems and their users are completely trustworthy.

ik666:
@SurfaceS 
> All things related to a file (metadata) can reside on the database

Audio tagging tools exists broadly in the wild for many years and they are used by people. I think this fact stands for itself. 

UMS is not the only, unbeaten omnipresent tool for managing / displaying / playing files, especially in regard to audio files. Tools like MediaMonkey and TwonkyMedia server pick up all tags from the file itself and therefore the information stored there is interoperable between tools. One obvious thing : If your UMS database is broken you lose all your month of work spend in tagging ... i had this once.

ik666:
@SubJunk Do you mind merging the UPnP service? Since I’ve moved away from the “HTTP-API” in favor of the UPnP service, my code is starting to drift out of sync.

SubJunk:
@ik666 I was busy at the end of the year but I will try to make progress on reviewing your PRs now. Sorry for the delay

ik666:
@SubJunk There seems to be an error in the build pipe. `/Users/runner/work/UniversalMediaServer/UniversalMediaServer/target/ffmpeg-macos-1015.zip` is not resolvable... maybe you can have a look at it.

SubJunk:
That error is fixed in main. A download link disappeared and I updated it

ik666:
@SubJunk All right this worked. I've made your suggested changes.

SubJunk:
@ik666 cool, I will do a release soon and then merge this for the next release

PR #5298, opened 2025-02-22
Update Kodi.conf - Set "TranscodeFastStart" to false

prescott66:
Revert "TranscodeFastStart" enabling, because it often caused non-compatible videos for Kodi player (tested on Android and Raspberry Pi).


SubJunk:
Great, thanks!

Issue #5211, opened 2024-12-30
x64 installer not working on Windows 10


SubJunk:
This is related to a comment on Patreon:

> Tested Windows x86_64 on Win10 22H2 bare metal and on a VM. Both simply do not start, tried updating and fresh install, Task manager shows nothing.

> Standard x86 works fine on both.

I have no idea why that would be happening. I am running it on Windows 11. I downloaded the build from Patreon and installed it, and it continued to work for me. Does anyone have any ideas?

Mik-S-UMS:
I did have an issue installing 14.8.0 after using the build script. 

It did not install the bin files into `C:\Program Files\Universal Media Server\` and I could not run UMS from there. 
The only way I could get it to run is with the UMS.bat file.
It also left the old `C:\Program Files (x86)\Universal Media Server\` folder alone.

Doing a clean install made it work properly but still have the old folder. 

I did not have a problem upgrading to 14.9.0 afterwards so may have something to do with the change in installation location, or may have just been because I was using that script instead of the official installer.

@SubJunk does that build script create the x86_64 installer? If not do I need to change anything for it to make this one?

SurfaceS:
> It also left the old C:\Program Files (x86)\Universal Media Server\ folder alone.

This need to be done if moving... I forgot that.

SubJunk:
@SurfaceS will you fix it? I can skip the x64 version in the auto updater and website for now to minimize the impact

SubJunk:
I have stopped UMS from updating to the x64 installer now, in https://github.com/UniversalMediaServer/UniversalMediaServer/pull/5212
I am making the same change to the website now, then I will remove the installer from the release post, until the problems are fixed.

SurfaceS:
> @SurfaceS will you fix it? I can skip the x64 version in the auto updater and website for now to minimize the impact

fixed

SubJunk:
Apparently it is not fixed, I posted this fix to Patreon and a user said:

> server doesn't launch by itself after installation even if it was checked. Tried open it manually, but it doesn't work. I reboot the pc and the server launch itself but with the 14.8.0 version :/

SubJunk:
@SurfaceS I forgot to tag you in the above post

SubJunk:
I have tested and my directory `C:\Program Files (x86)\Universal Media Server\` is still there after installing too

SurfaceS:
@SubJunk wait a bit for release as you updated the binaries.

[...]

PR #5305, opened 2025-02-26
Restore x86 installer

SubJunk:
# Description of code changes

Users complained about it being removed

# Checklist
- [x] Any relevant issues are [referenced](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#referencing-issues-and-pull-requests)
- [x] Any relevant [Knowledge Base](https://github.com/UniversalMediaServer/knowledge-base) articles are written/modified
- [x] Any relevant tests have been written/modified


SurfaceS:
@SubJunk I'll review it today.
I think it miss things.

SurfaceS:
@SubJunk 

The x86 build seems ok with that PR.

Just remember that it can only be temporary :
`ffmpeg` in windows x86 is version 4.2.2 from 5 years ago (ffmpeg fix the 4.2 version to rev 4.2.10).
It is a 4.2 version, when all other platforms use version 7.1.
As we use only 7.1 (I don't have any x86 windows anywhere), code changing on `ffmpeg` command may or not work with the 4.2 version.

Also, java had cut the Windows 32-bit x86 Port on their newer versions (like most of others).
[Remove the Windows 32-bit x86 Port (x86 processors)](https://openjdk.org/jeps/479) released, since all windows x86 reach Microsoft EOL support.

Latest x86 processors without 64-bit support are from 20 years ago. After that, they all can run 64-bit OS.
The same for performances : a 25 years processors can run a media server and transcode ?

I think windows x86 UMS can be given as this, without any support for few release, then we should stop releasing it.

@Mik-S-UMS 
Let me know if the x64 Windows build still has error in it. (I'll have a look on the Issues on the same time)

Mik-S-UMS:
@SurfaceS yeah there is still a problem with the first install. I updated my findings [here](https://github.com/UniversalMediaServer/UniversalMediaServer/issues/5211#issuecomment-2676909642)

SubJunk:
Yeah I agree it is rare for x86 to be used, but we did get about 3 comments about it 🤷‍♂️

PR #5332, opened 2025-03-06
v15 player

SurfaceS:
React GUI player :
Media card click will play directly
Added long press on media card to see metadata page
Moved media `edit` action under the media metadata page
Removed from bread

Added TV season metadata + localize on season metadata page Added db table for season localize data

React GUI : eslint more


SubJunk:
Nice changes!

Issue #5315, opened 2025-03-02
Videos in transcode folder appear empty


Mik-S-UMS:
The transcode folder is not showing any content on a video.

I noticed this in the `V15` branch so thought it was isolated to that but a user on the forum mentioned this is empty when I suggested playing from the transcode folder.
I reinstalled from `Main` (14.10.1-SNAPSHOT) to confirm and it is appearing the same.

It looks like the DIDL is not being populated for videos in the transcode folder

```
TRACE 2025-03-02 17:57:31.451 [jupnp-sync-protocol 317-1] DIDL-Lite result:
<DIDL-Lite xmlns="urn:schemas-upnp-org:metadata-1-0/DIDL-Lite/" xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:pv="http://www.pv.com/pvns/" xmlns:sec="http://www.sec.co.kr/" xmlns:upnp="urn:schemas-upnp-org:metadata-1-0/upnp/"/>

TRACE 2025-03-02 17:57:31.451 [jupnp-sync-protocol 317-1] Returning browse result
TRACE 2025-03-02 17:57:31.452 [jupnp-sync-protocol 317-1] Preparing HTTP response message: (OutgoingActionResponseMessage) 200 OK
TRACE 2025-03-02 17:57:31.452 [jupnp-sync-protocol 317-1] Sending HTTP response status: 200
TRACE 2025-03-02 17:57:31.453 [jupnp-sync-protocol 317-1] Response sent to 192.168.1.194:
============================= UPNP HTTP SERVER RESPONSE BEGIN ===============================
HTTP/1.1 200

HEADER:
  Server: Windows_11-amd64-10.0, UPnP/1.0 DLNADOC/1.50, UMS/14.10.1-SNAPSHOT
  Date: Sun, 02 Mar 2025 17:57:31 GMT
  Ext: 
  Content-type: text/xml;charset="utf-8"
  Content-Length: 663

CONTENT:
<s:Envelope xmlns:s="http://schemas.xmlsoap.org/soap/envelope/" s:encodingStyle="http://schemas.xmlsoap.org/soap/encoding/">
    <s:Body>
        <u:BrowseResponse xmlns:u="urn:schemas-upnp-org:service:ContentDirectory:1">
[...]

PR #5352, opened 2025-03-10
fix : discover children of children

ik666:
This is a fix for #5315. Transcode folder needs to have a special treatment. This PR fixed the reported error, without discovering children of children for other folder types.

### DIFF
--- a/src/main/java/net/pms/store/MediaStore.java
+++ b/src/main/java/net/pms/store/MediaStore.java
@@ -61,6 +61,7 @@ import net.pms.store.container.RealFolder;
 import net.pms.store.container.SearchFolder;
 import net.pms.store.container.ServerSettingsFolder;
 import net.pms.store.container.SevenZipFile;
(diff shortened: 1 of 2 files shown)


ik666:
@SubJunk 
Actually, the `TranscodeVirtualFolder` didn't discover it's children correctly. Probably it would be more clear to revert the revert #5334 and fix the error by adding a `discoverChildren` method to the `TranscodeVirtualFolder` as follow:

```
public class TranscodeVirtualFolder extends StoreContainer {

	public TranscodeVirtualFolder(Renderer renderer, UmsConfiguration configuration) {
		super(renderer, configuration.getTranscodeFolderName(), null);
	}

	// Sub class constructor
	protected TranscodeVirtualFolder(Renderer renderer, String name) {
		super(renderer, name, null);
	}

	@Override
	public void discoverChildren() {
		syncResolve();
	}
}
```

SubJunk (approved):

PR #5369, opened 2025-03-14
Audioaddict Network support

ik666:
This PR gives access to the AudioAddict network.

AudioAddict network is a set of internet radio stations
- Classical Radio
- Radio Tunes
- Rock Radio
- Jazz Radio
- Zen Radio
- DI.fm
covering all kinds of genres. 

**Account needed ?**
Yes, you need a (paid) premium account to access the radio stations. You only need to register on one site. Access is then granted to all sites with the same username / password.

**What is implemented ?**
Access all Radio sites, their channels and filters. Favorites channel is supported.

**Configuration**
Add to `UMS.conf` 
```
# TRUE = prefer European streaming server
audio_addict_europe = TRUE or FALSE

# your username 
audio_addict_pass = PASSWORD

# your password
audio_addict_user = USERNAME
```

**Where is it ?**
You find the Audio Addict folder in the `Media Library`

**Outlook (not yet implemented)**

- Support Favorites management and settings updates as UPnP UmsService actions, but first the service #5293 needs to be merged. 
- Show's support
- Show's history support
- Related radio stations support

### DIFF
--- a/src/main/java/net/pms/external/audioaddict/RadioNetwork.java
+++ b/src/main/java/net/pms/external/audioaddict/RadioNetwork.java
@@ -50,6 +50,7 @@ public class RadioNetwork {
 	private static String apiKey = null;
 
 	private static String listenKey = null;
+
 	private static boolean authenticated = false;
 
 	private HttpClient httpNonBlocking = new HttpClient();
@@ -249,8 +250,10 @@ public class RadioNetwork {
 		for (ChannelJson channelJson : allChannels) {
 			int errorConter = 0;
 			try {
-				ContentResponse response = httpBlocking.GET(channelJson.getPlaylist());
-				String url = getBestUrlFromPlaylist(response.getContentAsString());
+				String plsRequest = String.format("%s?listen_key=%s", channelJson.getPlaylist(), listenKey);
+				ContentResponse response = httpBlocking.GET(plsRequest);
+				String pls = response.getContentAsString();
+				String url = getBestUrlFromPlaylist(pls);
 				preferredChannels[i] = new Channel(channelJson.getId(), channelJson.getKey(), channelJson.getName(), url);
 			} catch (InterruptedException | ExecutionException | TimeoutException e) {
 				LOGGER.error("{} : convert playlist to url failed for item {} : {}", this.network.displayName, i, channelJson.getPlaylist(),


SubJunk:
Cool idea!

ik666:
Yeah, thank you. It's working quite descent. Had no issues yet. I can listen to all radio stations pretty stable (over some hours).

SurfaceS:
@ik666 you'll end up hating me :woozy_face:
It's not against you, I promise :wink:

I do not had full reviewed nor run this code... Just had a quick look in it.

## Promoting a paid service
This also concerns @SubJunk 
- I expect that you earn something from the AudioAddict network for advertising it on open source/free software.
- I don't know if you (UMS at end) can use their name, logo, implementation, etc (**copyright, trademark, etc**) and I doubt it, as all others media servers I know do not officially support it, and let it's implementation as non official plugin.

## about dependencies :
it add `com.squareup.okhttp3` (HTTP client) and `org.jetbrains.kotlin` (Runtime libraries for Kotlin reflection)...
- `org.jetbrains.kotlin` is not needed
- `com.squareup.okhttp3` should be replaced with `org.eclipse.jetty.client` (`HttpClient`, `Request`, `Response`, `ContentResponse`, etc) 

## about the code
I'll let you some things I suggest.
Like changing the `net/pms/store/StoreContainer.java` code is not the way to go.
`StoreContainer.java` is the base for all containers, so if you need to have a custom implementation of a member for your containers, you should use your Class that extend it (`AudioAddictNetwork` for example) and `@Override` that member.
Most of time, new implementation should not touch existing code used by all other working things.

`DLNAThumbnailInputStream.toThumbnailInputStream(HTTPResource.downloadAndSend(thumbnailIcon, false));`
This mean the Thumbnail will be downloaded **each time** the renderer will display the item. I don't think that is what you want.

`net/pms/store/container/MediaLibrary.java` code is not for non database item.
An implementation like that should sit where DbIdLibrary is.
Maybe another solution is to move it under shared content... but it will be a lot of work... I don't know.

`net/pms/io/OutputBufferConsumer.java` remove the typo created

`-Xmx4096M` this change should not be on this PR.
In v15 I'll remove it, and remove old needs `--add-opens java.base/jdk.internal.misc=ALL-UNNAMED -Dio.netty.tryReflectionSetAccessible=true"`

SubJunk:
@ik666 about the promotion of a paid service section of @SurfaceS response, are you affiliated with them or receiving money?

ik666:
@SubJunk No, I am not.

SubJunk:
Ok. Is this PR ready for another code review?

ik666:
Yes, or at least I wouldn't change anything anymore. I think it's pretty obvious, people have different understanding of architecture and how things should be done.

pom.xml
```diff
@@ -623,7 +623,23 @@
...
+				<exclusions>
+					<exclusion>
+						<groupId>org.jetbrains.kotlin</groupId>
+						<artifactId>kotlin-stdlib-jdk8</artifactId>
+					</exclusion>					
+				</exclusions>
+		</dependency>
+		<dependency>
+			<groupId>org.jetbrains.kotlin</groupId>
+			<artifactId>kotlin-stdlib-jdk8</artifactId>
```
SubJunk, line 639: Why do we need a kotlin lib?

pom.xml
```diff
@@ -623,7 +623,23 @@
 			<artifactId>directories</artifactId>
 			<version>26</version>
 		</dependency>
-	</dependencies>
+		<dependency>
+				<groupId>com.squareup.okhttp3</groupId>
+				<artifactId>okhttp</artifactId>
```
SubJunk, line 628: We already do HTTP requests throughout UMS, why is this extra lib being added instead of doing it the same way we usually do?

SubJunk (changes requested):
Asked some questions, also it would be great to add some tests for this feature. I don't use the service so tests would give me more confidence that it is working and won't break in the future

pom.xml
ik666: This library is needed by `okhttp3`

pom.xml
ik666: I also maintain the implementation in my control point at the moment. Switching implementation can be done, but makes more effort in keeping both implementation in sync. Maybe I'll habe a look and move to the jetty http client.

SubJunk: That would be great. I like the idea of this change so I want to support it. I hope it wouldn't be a lot of work to use our http client instead

ik666:
@SubJunk Actually, there was also an annoying blocking issue in the initialization phase. I have refactored the implementation to non-blocking access using the Jetty client.

pom.xml
ik666: No, it's not a lot of work. Have done it ...

ik666:
@SubJunk added some tests

SubJunk:
Thanks for adding tests!

SubJunk:
@ik666 would you add a document to the knowledge base for this? https://github.com/UniversalMediaServer/knowledge-base

ik666:
Yes, I'll add how to configure it.

ik666:
@SubJunk I added a PR [here](https://github.com/UniversalMediaServer/knowledge-base/pull/271).

PR #5372, opened 2025-03-14
main fix nsis exe

SurfaceS:
# Description of code changes

- removed `heapmem` from setup file and exe file
- added setup file properties `ProductName`, `CompanyName`, `Version`, etc
- added details about creating the uninstaller instead of leaving the last copied file, which makes the user think it takes a while

# Checklist

### DIFF
--- a/src/main/external-resources/UMS.bat
+++ b/src/main/external-resources/UMS.bat
@@ -6,4 +6,4 @@ echo You can try to reduce the Xmx parameter value if you keep getting "Cannot c
 echo Last word: You must have java installed ! http://www.java.com
 echo ------------------------------------------------
 pause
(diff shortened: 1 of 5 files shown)


SubJunk:
@SurfaceS after your comment in the Docker thread about Java managing its memory better after JRE 11, I was wondering whether we should default to no heapmem, what do you think?

SurfaceS:
- removed `heapmem` from setup file and exe file
- added setup file properties `ProductName`, `CompanyName`, `Version`, etc
- added details about creating the uninstaller instead of leaving the last copied file, which makes the user think it takes a while

SubJunk:
Do you think this should merge into v15, in case there are problems? Then we can get feedback on the stability via the v15 testers

SurfaceS:
@SubJunk this one was merged into v15 (no mem limit anymore).

but we still need to fix at least the `exe` bug. see #5390

PR #5374, opened 2025-03-15
WIP user switcher

SubJunk:
# Description of code changes

This is a WIP based on an old version of v15
I ran out of time today to work on it more, but if you want @SurfaceS you could continue it. Otherwise I will keep going when I have time. Please let me know if you start on it, so that we don't both do it.

For the UI I used the Mantine UI from https://ui.mantine.dev/component/user-info-action/

The idea of the feature is to use Local Storage to make it faster and look nicer for a user to log in, so they should be able to click "Select user" and it prepopulates the username field in the form.

In the future there could be the concept of multiple valid sessions, then there would be a "Switch User" menu item above "Log out" so the user can choose whether to logout or just switch. This is a first step towards that.

# Checklist


SurfaceS:
@SubJunk 
I'll have a look

SubJunk:
@dependabot recreate

PR #5375, opened 2025-03-15
V15 res

SurfaceS:
# Description of code changes

## React
implements user switcher
added login carousel
added login with pin code
added login without pass
uniformized login react
refactor `Permissions` to `UmsPermission` to avoid browser `Permissions` mistake
moved permissions check under `session`
adapt some `function`/`const` as `ReactNode` (avoid rendering different `ref` on re render, see #5389)
fully adapt `Accounts` as `ReactNode`
fully adapt `SharedSettings` as `ReactNode`
fully adapt `DirectoryChooser` as `ReactNode`
most of `ServerSettings` as `ReactNode`
auto disable `ServerSettings` inputs when user do not have perm to modify
get rid off some 'any' in `ServerSettings` and more
better handle connection-lost vs error

## playwright
added teardown to playwright test
(allow to run it twice or more from `npx playwright test --ui`)
removed playwright screenshoots

## java
fix TMDB `NPE`
nicely give empty result for `getFeedTitle` and `getFeedUrlFromWebpage` instead of `NPE`

## NSIS
fix NSIS update script
removed heapmem from setup and exe
added setup ProductName, etc
added detail on uninstaller creation instead of leaving the last file copied.

# Checklist


SubJunk:
Is this ready to test/review?

SurfaceS:
> Is this ready to test/review?

@SubJunk yes, it is a good start for review

SubJunk:
@SurfaceS I'm testing, how do I see the login carousel? If I log out, it just shows me the normal form

SurfaceS:
Server settings, under gui and player
As we present users, it is a security choice as end users may want to hide login usernames

SubJunk:
I see, but that's why I made it use local storage in the browser - that way the user's browser will remember them (with the option to forget them too) but anyone trying in a different browser will not see the username

SubJunk:
Also I think carousel isn't the right choice for the component because I can't click on the one I want, I have to cycle through them

I should have written about these details in the issue but I didn't realize you would change them

SubJunk:
Maybe we can have a blend of both ways:
By default, it uses local storage, but in the settings they can enable the users the way you did it.
Then we change the carousel to selectable elements so they can be clicked

What do you think?

SurfaceS:
I'll implement it today or tomorrow

SurfaceS:
@SubJunk done

SubJunk:
@SurfaceS this is great, thank you

It would also be cool to have a "Forget user" button below the "Log in" button, to remove that user from the local storage

SubJunk:
This is working great, thanks!

SubJunk (approved):

PR #5098, opened 2024-11-07
Recently added datefix

SubJunk:
# Description of code changes

Fixes https://github.com/UniversalMediaServer/UniversalMediaServer/issues/4836

# Checklist
- [ ] Any relevant issues are [referenced](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#referencing-issues-and-pull-requests)
- [ ] Any relevant [Knowledge Base](https://github.com/UniversalMediaServer/knowledge-base) articles are written/modified
- [ ] Any relevant tests have been written/modified

PR #5396, opened 2025-03-23
V15 - recently added + resource unique id

SurfaceS:
# Description of code changes

Be careful because it updates the version of the `FILES` table.

## media date discovered
implements #5098
`DATEADDED`
With `RUID` below, we will be able to keep this forever when it will be fully implemented.

## resource unique id
added resource unique id (`RUID`)

I created `ResourceIdentifier` class few months ago.
Then we are sure 2 files are the same file, even if we rediscover it after a move, rewrite, delete, copy, or rediscovered on share added and so on.

Hashing a file cost about 200 ms once when created/discovered.

I added this now, even if we do not yet use the `RUID`, as 200 ms is nothing when you add a file, but can be long when you upgrade the database of 1000+ media that need to be hashed.
This for avoiding a really long database upgrade, that may let users thinking the server is buggy and then they close the database (with potentially breaking of table).

<ins>In depth :</ins>
It is hashing the file with a lot of tweak especially for media files.
As media files are big (sometimes more than 10 GB), we cannot hash the all files (expensive in CPU resource, disk read, etc), and do not need to.
We know metadata sit at start or end of the files.
So basically, changing metadata will change either the first or last bytes, and changing a stream in the file will change the file size, or bytes positions on the file.

If `uri` is a file with less than 20MB size (small file), it hashes the file size, then all the bytes in the file.
If `uri` is a file with more than 20MB size (big file), it hashes the file size, then 5 x 4MB (start, 1/3, 1/2, 2/3 and end) of the file.
If `uri` is not a file (text), it hashes the `uri` as string.

# Checklist


SurfaceS:
@SubJunk 
I was thinking it force the table recreation, but in fact it only warn.

```
			if (version < TABLE_VERSION) {
				upgradeTable(connection, version);
			} else if (version > TABLE_VERSION) {
				LOGGER.warn(LOG_TABLE_NEWER_VERSION, DATABASE_NAME, TABLE_NAME);
			}
```

SubJunk (approved):

SubJunk:
Cool, I like this. It's similar to the OpenSubtitles hashing too, for similar reasons. Merge it if you want

Issue #5189, opened 2024-12-24
Directories that used to work now crash in the last few versions


jt-gilkeson:
The last few versions of UMS have started having this problem, I'll try to open a folder that I used to be able to view without any issue and now I'll get 401 errors.  Fresh install does not fix.

Both Windows x86 and now x86_64 have this problem.  

01:40:24 INFO  Language not specified, defaulting to OS language.
01:40:29 INFO  Starting Universal Media Server 14.8.0
01:40:29 INFO  Based on PS3 Media Server by shagrath, copyright 2008-2014
01:40:29 INFO  https://www.universalmediaserver.com
01:40:29 INFO  
01:40:29 INFO  Build: aeebf4a95 (2024-11-25)

02:08:46 ERROR Exception in result creation "Comparison method violates its general contract!"
java.lang.IllegalArgumentException: Comparison method violates its general contract!
	at java.base/java.util.TimSort.mergeHi(Unknown Source)
	at java.base/java.util.TimSort.mergeAt(Unknown Source)
	at java.base/java.util.TimSort.mergeCollapse(Unknown Source)
	at java.base/java.util.TimSort.sort(Unknown Source)
	at java.base/java.util.Arrays.sort(Unknown Source)
	at java.base/java.util.ArrayList.sort(Unknown Source)
	at java.base/java.util.Collections.sort(Unknown Source)
	at net.pms.store.utils.StoreResourceSorter.sortResourcesByModifiedDate(StoreResourceSorter.java:166)
	at net.pms.store.utils.StoreResourceSorter.sortResourcesByDefault(StoreResourceSorter.java:111)
	at net.pms.store.utils.StoreResourceSorter.sortResourcesByDefault(StoreResourceSorter.java:95)
	at net.pms.store.StoreContainer.sortChildrenIfNeeded(StoreContainer.java:460)
	at net.pms.store.container.VirtualFolder.discoverChildren(VirtualFolder.java:324)
	at net.pms.store.container.VirtualFolder.doRefreshChildren(VirtualFolder.java:373)
	at net.pms.store.StoreContainer.refreshChildren(StoreContainer.java:767)
	at net.pms.store.StoreContainer.refreshChildrenIfNeeded(StoreContainer.java:466)
	at net.pms.store.MediaStore.getResources(MediaStore.java:659)
[...]

PR #5468, opened 2025-04-17
main fix #5139

SurfaceS:
# Description of code changes

fix #5139

trace Invalid path for file
trace `IOException` & `SecurityException`
catch error and trace `sortResourcesByModifiedDate`
implements sort `#--TRANSCODE--#` folder children reflecting the base folder
do not sort `ChapterFileTranscodeVirtualFolder` (chapters order)
lock instance before sort children (may be filled by `runnable`)

catch `NullBoxIdException` on `JaudiotaggerParser`

https://github.com/UniversalMediaServer/UniversalMediaServer/issues/5189#issuecomment-2815265561

### DIFF
--- a/src/main/java/net/pms/parsers/JaudiotaggerParser.java
+++ b/src/main/java/net/pms/parsers/JaudiotaggerParser.java
@@ -40,14 +40,10 @@ import org.jaudiotagger.audio.AudioFile;
 import org.jaudiotagger.audio.AudioFileIO;
 import org.jaudiotagger.audio.AudioHeader;
 import org.jaudiotagger.audio.exceptions.CannotReadException;
-import org.jaudiotagger.audio.exceptions.InvalidAudioFrameException;
-import org.jaudiotagger.audio.exceptions.NullBoxIdException;
-import org.jaudiotagger.audio.exceptions.ReadOnlyFileException;
 import org.jaudiotagger.logging.ErrorMessage;
 import org.jaudiotagger.tag.FieldKey;
 import org.jaudiotagger.tag.KeyNotFoundException;
 import org.jaudiotagger.tag.Tag;
-import org.jaudiotagger.tag.TagException;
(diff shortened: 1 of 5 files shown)


SubJunk:
That was a lot of work, great job!

src/main/java/net/pms/parsers/JaudiotaggerParser.java
```diff
@@ -199,7 +200,7 @@ public static void parse(File file, MediaAudioMetadata audioMetadata) {
 			}
 			addMusicBrainzIDs(af, audioMetadata);
 			addAudioTrackRating(af, audioMetadata);
-		} catch (IOException | CannotReadException | InvalidAudioFrameException | ReadOnlyFileException | TagException e) {
+		} catch (IOException | CannotReadException | InvalidAudioFrameException | NullBoxIdException | ReadOnlyFileException | TagException e) {
```
SubJunk, line 203: This is why I like the old exception syntax of `Exception e`, and I often refactor this multicatch syntax to the old one - for example here recently https://github.com/UniversalMediaServer/UniversalMediaServer/commit/db76342bc677350f691a91eb5895fdde42feb002#diff-a290dd06f8ac9f7150c0d0cf38c4debd445772b457a5814b060f757c6ee9c474R180-R183. It's just nicer to know you will catch all exceptions IMO. What do you think?

SurfaceS: Thing is that custom exception implementation (eg `org.jaudiotagger.audio.exceptions.CannotReadException`) should **ideally** extends an internal Exception (eg `org.jaudiotagger.JaudioTaggerException`) then `java.lang.Exception` instead of directly extends `java.lang.Exception`

Then, you catch everything with `org.jaudiotagger.JaudioTaggerException`.

I changed code.

SubJunk (approved):

PR #5680, opened 2025-08-04
Add support for yt-dlp as a fallback in the YoutubeDl encoder

navaneedan07:
# Description of code changes

...

# Checklist
### Summary

This PR introduces three enhancements to the `YoutubeDl` encoder class:

1. **yt-dlp Fallback**: If `youtube-dl` is not found, the encoder now falls back to using `yt-dlp`, improving compatibility with modern YouTube formats.
2. **Video ID Extraction**: Added a helper method to extract the video ID from various YouTube URL formats (`watch?v=`, `youtu.be/`, `/embed/`) for better logging and traceability.
3. **Improved Logging**: Captures stderr output from the downloader process using `setErrorConsumer`, aiding in debugging failed downloads.

### Motivation

These changes improve robustness, maintainability, and observability of the webstreaming engine. They also prepare the encoder for future enhancements like subtitle support or metadata tagging.

### Testing

- Verified fallback to `yt-dlp` on systems without `youtube-dl`
- Confirmed correct video ID extraction across multiple URL formats
- Observed stderr logs during failed downloads

Let me know if you'd like this broken into smaller PRs or if there's a preferred way to expose downloader selection via configuration.

### DIFF
--- a/src/main/java/net/pms/encoders/YoutubeDl.java
+++ b/src/main/java/net/pms/encoders/YoutubeDl.java
@@ -17,6 +17,8 @@
 package net.pms.encoders;
 
 import com.sun.jna.Platform;
+
+import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
@@ -33,6 +35,31 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class YoutubeDl extends FFMpegVideo {
+
+	String extractVideoId(String url) {
+		if (url == null)
+			return null;
+
+		if (url.contains("v=")) {
+			int start = url.indexOf("v=") + 2;
+			int end = url.indexOf("&", start);
+			return end == -1 ? url.substring(start) : url.substring(start, end);
+		}
+
+		if (url.contains("youtu.be/")) {
+			int start = url.indexOf("youtu.be/") + 9;
+			int end = url.indexOf("?", start);
+			return end == -1 ? url.substring(start) : url.substring(start, end);
+		}
+
+		if (url.contains("/embed/")) {
+			int start = url.indexOf("/embed/") + 7;
+			int end = url.indexOf("?", start);
+			return end == -1 ? url.substring(start) : url.substring(start, end);
+		}
+
+		return null;
+	}
 
 	private static final Logger LOGGER = LoggerFactory.getLogger(YoutubeDl.class);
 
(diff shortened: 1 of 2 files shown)


src/main/java/net/pms/encoders/YoutubeDl.java
```diff
@@ -66,21 +93,29 @@ public boolean isTimeSeekable() {
...
+
 		setAudioAndSubs(resource, params);
 
 		// Build the command line
 		List<String> cmdList = new ArrayList<>();
 
-		cmdList.add(configuration.getYoutubeDlPath());
+		String downloaderPath = configuration.getYoutubeDlPath();
+		if (!new File(downloaderPath).exists()) {
+			downloaderPath = "/usr/bin/yt-dlp"; // fallback path
```
SubJunk, line 115: It would be good to make this work across platforms too, if you dig into that `getYoutubeDlPath` you'll see how we usually do that.

src/main/java/net/pms/encoders/YoutubeDl.java
```diff
@@ -66,21 +93,29 @@ public boolean isTimeSeekable() {
 
 	@Override
 	public synchronized ProcessWrapper launchTranscode(
-		StoreItem resource,
-		MediaInfo media,
-		OutputParams params
-	) throws IOException {
+			StoreItem resource,
+			MediaInfo media,
+			OutputParams params) throws IOException {
```
SubJunk, line 98:
```suggestion
		StoreItem resource,
		MediaInfo media,
		OutputParams params
	) throws IOException {
```

src/main/java/net/pms/encoders/YoutubeDl.java
```diff
@@ -34,6 +36,31 @@
 
 public class YoutubeDl extends FFMpegVideo {
 
+	String extractVideoId(String url) {
```
SubJunk, line 39: It would be good to rename this to indicate it is YouTube-specific, because youtube-dl supports a lot of other websites too (a future goal could be to add support for all of the sites)

src/main/java/net/pms/encoders/YoutubeDl.java
```diff
@@ -123,7 +155,8 @@ public synchronized ProcessWrapper launchTranscode(
 
 		if (!directPipe) {
 			ProcessWrapper mkfifoProcess = pipe.getPipeProcess();
-			pw.attachProcess(mkfifoProcess); // Clean up the mkfifo process when the transcode ends
+			((ProcessWrapperImpl) pw).attachProcess(mkfifoProcess); // Clean up the mkfifo process when the transcode
+																	// ends
```
SubJunk, line 159:
```suggestion
			((ProcessWrapperImpl) pw).attachProcess(mkfifoProcess); // Clean up the mkfifo process when the transcode ends
```

src/main/java/net/pms/encoders/YoutubeDl.java
```diff
@@ -100,11 +135,8 @@ public synchronized ProcessWrapper launchTranscode(
 			params.setInputPipes(new IPipeProcess[2]);
 		} else {
 			// basename of the named pipe:
-			String fifoName = String.format(
-				"youtubedl_%d_%d",
-				Thread.currentThread().getId(),
-				System.currentTimeMillis()
-			);
+			String fifoName = String.format("youtubedl_%d_%d", Thread.currentThread().threadId(),
+					System.currentTimeMillis());
```
SubJunk, line 139:
```suggestion
			String fifoName = String.format(
				"youtubedl_%d_%d",
				Thread.currentThread().threadId(),
				System.currentTimeMillis()
			);
```

SubJunk (changes requested):
Thanks for this PR! There's good work in here, and especially thanks for adding tests. I left more specific comments in the code

src/main/java/net/pms/configuration/UmsConfiguration.java
```diff
@@ -1605,7 +1605,20 @@ public String getYoutubeDlPath() {
...
+
+		//Fallback paths for OS
+		if (Platform.isWindows()) {
+			return "C:\\Program Files\\yt-dlp\\yt-dlp.exe";
+		}
+		else if (Platform.isMac()) {
+			return "/usr/local/bin/yt-dlp";
+		}
+		else {
+			return "/usr/bin/yt-dlp";
```
SubJunk, line 1620: I think it would be good to make this more standardized, so there should be `ytDlpPaths.getDefaultPath();` in this file, and then you can do the same checking that is done on line 1608 for yt-dlp.
I would be happy to help if you want

SubJunk (changes requested):

src/main/java/net/pms/configuration/UmsConfiguration.java
navaneedan07: Thanks for the suggestion! I've updated the method to use _**ytDlpPaths.getDefaultPath()**_ as the primary path, and added fallback logic similar to line 1608. This should make the behavior more consistent and easier to maintain.

SubJunk:
@navaneedan07 I think youtube-dl is broken now so I will switch to yt-dlp completely. I have already started that, so this PR isn't needed anymore. Thank you :)

Issue #5689, opened 2025-08-11
Devices spamming requests causing "[Fatal Error] :1:2: Reference is not allowed in prolog"


Mik-S-UMS:
Got this interesting problem on the [forum](https://www.universalmediaserver.com/forum/viewtopic.php?t=16655) where this error message would keep coming up but besides that everything else looked fine. The one weird thing was the size of the logs being over 200mb for a few seconds of logging.

After looking to see why I saw repeated messages to the web UI every few milliseconds
```
ServerSentEvents send message: {"id":3,"name":"APTV_9","address":"192.168.20.9","uuid":"898f9738-d930-4db4-a3cf-d80f99888515","icon":"wmp.png","iconOverlays":"","playing":" ","time":" ","progressPercent":0,"userId":-1,"state":{"playback":0,"mute":false,"volume":50,"buffer":0,"name":" "},"isActive":true,"isAllowed":true,"isAuthenticated":false,"controls":3,"action":"renderer_update"}
```

and looking further there was an equal number of requests coming from the renderer

```
Received a request from 192.168.20.9:
============================= UPNP HTTP SERVER REQUEST BEGIN ================================
NOTIFY /dev/898f9738-d930-4db4-a3cf-d80f99888515/svc/upnp-org/RenderingControl/event/cb HTTP/1.1

HEADER:
...

CONTENT:
<e:propertyset xmlns:e="urn:schemas-upnp-org:event-1-0">
    <e:property>
        <LastChange>&lt;Event xmlns="urn:schemas-upnp-org:metadata-1-0/RCS/"&gt;&lt;InstanceID val="0"&gt;&lt;Volume Channel="Master" val="50"/&gt;&lt;Mute Channel="Master" val="0"/&gt;&lt;PresetNameList val="FactoryDefaults"/&gt;&lt;VolumeDB Channel="Master" val="0"/&gt;&lt;/InstanceID&gt;&lt;/Event&gt;</LastChange>
    </e:property>
</e:propertyset>

============================= UPNP HTTP SERVER REQUEST END ==================================
```
and this may have been tripping up UMS as it cannot keep up.

This apparently came from a screen mirroring app on the TV and removing it fixed the problem.

I don't know if the app was responding to a request from UMS first though

I think UMS needs some protection against these apps or renderers that constantly spam requests and either ignore them if there are too frequent or block them.

SubJunk:
@Mik-S-UMS did the user try blocking that app in UMS?

[...]

PR #5700, opened 2025-08-15
Do not subscribe to UPnP events while getting DeviceControls.

ik666:
Subscribing to UPnP events shall not be done while getting the DeviceControls but only if the renderer needs a renewal. This might also help to reduce log messages (not checked) reported in #5689.


SubJunk:
The change looks correct to me. I can see we are still doing the subscription when `addRenderer` and `rendererUpdated` are called. I will test my devices with this change before merging

SubJunk (approved):

SubJunk:
I tested this on my TV and VLC iOS app and it all works with reconnections etc so I will merge it

PR #5570, opened 2025-05-28
Revert "fix : discover children of children"

SubJunk:
Reverts UniversalMediaServer/UniversalMediaServer#5352

@ik666 I think this is responsible for a failure in the web tests in v15 which might mean it introduces a significant performance problem or something

PR #5717, opened 2025-08-28
do not discover children of children

ik666:
Picking up discussion in #5570.

The E2E test fails because of an UI issue.

The added Youtube channel was given twice in the `SharedContentArray` which is later processed by `SharedContentConfiguration.updateSharedContent`. The JSON content comes directly from the UI. There is no further JAVA processing.

The playwright test strictly checks that exact one item exists. However, the UI has send 2 items with the same name resulting in an error on the strict check. Brief check on v15 looked ok for me though!

This issue reported on the [forum](https://www.universalmediaserver.com/forum/viewtopic.php?t=16671) is most likely related to this PR. Speed improvement is noticeably.

### DIFF
--- a/src/main/java/net/pms/store/container/Feed.java
+++ b/src/main/java/net/pms/store/container/Feed.java
@@ -60,6 +60,7 @@ public class Feed extends StoreContainer {
 		super(renderer, name, null);
 		childSpecificType = type;
 		this.url = getFeedUrl(url);
+		discoverChildren();
 	}
 
 	public void parse() throws Exception {
@@ -341,5 +342,11 @@ public class Feed extends StoreContainer {
 		}
 
 		return url;
+	}
+
+	@Override
+	public void discoverChildren() {
+		super.discoverChildren();
+		syncResolve();
 	}
 }


SubJunk:
> The playwright test strictly checks that exact one item exists

Are you sure? The test failure says `waiting for getByText('Kurzgesagt – In a Nutshell')` and hits the timeout. That seems like a different error

ik666:
This is another error. I was on the other branch you've mentioned. The test initially failed in the stage `should be able to add a YouTube channel as a video feed`. With current v15 this test stage seems to pass without error but failing playing the video.

ik666:
@SubJunk I fixed the tests for current main v15 branch.

SubJunk:
Have you tested much with this fix? Does the MediaStore still populate itself properly if you keep browsing different levels?

SubJunk:
I am testing it now

ik666:
Good idea. Potentially, all StoreResources could be affected. However, I don't know how i.e. "ZippedFolder" oder "ZippedFile" actually work with UMS.

SubJunk:
I haven't noticed any problems during my normal use. But I didn't have any problems before this either.

ik666:
In case of an issue with other objects: This can be fixed quickly if needed. We’d just need to call `discoverChildren();` at the end of the constructor to ensure child content is discovered during initialization.

SubJunk (approved):

Issue #5736, opened 2025-09-10
Plain list m3u playlists appear empty unless #EXTM3U is present


CyDragon80:
Observed in UMS Version 14.13.0 and 14.13.1 under Windows 11
Observed when viewed via Heos app, VLC android (I could test some others if needed though)

Took a little fiddling around to figure this out, but it appears that a plain file list m3u playlist will show as empty, unless #EXTM3U is present at the top of the file.  While this header seems to be required as part of the "Extended M3U" specification, m3u itself has no real spec other than an expectation for a list of path names. "Extended M3U" seems to be considered an optional extension within m3u and not all programs that save playlists as m3u include it. This can lead to some confusion.

(Apologies if this was addressed elsewhere, my preliminary searches did not come up with anything. However I may have missed the right keywords.)

CyDragon80:
A tip for foobar2000 users who might stumble across this. Under Advanced in Preferences, there appears to be a "Write EXTM3U playlists" option that you can select.

SubJunk:
Merged the fix, thanks again!

PR #5738, opened 2025-09-13
Issue 5736: If play list has an extension and it is a known m3u variant, assume it is m3u without checking for headers.

CyDragon80:
Crude workaround for https://github.com/UniversalMediaServer/UniversalMediaServer/issues/5736

It does feel a little redundant to recheck the extension when it has potentially been extracted before, but it gets around the immediate issue. It appears the code is wholly reliant on detecting a header line to determine m3u vs pls parsing, which means if the m3u has no extension header, then it does not parse. PlaylistManager technically seems to create based on extension already, but there are other forms of construction.

I did not want to go touching a lot of the code, but if someone more familiar wishes to address the issue in a deeper way, then by all means. At the least, I hope this PR helps in narrowing down the area of the issue mentioned or act as a short term stop gap.

I did not see any direct test cases to modify or documentation to update. However I did not search exhaustively, my apologies if missed. I was tempted to create a test for the m3u, but was not sure how much of the repo I should disturb.

On a side note, it was not related to the issue, but this conditional includes a comparison to "cue" with the dot which on the surface seems like bug given the util functions exclude the dot. so when I move the checks to be together I removed the dot. However it seems like there is another class CueFolder so perhaps the check is unused anyway.
> if (extension != null && (extension.equals("m3u8") || extension.equals(".cue"))) {

...

# Checklist
- [X] Any relevant issues are [referenced](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#referencing-issues-and-pull-requests)
- [X] Any relevant [Knowledge Base](https://github.com/UniversalMediaServer/knowledge-base) articles are written/modified
- [X] Any relevant tests have been written/modified

### DIFF
--- a/src/main/java/net/pms/store/container/PlaylistFolder.java
+++ b/src/main/java/net/pms/store/container/PlaylistFolder.java
@@ -100,9 +100,9 @@ public final class PlaylistFolder extends StoreContainer {
 		resolve();
 	}
 
-	private BufferedReader getBufferedReader() throws IOException {
+	/** Attempts to return a lowercase file extension from uri (excluding the dot) or null */
+	private String getExtension() {
 		String extension;
-		Charset charset;
 		if (FileUtil.isUrl(uri)) {
 			extension = FileUtil.getUrlExtension(uri);
 		} else {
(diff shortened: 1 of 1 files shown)


ik666:
@CyDragon80 This PR makes the code more clear and reliable.
@SubJunk I would merge it.

SubJunk:
@CyDragon80 thanks for this

PR #5777, opened 2025-09-30
FileWatcher UPDATE event / invalidate cache

ik666:
This PR fixes an issue where file updates (e.g. metadata changes) weren’t reflected at the control point.

Previously, the database was updated correctly, but because `MediaInfo` was cached, the control point was still receiving a stale object.

The changes address this by refreshing the cached MediaInfo when a file is rescanned, so both control points and the UMS UI always see the latest metadata.

### DIFF
--- a/src/main/java/net/pms/store/MediaScanner.java
+++ b/src/main/java/net/pms/store/MediaScanner.java
@@ -40,10 +40,8 @@ import net.pms.renderers.devices.MediaScannerDevice;
 import net.pms.store.container.DVDISOFile;
 import net.pms.store.container.PlaylistFolder;
 import net.pms.store.container.RealFolder;
-import net.pms.store.item.RealFile;
 import net.pms.util.FileUtil;
 import net.pms.util.FileWatcher;
-import net.pms.util.InputFile;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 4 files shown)


SubJunk:
Looks good, is it ready for review/merge?

src/main/java/net/pms/store/MediaScanner.java
```diff
@@ -504,19 +525,24 @@ private static void removeFileEntry(String filename) {
...
 					removeFolderEntry(filename);
 				}
 			} else {
 				if (ENTRY_CREATE.equals(event)) {
-					parseFileEntry(new File(filename), true, false);
+					parseFileEntry(f, true, false);
 				} else if (ENTRY_DELETE.equals(event)) {
 					removeFileEntry(filename);
+				} else if (ENTRY_MODIFY.equals(event)) {
+					parseFileEntry(f, false, false);
```
ik666, line 541: @SubJunk Is Line 541 still needed? Not sure, what's exactly suppose to do.

ik666:
@SubJunk I think it can be merged now.

src/main/java/net/pms/store/MediaScanner.java
SubJunk: Looks like it re-parses a modified file, but I'm not sure if it works or just hits the cache

SubJunk (approved):

Issue #5295, opened 2025-02-21
Playback history


SubJunk:
At the moment we only somewhat have long-term playback history - you can go into the `Recently Played` folder in the `Media Library` and see the last 100 items. That's good, but it only has _existing_ items, not ones you removed or moved.

We should have a feature on the web player that can show all media you ever played, including the API metadata and ideally whether it was fully played.

It could also be built on later to have things like ratings and comments, which would also show in the normal web player. But the first step is to add this concept of storing info about media even after it has been removed.

Issue #4979, opened 2024-09-20
Add database repair/deletion options


Mik-S-UMS:
It is not uncommon for users to have database problems after upgrading and they find they are unable to find any content on their renderers. Just had another one [here](https://www.universalmediaserver.com/forum/viewtopic.php?t=16143).

I have noticed in the logs that it mentions you can use a repair tool to recover it but I have no idea what this is or how to use it, so the only solution I know is to delete the database so a new one is made.

`
net.pms.database.DatabaseEmbedded Database migration failed: File corrupted while reading record: "C:/ProgramData/UMS/database/medias.mv.db". Possible solution: use the recovery tool [90030-214]
`

I think there should be some notification or something obvious in the settings that shows if there is a problem connecting to the database and gives the option to backup, restore, delete, or repair it without needing to manually doing it.

SubJunk:
Yeah this is a good idea. I'm disappointed to see how often this is happening. It also makes me wonder whether h2database is the right tool for us. We don't do a lot of technical things with it

PR #5778, opened 2025-09-30
rating backup/restore support

ik666:
This PR introduces a Backup Manager for saving and restoring rating information (currently for audio resources).

- Backup: Stores the rating data in the file system.
- Restore: Reads backup file and updates the database. Supports multiple copies of the same file within the filesystem.

This feature is especially useful after a database restore/recreation, since rating information are currently lost in that process. Resources are now identified using the newly introduced resource ID.


Mik-S-UMS:
It would be good if it can also backup played status too for those times the database is corrupted on an update. 
I think everything else can be rebuilt from a rescan.

I made #4979 a while ago that some database management could be added to UMS to back up, repair, or delete it if needed.

SubJunk:
@ik666 cool feature!

@Mik-S-UMS I like that suggestion too

ik666:
@SubJunk Yes, I thinks it works as expected. It has to be linked in the UMS UI. Maybe you could do it. And I would like to connect it to the UMS service PR #5293 ... 

@Mik-S-UMS I'll have a look, where the "played" status is held and provide a solution for that. You refer to a video that is fully played, correct?

Mik-S-UMS:
Yeah fully played status might be enough if there is a table that holds this and is easy to implement, but if possible partially played information would be good too.

In the logs it does mention a `getLastStartPosition` when making the decision if something has been fully played so I assume this is stored somewhere in the database.

ik666:
I can make a BackupManager for video files as well. There is a `files_status` table, which holds those information. However, the information are associated with the filename. I think some issues were reported, where all status information were lost after a file movement. Therefore, the code should probably refactored to use the resource id of a file.

SubJunk:
> Therefore, the code should probably refactored to use the resource id of a file.

Yes, and that refactoring needs to happen for this issue too https://github.com/UniversalMediaServer/UniversalMediaServer/issues/5295

So not just a resource ID, but also a TMDB ID, so that if the user gets another version of a file (maybe they had 1080p but then upgraded to 2160p), their UMS still knows that they fully played that TMDB ID

ik666:
@SubJunk  Added a files status backup manager without the mentioned refactoring. I've tested it briefly, but there is no interaction with the UI yet.

SubJunk:
Is this ready for testing/review?

ik666:
Yes. Please remember, that it's not linked into the UI yet. I've linked it to the UPnP service which is not merged yet.

SubJunk:
@ik666 how can I test it?

ik666:
Maybe you could create 2 buttons in the UI and point to the "backup" and "restore" methods of the BackupManager?

SubJunk:
Ok, I'll do that

ik666:
@SubJunk I added a test case in `MediaTableFilesStatusTest.testBackupOfIsFullyPlayed` which tests the FilesStatusBackupManager by running the "IsFullyPlayed" test enhanced in this order
- insert "fullyPlayed" data
- backups the files table
- drops all tables
- creates new tables
- restores the backup

and checks if the "IsFullyPlayed" status are still available.

PR #5780, opened 2025-10-02
Docs: Workaround for UMS ignoring SHARED.conf on reboot

gms004a:
This is a user+AI assisted solution to preserve and restore shared folders on UMS restarts/reboots. 
Includes scripts and systemd units to automatically repopulate UMS.conf from SHARED.conf.


Mik-S-UMS:
The devs will be able to give more information but UMS does not use the shared folder configuration from UMS.conf any more. 

This was migrated to SHARED.conf along with virtual folders and web content and should be ignoring what was set in UMS.conf if SHARED.conf exists. If it doesn't and `shared_folders` is present in UMS.conf then a new SHARED.conf will be made with this information.

If UMS is only reading the shares in UMS.conf then this could be a bug. 
Try deleting `shared_folders =` from UMS.conf so it should only be using SHARED.conf.

I think I have seen mention of network shares not working because they are not ready when UMS is started up with the system and the fix was to delay UMS running till they are active. I can't find that on the forum at the minute though. 
I think this is the reason why this method is working as it essentially delays UMS by restarting it again.

SubJunk:
I think the way forward for this is to make an issue for the problem so we can fix it instead, can you please make the issue?

PR #5784, opened 2025-10-05
Revert "FileWatcher UPDATE event / invalidate cache"

SubJunk:
@ik666 I have been getting an unresponsive server over both UPnP and the web player since this merged, it has happened 3 times now. It works at first, but then after I add or delete files, the server can't be browsed.

Reverts UniversalMediaServer/UniversalMediaServer#5777

### DIFF
--- a/src/main/java/net/pms/renderers/ConnectedRenderers.java
+++ b/src/main/java/net/pms/renderers/ConnectedRenderers.java
@@ -16,7 +16,6 @@
  */
 package net.pms.renderers;
 
-import java.io.File;
 import java.net.InetAddress;
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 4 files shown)


ik666:
OK, I'll have a look at it.

ik666:
@SubJunk Do you possible have a log or can explain steps to reproduce this issue? I copy / delete / update files 10 minutes in a row, but all seems to be ok ...

SubJunk:
There was nothing in the logs when I looked but it might be related to the "fully played action" because I use "move file to a different folder and mark it", can you try that? Otherwise I can upload logs in case you see something I missed

ik666:
I think I got it. When I did bulk updated of files (like updating album information to audio files), the system got stuck.

It's related to the review line I've asked you (parseFileEntry, updateFileEntry), if it's still needed. They've blocked each other when many change have been made. But we don't need both lines, there doing the same ... I am making a local build and check if the blocking issue is gone.

PR #5792, opened 2025-10-09
Check/filescan update

ik666:
Should fix issues discovered in #5784 and updates #5777.

I made bulk file updates and it didn't block on my side. @SubJunk Maybe you could also do a check.

### DIFF
--- a/src/main/java/net/pms/store/item/RealFile.java
+++ b/src/main/java/net/pms/store/item/RealFile.java
@@ -343,7 +343,8 @@ public class RealFile extends StoreItem implements SystemFileResource {
 	 * @param inputFile
 	 */
 	protected void checkCoverThumb() {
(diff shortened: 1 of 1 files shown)


SubJunk:
LGTM, thanks

SubJunk:
@ik666 I am still getting the problem sometimes, it is more rare but I have had it happen 2 times since this merged again, and the only way to get the server responsive again is to restart it

My logs show:
```
[File event] Database error while trying to add parsed information for "file" to the cache: Cannot invoke "java.sql.Connection.prepareStatement(String)" because "connection" is null
```

SubJunk:
Sorry I have to revert it again because it will block me from doing a release

ik666:
Too bad ... Maybe some deadlock situation. I'll check.

PR #5868, opened 2025-11-20
AudioAddict / network timeout / init failure

ik666:
In case of a timeout while reading network data, channels are left uninitialized. This PR tries to recover from an exception and updates the children concurrently.

### DIFF
--- a/src/main/java/net/pms/store/container/audioaddict/INetworkInitilized.java
+++ b/src/main/java/net/pms/store/container/audioaddict/INetworkInitilized.java
@@ -1,6 +1,0 @@
-package net.pms.store.container.audioaddict;
-
-public interface INetworkInitilized {
-
-	public void networkInitilized();
-}
--- a/src/main/java/net/pms/external/audioaddict/RadioNetwork.java
+++ b/src/main/java/net/pms/external/audioaddict/RadioNetwork.java
@@ -35,7 +35,7 @@ import net.pms.external.audioaddict.mapper.ChannelFilter;
 import net.pms.external.audioaddict.mapper.ChannelJson;
 import net.pms.external.audioaddict.mapper.Favorite;
 import net.pms.external.audioaddict.mapper.Root;
-import net.pms.store.container.audioaddict.INetworkInitilized;
+import net.pms.store.container.audioaddict.INetworkInitialized;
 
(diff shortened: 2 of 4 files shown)


SubJunk:
@ik666 is this tested and ready to review/merge?

ik666:
@SubJunk No, please wait ...

ik666:
Works now as expected but : @SubJunk Do you know, why the ui test fails?

SubJunk:
I'm not sure from looking at it but the commit that caused the failure (https://github.com/UniversalMediaServer/UniversalMediaServer/pull/5868/commits/6cf29d89634c66bbdcd09874dc32d978c3a20700) is small so maybe try reverting part of it to see if it is fixed. I would guess it is related to the changes in `src/main/java/net/pms/renderers/Renderer.java`

ik666:
@SubJunk Looks ok now, but 1 test is canceled / failed with : 

> The macOS-13 based runner images are now retired. For more details, see https://github.com/actions/runner-images/issues/13046.

I think this is more pipeline related.

SubJunk:
@ik666 I fixed that now in `main`, can you update the branch?

ik666:
@SubJunk That did it ...

src/main/java/net/pms/external/audioaddict/AudioAddictService.java
```diff
@@ -53,6 +53,7 @@ private synchronized RadioNetwork getNetwork(Platform platform) {
 		RadioNetwork network = networkSettings.get(platform);
 
 		if (network == null) {
+			LOGGER.info("{} : not yet initilized.", platform.displayName);
```
SubJunk:
```suggestion
			LOGGER.debug("Initializing network settings for {}", platform.displayName);
```

src/main/java/net/pms/store/container/audioaddict/INetworkInitilized.java
```diff
@@ -0,0 +1,6 @@
+package net.pms.store.container.audioaddict;
+
+public interface INetworkInitilized {
```
SubJunk: `Initilized` should be `Initialized` in all of these files

SubJunk (dismissed):

PR #5967, opened 2026-01-12
docs: clarify default values in UI tooltips

GitOn1t:
This PR adds a short clarification to the README explaining that some user interface options include tooltips indicating default values.

This helps users better understand initial configuration behavior without changing any existing functionality.


SubJunk:
Thanks, but I don't think this belongs in the readme

PR #5998, opened 2026-02-05
make multi module build (linux x86_64/arm64)

### LINKED ISSUE #4770: official ums docker-image for arm64
Hi Guys,
I found your docker image here.  https://hub.docker.com/r/universalmediaserver/ums/tags
Unfortunately only for amd64. Are there plans to provide an official docker image for arm64 as well?

ik666:
This PR make a docker multi-platform build and adds an arm64 target. If this works, it should resolve #4770. However, I'm not sure, if further steps to the build pipeline are needed ... we have to check it.
.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3706,7 +3706,7 @@
 								<configuration>
 									<target>
 										<exec executable="docker" dir="${project.basedir}/target/docker">
-											<arg line="build . -f Dockerfile -t universalmediaserver/ums:${project.version} -t universalmediaserver/ums:latest" />
+											<arg line="buildx build --platform linux/amd64,linux/arm64 . -f Dockerfile -t universalmediaserver/ums:${project.version} -t universalmediaserver/ums:latest" />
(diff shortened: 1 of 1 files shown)


SubJunk:
Oh cool! I hope it's this easy. Will try it

SubJunk:
Ideally it will also use different binaries for arm64 but this can be a first step

ik666:
> Ideally it will also use different binaries for arm64 but this can be a first step

Docker should do a cross platform build for the arm64 platform using the native arm64 binaries. Do you have any further binaries in mind?

SubJunk:
For our Linux ARM distributions we use different binaries for FFmpeg and yt-dlp so those would be good

ik666:
OK I see ...

SubJunk:
This seems to have worked well. I had to enable an extra setting in Docker to enable `containerd` but after that it was good and we now have a multi-platform image

PR #6012, opened 2026-02-12
Performance / content delivery

ik666:
I observed laggy behavior in UMS when browsing folders containing a large number of thumbnails (more than 1,000). Profiling identified the `copyStream` method in the `HttpServletHelper` class as the primary bottleneck.  

Most likely, it's only the flush call in the loop, but I've updated the implementation to use direct buffers and NIO channels, which reduces the overhead of the copy operation and results in faster response times.

### DIFF
--- a/src/main/java/net/pms/network/HttpServletHelper.java
+++ b/src/main/java/net/pms/network/HttpServletHelper.java
@@ -66,6 +66,7 @@ public abstract class HttpServletHelper extends HttpServlet {
 
 	protected static final UmsConfiguration CONFIGURATION = PMS.getConfiguration();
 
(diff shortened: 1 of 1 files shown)


SubJunk:
@ik666 cool, good idea. Have you tested it?

ik666:
I'm using it on my system for 2 or 3 days in many situations. Looks stable so far. The `copyStream` method doesn't show up in the profilers hotspot section any more.

SubJunk:
Great. I remember a long time ago I did a similar enhancement to our file reading methods to use `nio` and it had good performance effects. Nice one

PR #6038, opened 2026-02-23
Fix: Save button disappears instead of showing loading state in Shared Content settings

### LINKED ISSUE #6037: Add a loading indicator to the Shared Content area on the web settings
There is currently no loading indicator when you save settings, the Save button just disappears and then the message pops up in the lower right when it has finished saving. If my hard drive has gone to sleep so it needs a few seconds to wake up, that can take a while to appear so it seems frozen.

Acceptance criteria:
- [ ] The "Save" button in the Shared Content area in the web settings has a loading state instead of disappearing when clicked

WarisMaqbool-ship-it:
Fixes #6037

The Save button in the Shared Content area was disappearing when clicked 
because the button visibility was tied only to the `modified` state. 
When save is triggered, `modified` becomes false before the loading 
completes, causing the button to vanish.

The fix adds `isLoading` to the visibility condition so the button 
stays visible with a loading spinner while the save operation is in progress.

Changed:
{canModify && modified && (
to:
{canModify && (modified || isLoading) && (

Note: Unable to run the full Java backend locally for testing, 
but the UI logic has been verified.

### DIFF
--- a/react-client/src/components/SharedContent/SharedContent.tsx
+++ b/react-client/src/components/SharedContent/SharedContent.tsx
@@ -131,7 +131,7 @@ export default function SharedContent({ i18n, session }: { i18n: I18nInterface, 
           {sharedContents && configuration && (
             <SharedContentSettings i18n={i18n} canModify={canModify} scan={session.mediaScan} sharedContents={sharedContents} setSharedContents={setSharedContents} configuration={configuration} />
           )}
-          {canModify && modified && (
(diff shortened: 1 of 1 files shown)


SubJunk:
@WarisMaqbool-ship-it that was fast work, thanks. It failed on a linter check

SubJunk:
@WarisMaqbool-ship-it will you fix it?

PR #6087, opened 2026-03-23
Added Lucene search capabilities

ik666:
Added Apache Lucene search support.
=======================================================

This is a significant change in the way user can search for a term.

For the person browsing their media through a DLNA/UPnP client, Lucene-backed search feels "smarter" and faster:

- Typo Tolerance (Fuzzy Search): If a user misspells an album name (e.g., "Led Zeplin" instead of "Led Zeppelin"), Lucene's fuzzy search capabilities can still find the correct match.
- Near-Instant Autocomplete: Its high-speed lookup enables real-time "search-as-you-type" functionality, helping users find specific filenames or artists before they finish typing.
- Relevance Ranking: Rather than just returning an alphabetical list, Lucene can rank results by relevance. If a user searches for a specific track title, that exact match appear above albums that do not perfectly match.
- Linguistic Analysis (Stemming): Lucene can use stemming to treat related words similarly (e.g., a search for "running" might also find "run"). This is particularly useful for finding media with descriptive filenames.

DB changes:
- Switched thumbnail serialization from Java objects to byte[] for better efficiency and compatibility with lucene.


ik666:
I've played around with a few control points. The LINN app makes in some cases more complex UPnP searches, which wasn't addressed yet. I cleaned up the code by implementing an AST and parser. I also added "genre" to the lucene index. So a request like this one

```
upnp:class derivedfrom "object.item.audioItem" and ( dc:title contains "Love don" or upnp:album contains "Love don" 
or upnp:artist contains "Love don" or upnp:genre contains "Love don" )
```
or even nested queries will be interpreted as expected.

ik666:
I've added a test class `SearchRequestDatabaseTest` which performs all kind of real database tests. Real means that the Tables `StoreIds`, `Files`, and `AudioMetadata` are filled and the UPnP search requests are performed against the database using the same mechanism a real renderer would do it.

Tests include so far 
- "global" searches (search directory is "0" = root, which means a search over the entire database is performed
- "tree" searches which means a subtree search where the search directory is not "0"
for audio metadata and for playlist names

src/test/java/net/pms/network/mediaserver/handlers/SearchRequestDatabaseTest.java
```diff
@@ -14,61 +21,148 @@
...
-		new String[] {"OK Computer", "Radiohead", "Radiohead", "Paranoid don't Android", "Alternative", "1997", "2", "1", "Thom Yorke", null},
-		new String[] {"In Rainbows", "Radiohead", "Radiohead", "Nude", "Alternative", "2007", "3", "1", "Thom Yorke", null},
-		new String[] {"Requiem", "W. A. Mozart", "Various Artists", "Lacrimosa", "Classical", "1791", "8", "1", "W. A. Mozart", "Herbert von Karajan"},
-		new String[] {"Symphony No. 5", "Ludwig van Beethoven", "Various Artists", "I. Allegro con brio", "Classical", "1808", "1", "1", "Ludwig van Beethoven", "Carlos Kleiber"},
-		new String[] {"The Four Seasons", "Antonio Vivaldi", "Various Artists", "Spring: Allegro", "Classical", "1725", "1", "1", "Antonio Vivaldi", "Claudio Abbado"},
-		new String[] {"Appetite for Destruction", "Guns N' Roses", "Guns N' Roses", "Welcome to the Jungle", "Hard Rock", "1987", "1", "1", "Slash", null},
-		new String[] {"Master of Puppets", "Metallica", "Metallica", "Battery", "Metal", "1986", "1", "1", "James Hetfield", null}
+		// Bucket 1
+		new String[] {"The Wall", "Pink Floyd", "Pink Floyd", "In the Flesh?", "Rock", "1979", "1", "1", "Roger Waters", null, "/music/1970/In the Flesh?"},
```
valib: "?" is an illegal character in the path name in the Win OS. The same for ":" in line 76.

ik666: I see. Thanks for comment. I've updated the test case.

ik666:
Hmmm ... good to have the test cases, but the subtree search test fail on windows. Not sure why, since I don't have a system to test with. 

@SubJunk Is there a way to access the logfile of the windows build `C:\ProgramData\UMS-tests\debug.log` ?

@valib Are you able to debug/test it on windows, or can you supply the database (`medias.mv.db` and the folder `medias`) from the test build after the test was run?

Edit:
@valib Could you look into the h2 database and see if filenames in the table STORE_IDS and in the table FILES are the same (not mixed like "\\" and "/")?

ik666:
OK guys. Have figured it out by thinking about the problem :)

SubJunk:
This is exciting work!

Mik-S-UMS:
Could this be accessed from the web interface or is it only for clients and control points?

It was recently suggested on the [forum](https://www.universalmediaserver.com/forum/viewtopic.php?t=16826) if a search functionality could be added to the web interface and this seems to be doing most of the work towards it.

sf666:
Hello @Mik-S-UMS. Currently, this implementation is tied to the UPnP Search Request, making it accessible only to control points or 2-box setup devices (like TVs with embedded control points). However, adapting it for the UMS UI should be relatively straightforward.

I've created a reference implementation in [my web-based control point](https://sf666.github.io/nextcp2/user_interface/music_library/#upnp-search). While it lacks extended video support, it’s a solid alternative for users focused on audio playback with large libraries.

ik666:
@SubJunk I think I'm finished with this PR now.

SubJunk:
@ik666 can you please resolve conflicts?

ik666:
@SubJunk It's fixed. I think we should merge it since bigger changes have been done on the main branch.

src/main/java/net/pms/database/lucene/UmsFullTextLucene.java
```diff
@@ -0,0 +1,799 @@
...
+import org.h2.message.DbException;
+import org.h2.store.fs.FileUtils;
+import org.h2.tools.SimpleResultSet;
+import org.h2.util.StringUtils;
+import org.h2.util.Utils;
+
+/**
+ * COPIED from h2 version 2.2.240 - Moved all references from FullTextLucene to
+ * UmsFullTextLucene - Added UmsAnalyzer for indexing
+ */
```
SubJunk, line 54: Can you explain why this was needed? Why couldn't we just use the h2database one?

src/main/java/net/pms/database/DatabaseHelper.java
```diff
@@ -61,6 +61,7 @@ public abstract class DatabaseHelper {
 	protected static final String LOG_ERROR_WHILE_VAR_IN_FOR = "Database \"{}\" error while {} \"{}\" in \"{}\" for \"{}\": {}";
 
 	// Generic constant for the maximum string size: 255 chars
+	protected static final int SIZE_2048 = 1024;
```
SubJunk: This should be 2048

src/main/java/net/pms/database/MediaTableFiles.java
```diff
@@ -772,6 +799,19 @@ public static MediaInfo getMediaInfo(final Connection connection, String filenam
 		return media;
 	}
 
+	// Hilfsmethode zum Deserialisieren
```
SubJunk: Can you add this comment in English too please, it's ok if other languages are here too

SubJunk (changes requested):

src/main/java/net/pms/database/lucene/UmsFullTextLucene.java
ik666: The default Analyzer would split filenames in a way, that is not usable for common audio and video files. Thats's why I added `UmsAnalyzer`. There is no "standard" way in telling h2 FullText implementation to use this Analyzer. That's why the copy is needed.

src/main/java/net/pms/database/DatabaseHelper.java
ik666: right.

src/main/java/net/pms/database/MediaTableFiles.java
ik666: Ups. Actually the method name says everything. I'll removed the comment.

SubJunk (approved):
Cool that you added tests

I don't use UMS for music personally, so I am putting trust in you that you have tested it. Good work

I also see your note about JDK 21+, maybe we should look at updating. We haven't had a great reason for that until now

ik666:
You're welcome. I actually only use UMS for music, so we're a perfect match. :)

Moving to JDK 21+ makes a lot of sense, especially for features like memory mapping (while generating resource IDs (RUID)).

IMO, staying on JDK 17 only makes sense if anyone still needs 32-bit OS support.

PR #6089, opened 2026-03-24
fix: use JRE instead of JDK in Docker image

### LINKED ISSUE #6044: Docker JDK should be JRE
In our Docker images we use JDK, but it should be JRE to save space

mvanhorn:
Swaps JDK for JRE in the Ubuntu Docker image.

## Why this matters

The server doesn't compile Java. It just runs it. JDK adds ~200MB for tools nobody uses in production. The Alpine Dockerfile already uses JRE.

## Changes

- **docker/Dockerfile**: Changed `openjdk-17-jdk-headless` to `openjdk-17-jre-headless`

## Testing

Built locally. Container starts and serves media.

Fixes #6044

This contribution was developed with AI assistance (Claude Code).


SubJunk:
Thanks!

PR #5938, opened 2025-12-27
Docs: Update ffmpeg_gpu_decoding_acceleration_method default to none

chaitjosh:
Fixes #5682 
Corrected the documentation in UMS.conf. The comment for `ffmpeg_gpu_decoding_acceleration_method`, changed the default from "auto" to "none"

This is my first time contributing to Open Source Organizations! Excited to be part of the community.

# Checklist
- [x] Any relevant issues are referenced
- [ ] Any relevant Knowledge Base articles are written/modified
- [ ] Any relevant tests have been written/modified

PR #6121, opened 2026-04-19
Docs: Fix and clarify ffmpeg_gpu_decoding_acceleration_method default and usage

### LINKED ISSUE #5682: ffmpeg_gpu_decoding_acceleration_method Does Nothing at All
Hello Everyone,

First let me thank everyone who contributed for this amazing piece of software.
I've been using it for years now.

However, now I'v got a new server with a modern GPU and wanted to get hardware acceleration running.
Everything works fine when running: 
`sudo -u ums-user /home/ums-user/ums-14.12.0/linux/ffmpeg -y -hide_banner -hwaccel qsv -hwaccel_device /dev/dri/renderD128 -hwaccel_output_format qsv -i "/InputFile_AV1_1080p.mkv" -maxrate 50000k -crf 23 -ab 448k -c:a copy -c:v hevc_qsv -f mpegts /tmp/test2`
I get around 550 FPS. Great :)

However, setting
`ffmpeg_gpu_decoding_acceleration_method = -hwaccel qsv -hwaccel_device /dev/dri/renderD128 -hwaccel_output_format qsv`
did not change the command line used at all so I downloaded the source code and found no place where my command line actually is being used.
Furthermore UMS.conf is incorrectly documented as ffmpeg_gpu_decoding_acceleration_method defaults to "none", not "auto".
`src\main\java\net\pms\configuration\UmsConfiguration.java:
return getString(KEY_FFMPEG_GPU_DECODING_ACCELERATION_METHOD, "none");`

Cheers,
Mark

Harshit-dell:
Fixes #5682

There was previous confusion about whether the default for
ffmpeg_gpu_decoding_acceleration_method should be "auto" or "none".

After checking the code in UmsConfiguration.java:

getString(KEY_FFMPEG_GPU_DECODING_ACCELERATION_METHOD, "none")

The default is explicitly set to "none".

This appears intentional:
- GPU decoding is disabled by default (safe fallback)
- Similar to encoding default "libx264"
- Avoids GPU-related issues on unsupported systems

Changes in this PR:
- Update documentation to reflect correct default ("none")
- Clarify behavior and interaction with gpu_acceleration
- List possible values
- Improve formatting and readability

This PR does not change behavior, only documentation.

Supersedes #5938

### DIFF
--- a/src/main/external-resources/UMS.conf
+++ b/src/main/external-resources/UMS.conf
@@ -914,15 +914,22 @@ ffmpeg_sox =
 
 # GPU decoding acceleration method
 # -----------------------
-# Recommendation is "auto" because other methods don't fallback to software
-# decoding when errors appear.
-# Also accepts custom options, e.g.
+# Controls the hardware acceleration method used for video decoding.
+#
+# Default: none
+# Possible values: none, auto, vaapi, cuda, dxva2, videotoolbox (platform dependent)
+#
+# Note:
+# - "none" disables GPU decoding (safe default)
+# - "auto" enables automatic GPU acceleration with fallback to software decoding
+# - Only used when gpu_acceleration=true
+#
+# Advanced:
+# Custom FFmpeg options are also supported, e.g.
 # -hwaccel vaapi -hwaccel_device /dev/dri/renderD128 -hwaccel_output_format vaapi
-# for not only GPU decoding but also GPU encoding.
-# -init_hw_device vaapi=amd:/dev/dri/renderD129 -hwaccel vaapi -hwaccel vaapi -hwaccel_device amd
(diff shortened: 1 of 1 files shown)


Harshit-dell:
verified the default in UmsConfiguration.java and updated the documentation accordingly.

If the intended behavior is to use "auto" instead, I’m happy to adjust this PR or update the code in a separate change.

Harshit-dell:
The default is explicitly set to "none", and it’s also used consistently when serializing the configuration.So in this PR I kept the behavior unchanged and updated the documentation to match the current implementation.

This seems intentional and consistent with other safe defaults  where GPU features are not enabled automatically to avoid issues on unsupported systems

SubJunk:
@Harshit-dell thanks!

PR #6138, opened 2026-04-28
This PR fixes a logic error in playlists

ik666:
Fixes a concurrent modification exception in playlist handling , when adding items.


ik666:
wrong branch root

PR #6140, opened 2026-04-29
fix: concurrent modification exception in playlist folder

ik666:
fixes concurrent modification exception in playlist folder, when modifying playlist file by external process.

### DIFF
--- a/src/main/java/net/pms/store/container/PlaylistFolder.java
+++ b/src/main/java/net/pms/store/container/PlaylistFolder.java
@@ -260,7 +260,9 @@ public final class PlaylistFolder extends StoreContainer {
 			Collections.shuffle(getChildren());
 		}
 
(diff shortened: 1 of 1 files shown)


SubJunk (dismissed):

SubJunk (changes requested):
I wonder whether this could have any unseen consequences - are we relying on that list during sync for anything, like to ensure the child-parent relationship is kept? I think the answer is no, but I want to check before merging because that kind of bug can be hard to solve later

ik666:
I think there is some timing issues or collision between the FileWatcher and how the PlaylistFolder is being updated. I haven't gone deep, since this fix worked for me. New entries in the playlist show up. All other play well.

SubJunk:
Ok, I will merge it. Let's try to remember this if we get a weird bug later on

SubJunk (approved):

PR #6150, opened 2026-05-06
fixed album art issue after creating new resources (UPnP)

ik666:
Album Art wasn't updated correctly, when new files in playlists where added or new folders or files in new folders were created / added.

### DIFF
--- a/src/main/java/net/pms/store/MediaStore.java
+++ b/src/main/java/net/pms/store/MediaStore.java
@@ -739,6 +739,13 @@ public class MediaStore extends StoreContainer {
 			if (storeResource instanceof RealFile rf) {
 				rf.setMediaInfo(null);
 				rf.resolve();
+				rf.getParent().doRefreshChildren();
+				LOGGER.debug("File {} updated, media info refreshed and parent folder refreshed.", file.toString());
+			} else if (storeResource instanceof RealFolder rf) {
+				rf.setMediaInfo(null);
+				rf.resolve();
+				rf.doRefreshChildren();
+				LOGGER.debug("Folder {} updated, media info refreshed and children refreshed.", file.toString());
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


src/main/java/net/pms/store/MediaStore.java
```diff
@@ -739,7 +739,14 @@ public void fileUpdated(File file) {
...
 				rf.setMediaInfo(null);
 				rf.resolve();
+				LOGGER.debug("File {} updated, media info refreshed and parent folder refreshed.", file.toString());
+			} else if (storeResource instanceof RealFolder rf) {
+				rf.setMediaInfo(null);
+				rf.resolve();
+				rf.doRefreshChildren();
+				LOGGER.debug("Folder {} updated, media info refreshed and children refreshed.", file.toString());
 			}
+			storeResource.getParent().discoverChildren();
```
SubJunk, line 749: Why is this line needed? It seems like a lot of extra overhead

ik666: For existing resources: Because the parent holds the reference to the "old folder" showing the old image. Therefore, the parent has to be notified. If a new folder was created the scan is needed to add it to the MediaStore.

ik666: One other option would be to update the ThumbStore directly for the specific objectId. But this implementation works so far fast enough, even for folder containing many thousand entries.

SubJunk: Thanks for explaining, let's try it

SubJunk (approved):

src/main/java/net/pms/store/MediaStore.java
ik666: Thanks for your confidence :)

PR #6195, opened 2026-06-08
Switched RUID hash implementation to xx3hash

ik666:
# Switch FILES table hashing from BLAKE3 to xxHash3

## Summary

This PR replaces the resource-identifier (RUID) hash used by the `FILES` table from **BLAKE3** to **xxHash3 (XXH3)**. The RUID is used to recognise media files — it is not a security primitive — so a fast non-cryptographic hash suits better.

## Motivation

The RUID exists only to identify and de-duplicate local media, where the inputs are ordinary files.

**xxHash3 is much faster for this workload.** XXH3 is a non-cryptographic hash built for raw throughput. This directly cuts scan time and CPU usage (empty database startup phase)

## Implementation notes
- A documented migration path is included for when UMS moves to JDK 21+: the chunking workaround can be deleted in favour of a single FFM-API (`MemorySegment`) implementation.
- Test coverage added (`ResourceIdentifierTest`).

## Migration / compatibility

Because the hash algorithm changes, RUIDs are **not** comparable to previously stored BLAKE3 values — existing identifiers are recomputed.


src/main/java/net/pms/util/ResourceIdentifier.java
```diff
@@ -80,76 +64,137 @@ public static String getResourceIdentifier(final String uri) {
...
+			// not a path
 		}
 		return null;
 	}
 
 	/**
-	 * Calculates the Blake3 digest and returns the value as a 32 character hex
-	 * string.
+	 * Calculates the XXH3 hash and returns the value as a 16 character hex
+	 * string. File has to be smaller than 2 GB for correct hashing.
```
SubJunk, line 97: Smaller than 2GB? Every video I add lately is like 8-20GB+

SubJunk (changes requested):

src/main/java/net/pms/util/ResourceIdentifier.java
ik666: You're just looking at one method : `getBigFileIdentifier`. 

We have 3 cases:
- small files
- large files (<2GB)
- very large files (>2GB)

JDK 17 doesn't allow direct reading / mapping files larger than 2GB. That's why we distinguish between these cases in `public static String getResourceIdentifier`. 

- Large files are handled by `getBigFileIdentifier`
- Very large files are handled by `getVeryBigFileIdentifier` method.

Both methods can be merged in case we switch to JDK21+. The migration code is already in the class as comment.

SubJunk: I see, I should have finished looking, sorry. Will look again soon

SubJunk (approved):

SubJunk:
@ik666 do you have an idea about how much time will be saved with this?

sf666:
Yes. the hash algorith is appr. 3-10x faster than blacke3. Especially in the startup phase with empty DB I saved up to 20 minutes. On large files you'll be on the upper speed improvement side.

SubJunk:
Wow that's a lot. Nice job

SubJunk:
@ik666 I get some instances of this error after using this:

```
21:09:44 INFO  Exception in thread "MediaScanner File Parser" java.lang.IllegalAccessError: class net.openhft.hashing.LongHashFunction (in unnamed module @0x757acd7b) cannot access class sun.nio.ch.DirectBuffer (in module java.base) because module java.base does not export sun.nio.ch to unnamed module @0x757acd7b
21:09:44 INFO  	at net.openhft.hashing.LongHashFunction.hashByteBuffer(LongHashFunction.java:589)
21:09:44 INFO  	at net.openhft.hashing.LongHashFunction.hashBytes(LongHashFunction.java:562)
21:09:44 INFO  	at net.pms.util.ResourceIdentifier.getVeryBigFileIdentifier(ResourceIdentifier.java:140)
21:09:44 INFO  	at net.pms.util.ResourceIdentifier.getResourceIdentifier(ResourceIdentifier.java:73)
21:09:44 INFO  	at net.pms.store.MediaInfoStore.updateMediaInfoFromFile(MediaInfoStore.java:183)
21:09:44 INFO  	at net.pms.store.MediaInfoStore.getMediaInfo(MediaInfoStore.java:156)
21:09:44 INFO  	at net.pms.store.item.RealFile.resolve(RealFile.java:254)
21:09:44 INFO  	at net.pms.store.StoreItem.syncResolve(StoreItem.java:1470)
21:09:44 INFO  	at net.pms.store.item.RealFile.syncResolve(RealFile.java:333)
21:09:44 INFO  	at net.pms.store.MediaScanner.lambda$parseFileEntry$3(MediaScanner.java:477)
21:09:44 INFO  	at java.base/java.lang.Thread.run(Unknown Source)
```

ik666:
For JDK 17 we need to add the VM Parameter `--add-opens java.base/sun.nio.ch=ALL-UNNAMED`. I've done it for the Docker images, but missed the app bundles. I'll fix it now for Windows and Mac.

PR #6233, opened 2026-07-02
Add "Delete all thumbnails" action to Tools

ik666:
### Background / Problem
Once thumbnails have been generated and cached, there was previously no way to get rid of "wrong" or outdated thumbnails (e.g. when the generated thumbnail isn't the one the user expects) other than deleting the whole database. 

### What this PR does
Adds a dedicated "Delete all thumbnails" action (Settings → Tools → Actions) that clears only the cached thumbnails. The thumbnails are regenerated on demand the next time the content is browsed.


Mik-S-UMS:
Maybe you could add the option to do this for just one video or for a series for those cases where only a few thumbnails are incorrect.

Something similar could be used for when videos are mis-identified and need re-scanning to update their metadata.
I recently had Spiderman Noir (colour) being identified as the series "In living colour" so I had to manually rename and re-add them for them to be correctly identified.

ik666:
Good point. This PR was easy to integrate, deleting specific thumbs and metadata sounds like a good idea to remember.

SubJunk (approved):

SubJunk:
Cool, good idea. I agree about Mik's suggestion too, can someone make an issue for it?

PR #6240, opened 2026-07-08
Support flexible 3-box UPnP setups when the authentication service is disabled

ik666:
### Problem

This is a follow-up to the discussion in [#4886](https://github.com/UniversalMediaServer/UniversalMediaServer/pull/4886). Rather than choosing one model over the other, it **unifies both concepts** behind a single configuration switch: the UUID-based authentication/authorization model stays in place when the authentication service is enabled, and the just-in-time, browsing-agnostic 3-box model takes over when it is disabled.

Renderer authorization currently couples **browsing** and **rendering**: a resource URL embeds the UUID of the renderer that built the CDS tree (`StoreItem.getMediaURL()` → `renderer.getUUID()`), and `MediaServerServlet` looks the renderer up by that UUID when the resource is fetched.

This only holds in a **2-box setup**, where control point and media renderer are the same device (same IP, same UUID). In a classic **3-box UPnP setup** — a control point browsing on behalf of a *separate* renderer — the URL carries the control point's UUID, so the servlet resolves the **wrong renderer** (the control point instead of the actual player). Content and any transcoding decisions are then produced for the wrong device.

The authenticated model deliberately relies on this coupling: it maps the browse UUID to a user account and enforces per-group shared-folder restrictions and the renderer allow-list. Dropping the coupling outright would break that model, which is why this PR gates the two behaviors instead of replacing one with the other.

### Solution

Make the behavior conditional on the existing **authentication service** toggle (`UmsConfiguration.isAuthenticationEnabled()`):

| Authentication service | CDS tree / renderer |
|---|---|
| **enabled** (unchanged) | bound to the browse UUID → account → group-filtered shared folders | 
| **disabled** | one neutral `ControlPoint` renderer holds the tree (group 0 = default shared folders); the **real** renderer is identified just-in-time from request headers (User-Agent) + IP |

When authentication is disabled, the servlet resolves the resource via `renderer.getMediaStore().getResource(id)` against the **just-in-time identified** renderer, so transcoding is automatically decided for the correct device (resource IDs are global across media stores).

### Runtime toggle

Toggling the authentication service at runtime rebuilds all renderer media stores (including the control-point tree), so switching between modes doesn't require a restart.

### DIFF
--- a/src/main/java/net/pms/iam/AuthService.java
+++ b/src/main/java/net/pms/iam/AuthService.java
@@ -196,7 +196,15 @@ public class AuthService {
 	}
 
 	public static void setEnabled(boolean value) {
+		boolean changed = CONFIGURATION.isAuthenticationEnabled() != value;
 		CONFIGURATION.setAuthenticationEnabled(value);
+		if (changed && PMS.isReady()) {
+			// Renderer identification and the CDS object tree depend on whether authentication is
+			// enabled (UUID-bound per-account renderer vs. the neutral control-point renderer with
+			// just-in-time identification). Rebuild the renderers' media stores so the change takes
+			// effect without requiring a restart.
+			PMS.get().resetRenderersMediaStore();
+		}
 	}
 
 	public static boolean isLocalhostAsAdmin() {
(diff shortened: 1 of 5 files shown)


SubJunk:
You've been working hard! Maybe this set of features should be v16

ik666:
@Subjunk. Thanks :) And I agree, we can merge it in the new v16. Then we have some nice features from the beginning. May also apply to AudioAddict stuff.

SubJunk (approved):

SubJunk:
Good idea

PR #6241, opened 2026-07-09
AudioAddict: curated playlists and events support

ik666:
Extends the AudioAddict integration beyond plain radio stations with two new content types exposed as browsable folders: **curated playlists** and **events** (live shows + on-demand episodes). These are currently offered by **DI.fm** only, so they are added exclusively under the DI.fm network; the other AudioAddict networks keep their existing flat station list.

### Browse tree

- **Playlists** — AudioAddict's curated playlists, played back as continuous audio. Track/album/artist metadata is resolved per playlist and mapped to UPnP; the current track title is updated live during playback (see ICY below). An optional endless-loop mode is available.
- **Events** — upcoming/running shows. A live event exposes its currently playing episode directly; shows with on-demand history additionally get a child folder listing their recent episodes (paged).

### Live "now playing" metadata (ICY)

Curated playlists and radio stations serve SHOUTcast/Icecast (ICY) in-band metadata, so a renderer that requests it (`Icy-MetaData: 1`) sees the current track title update while playing.

### Configuration (UMS.conf)

| Key | Default | Meaning |
|---|---|---|
| `audio_addict_icy_metadata` | `true` | Send ICY in-band track-title metadata |
| `audio_addict_playlist_loop` | `false` | Loop curated playlists endlessly instead of stopping at the end |
| `audio_addict_episodes_per_container` | `20` | number of recent on-demand episodes listed per folder |
| `audio_addict_tree_cache_ttl_minutes` | `60` | How long fetched events/episodes trees are cached before refetch |


ik666:
merged into v16

PR #6273, opened 2026-08-04
Fix "MemorySegmentIndexInputProvider is missing" on JDK 21+ (Multi-Release fat jar)

ik666:
fix(build): mark jar-with-dependencies as Multi-Release

The assembly plugin unpacks and repackages all dependencies into the
fat jar but does not set `Multi-Release: true` in the merged manifest.
As a result the JVM ignores META-INF/versions/, so on JDK 21+ Lucene's
MMapDirectory cannot find MemorySegmentIndexInputProvider and startup
fails with:

  LinkageError: MemorySegmentIndexInputProvider is missing in Lucene JAR file

Add <manifestEntries><Multi-Release>true</Multi-Release> to every
assembly <archive> so the versioned classes shipped inside the fat jar
are honored again.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1030,6 +1030,10 @@
 									-->
 									<addDefaultImplementationEntries>true</addDefaultImplementationEntries>
 								</manifest>
+								<manifestEntries>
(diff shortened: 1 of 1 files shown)


SubJunk:
Is this intended for main or v16?

ik666:
This PR is the main (v15) fix. I committed the changes already to v16.

PR #6281, opened 2026-08-06
Fix endless rescan loop when a new folder appears

ik666:
I've seen that a new directory (with some content) in a shared folder makes UMS stop answering browse requests, because of an error in duplicate check. Then the scanner logs the same paths for ever.

I also added some test to make usage of FileWatcher.Watch clear for future use.

### DIFF
--- a/src/main/java/net/pms/store/MediaScanner.java
+++ b/src/main/java/net/pms/store/MediaScanner.java
@@ -309,13 +309,8 @@ public class MediaScanner implements SharedContentListener {
 			} else {
 				LOGGER.debug("Scanning folder \"{}\"", file.getAbsolutePath());
 			}
+
 			List<StoreResource> systemFileResources = RENDERER.getMediaStore().findSystemFileResources(file);
-			if (systemFileResources.isEmpty()) {
-				if (isInSharedFolders(filename)) {
-					internalScanFileOrFolder(filename);
-					systemFileResources = RENDERER.getMediaStore().findSystemFileResources(file);
-				}
-			}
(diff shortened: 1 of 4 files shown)


src/main/java/net/pms/store/MediaScanner.java
```diff
@@ -435,8 +462,16 @@ private static void parseFileEntry(File file, boolean advise, boolean isCrawling
 				//wait 500 ms
 				Thread.sleep(500);
 				//Check if size changed (copying, downloading)
+				int waits = 0;
 				while (file.exists() && (currentSize != file.length() || FileUtil.isLocked(file))) {
 					//loop until file size is not changing anymore and file is unlocked.
+					if (waits++ >= MAX_FILE_SETTLE_WAITS) {
```
Entropy512: I'm in the process of digging into UMS' scanning code because I've found it to be extremely fragile, but a question:

Will this file eventually be revisited after some time, or is "giving up" permanent?

If the latter, this will likely break in the scenario of files being written over a wireless network, or even worse, files that are the result of an ongoing transcode job where the file could take a significant time to write.  UMS already appears to be extremely fragile in these use cases (as in I always have to restart UMS after running transcode jobs, or sit and wait for minutes as it sits and digests a folder when I try to access it from Kodie)

SubJunk:
@ik666 is this ready to merge?

ik666:
@SubJunk Yes, that looks stable.

SubJunk:
Cool, I haven't tested it but I am doing a release now because of the performance fix, so will include this too

SubJunk (approved):

Issue #5974, opened 2026-01-22
Level parameter in supported formats


Mik-S-UMS:
Just discovered that Samsung TVs have different level limits depending on FHD or UHD and there is no way to currently set this in the conf.

> H.264 FHD: Level 4.2 supported (FMO/ASO/RS not supported).
> H.264 UHD: Level 5.1 supported, seamless resolution change supported up to 3840x2160.
> HEVC FHD: Level 4.1 supported.
> HEVC UHD: Level 5.2 supported, seamless resolution change supported with Level 5.1 up to 3840x2160.
> HEVC 8K : Level 6.1 supported. (8K only)

Setting `H264LevelLimit` and `H265LevelLimit` to 4.2 and 4.1 respectively would allow more videos to be able to play but would mean unnecessarily transcoding the UHD videos with higher levels.

I suggest that this could be a parameter in supported formats like width, height, bitrate etc.

PR #5826, opened 2025-10-22
docs: added Contributing and License sections in README

niveshpadamata1218:
### Summary
Added a short **Contributing** section and a **License** section to the README to help new contributors find contribution guidelines and the GPL license.

### Changes
- Added `## Contributing` with basic fork/branch/PR instructions and a link to `CONTRIBUTING.md`.
- Added `## License` referencing the repository `LICENSE` file.
- Minor formatting improvements.

### Why this helps
This improves onboarding for new contributors and makes the repository README more immediately useful.

### Checklist
- [x] Documentation change only (no code)
- [ ] Any relevant issues referenced
- [ ] Any relevant Knowledge Base articles updated
- [ ] Any tests written/modified (not applicable)

Issue #3723, opened 2022-12-22
(FEATURE) 2D to 3D Algorithmic Enhancement


threedguru:
Hi Guys,

This is a bit of a heads up. I'm hoping we can get this into the next release. I've developed a significant enhancement to the 2D to 3D algorithm that iscurrently in UMS Release 13. I've described the enhancement here:

https://iwantaholodeck.com/enhanced-2d-to-3d-algorithm/

I've already proven it against my development version of UMS. I need to create a pull request - just need a little time. There are changes to the conversion script, but the main application addition is a new AviSynth library masktools2, which is added in exactly the same way as mvtools2 library currently in UMS. 

More specifically, the code uses the motion vector analysis (MAnalyse) and scene change detection routines (MSCDetection) from the mvtools2 library, see documentation and github repository below:

http://avisynth.org.ru/mvtools/mvtools2.html#version2
https://github.com/pinterf/mvtools
 
In addition, it uses the mt_merge routine from the masktools2 library,
see documentation and github repository below:
 
http://avisynth.nl/index.php/MaskTools2/Mt_merge
https://github.com/pinterf/masktools

SubJunk:
@threedguru that sounds good, I'm familiar with those libraries and those functionalities. Thanks for the heads up

threedguru:
This is now in the main branch so closing.

SubJunk:
@threedguru great, I'm planning to do a release in the next few days

Issue #3486, opened 2022-10-08
(FEATURE) On-the-fly 2D to 3D Video Conversion for VR Headsets 


threedguru:
Hi,

I posted this as a new feature suggestion:

https://www.universalmediaserver.com/forum/viewtopic.php?f=13&t=15369

mik_s suggested that I come over here to Git hub. I am not sure if this is the right place. I have already developed the solution, but it needs broader proving (e.g. on AMD GPU). It's in my local git repository. I developed it against 11.4.1. 

https://www.universalmediaserver.com/forum/viewtopic.php?f=14&t=15377

I am trying to establish the best way to take it forward.

Any suggestions?

SubJunk:
@threedguru great idea, I would love to see the code, can you submit a pull request so we can see it?

threedguru:
Hi @SubJunk, thanks for replying. I have never done a "pull request"
before, but it sounds like the right next step though. I have just read a
brief overview of how a "pull request" works. I can certainly have a go. I
git cloned the 11.4.1 version of UMS as a starting point. Is there a better
link to follow on "pull requests" than this one?

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-pull-requests

I have been thinking/working on how to share this with the VR community. To
that extent I have been working on the video linked below. It is a work in
progress from me, but it provides a demo of the solution and contains
screenshots and descriptions of most of the changes that I have made to UMS
so far.

https://streamable.com/sjwdqy

Comments on the video are very much appreciated.

threedguru:
Hi @SubJunk. I have created a pull request as requested. I have created a local fork to perform the pull request. I could not figure out how to make the request against tag 11.4.1 so there are probably some conflicts with main. Anyway, you can see the code. Comments appreciated. The screen shots and my commentary in the video will help to understand what I have done. See video link again below:

https://streamable.com/sjwdqy

threedguru:
Hello again @SubJunk. I have updated the pull request after merging with main, testing and fixing the conflicts. I have provided a more detailed overview in the description of the pull request, a video link and some screen shots. See here:

https://github.com/UniversalMediaServer/UniversalMediaServer/pull/3490

threedguru:
[...]

PR #6330, opened 2026-09-19
Signal frame-packed 3D in H.264 transcodes, and correct the 2011-2012 Bravia profiles

### LINKED ISSUE #6329: Preserve/inject the H.264 frame_packing SEI so frame-packed 3D auto-engages over DLNA
### Summary

Frame-packed 3D files (SBS / Top-Bottom) served or transcoded by UMS do not auto-engage 3D on 3D TVs that auto-detect from the stream, they play flat. The cause is a signalling detail worth handling explicitly.

### Mechanism

3D TVs of the frame-compatible era auto-engage 3D from the H.264 `frame_packing_arrangement` SEI (payload type 45, Annex D) carried **in the elementary stream**, not from the Matroska `StereoMode` container tag. The two are independent:

- **Remux / stream-copy (MKV to TS):** the container `StereoMode` tag is dropped and only an in-stream SEI survives. If the source file has no SEI (many frame-compatible rips carry only the container tag), nothing signals 3D and the TV shows it flat.
- **Re-encode:** libx264 writes the SEI only when passed `--frame-packing 3|4` (3 = SBS, 4 = TAB); without it, a freshly encoded 3D file carries no SEI.

libavcodec exposes the in-stream signal as `AV_FRAME_DATA_STEREO3D`. In the DVB frame-compatible 3DTV spec (ETSI TS 101 547-2) this in-stream SEI is defined to take precedence over container-level signalling.

### Request

On 3D-flagged sources: (a) inject the `frame_packing_arrangement` SEI on stream-copy delivery paths (a bitstream filter, no re-encode, negligible size cost), and (b) pass `-x264-params frame-packing=3|4` on the re-encode path. The SEI is ignorable metadata, so non-3D renderers are unaffected.

### Context

Full write-up, evidence and the same problem documented across several media servers: https://github.com/danielcamposramos/sony-bravia-linux (`docs/3d-signalling-ecosystem.md`). HandBrake has merged SEI writing (PR #8100); mpv is adding read support (#18490). Restored fully automatic 3D over DLNA on stock free Serviio with a profile plus SEI injection.

Written with AI assistance; filed under my own name, taking responsibility for the content.

danielcamposramos:
Closes #6329.

Two changes, both from measurements on real hardware: a KDL-46EX725 (2011) and a KDL-46HX855 (2012).

## 1. Frame-packed 3D is never signalled in the elementary stream

3D TVs of the frame-compatible era auto-engage 3D from the H.264 `frame_packing_arrangement` SEI (Annex D, payload type 45) carried **in the elementary stream**. x264 writes it only when asked (`frame-packing=N`). The container-level stereoscopic tag is a separate signal: it does not survive a remux to MPEG-TS, and in the DVB frame-compatible 3DTV spec (ETSI TS 101 547-2) the in-stream SEI takes precedence over it.

So a 3D transcode currently arrives at those renderers as perfectly correct frame-packed video that they have no way to recognise, and it plays flat.

`FFMpegVideo.getVideoTranscodeOptions()` now adds `-x264-params frame-packing=N` on libx264 transcodes whose output is frame-packed:

| Output layout | `frame-packing` |
|---|---|
| `sbsl` / `sbsr` / `sbs2l` / `sbs2r` | 3 (side by side) |
| `abl` / `abr` / `ab2l` / `ab2r` | 4 (top bottom) |
| `irl` / `irr` | 2 (row alternation) |
| `ml` / `mr` (2D output), anaglyph | not signalled |

The value follows `Output3DFormat` when the renderer sets one, since that is the layout actually leaving the encoder, and otherwise the source layout. It is skipped when `CustomFFmpegOptions` already contains `-x264-params`, and it only applies on the libx264 path (GPU encoders do not take x264 parameters).

This mirrors the equivalent change in HandBrake, [PR #8100](https://github.com/HandBrake/HandBrake/pull/8100), merged 16 September 2026.

## 2. The Bravia profiles transcode things these sets play natively

Measured with a pass-through DLNA server, files byte-exact on the wire, tracks copied without re-encoding:

- **MP4 is played directly.** `Sony-BraviaEX725.conf` had no `f:mp4` line at all, so every MP4 was transcoded on a set that advertises `http-get:*:video/mp4:*` as a wildcard and plays H.264 in MP4 without help.
- **E-AC3 is decoded from MP4, including 7.1**, and the set reports it on screen as "Dolby Digital Plus". Transcoding it to AC-3 is a real loss. Worth noting because it is counter-intuitive: the sets' HDMI *input* does not accept E-AC3 as a bitstream at all, so decoding a file and accepting a bitstream are genuinely different capabilities here.
- **DTS is silent** in MP4 while the video plays, which the existing "DTS is not supported" note now states precisely.
- **Matroska is not advertised at all**, so an MKV is filtered out of the browse listing rather than merely failing to play. Recorded as a comment so the next person does not try to add it.

Changed: `Sony-BraviaEX725.conf`, `Sony-BraviaHX.conf`, `Sony-BraviaHX75.conf`.

## Scope, stated honestly

The two sets above are what I measured. The profile changes are applied to the configs whose `UpnpDetailsSearch` matches those same chassis generations, because Sony shipped one software platform per chassis generation rather than per model — one application catalog and one firmware line served every set in a generation, which is documented [here](https://github.com/danielcamposramos/sony-bravia-linux). If you would rather keep the changes to the exact models measured, say so and I will narrow them.

I have not extended this to `Sony-BraviaEX.conf`, `Sony-BraviaNX70x.conf` or `Sony-BraviaNX800.conf`: those cover older models I have no hardware for, and adding MP4 support to a set I cannot test seemed the wrong side of the line.

## Testing

The 3D signalling half is verified end to end on both sets, through two independent servers (Serviio and Kodi's UPnP, which never transcodes): an MP4 carrying the SEI switches the TV into 3D automatically, the identical file with the SEI stripped — a difference of 791 bytes, picture data untouched — plays flat, and a Matroska carrying only the container `StereoMode` tag is not listed at all. Method and files: https://github.com/danielcamposramos/sony-bravia-linux/blob/main/docs/3d-signalling-ecosystem.md

I could not run the UMS test suite locally (no Maven, and Java 25 here against the project's 17), so the Java change has had a syntax check only and needs CI. It is a small, isolated addition to one options block, but please treat it as untested by me.


danielcamposramos:
For triage, I searched this repo for anything else either half of the PR would close, so you do not have to.

**Closed by this PR:** #6329 only. That is the honest answer — I searched open issues for `3D`, `stereoscopic`, `frame packing`, `SBS`, `side by side`, `Output3DFormat`, `anaglyph`, `Bravia`, `Sony`, `E-AC3` and `Dolby Digital Plus`, and nothing else open is about either the in-stream 3D signal or these sets' audio handling.

**Not closed by this PR, listed so they are not conflated with it:**

- #6099 — MP4 not visible, Sony KDL-43WF665, reported as a v15 regression from v13. Different set (2017 W series, Android era), different cause, and this PR does not touch that profile. Mentioning it only because "Sony" and "MP4" both appear and I would rather rule it out explicitly than leave you guessing.
- The older 3D issues in the tracker are all a different subject: #587, #878, #893 and #745 are subtitle handling on 3D video, and #3486 and #3723 were 2D-to-3D conversion feature requests. None involve the frame-packing SEI.

**Wider context, if it is useful.** This is a signalling gap that runs across the whole chain rather than a UMS quirk, and the same finding has been taken to the other projects in it. HandBrake merged the equivalent encoder change ([PR #8100](https://github.com/HandBrake/HandBrake/pull/8100), 16 September 2026, closing their #5826). Reports are open at FFmpeg ([#24530](https://code.ffmpeg.org/FFmpeg/FFmpeg/issues/24530), [#24531](https://code.ffmpeg.org/FFmpeg/FFmpeg/issues/24531)), mpv ([#18489](https://github.com/mpv-player/mpv/issues/18489) with [PR #18490](https://github.com/mpv-player/mpv/pull/18490)), MKVToolNix ([#6309](https://codeberg.org/mbunkus/mkvtoolnix/issues/6309), for the container side), Jellyfin ([PR #18060 comment](https://github.com/jellyfin/jellyfin/pull/18060#issuecomment-5726381078)), Gerbera ([#3937](https://github.com/gerbera/gerbera/issues/3937)) and Kodi ([#29337](https://github.com/xbmc/xbmc/issues/29337)).

Each project owns a different link: the encoder writes the SEI, the muxer keeps or drops the container tag, the server decides whether to remux, and the TV reads only the SEI. UMS sits at the point where a correct file can still lose the signal, which is why it is worth having here.

danielcamposramos:
Following up on the triage above with a deeper sweep, since the first pass only covered 3D wording. I also searched the open issues for `Dolby`, `AC3`, `E-AC3`, `passthrough`, `5.1`, `transcode audio`, `Bravia` and `KDL`.

**Still only #6329 is closed by this PR.** Nothing else open turns on the frame-packing SEI or on these sets' audio handling.

**Same family, different mechanism — not closed by this PR, but worth reading alongside it:**

- #1775 — UMS transcodes video *and* audio when only the audio is unsupported (reported against Chromecast). Same underlying waste as the E-AC3 half of this PR, where the fix here is simply to stop declaring a codec unsupported when the device decodes it. That issue needs audio-only transcoding logic, which this PR does not touch.
- #5974 — no way to express per-resolution level limits in a profile. Adjacent in spirit: both are about a profile describing a device accurately enough that UMS stops guessing. The MP4 line I added leans on H.264 level support that today can only be stated coarsely.

**Ruled out, so they are not conflated:**

- #6099 — Sony KDL-43WF665, MP4 not visible, reported as a v15 regression. Different generation (2017, Android TV era), different cause, and this PR does not touch that profile.
- #1804 — always-transcoded UDP stream via web.conf, Samsung. Unrelated path.
- The older 3D tickets (#587, #878, #893, #745) are subtitle handling on 3D video, and #3486/#3723 were 2D-to-3D conversion requests. None involve the in-stream signal.

**One observation you may want, offered as a hypothesis rather than a measurement.** The E-AC3 finding is probably not specific to these two sets: a TV of that era can decode Dolby Digital Plus from a file while its HDMI input refuses the same codec as a bitstream, and those are separate capabilities that a single profile line tends to blur. If other profiles in the repo were written from HDMI specs rather than from playback tests, they may under-declare in the same way. I have only measured the two Sony sets I own, so I am not going to touch anyone else's profile on a guess — but if you want the same method applied to a specific renderer and someone can run the files, I am happy to write up the test set.

SubJunk (approved):

SubJunk:
> If other profiles in the repo were written from HDMI specs rather than from playback tests, they may under-declare in the same way

Unfortunately this is the reality of our project - we make a first pass at renderer configs based on their manuals, and then it is up to individual people to test and optimize if they want the best experience for their renderer, as you have.

Thanks for the contribution, I will merge it now

PR #6331, opened 2026-09-19
Optimize media browsing and metadata loading in V16

valib:
Reduce repeated database and parsing work, defer localized metadata and image JSON, and isolate concurrent cache loads. Coalesce background API requests and back off transient MusicBrainz failures.

Add phase timing diagnostics and regression tests for caching, concurrency, metadata and folder queries.

Just for fun made by using ChatGpt/ Codex which seems to be usefull. Tested by me.


ik666:
@valib  I had a look at this and I think it covers some points! However, two things I would change before merging:
- The phase timers should not ship: They run per file and per row and they log at INFO, so they flood the log.
- The async translations need to bump the store id. A title that arrives after the browse response has been sent is invisible to the client until something changes the resource's UpdateID

If you don't mind, I'll do the update myself.

